Issue #217, opened 2016-07-13
Refactor packet packing


lucas-clemente:
I propose we refactor the packet packer & framer for reducing allocations (#129) and general performance improvement.

The goals should be:

- Reduce the number of iterations over the streams map in the framer, ideally to 1 per packing.
- Reduce the number of slice allocations, in particular due to growing `bytes.Buffer`s.
- If possible, reduce the number of heap-allocated frames.

A possible solution to most of these goals:

- [x] Remove `framer.HasData` and all calls.
- [x] Replace `framer.EstimatedDataLen` with a function that returns the stream frames for packing (with a threshold for the total length).
- [x] We could think about using a `sync.Pool` of max-packet-sized byte slices. That would increase RAM only if we mostly send small packets, but could possibly reduce allocations a lot.
- [x] We should use in-place encryption.

Maybe we could think about dropping the small packet optimization for stream frames?

lucas-clemente:
As a baseline, `maybeSendPacket` currently takes ~46% of our runtime when profiling on OS X with [the instructions](https://github.com/lucas-clemente/quic-go/wiki#profiling) in the wiki.

<img width="1194" alt="screenshot 2016-07-25 12 36 44" src="https://cloud.githubusercontent.com/assets/918034/17098773/8422cdba-5264-11e6-9a20-cc71656b367b.png">

lucas-clemente:
FFR, @marten-seemann and I decided to drop the small frame optimization for now. Let's think later about re-adding something similar, if we find it's a problem during profiling.

lucas-clemente:
We now have a peak download speed of 66MB/s, a 43% improvement from before.

Current flame graph: https://dl.dropboxusercontent.com/u/1428776/dtrace_after_217.svg

<img width="1190" alt="screenshot 2016-07-26 17 16 33" src="https://cloud.githubusercontent.com/assets/918034/17143782/cebd2f5a-5354-11e6-9372-bfd0528c0bf8.png">

Issue #201, opened 2016-07-05
Switch to using chacha if the client does not support AES


lucas-clemente:
Since currently all clients support AES, this is not a pressing issue.

lucas-clemente:
I'm gonna close this, as we probably don't need chacha before TLS 1.3, and the difficult part here (12 byte tag sizes) are going away with that.

PR #221, opened 2016-07-17
Replace ChaCha20Poly1305 implementation

aead:
Improve AEAD speed with slightly faster poly1305 implementation.
Avoid memory allocations whenever possible. (AEAD)
But currently missing AVX2 support.

BenchmarkSeal64B-8  &nbsp;        &nbsp;  &nbsp;40.97 MB/s
BenchmarkSeal1K-8 &nbsp; &nbsp;  &nbsp; 183.82 MB/s
BenchmarkSeal64K-8  &nbsp;     &nbsp;406.37 MB/s
BenchmarkOpen64B-8    &nbsp; &nbsp; 45.79 MB/s
BenchmarkOpen1K-8  &nbsp;    &nbsp;   181.14 MB/s
BenchmarkOpen64K-8   &nbsp;  417.22 MB/s

### DIFF
--- a/crypto/chacha20poly1305_aead.go
+++ b/crypto/chacha20poly1305_aead.go
@@ -5,7 +5,7 @@ import (
 	"encoding/binary"
 	"errors"
 
-	"github.com/lucas-clemente/chacha20poly1305"
+	"github.com/aead/chacha20"
 
 	"github.com/lucas-clemente/quic-go/protocol"
 )
@@ -22,11 +22,16 @@ func NewAEADChacha20Poly1305(otherKey []byte, myKey []byte, otherIV []byte, myIV
 	if len(myKey) != 32 || len(otherKey) != 32 || len(myIV) != 4 || len(otherIV) != 4 {
 		return nil, errors.New("chacha20poly1305: expected 32-byte keys and 4-byte IVs")
 	}
-	encrypter, err := chacha20poly1305.New(myKey, 12)
+	// copy because ChaCha20Poly1305 expects array pointers
+	var MyKey, OtherKey [32]byte
+	copy(MyKey[:], myKey)
+	copy(OtherKey[:], otherKey)
+
+	encrypter, err := chacha20.NewChaCha20Poly1305WithTagSize(&MyKey, 12)
 	if err != nil {
 		return nil, err
 	}
-	decrypter, err := chacha20poly1305.New(otherKey, 12)
+	decrypter, err := chacha20.NewChaCha20Poly1305WithTagSize(&OtherKey, 12)
 	if err != nil {
 		return nil, err
 	}


aead:
The Travis-CI build seems to fail because of a network error...
> unable to access 'https://go.googlesource.com/tools/

lucas-clemente:
Thanks for this PR! I restarted the travis build and it worked, so I'm going ahead with merging this now.

I've had a quick look at your implementation, and it indeed looks better with regard to allocations. It also allows us to use in-place encryption, which will probably be done in our current milestone (see #217).

However, I'm not sure if you're aware that we're only using AES-GCM right now. We switched since it is around an order of magnitude faster (with the current implementations, see below). Due to a lack of time I didn't build proper negotiation for chacha yet (the client gets to decide, see #201), but I will do this soon-ish.

Just pasting my benchmark results for future reference (of course they don't cover GC pressure):

```
Old implementation:
BenchmarkChaCha20Poly1305Seal1K-8   	  500000	      2804 ns/op	 365.13 MB/s
BenchmarkChaCha20Poly1305Open1K-8   	  500000	      2808 ns/op	 364.60 MB/s

New implementation:
BenchmarkSeal1K-8            	  500000	      2745 ns/op	 373.03 MB/s
BenchmarkOpen1K-8            	  500000	      2732 ns/op	 374.76 MB/s

AES-GCM:
BenchmarkAESGCMSeal1K-8   	 3000000	       466 ns/op	2196.10 MB/s
BenchmarkAESGCMOpen1K-8   	 3000000	       440 ns/op	2326.29 MB/s
```

aead:
Thanks for merging!  

It's quite hard to beat AES-GCM on amd64 with AES-NI - Maybe AVX2 has a chance (but both poly1305 and chacha20 must use it).
I'm working on that as soon as I have some time left.  

Thanks for the benchmark reference - quite helpful!

aead:
Significant performance improvement for poly1305 through amd64 assembly.
But still not comparable with AES-NI:

BenchmarkSeal64B-8 &nbsp; &nbsp; &nbsp; 57.89 MB/s 
BenchmarkSeal1K-8 &nbsp; &nbsp; &nbsp; 298.51 MB/s 
BenchmarkSeal64K-8 &nbsp; &nbsp;  688.23 MB/s 
BenchmarkOpen64B-8 &nbsp; &nbsp;  47.62 MB/s 
BenchmarkOpen1K-8 &nbsp; &nbsp; 302.07 MB/s 
BenchmarkOpen64K-8 &nbsp; 678.04 MB/s

lucas-clemente:
Very cool work! :+1: On my machine it's around 420, up from 370.

```
BenchmarkSeal1K-8            	 1000000	      2435 ns/op	 420.48 MB/s
BenchmarkOpen1K-8            	  500000	      2431 ns/op	 421.14 MB/s
```

Once QUIC gets adopted by mobile browsers (who like choosing chacha over AES) this will become really useful :)

aead:
Thanks :smile:
I've some working (experimental) AVX2 code for ChaCha20 giving a great speedup again, but only go-1.7 will support AVX2 asm instructions (VPSHUFD etc.). The current go 1.7beta-3 have still some problems with AVX2 instructions - so readable asm-code requires the final go 1.7 release. 

The main problem for you now is the compilation. If I add the AVX2 code to the repo (as soon as the release is available), only the 1.7 builds will pass ( <= 1.6.3 will fail). As far as I know Go does not support something like "version depended compilation", so I cannot fix this...
Do you want to keep support for older versions of Go ( <= 1.6.3) ?  

Currently i think i'll create a AVX2 branch for chacha20 and as far as possible (without breaking other's code) make it the master...

Some benchmarks for AVX2 code on a Skylake i7-6500U:
Without AVX2:
- BenchmarkSeal1K-4 565.44 MB/s
- BenchmarkOpen1K-4 559.03 MB/s

With AVX2:
+ BenchmarkSeal1K-4 791.02 MB/s
+ BenchmarkOpen1K-4 779.34 MB/s

lucas-clemente:
I would like to support keep support for 1.6 for the near future.

However, I think it's possible to use Go's build tags for that purpose, see https://golang.org/pkg/go/build/#hdr-Build_Constraints.

In short, if you put `// +build go1.7` into the AVX2 files, they will only be compiled on go 1.7 or up. Similarly for the old implementation.

aead:
Thanks for the info - perfect :+1:
Okay, AVX2 will be supported as soon as 1.7 is stable...

aead:
SSSE3 support available: aead/chacha20@b2542acc9b7dab45d6ac7f7ecf4fce58a06eb79b
Performance improvement about 8-10% on my machine  
Further fixed a critical bug detected by [v2ray](https://github.com/v2ray/v2ray-core/pull/215#issuecomment-235071183) See  aead/chacha20@b9b9fe0318a29cfdf8024f4e2e92a735c23529a3 
Dependency update necessary...

PR #269, opened 2016-08-09
use linked list to store sent packets in new AckHandler

### LINKED ISSUE #246: fix DOS against SentPacketHandler when ACKing too many packets
An attacker could send ACK frames and never raise the `LowestAcked` value. Then we have to iterate over all packets between `LowestAcked` and `LargestAcked` in `SentPacketHandler.ReceivedAcked`.
 This attack would only be limited (after implementing #186) when the number of randomly skipped packets gets larger than 255 (the maximum number of `AckRanges` in an ACK), and since we might skip one packet out of 1000, this is not an effective measure to mitigate this attack.

Ignoring all ACKed packets below the `SentPacketHandler.LargestInOrderAcked` might be a good idea for performance reasons in the case of a non-malicious client, but probably is not a suitable DOS mitigation, since by not ACKing a packet with low packet number (e.g. packet 1), an attacker could prevent this value from increasing (fast enough).

What I'd suggest is to introduce a maximum value for `AckFrame.LargestAcked - AckFrame.LowestAcked`, around O(5000) or so.

marten-seemann:
fixes #246


lucas-clemente:
The patch to the codecov file has a wrong filename, should be `ackhandlerlegacy/packet_linkedlist.go`.

marten-seemann:
@lucas-clemente : I know, and I fixed this already in another commit, but somehow it doesn't show up here. Anyway, I rebase-fixed everything properly now.

PR #294, opened 2016-08-17
Fix benchmark tests

lucas-clemente:


marten-seemann:
This is not a good idea, for (at least) two reasons:

1. running `ginkgo` for the quic package takes a long time and consumes a lot of battery.
2. as @codecov-io is telling us, we're artificially increasing our test coverage. Test coverage should be generated by unit tests only, not by integration tests.

PR #316, opened 2016-09-03
Add support for version 36

### LINKED ISSUE #312: Add support for v36
Just add it to the supported versions list :)

lucas-clemente:
Fixes #312. This PR is blocked until Chrome beta adds support for v36.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ quic-go is an implementation of the [QUIC](https://en.wikipedia.org/wiki/QUIC) p
 
 Done:
 
-- Basic protocol with support for QUIC version 34-35
+- Basic protocol with support for QUIC version 34-36
 - HTTP/2 support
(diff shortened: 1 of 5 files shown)


Aron-Schats:
I think that supporting Q036 means supporting forced head-of-line blocking:

  https://groups.google.com/a/chromium.org/forum/#!topic/proto-quic/NJv1APbKb9g

lucas-clemente:
I saw your question on the list, waiting for an answer too… If it really changes what data comes on which stream we will have to do more stuff. If it doesn't (what I thought so far) we don't have to do anything else. v36 seems to be just about statistics, and I don't think we're gonna mess up Google's data there ;)

Aron-Schats:
It does not seem like they're in a hurry to answer.  Maybe you could prod them by posting a "me, too!" to the thread. :)

PR #328, opened 2016-10-04
fix writing of multiple headers with the same name

chestnutprog:


lucas-clemente:
Thanks for the PR and finding the bug :) I'm happy to merge this with a matching test case. Let me know if you need help writing one.

Also, when rebasing, could you maybe edit the commit message to something more descriptive? E.g. "fix writing of multiple headers with the same name".

PR #353, opened 2016-11-12
Don't rely on GOPATH to load the certificates or error codes

carlosmn:
GOPATH is a list of paths, similar to PATH. If someone does have a list
set, the tests will try to use the full list as a path prefix to load
the certificates, which won't work.

But even handling GOPATH as a list does not solve the issue of what
certificate or error_codes.go file to load, since the code might not be
under the first one it finds.

Instead, use the runtime functionality to get the filename of the path
of the project at compilation time and perform the lookups relative to
that, which guarantees that we're loading the files from the path of the
code that is running.

### DIFF
--- a/testdata/cert.go
+++ b/testdata/cert.go
@@ -2,19 +2,30 @@ package testdata
 
 import (
 	"crypto/tls"
-	"os"
+	"path"
+	"runtime"
 )
 
 var certPath string
 
 func init() {
-	certPath = os.Getenv("GOPATH")
-	certPath += "/src/github.com/lucas-clemente/quic-go/example"
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		panic("Failed to get current frame")
+	}
+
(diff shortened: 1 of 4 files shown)


testdata/cert.go
```diff
@@ -2,19 +2,32 @@ package testdata
 
 import (
 	"crypto/tls"
-	"os"
+	"path"
+	"runtime"
 )
 
+const examplePath = "src/github.com/lucas-clemente/quic-go/example"
```
lucas-clemente: I think this line can be removed, right?

lucas-clemente:
Thanks for the PR and sorry for taking a few days to review it! I'm happy to merge this, especially with Go 1.8 coming up (where you don't need an explicit GOPATH).

testdata/cert.go
carlosmn: Indeed. Removed.

PR #373, opened 2016-12-19
wait until the run loop has stopped before returning Session.Close()

### LINKED ISSUE #371: make sure the ConnectionClose is sent before closing UDP connection
In the `Session.closeImpl` we put the error in the `closeChan`, and hope that it is sent by the `run` loop before the `closeImpl` returns. However, this is not guaranteed, because `run` is running in a separate go routine.

marten-seemann:
fixes #371

### DIFF
--- a/session.go
+++ b/session.go
@@ -32,6 +32,7 @@ type receivedPacket struct {
 var (
 	errRstStreamOnInvalidStream   = errors.New("RST_STREAM received for unknown stream")
 	errWindowUpdateOnClosedStream = errors.New("WINDOW_UPDATE received for an already closed stream")
+	errSessionAlreadyClosed       = errors.New("Cannot close Session. It was already closed before.")
 )
 
 // StreamCallback gets a stream frame and returns a reply frame
@@ -153,7 +154,7 @@ func (s *Session) run() {
 	// Start the crypto stream handler
 	go func() {
 		if err := s.cryptoSetup.HandleCryptoStream(); err != nil {
-			s.close(err)
+			s.Close(err)
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -407,7 +413,22 @@ func (s *Session) handleAckFrame(frame *frames.AckFrame) error {
...
+	err := s.closeImpl(e, false)
+
+	if atomic.LoadUint32(&s.closed) == 1 {
+		return err
+	}
+
+	select {
+	case <-s.runClosed:
+		return err
+	}
```
lucas-clemente: You can drop the select statement here and just simply say `<-s.runClosed`.

session.go
```diff
@@ -407,7 +413,22 @@ func (s *Session) handleAckFrame(frame *frames.AckFrame) error {
 }
 
 // Close the connection. If err is nil it will be set to qerr.PeerGoingAway.
+// It waits until the run loop has stopped before returning
 func (s *Session) Close(e error) error {
+	err := s.closeImpl(e, false)
+
+	if atomic.LoadUint32(&s.closed) == 1 {
```
lucas-clemente: Can this ever evaluate to false? The first statement in `closeImpl` sets this atomic to true, so we will never reach the code below, right?

lucas-clemente: You could test for that using `Expect(s.runClosed).To(NotReceive())` or similar, since the channel should have been drained here.

session.go
marten-seemann: True. Thanks.

session.go
marten-seemann: Right. I was trying to catch multiple calls to `Close` there, but it doesn't work that way.
I added a fix, that now works with an explicit error returned by the `closeImpl` in case the session is already closed.

Issue #315, opened 2016-09-01
Fix only-ack detection on windows


lucas-clemente:
I suspect this is an issue with the windows timer resolution, leading to us not sending ACK frames. Log: https://dl.dropboxusercontent.com/u/1428776/log_windows_timer_ack.txt. It looks like sometimes timers are not firing properly? We should probably test this in isolation. For now, I've simply disabled the detection on windows.

Apparently the improved timers in 1.7 were reverted, see https://github.com/golang/go/issues/8687, maybe this is related?

PR #388, opened 2017-01-09
implement a better ACK algorithm

### LINKED ISSUE #354: improve ACK algorithm
We currently send ACKs only with other frames, or after 5ms have expired. Under some situations this can cause problems, since it massively delays the congestion window increase, see the attached traces. #340 is a workaround, but we should still fix this issue.

screenshot 2016-11-17 15 50 52
screenshot 2016-11-17 15 50 59

marten-seemann:
fixes #354 and #315.

### DIFF
--- a/ackhandler/interfaces.go
+++ b/ackhandler/interfaces.go
@@ -28,8 +28,8 @@ type SentPacketHandler interface {
 
 // ReceivedPacketHandler handles ACKs needed to send for incoming packets
 type ReceivedPacketHandler interface {
-	ReceivedPacket(packetNumber protocol.PacketNumber) error
+	ReceivedPacket(packetNumber protocol.PacketNumber, shouldInstigateAck bool) error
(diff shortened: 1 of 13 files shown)


marten-seemann:
This PR implements an ACK algorithm similar to the one Chromium is using (in TCP_ACKING mode).

Note that the transfer speed in the benchmark tests significantly decreases, on my Macbook from about 200 MB/s to 135 MB/s and on Travis from 11.5 MB/s to 10.7 MB/s. 

The reason for this is *not* a timer issue in the `Session`, but seems to be caused by us sending more ACKs for retransmittable packets: If the packets are arriving fast enough (i.e. less than 5ms delay between two packets), we're now sending an ACK for every second packet. Before, we only sent one ACK every 5ms.
Increasing `protocol.RetransmittablePacketsBeforeAck`, which is now equal to 2, and determines every how many retransmittable packets we send an ACK, to about 20 increases the benchmark to 205 MB/s.

PR #420, opened 2017-02-16
improved API

marten-seemann:


streams_map.go
```diff
@@ -69,14 +71,42 @@ func (m *streamsMap) GetOrOpenStream(id protocol.StreamID) (*stream, error) {
...
 		return s, nil
 	}
+
+	if id <= m.highestStreamOpenedByPeer {
+		return nil, nil
+	}
+
+	highestOpened := m.highestStreamOpenedByPeer
+	sid := id
+	for sid > highestOpened {
```
lucas-clemente: Rewrite this as `for sid := id; sid > highestOpened; sid-=2 { ... }`

streams_map.go
```diff
@@ -69,14 +71,42 @@ func (m *streamsMap) GetOrOpenStream(id protocol.StreamID) (*stream, error) {
...
+		_, err := m.openRemoteStream(sid)
+		if err != nil {
+			return nil, err
+		}
+		if sid == 1 {
+			break
+		}
+		sid -= 2
```
lucas-clemente: Is it obvious that sid will never underflow?

streams_map.go
```diff
@@ -164,17 +165,19 @@ func (m *streamsMap) OpenStream() (*stream, error) {
...
 // AcceptStream returns the next stream opened by the peer
 // it blocks until a new stream is opened
-// TODO: implement error conditions
 func (m *streamsMap) AcceptStream() (utils.Stream, error) {
 	m.mutex.Lock()
 	var str utils.Stream
 	for {
 		var ok bool
+		if m.closeErr != nil {
+			return nil, m.closeErr
```
lucas-clemente: The mutex is not unlocked in this return path. Insert a `defer m.mutex.Unlock()` right after locking and remove the `Unlock()` below.

streams_map.go
marten-seemann: Yes, isn't it?

streams_map.go
lucas-clemente: The mutex is not unlocked in this return path. Insert a `defer m.mutex.Unlock()` right after locking and remove the `Unlock()` below.

streams_map.go
marten-seemann: I wanted to rewrite this loop anyway, such that streams are opened in order. It just seemed easier to me to implement this way first.

streams_map.go
lucas-clemente: Maybe add a comment that highestStreamOpenedByPeer (and therefore sid) is always odd.

streams_map.go
lucas-clemente: sgtm

lucas-clemente:
LGTM modulo the mutex comment.

h2quic/server.go
```diff
@@ -20,10 +20,8 @@ import (
 )
 
 type streamCreator interface {
-	AcceptStream() (utils.Stream, error)
+	quic.Session
```
lucas-clemente: I think the normal pattern is to have only one function in these types of interfaces, and to type assert onto it just before the method call. It doesn't matter, just fyi.

h2quic/server.go
```diff
@@ -72,31 +70,41 @@ func (s *Server) serveImpl(tlsConfig *tls.Config, conn *net.UDPConn) error {
 	if s.Server == nil {
 		return errors.New("use of h2quic.Server without http.Server")
 	}
-	s.serverMutex.Lock()
-	if s.server != nil {
-		s.serverMutex.Unlock()
+	s.listenerMutex.Lock()
```
lucas-clemente: defer Unlock() here please, no reason not to (especially since defer is fast as of 1.8)

interface.go
```diff
@@ -20,3 +21,34 @@ type Session interface {
...
+// ConnState is the status of the connection
+type ConnState int
+
+const (
+	// ConnStateVersionNegotiated means that version negotiation is complete
+	ConnStateVersionNegotiated ConnState = iota
+	// ConnStateSecure means that the connection is encrypted
+	ConnStateSecure
+	// ConnStateForwardSecure means that the connection is forward secure
+	ConnStateForwardSecure
```
lucas-clemente: I think these need to be directional, i.e. ConnStateSendingSecure, ConnStateReceivingSecure. Otherwise you cannot do 0-RTT properly, right?

h2quic/client_test.go
```diff
@@ -4,97 +4,102 @@ import (
 	"bytes"
 	"compress/gzip"
 	"errors"
+	"net"
 	"net/http"
 
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
 
+	quic "github.com/lucas-clemente/quic-go"
```
lucas-clemente: Unnecessary `quic`

h2quic/request_body.go
```diff
@@ -3,18 +3,18 @@ package h2quic
 import (
 	"io"
 
-	"github.com/lucas-clemente/quic-go/utils"
+	quic "github.com/lucas-clemente/quic-go"
```
lucas-clemente: Unnecessary `quic`

h2quic/request_writer.go
```diff
@@ -12,21 +12,22 @@ import (
 	"golang.org/x/net/http2/hpack"
 	"golang.org/x/net/lex/httplex"
 
+	quic "github.com/lucas-clemente/quic-go"
```
lucas-clemente: Unnecessary `quic`

h2quic/response_writer.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 
+	quic "github.com/lucas-clemente/quic-go"
```
lucas-clemente: Unnecessary `quic`

h2quic/server_test.go
```diff
@@ -13,10 +13,10 @@ import (
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
 
+	quic "github.com/lucas-clemente/quic-go"
```
lucas-clemente: Unnecessary `quic`

client.go
```diff
@@ -84,7 +67,13 @@ func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config
 	go c.listen()
 
 	c.mutex.Lock()
-	for !c.versionNegotiated {
+	for {
```
lucas-clemente: This loop will never terminate if the connection errors.

interface.go
marten-seemann: I think the server-side is pretty clear here (`ConnStateVersionNegotiated` and `ConnStateSecure` happen at the same time, after receiving the first packet from the client. So I guess your comment is referring to the client-side of a 0-RTT.

In that case `ConnStateSecure` would happen immediately (i.e. before the first packet is sent), whereas `ConnStateVersionNegotiated` only fires after receiving the first packet from the server. This will require some changes in the client (I'm assuming an ordering of the `ConnState`s at some places), but I don't think any API change is needed.
In case of a rejected 0-RTT `ConnStateSecure` would fire twice though. This is something we probably should document sometime somewhere.

h2quic/client_test.go
marten-seemann: Can you please clarify? `quic` is needed because we're implementing mock interfaces, e.g. for the `Session`, in the `h2quic` tests. How do you intent to get rid of importing `quic` then?

h2quic/server.go
marten-seemann: Unfortunately, this doesn't work here. `quic.Listener.Serve`blocks indefinitely, which means that the `defer` will never be executed, and the mutex would never be released.

I created a minimal working example in the Go Playground (https://play.golang.org/p/uRP3khNBBF).

h2quic/client_test.go
marten-seemann: Maybe you mean that it would be sufficient to just write `"github.com/lucas-clemente/quic-go"` instead of `quic "github.com/lucas-clemente/quic-go"`. That's true, but `gofmt` always adds the `quic`, so I'd rather leave it in here.

client.go
marten-seemann: Nice catch!

h2quic/server.go
lucas-clemente: Right, missed that.

h2quic/client_test.go
lucas-clemente: sgtm

interface.go
lucas-clemente: Server:

There's a period of time between ConnStateVersionNegotiated and ConnStateSecure, e.g. if the client has to present STKs before the server sends it the full SCFG. During this phase, the server sends insecure data.

I agree that for the server there's no period between receiving the first (fw) secure data and sending the first (fw) secure data.

Client:

I don't think ConnStateSecure should fire twice. Our API hides the effects of 0-RTT (we don't need the user to re-`Write` stream data after all), so I don't see why we should do this here.

Still, there's a period of time during a slow handshake where we are sending (fw) secure data but receiving insecure data. Are you saying that doesn't convey any information that warrants escalating through our API?

lucas-clemente:
LGTM

PR #421, opened 2017-02-17
get the respone context into files

keith2008:
the origin example just print the response context in terminal. i think get the response context into files, will be more available when the content length is huge.

### DIFF
--- a/example/client/main.go
+++ b/example/client/main.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"net/http"
 	"sync"
+  "strings"
+  "os"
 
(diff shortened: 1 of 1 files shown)


lucas-clemente:
Thanks for the PR. I'm okay with accepting this feature behind a command line flag that specifies the filename. Could you update your PR to do that, and also treat the error value you're currently ignoring (panicing is fine)?

lucas-clemente:
Closing this as there hasn't been any activity in a month. Feel free to reopen, taking into account my comment above.

PR #422, opened 2017-02-17
Update main.go

keith2008:
make certFile&keyfile as input parameter,which will be more  flexible.


lucas-clemente:
I'm not sure I'm a fan of this much configurability. Could you explain why you need that in particular?

Also, note that these are just examples. You can very easily build your own server by copying the code.

lucas-clemente:
Closing this as there hasn't been any activity in a month. Feel free to reopen, taking into account my comment above.

PR #427, opened 2017-02-18
move Public Header to separate package

### LINKED ISSUE #412: move PublicHeader to separate packet
It could be useful e.g. for building a QUIC proxy.

We have to think about how we make the packet number accessible, since the header itself just contains the shortened packet number.

marten-seemann:
The only thing I don't like about this is that a PH is now a `publicheader.PublicHeader`, but I don't see a nice way to solve this.

Fixes #412.


marten-seemann:
@codecov-io is complaining about *steams_map.go*, which wasn't changed at all. Fail!

marten-seemann:
We decided to not do this, in #412.

PR #428, opened 2017-02-18
support unset GOPATH in Go 1.8 in integrationtests

### LINKED ISSUE #415: Support unset GOPATH in our tests on Go 1.8

marten-seemann:
fixes #415

### DIFF
--- a/integrationtests/drop_test.go
+++ b/integrationtests/drop_test.go
@@ -3,9 +3,7 @@ package integrationtests
 import (
 	"bytes"
 	"fmt"
-	"os"
 	"os/exec"
-	"runtime"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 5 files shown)


lucas-clemente:
From the docs at https://golang.org/pkg/go/build/:

> Default is the default Context for builds. It uses the GOARCH, GOOS, GOROOT, and GOPATH environment variables if set, or else the compiled code's GOARCH, GOOS, and GOROOT.

This sounds like the value you're using is the GOPATH at buildtime, not at runtime.

Anyway, I think a better solution would be the thing we do in testdata (use `runtime.Caller(0)`). Would you mind changing this commit to do that?

marten-seemann:
Well, considering that this a test, I don't expect anyone to build a binary and execute it later (or on another machine) with a different GOPATH. I'll change it anyway.

PR #432, opened 2017-02-21
run benchmark tests with integration tests on travis

### LINKED ISSUE #430: limit code coverage to unit tests
At the moment it includes the benchmark test as well.

marten-seemann:
Fixes #430 

Considering that the benchmark test is more like an integration test than a unit test, I think it makes sense to run it in the *integration* Travis job.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -4,9 +4,10 @@ set -e
 
 go get -t ./...
 if [ ${TESTMODE} == "unit" ]; then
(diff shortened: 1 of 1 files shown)


marten-seemann:
The decreased code coverage reported by @codecov-io is expected and exactly the point of this PR. It's not a lot, but we should consider writing real unit tests for the now uncovered code parts.

PR #448, opened 2017-02-25
Improve documentation for our new API

lucas-clemente:


interface.go
```diff
@@ -45,20 +50,24 @@ const (
 	ConnStateForwardSecure
 )
 
-// ConnStateCallback is called every time the connection moves to another connection state
-// the callback is called in a new go routine
+// ConnStateCallback is called every time the connection moves to another connection state.
 type ConnStateCallback func(Session, ConnState)
 
-// Config is the configuration for QUIC
+// Config contains all configuration data needed for a QUIC server or client.
```
marten-seemann: Maybe we should state here that new configuration options will be added soon.

interface.go
```diff
@@ -45,20 +50,24 @@ const (
...
 type Config struct {
 	TLSConfig *tls.Config
-	// will be called in a separate goroutine
+	// ConnStateCallback will be called when the QUIC version is successfully negotiated or when the encryption level changes.
+	// If this field is not set, the Dial functions will return only when the connection is forward secure.
+	// The callback may be called in a separate goroutine.
```
marten-seemann: I don't like that sentence, the "may" sounds like we're allowing the user to do something, where in fact we're reserving us the right to call the callback in a goroutine.
Maybe we could write something along these lines:

> Callbacks have to be thread-safe, since they might be called in a separate goroutine.

or "executed concurrently".

public_header.go
```diff
@@ -17,7 +17,7 @@ var (
 	errGetLengthNotForVersionNegotiation = errors.New("PublicHeader: GetLength cannot be called for VersionNegotiation packets")
 )
 
-// The PublicHeader of a QUIC packet
+// The PublicHeader of a QUIC packet.
```
marten-seemann: It's the whole `PublicHeader` that will change soon, and we should state that here, not in the member functions.

If Jana's proposal for a new header is accepted, we might also want to rename the fields here. For example, there won't be a dedicated version flag anymore, so a more appropriate name might be `ContainsVersion` instead of `VersionFlag`.

public_header.go
```diff
@@ -31,7 +31,7 @@ type PublicHeader struct {
 	DiversificationNonce []byte
 }
 
-// Write writes a public header
+// Write writes a public header. Warning: This API should not be considered stable and will change soon.
```
marten-seemann: See comment for above.

public_header.go
```diff
@@ -109,8 +109,9 @@ func (h *PublicHeader) Write(b *bytes.Buffer, version protocol.VersionNumber, pe
 	return nil
 }
 
-// ParsePublicHeader parses a QUIC packet's public header
-// the packetSentBy is the perspective of the peer that sent this PublicHeader, i.e. if we're the server, packetSentBy should be PerspectiveClient
+// ParsePublicHeader parses a QUIC packet's public header.
+// The packetSentBy is the perspective of the peer that sent this PublicHeader, i.e. if we're the server, packetSentBy should be PerspectiveClient.
+// Warning: This API should not be considered stable and will change soon.
```
marten-seemann: See comment for above.

public_header.go
```diff
@@ -216,8 +217,8 @@ func ParsePublicHeader(b *bytes.Reader, packetSentBy protocol.Perspective) (*Pub
 	return header, nil
 }
 
-// GetLength gets the length of the publicHeader in bytes
-// can only be called for regular packets
+// GetLength gets the length of the publicHeader in bytes.
+// It can only be called for regular packets.
```
marten-seemann: Does this need to be public?

interface.go
```diff
@@ -8,26 +8,31 @@ import (
...
+	// New streams always have the smallest possible stream ID.
+	// TODO: Enable testing for the special error
 	OpenStream() (Stream, error)
-	// blocks until a new stream can be opened, if the maximum number of stream is opened
+	// OpenStreamSync opens a new QUIC stream, blocking until the peer's concurrent stream limit allows a new stream to be opened.
+	// It always picks the smallest possible stream ID.
 	OpenStreamSync() (Stream, error)
+	// RemoteAddr returns the address of the peer.
 	RemoteAddr() net.Addr
+	// Close closes the connection. An error value of nil is allowed.
```
marten-seemann: We should explain the error parameter.

interface.go
lucas-clemente: done

interface.go
lucas-clemente: done

public_header.go
lucas-clemente: done

public_header.go
lucas-clemente: added the comment above and left it here too.

public_header.go
lucas-clemente: added the comment above and left it here too.

public_header.go
lucas-clemente: I don't have an opinion on that, but since we are stating quite clearly that it's unstable I would leave it in for now, since I don't see any advantage in privatizing it.

interface.go
lucas-clemente: done

PR #449, opened 2017-02-25
immediately switch to forward-secure crypto after sending the SHLO

### LINKED ISSUE #391: immediately switch to forward-secure encryption after sending the SHLO
Once the server sent the SHLO, it can encrypt the next packet it sends with forward-secure encryption. The client will be able to decrypt those packets once it has read the SHLO.

The reason why we haven't implemented this yet is that if the SHLO gets lost, we need to retransmit it with the initial encryption level, since the client will never be able to encrypt it if we encrypt it forward-secure. This mean, we need to implement some more logic for the retransmission of handshake packets.

An on-path attacker can drop the SHLO and prevent an upgrade to forward-secure crypto altogether. The session will be closed after the MaxTimeForCryptoHandshake (set to 10s), but in these 10s all data is only encrypted with initial encryption.

marten-seemann:
This requires us to pay more attention to the encryption level of retransmission. The SHLO (which is always sent with initial encryption) must be retransmitted with initial encryption level.

Fixes #391.

### DIFF
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -104,7 +104,8 @@ func (p *packetPacker) packPacket(stopWaitingFrame *frames.StopWaitingFrame, lea
 	if isConnectionClose {
 		payloadFrames = []frames.Frame{p.controlFrames[0]}
 	} else {
-		payloadFrames, err = p.composeNextPacket(stopWaitingFrame, publicHeaderLength)
+		maxSize := protocol.MaxFrameAndPublicHeaderSize - publicHeaderLength
+		payloadFrames, err = p.composeNextPacket(stopWaitingFrame, maxSize)
 		if err != nil {
 			return nil, err
 		}
@@ -164,11 +165,9 @@ func (p *packetPacker) packPacket(stopWaitingFrame *frames.StopWaitingFrame, lea
 	}, nil
 }
 
-func (p *packetPacker) composeNextPacket(stopWaitingFrame *frames.StopWaitingFrame, publicHeaderLength protocol.ByteCount) ([]frames.Frame, error) {
+func (p *packetPacker) composeNextPacket(stopWaitingFrame *frames.StopWaitingFrame, maxFrameSize protocol.ByteCount) ([]frames.Frame, error) {
 	var payloadLength protocol.ByteCount
 	var payloadFrames []frames.Frame
-
-	maxFrameSize := protocol.MaxFrameAndPublicHeaderSize - publicHeaderLength
 
 	if stopWaitingFrame != nil {
 		payloadFrames = append(payloadFrames, stopWaitingFrame)
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/lucas-clemente/quic-go/ackhandler"
 	"github.com/lucas-clemente/quic-go/frames"
 	"github.com/lucas-clemente/quic-go/handshake"
 	"github.com/lucas-clemente/quic-go/protocol"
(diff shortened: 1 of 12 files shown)


protocol/server_parameters.go
```diff
@@ -2,6 +2,10 @@ package protocol
 
 import "time"
 
+// NonForwardSecurePacketSizeReduction is the number of bytes a non forward-secure packet has to be smaller than a forward-secure packet
+// This makes sure that those packets can always be retransmitted without splitting the contained StreamFrames
+const NonForwardSecurePacketSizeReduction = 50
```
lucas-clemente: I would prefer to not have a magic number, but construct this value out of other constants we already have.

packet_packer.go
```diff
@@ -51,18 +52,31 @@ func (p *packetPacker) PackConnectionClose(ccf *frames.ConnectionCloseFrame, lea
 	// in case the connection is closed, all queued control frames aren't of any use anymore
 	// discard them and queue the ConnectionCloseFrame
 	p.controlFrames = []frames.Frame{ccf}
-	return p.packPacket(nil, leastUnacked)
+	return p.packPacket(nil, leastUnacked, nil)
+}
+
+// RetransmitHandshakePacket retransmits a handshake packet, that was sent with less than forward-secure encryption
```
lucas-clemente: Update comment to match function name

packet_packer.go
```diff
@@ -51,18 +52,31 @@ func (p *packetPacker) PackConnectionClose(ccf *frames.ConnectionCloseFrame, lea
...
 // the other controlFrames are sent in the next packet, but might be queued and sent in the next packet if the packet would overflow MaxPacketSize otherwise
 func (p *packetPacker) PackPacket(stopWaitingFrame *frames.StopWaitingFrame, controlFrames []frames.Frame, leastUnacked protocol.PacketNumber) (*packedPacket, error) {
 	p.controlFrames = append(p.controlFrames, controlFrames...)
-	return p.packPacket(stopWaitingFrame, leastUnacked)
+	return p.packPacket(stopWaitingFrame, leastUnacked, nil)
 }
 
-func (p *packetPacker) packPacket(stopWaitingFrame *frames.StopWaitingFrame, leastUnacked protocol.PacketNumber) (*packedPacket, error) {
+func (p *packetPacker) packPacket(stopWaitingFrame *frames.StopWaitingFrame, leastUnacked protocol.PacketNumber, packetToRetransmit *ackhandler.Packet) (*packedPacket, error) {
+	isHandshakeRetransmission := (packetToRetransmit != nil)
```
lucas-clemente: I find it non-obvious that this is the case. Either add a comment why (i.e. packetToRetransmit is only ever set for handshake packets), or, better, rename it to handshakePacketToRetransmit or similar.

packet_packer.go
```diff
@@ -150,7 +176,16 @@ func (p *packetPacker) packPacket(stopWaitingFrame *frames.StopWaitingFrame, lea
...
-	_, encryptionLevel := p.cryptoSetup.Seal(raw[payloadStartIndex:payloadStartIndex], raw[payloadStartIndex:], currentPacketNumber, raw[:payloadStartIndex])
+	var encryptionLevel protocol.EncryptionLevel
+	if isHandshakeRetransmission {
+		var err error
+		_, encryptionLevel, err = p.cryptoSetup.SealWith(raw[payloadStartIndex:payloadStartIndex], raw[payloadStartIndex:], currentPacketNumber, raw[:payloadStartIndex], packetToRetransmit.EncryptionLevel)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		_, encryptionLevel = p.cryptoSetup.Seal(raw[payloadStartIndex:payloadStartIndex], raw[payloadStartIndex:], currentPacketNumber, raw[:payloadStartIndex])
```
lucas-clemente: Just a thought: Could we implement the latching logic for the encrypter here, so that we only need the SealWith function and not the generic Seal?

protocol/server_parameters.go
marten-seemann: I'll create an issue to refactor the retransmission logic of handshake packets, such that we can get rid of this `NonForwardSecurePacketSizeReducton`. It works for now, but it's not a clean solution anyway, and it won't get a lot cleaner by calculation the exact value here.

packet_packer.go
marten-seemann: I doubt it, because the latching logic is different for server and client. I think it's better to hide it behind the `CryptoSetup` interface.

protocol/server_parameters.go
marten-seemann: See #457.

PR #453, opened 2017-02-27
reject stream frames that overflow the offset

### LINKED ISSUE #452: reject stream frames that overflow the offset
i.e. a `StreamFrame` with a `ByteOffset` close to `MaxByteCount`.

A frame where `frame.Offset + len(frame.Data)` overflows `MaxByteCount` is currently accepted (and saved in the `streamFrameSorter`.

marten-seemann:
fixes #452


marten-seemann:
Don't trust @codecov-io here, nothing was changed in *client.go*.

PR #465, opened 2017-03-05
run Go scheduler before doing trial decryption

marten-seemann:
In the integration test of the https://github.com/marten-seemann/quic-conn/ I'm seeing the client send Public Resets because it received too many undecryptable packets, although it already received the packet containing the SHLO. The problems seems to be that the packets are received in a burst (the RTT is tiny on the local interface), such that the Go routine handling the crypto stream is sometimes not scheduled during the time when many packets (more than `MaxUndecryptablePackets`) are received.

Calling `runtime.Gosched()` should solve this issue (or at least make it *a lot* less likely). The performance impact is expected to be negligible, since it's only called when the decryption of a packet fails.

This issue would fix https://github.com/marten-seemann/quic-conn/issues/3. We can also think about increasing `MaxUndecryptablePackets` (currently set to 10) for the client.

### DIFF
--- a/session.go
+++ b/session.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"runtime"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
This is a real issue, and there needs to be better solution than poking the Go scheduler.

Issue #468, opened 2017-03-06
detection of too large packets not working


marten-seemann:
We're reading packets from the UDPConn into a buffer of the size `MaxPacketSize`. Thus, if a peer sends out a packet exceeding the `MaxPacketSize`, we'll cut the packet when reading it, never triggering the error condition for the `PacketTooLarge` error.

This applies to both server and client.

lucas-clemente:
There's the possibility to do a second readfrom with MSG_PEEK set just ahead of the current one. But I'm not sure detecting this error is actually worth the double amount of syscalls, which would probably have a noticeable performance impact.

marten-seemann:
I think it's definitely not worth that many syscalls, however, we could make the buffers we read into one byte bigger than the maximum allowed packet size, and see if a packet completely fills that buffer (and then send a `PacketTooLarge` error).

marten-seemann:
We need to decide if we properly want to implement this check (as described above?). Otherwise we should remove the useless code.

lucas-clemente:
I don't think we need to add that check, but I'd like to keep the if as a sanity check anyway.

lucas-clemente:
This was bugging me, so I did some more research, and found a solution that doesn't require a second syscall: The syscall.MSG_TRUNC flag is set in the data returned by [ReadMsgUdp](https://golang.org/pkg/net/#UDPConn.ReadMsgUDP). That however doesn't work with net.PacketConn. Tbh I think this is an issue with net.PacketConn, but there's nothing we can do about that.

marten-seemann:
I'll remove the check for too large packets. If a peer really decides to send us a too large packet, we will read the truncated version of it and process it normally. That means, this packet will end up being undecryptable, and will be dropped anyway. This way, a peer doing PMTUD will learn about our maximum packet size.

There's a better signaling for this on the horizon (see https://github.com/quicwg/base-drafts/issues/383), so we'll come back to this when the WG makes a decision here.

PR #469, opened 2017-03-06
read packets with the maximum packet size

### LINKED ISSUE #467: too small MaxPacketSize for receiving packets
We're currently defining a `MaxPacketSize` of 1350 bytes.

According to the Wire Spec, the maximum packet size is 1350 bytes for IPv6 and 1370 bytes for IPv4. That means that for sending packets, we're complying with the spec. However, we're only allocating `MaxPacketSize` bytes for reading packets in the [server.go:L86](https://github.com/lucas-clemente/quic-go/blob/eff9c6f90649c8aa63b155b1143237c6120fba96/server.go#L86), which means that we're not able to read packets using the full 1370 bytes allowed on an IPv4 connection.

marten-seemann:
fixes #467

### DIFF
--- a/client.go
+++ b/client.go
@@ -115,7 +115,7 @@ func (c *client) listen() {
 		var n int
 		var addr net.Addr
 		data := getPacketBuffer()
-		data = data[:protocol.MaxPacketSize]
+		data = data[:protocol.MaxReceivePacketSize]
 
 		n, addr, err = c.conn.Read(data)
 		if err != nil {
@@ -141,7 +141,7 @@ func (c *client) listen() {
 }
 
 func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
-	if protocol.ByteCount(len(packet)) > protocol.MaxPacketSize {
+	if protocol.ByteCount(len(packet)) > protocol.MaxReceivePacketSize {
 		return qerr.PacketTooLarge
 	}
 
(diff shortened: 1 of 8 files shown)


client.go
```diff
@@ -141,7 +141,7 @@ func (c *client) listen() {
 }
 
 func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
-	if protocol.ByteCount(len(packet)) > protocol.MaxPacketSize {
+	if protocol.ByteCount(len(packet)) > protocol.MaxReceivePacketSize {
```
lucas-clemente: Can this actually ever happen if we only read packets of a certain size? Can you look into whether there's some kind of signal from the kernel when packets got truncated?

server.go
```diff
@@ -122,7 +122,7 @@ func (s *server) Addr() net.Addr {
 }
 
 func (s *server) handlePacket(pconn net.PacketConn, remoteAddr net.Addr, packet []byte) error {
-	if protocol.ByteCount(len(packet)) > protocol.MaxPacketSize {
+	if protocol.ByteCount(len(packet)) > protocol.MaxReceivePacketSize {
```
lucas-clemente: see above

client.go
marten-seemann: It cannot, and I already opened a new issue for that: #468. Afaik there's no signal, so we'll have to allocate 1 more byte in the packet buffer if we want to perform this check.

lucas-clemente:
lgtm (the review button vanished).

PR #473, opened 2017-03-08
only send Public Reset after a timeout

### LINKED ISSUE #471: send public reset after a timeout, after receiving too many undecryptable packets
This is a better solution than #465.

marten-seemann:
fixes #471

### DIFF
--- a/session.go
+++ b/session.go
@@ -76,8 +76,12 @@ type session struct {
 	runClosed chan struct{}
 	closed    uint32 // atomic bool
 
-	undecryptablePackets []*receivedPacket
-	aeadChanged          chan protocol.EncryptionLevel
+	// when we receive too many undecryptable packets during the handshake, we send a Public reset
+	// but only after a time of protocol.PublicResetTimeout has passed
+	undecryptablePackets                   []*receivedPacket
+	receivedTooManyUndecrytablePacketsTime time.Time
+
+	aeadChanged chan protocol.EncryptionLevel
 
 	nextAckScheduledTime time.Time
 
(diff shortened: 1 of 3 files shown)


marten-seemann:
The AppVeyor failures have nothing to do with this PR, but I can't restart the build there 👎

lucas-clemente:
Restarted

session.go
```diff
@@ -250,6 +254,9 @@ runLoop:
 		if err := s.sendPacket(); err != nil {
 			s.close(err)
 		}
+		if !s.receivedTooManyUndecrytablePackets.IsZero() && s.receivedTooManyUndecrytablePackets.Add(protocol.PublicResetTimeout).Before(time.Now()) {
```
lucas-clemente: Add `now := time.Now()` just above this line and use that variable here and in the next 2 conditions.

session.go
```diff
@@ -276,6 +283,9 @@ func (s *session) maybeResetTimer() {
 		handshakeDeadline := s.sessionCreationTime.Add(protocol.MaxTimeForCryptoHandshake)
 		nextDeadline = utils.MinTime(nextDeadline, handshakeDeadline)
 	}
+	if !s.receivedTooManyUndecrytablePackets.IsZero() {
+		nextDeadline = utils.MinTime(nextDeadline, time.Now().Add(protocol.PublicResetTimeout))
```
lucas-clemente: time.Now() is wrong here, replace with receivedTooManyUndecrytablePackets.

session.go
```diff
@@ -76,8 +76,12 @@ type session struct {
 	runClosed chan struct{}
 	closed    uint32 // atomic bool
 
-	undecryptablePackets []*receivedPacket
-	aeadChanged          chan protocol.EncryptionLevel
+	// when we receive too many undecryptable packets during the handshake, we send a Public reset
+	// but only after a time of protocol.PublicResetTimeout has passed
+	undecryptablePackets               []*receivedPacket
+	receivedTooManyUndecrytablePackets time.Time
```
lucas-clemente: I don't like the variable name, it sounds like a bool. Would you mind adding "Time" as a suffix?

session.go
```diff
@@ -774,8 +784,9 @@ func (s *session) tryQueueingUndecryptablePacket(p *receivedPacket) {
 		return
 	}
 	utils.Infof("Queueing packet 0x%x for later decryption", p.publicHeader.PacketNumber)
-	if len(s.undecryptablePackets)+1 >= protocol.MaxUndecryptablePackets {
-		s.close(qerr.Error(qerr.DecryptionFailure, "too many undecryptable packets received"))
+	if len(s.undecryptablePackets)+1 >= protocol.MaxUndecryptablePackets && s.receivedTooManyUndecrytablePackets.IsZero() {
+		s.receivedTooManyUndecrytablePackets = time.Now()
+		s.maybeResetTimer()
```
lucas-clemente: I think we should return here and not append the packet to the list, so that the size is always constrained by MaxUndecryptablePackets.

session.go
marten-seemann: You're right, but it's already that long :(

lucas-clemente: To quote effective Go: Don't worry about overflowing a punched card 😜

Issue #343, opened 2016-11-02
download stalling because congestion limited


marten-seemann:
When running the Chrome small file download integration test, about every 1 in 100 attempts the download stalls on Ubuntu, because we're congestion limited.

See attached [log.txt](https://github.com/lucas-clemente/quic-go/files/566188/log.txt).

marten-seemann:
I ran a couple of tests, and this is what I found out: The download begins to stall once we start queueing RTOs. This is a local, lossless connection, so that isn't expected to happen at all.

Queueing of RTOs begins once the client sends an ACK with a very high delay time. Usually, the ACK frames received in the tests have delay times of about 10 μs, the ACK frame that triggers the RTOs has a delay time of 23 ms (!). The RTT estimate at that point was 5 ms.

f3z0:
Hi Marten - can you re-run chrome small file download integration test after applying PR #492 to see if the issue is resolved?

lucas-clemente:
This should be fixed by #502.

PR #474, opened 2017-03-09
fixed bug where tls.config was ignored in h2quic.client, removed quic…

f3z0:
….clemente.io private key and configured example client to skip verification to allow self signing the localhost certificate


f3z0:
I signed it!

f3z0:
Do we want to move to self-signed localhost certificates for the purpose of example client? The benefit is it will just work for new devs without having to setup a reverse proxy to the clemente domain or setting up a private domain with certificates that are ca signed. If so, then how should I proceed with the cert_chain_test.go integration testing that is failing due to removing the clemente pk? Thanks!

marten-seemann:
Hi @f3z0,

Thanks for your PR. The reason we're using a real certificate for quic.clemente.io is that at the moment, it's needed in our integration tests. We're using the quic_client and Chromium to do some up- and downloads to our QUIC server. The problem is that both of them don't send a SNI in their CHLO when connecting to localhost. However, we need the SNI to select the certificate. So we can't run our integration tests on localhost, but we have to map them to a real host name.
I thought about our integration test setup, and we might be able modify it such that we use a self-signed certificate, if we make sure that the CA cert is added to the Root Certs of our CIs. I created #475 to track this.

The issue in the h2quic client is a separate issue, and I verified that your fix works.
Can you please create a new PR that only fixes this issue (I created #476 to track that). Please also add a test.

f3z0:
I get it, I'll submit new PR as suggested. I may have a fix for https://github.com/lucas-clemente/quic-go/issues/343. If so, it was not RTOs/congestion but instead a racetime with aead state changes and a few missing mutex locks, possible?

marten-seemann:
Thanks @f3z0. Looking forward to your PR.

Regarding #343, I used to think this is a result of our incomplete implementation of the loss recovery algorithm (there are a bunch of issues related to this). I'd be very happy to fix any race condition, can you tell me where you found one (or send another PR)?

marten-seemann:
I just fixed the h2quic client issue in #480. Thanks again @f3z0 for reporting.

f3z0:
Sorry for not pushing updated pr, last week was under the gun. So I'm working with streaming high resolution MP4 which stresses the system due to high throughput and I was definitely facing the stalled dl which sounds a lot like 343 but it was due to dead locking on that specific quic connection (I could still create new connections). There were a few spots this was happening, and in all cases it was due to a property being mutated outside of its mutex lock/unlock, it now streams uninterrupted for extended periods so I think it resolves that issue. I will diff a patch today or tomorrow at the latest so you can review.

PR #481, opened 2017-03-15
remove ineffective check for too large packets

### LINKED ISSUE #468: detection of too large packets not working

marten-seemann:
closes #468


marten-seemann:
Appveyor failed again, because of another test. @lucas-clemente please restart.
I'll try to figure out what is flaky there.

PR #485, opened 2017-03-16
return an error when the network conn is closed in Listener.Serve()

### LINKED ISSUE #483: Should we return an error from Listener.Serve() when it is closed?
The following code for a TCP connection

```go
addr, _ := net.ResolveTCPAddr("tcp", "localhost:1234")
ln, _ := net.ListenTCP("tcp", addr)

go func() {
	time.Sleep(time.Second)
	ln.Close()
}()

_, err = ln.Accept()
fmt.Println(err)
```

prints `accept tcp 127.0.0.1:1234: use of closed network connection`.

The equivalent for a QUIC connection returns a `nil` from `Listener.Serve()` (i.e. after `Listener.Close()` is called), because we're replacing all errors containing the substring "use of closed network connection" with `nil`.

marten-seemann:
fixes #483

### DIFF
--- a/server.go
+++ b/server.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"net"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


marten-seemann:
-6% in *client.go*
@codecov-io is making jokes again...

PR #489, opened 2017-03-20
send a Public Reset when receiving a CHLO with the FHL2 tag

### LINKED ISSUE #411: Send public reset when receiving FHOL from client

marten-seemann:
Fixes #411.

Chrome sends the FHL2 when it wants to perform a head-of-line blocking
experiment, introduced in QUIC version 36 (see
https://codereview.chromium.org/2115033002). We don’t support this
experiment. By sending a Public Reset when receiving this tag we force
Chrome to use the TCP fallback.

### DIFF
--- a/handshake/crypto_setup_server.go
+++ b/handshake/crypto_setup_server.go
@@ -47,6 +47,11 @@ type cryptoSetupServer struct {
 
 var _ CryptoSetup = &cryptoSetupServer{}
 
+// ErrHOLExperiment is returned when the client sends the FHL2 tag in the CHLO
+// this is an expiremnt implemented by Chrome in QUIC 36, which we don't support
+// TODO: remove this when dropping support for QUIC 36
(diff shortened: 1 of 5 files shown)


marten-seemann:
I tested that
* Chrome actually uses the TCP fallback when receiving a Public Reset during the handshake.
* The Public Reset is actually only sent when the FHL2 tag is present (using our client, I didn't dig deep enough into the Chrome code to compile a quic_client that sends that tag).

handshake/crypto_setup_server.go
```diff
@@ -47,6 +47,11 @@ type cryptoSetupServer struct {
 
 var _ CryptoSetup = &cryptoSetupServer{}
 
+// ErrHOLExperiment is returned when the client sends the FHL2 tag in the CHLO
+// this is an expiremnt implemented by Chrome in QUIC 36, which we don't support
+// TODO: remove this when dropping support for QUIC 36
```
lucas-clemente: In the future: s/QUIC 36/Version36/ to make grepping easier

Issue #486, opened 2017-03-18
Example client returns 502 Bad Gateway


fl43:
Installed Go 1.8 on a Windows 7 machine.
Cloned version 9a1e86cf5e92f3439eb22f1ede8240f33955e2a9 of quic-go.

Tried to execute the test client against https://quic.clemente.io and get a 502 bad Gateway error.

Here the full log:
C:\dev\go\src\edge\quic-go>go run example/client/main.go https://quic.clemente.io
GET https://quic.clemente.io
Starting new connection to quic.clemente.io (188.68.52.107:443), connectionID 4d86327485f1ab8f, version 36
Got response for https://quic.clemente.io: &http.Response{Status:"502 Bad Gateway", StatusCode:502, Proto:"HTTP/2.0", ProtoMajor:2, ProtoMinor:0, Header:http.Header{"X-Content-Type-Options":[]string{"
nosniff"}, "Alternate-Protocol":[]string{"7000:quic"}, "Alt-Svc":[]string{"quic=\":7000\"; ma=2592000; v=\"36,35,34\""}, "Server":[]string{"Caddy"}, "Content-Type":[]string{"text/plain; charset=utf-8"
}}, Body:(*quic.stream)(0xc042063680), ContentLength:-1, TransferEncoding:[]string(nil), Close:false, Uncompressed:false, Trailer:http.Header(nil), Request:(*http.Request)(0xc042032100), TLS:(*tls.Con
nectionState)(nil)}
Request Body:
502 Bad Gateway

lucas-clemente:
Yep, that's expected, as the demo page is currently down. Fun fact though: that error has been loaded via QUIC (as you can see by the `Body:(*quic.stream)`). Try pointing it to clemente.io (or google.com, but I believe their quic is currently disabled).

fl43:
Sorry, for the issue. Closing it now.

PR #491, opened 2017-03-20
Change client example to point to clemente.io

lucas-clemente:
This caused confusion in #486.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Using Chrome:
 
 ### Using the example client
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
Why not fix quic.clemente.io instead?

lucas-clemente:
Because I don't have the time and the benchmarks there don't work properly anymore (since we don't support 0-RTT and loss recovery is still an issue).

PR #492, opened 2017-03-20
Fix some racetime conditions

f3z0:
Re-arranged mutex locks and unlocks to avoid some racetime conditions,
this potentially resolves issue #343, download stalling because
congestion limited

### DIFF
--- a/client.go
+++ b/client.go
@@ -150,11 +150,14 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
 		return qerr.Error(qerr.InvalidPacketHeader, err.Error())
 	}
 	hdr.Raw = packet[:len(packet)-r.Len()]
-
+	
+	c.mutex.Lock()
 	// ignore delayed / duplicated version negotiation packets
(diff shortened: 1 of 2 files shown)


lucas-clemente:
Thanks for looking into this! :) Since you've only touched the client this probably won't affect the integration test, which only tests the server part.

I'm wondering why the `[Un]LockForSealing` functions are empty for the client and whether the mutex should be put there, but I'll talk this over with Marten tomorrow and we'll get back to you.

lucas-clemente:
Also: the crypto_setup_server also locks in Open() which the client doesn't. That might be another spot to change.

marten-seemann:
@f3z0: Thanks for finding these race conditions!

I had a closer look now, and I found some more race conditions in the `CryptoSetupClient`. I'm now working on a fix that will kill (hopefully!) all of them, by rewriting the way packets are sealed (which involves several changes to the `CryptoSetup` interface).

marten-seemann:
All these race conditions were fixed in #493. Thanks again, @f3z0.

lucas-clemente:
Thanks! :)

PR #493, opened 2017-03-22
improve the CryptoSetup interface

marten-seemann:
Removes race conditions in the `cryptoSetupClient` and simplifies how packets are sealed.


client.go
```diff
@@ -222,6 +186,52 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
...
+		// check if the server sent the offered version in supported versions
+		if v == c.version {
+			return qerr.Error(qerr.InvalidVersionNegotiationPacket, "Server already supports client's version and should have accepted the connection.")
+		}
+		if v != protocol.VersionUnsupported {
+			hasCommonVersion = true
+		}
+	}
+	if !hasCommonVersion {
+		utils.Infof("No common version found.")
```
lucas-clemente: This log doesn't really offer much infos. Should we remove it?

handshake/crypto_setup_client.go
```diff
@@ -92,7 +92,10 @@ func (h *cryptoSetupClient) HandleCryptoStream() error {
 		}
 
 		// send CHLOs until the forward secure encryption is established
-		if h.forwardSecureAEAD == nil {
+		h.mutex.Lock()
```
lucas-clemente: This could be a RLock & RUnlock.

handshake/interface.go
```diff
@@ -0,0 +1,19 @@
+package handshake
+
+import "github.com/lucas-clemente/quic-go/protocol"
+
+// Sealer seals a packet
+type Sealer func(dst, src []byte, packetNumber protocol.PacketNumber, associatedData []byte) []byte
```
lucas-clemente: I think I would prefer to define a Sealer interface with a single Seal function. Then we can just return the implementations of the various Sealers directly. If we need a lambda somewhere (although I don't see any place right now), we can to something similar to how https://golang.org/pkg/net/http/#HandlerFunc works. What to you think?

lucas-clemente: Actually, we could even split up crypto.AEAD into a Sealer and a Opener and reuse the Sealer interface from there (or embed it in a handshake.Sealer).

packet_packer.go
```diff
@@ -78,10 +78,20 @@ func (p *packetPacker) PackPacket(stopWaitingFrame *frames.StopWaitingFrame, con
 func (p *packetPacker) packPacket(stopWaitingFrame *frames.StopWaitingFrame, leastUnacked protocol.PacketNumber, packetToRetransmit *ackhandler.Packet) (*packedPacket, error) {
 	// packetToRetransmit is only set for handshake retransmissions
 	isHandshakeRetransmission := (packetToRetransmit != nil)
```
lucas-clemente: I think I made this comment previously: This would be much clearer with s/packetToRetransmit/handshakePacketToRetransmit/g ;)

client.go
```diff
@@ -222,6 +186,52 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
...
 
+func (c *client) handlePacketWithVersionFlag(hdr *PublicHeader) error {
+	var hasCommonVersion bool // check if we're supporting any of the offered versions
+	for _, v := range hdr.SupportedVersions {
+		// check if the server sent the offered version in supported versions
+		if v == c.version {
+			return qerr.Error(qerr.InvalidVersionNegotiationPacket, "Server already supports client's version and should have accepted the connection.")
+		}
+		if v != protocol.VersionUnsupported {
+			hasCommonVersion = true
```
marten-seemann: True. But this doesn't really belong in this PR.

client.go
```diff
@@ -151,67 +151,31 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
 	}
 	hdr.Raw = packet[:len(packet)-r.Len()]
 
+	c.mutex.Lock()
```
marten-seemann: The defer would include the `session.handlePacket`.

lucas-clemente: Which should be fine, since that only pushed the packet into the chan.

marten-seemann: You're right. I'll change it.

client.go
marten-seemann: I'll make a separate commit for the version stuff after we merge this.

lucas-clemente: sgtm

PR #502, opened 2017-03-29
Implement loss recovery from the current WG draft

### LINKED ISSUE #498: Reimplement loss recovery from the WG draft

lucas-clemente:
Fixes #498 and will hopefully go a long way towards fixing the many
flaky tests.

### DIFF
--- a/ackhandler/sent_packet_handler.go
+++ b/ackhandler/sent_packet_handler.go
@@ -12,6 +12,18 @@ import (
 	"github.com/lucas-clemente/quic-go/utils"
 )
 
+const (
+	// Maximum reordering in time space before time based loss detection considers a packet lost.
+	// In fraction of an RTT.
+	timeReorderingFraction = 1.0 / 8
+	// defaultRTOTimeout is the RTO time on new connections
+	defaultRTOTimeout = 500 * time.Millisecond
+	// Minimum time in the future an RTO alarm may be set for.
+	minRTOTimeout = 200 * time.Millisecond
+	// maxRTOTimeout is the maximum RTO time
+	maxRTOTimeout = 60 * time.Second
+)
+
 var (
 	// ErrDuplicateOrOutOfOrderAck occurs when a duplicate or an out-of-order ACK is received
 	ErrDuplicateOrOutOfOrderAck = errors.New("SentPacketHandler: Duplicate or out-of-order ACK")
@@ -22,11 +34,10 @@ var (
 	errAckForUnsentPacket  = qerr.Error(qerr.InvalidAckData, "Received ACK for an unsent package")
 )
 
-var errPacketNumberNotIncreasing = errors.New("Already sent a packet with a higher packet number.")
+var errPacketNumberNotIncreasing = errors.New("Already sent a packet with a higher packet number")
 
 type sentPacketHandler struct {
 	lastSentPacketNumber protocol.PacketNumber
-	lastSentPacketTime   time.Time
 	skippedPackets       []protocol.PacketNumber
 
 	LargestAcked protocol.PacketNumber
@@ -40,10 +51,17 @@ type sentPacketHandler struct {
 
 	bytesInFlight protocol.ByteCount
 
-	rttStats   *congestion.RTTStats
 	congestion congestion.SendAlgorithm
+	rttStats   *congestion.RTTStats
 
-	consecutiveRTOCount uint32
+	// The number of times an RTO has been sent without receiving an ack.
+	rtoCount uint32
+
+	// The time at which the next packet will be considered lost based on early transmit or exceeding the reordering window in time.
+	lossTime time.Time
+
+	// The alarm timeout
+	alarm time.Time
 }
 
 // NewSentPacketHandler creates a new sentPacketHandler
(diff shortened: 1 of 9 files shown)


marten-seemann:
I think we need to check if a packet is retransmittable in `sentPacketHandler.SentPacket()` (it probably makes sense to define this as a method on the `Packet`.

lucas-clemente:
Agreed, but I'd rather do this as a separate PR later. Filed #505.

ackhandler/sent_packet_handler.go
```diff
@@ -12,6 +12,18 @@ import (
 	"github.com/lucas-clemente/quic-go/utils"
 )
 
+const (
+	// Maximum reordering in time sapce before time based loss detection considers a packet lost.
```
marten-seemann: Nit: typo.

ackhandler/sent_packet_handler.go
```diff
@@ -12,6 +12,18 @@ import (
 	"github.com/lucas-clemente/quic-go/utils"
 )
 
+const (
```
marten-seemann: Maybe move all the Loss Recovery constants to a separate file.

ackhandler/sent_packet_handler.go
```diff
@@ -211,59 +204,124 @@ func (h *sentPacketHandler) ReceivedAck(ackFrame *frames.AckFrame, withPacketNum
...
+		// Packets are sorted by number, so we can stop searching
+		if packet.PacketNumber > largestAcked {
+			break
+		}
 	}
+	return false
+}
 
-	h.garbageCollectSkippedPackets()
+func (h *sentPacketHandler) updateLossDetectionAlarm() {
```
marten-seemann: Why did you rename `setLossDetectionAlarm()` from the WG draft?

ackhandler/sent_packet_handler.go
```diff
@@ -211,59 +204,124 @@ func (h *sentPacketHandler) ReceivedAck(ackFrame *frames.AckFrame, withPacketNum
...
+	// Cancel the alarm if no packets are outstanding
+	if h.packetHistory.Front() == nil {
```
marten-seemann: Better: 
```go
if h.packetHistory.Len() == 0
```

ackhandler/sent_packet_handler.go
```diff
@@ -211,59 +204,124 @@ func (h *sentPacketHandler) ReceivedAck(ackFrame *frames.AckFrame, withPacketNum
...
+	}
+}
 
-	h.congestion.OnCongestionEvent(
-		rttUpdated,
-		h.BytesInFlight(),
-		ackedPackets,
-		lostPackets,
-	)
+// TODO(lclemente): Introducing congestion.MaybeExitSlowStart() would allow us to call through for each packet and eliminate both the rttUpdated param and the packet slices passed to the congestion
```
marten-seemann: Please use your Github username for TODOs.

ackhandler/sent_packet_handler.go
```diff
@@ -211,59 +204,124 @@ func (h *sentPacketHandler) ReceivedAck(ackFrame *frames.AckFrame, withPacketNum
...
+
+		if packet.PacketNumber > h.LargestAcked {
+			break
+		}
+
+		timeSinceSent := now.Sub(packet.SendTime)
+		if timeSinceSent > delayUntilLost {
+			lostPackets = append(lostPackets, el)
+		} else if h.lossTime.IsZero() {
+			// Note: This conditional is only entered once per call
```
marten-seemann: This note is not obvious to me. Can you add a reason?

ackhandler/sent_packet_handler.go
```diff
@@ -149,54 +138,58 @@ func (h *sentPacketHandler) ReceivedAck(ackFrame *frames.AckFrame, withPacketNum
...
+		return ErrAckForSkippedPacket
+	}
 
-	// check if it acks any packets that were skipped
-	for _, p := range h.skippedPackets {
-		if ackFrame.AcksPacket(p) {
-			return ErrAckForSkippedPacket
+	rttUpdated := h.maybeUpdateRTT(ackFrame.LargestAcked, ackFrame.DelayTime, rcvTime)
+
+	ackedPackets, err := h.determineNewlyAckedPackets(ackFrame)
```
marten-seemann: I think there's a better way than creating a slice just to loop over it later.

marten-seemann:
Does this resolve the flakiness of our integration tests? If so, I'd be very happy if we could remove the `travis_retry` hack.

ackhandler/sent_packet_handler.go
lucas-clemente: Done

ackhandler/sent_packet_handler.go
lucas-clemente: I don't see a benefit, and they are not too many here. For constants that are only used in one place, I'd rather have them as local ad as close to their use as possible.

ackhandler/sent_packet_handler.go
lucas-clemente: Because we don't set the alarm here, only update one particular alarm (h.alarm) that is then set in the session.

ackhandler/sent_packet_handler.go
lucas-clemente: Done

ackhandler/sent_packet_handler.go
lucas-clemente: 🙄

ackhandler/sent_packet_handler.go
marten-seemann: Actually, this is pretty obvious ;)

ackhandler/sent_packet_handler.go
lucas-clemente: True, this is slightly suboptimal. But I really like this structure, as it makes determineNewlyAckedPackets pure and therefore much easier to test (and isolates concerns between the functions better). I also want to try applying the same pattern to detectLostPackets, but it's a bit trickier to make that pure as well.

ackhandler/sent_packet_handler.go
lucas-clemente: Leaving it as is after OOB discussion.

PR #503, opened 2017-03-29
Fix flaky appveyor tests

### LINKED ISSUE #484: flaky session test on AppVeyor
see here: https://ci.appveyor.com/project/lucas-clemente/quic-go/build/1261/job/lcr679woeucgefuo

lucas-clemente:
We were relying on the number of goroutines in those tests. This commit
replaces that logic with grepping through the current callstack.

Fixes #484

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -6,8 +6,6 @@ import (
 	"errors"
 	"net"
 	"reflect"
-	"runtime"
-	"time"
 	"unsafe"
 
 	"github.com/lucas-clemente/quic-go/protocol"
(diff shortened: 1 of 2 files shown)


marten-seemann:
Where do the additional goroutines come from anyway?

lucas-clemente:
I don't know, but it's really not something we should rely on.

The problem here is that the client tests are very leaky when it comes to sessions, so I'm gonna have to change a lot of them.

lucas-clemente:
I think more assertions in tests are almost always better (especially if they're easy wins, such as assertions in AfterEach). In this particular case, it also causes us to look for rough edges around our shutdown APIs.

I agree that different tests shouldn't affect each other, but that's really down to how testing in Go works (i.e. single process for all tests in a suite and no forking).

PR #508, opened 2017-04-05
Simplify ackhandler and congestion by splitting up OnCongestionEvent

lucas-clemente:
No functional change.


lucas-clemente:
Unfortunately this breaks the go1.7 build. I will remove the problematic changes and merge then.

PR #509, opened 2017-04-05
Read the initial log level from the QUIC_GO_LOG_LEVEL env if present

### LINKED ISSUE #501: Proposal: Read an env to set the logging level in init()
This would allow debugging logs to be enabled without recompiling, e.g. for existing caddy binaries, and means we don't have to define a stable logging API.

Also, this would make it easier to enable logging by default on the CIs, which would make flaky tests easier to debug.

lucas-clemente:
We deliberately ignore any parsing errors.

Fixes #501.

### DIFF
--- a/utils/log.go
+++ b/utils/log.go
@@ -4,15 +4,16 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"sync"
 )
-
-var out io.Writer = os.Stdout
 
 // LogLevel of quic-go
 type LogLevel uint8
 
 const (
+	logEnv = "QUIC_GO_LOG_LEVEL"
+
 	// LogLevelDebug enables debug logs (e.g. packet contents)
 	LogLevelDebug LogLevel = iota
 	// LogLevelInfo enables info logs (e.g. packets)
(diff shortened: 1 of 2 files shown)


utils/log.go
```diff
@@ -4,15 +4,16 @@ import (
...
 	"sync"
 )
 
-var out io.Writer = os.Stdout
-
 // LogLevel of quic-go
 type LogLevel uint8
 
 const (
+	logEnv = "QUIC_GO_LOG_LEVEL"
```
marten-seemann: Or `QUIC_LOG_LEVEL`. Which one do you prefer?

lucas-clemente: I think we should make colissions as unlikely as possible, so I'd prefer our full name.

marten-seemann: Makes sense.

utils/log.go
```diff
@@ -68,3 +72,15 @@ func Errorf(format string, args ...interface{}) {
 func Debug() bool {
 	return logLevel == LogLevelDebug
 }
+
+func init() {
```
marten-seemann: Is it possible to write a test for that? Or is `init` executed before we can set an env variable in the tests?

lucas-clemente: I can factor the implementation into a separate function and test that individually, but I don't see a clean way to test init things.

lucas-clemente: Done

marten-seemann:
We should inform mholt, so that he can update his QUIC wiki page.

lucas-clemente:
I'd wait until we know this PR is in the stable caddy release.

PR #513, opened 2017-04-08
rewrite the benchmark test

### LINKED ISSUE #325: Fix timeout in benchmark test on Windows

### LINKED ISSUE #512: rewrite benchmark test
Now that we have a client implementation, there's no need to have two QUIC servers talk to each other (as in our current benchmark test).

marten-seemann:
The completely rewritten benchmark tests sets up a `quic.Listener` and downloads a file from it. This saves us the rather involved setup of the old benchmark test.
It uses a normal UDP connection on localhost, which hopefully saves us some of the artifacts we saw when doing direct in-memory transfers. 
However, I suspect this is the reason why the new benchmark test achieves data rates that are about 30% slower than those of the old test (I didn't do any measurements if this is the real reason).

Closes #512.
Also closes #325. I'm not sure what the problem was previously, but the new benchmark seems to work fine on AppVeyor.

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -2,172 +2,76 @@ package quic
 
 import (
 	"bytes"
-	"crypto/rand"
-	"encoding/binary"
+	"crypto/tls"
 	"fmt"
 	"io"
-	mrand "math/rand"
-	"net"
-	"reflect"
-	"runtime"
-	"runtime/debug"
+	"math/rand"
 	"time"
-	"unsafe"
 
+	"github.com/lucas-clemente/quic-go/protocol"
+	"github.com/lucas-clemente/quic-go/testdata"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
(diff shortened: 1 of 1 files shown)


benchmark_test.go
```diff
@@ -2,172 +2,76 @@ package quic
...
-					Skip("benchmark tests disabled on windows, see #325")
-				}
-
-				connID := protocol.ConnectionID(mrand.Uint32())
-
-				c1 := newLinkedConnection(nil)
-				session1I, err := newSession(c1, version, connID, nil, func(id protocol.ConnectionID) {}, func(Session, bool) {})
-				if err != nil {
-					Expect(err).NotTo(HaveOccurred())
+			Measure("better benchmarks", func(b Benchmarker) {
```
lucas-clemente: replace "better benchmarks" with something more meaningful.

marten-seemann: Oh, I only put that in during dev. Meant to remove it later, but forgot...

lucas-clemente:
LGTM module comments

PR #537, opened 2017-04-14
Replace docker with a directly launched Chrome for integration tests

### LINKED ISSUE #417: Simplify integration test setup
I think with https://github.com/knq/chromedp and https://chromium.googlesource.com/chromium/src/+/master/headless/README.md we will be able to remove docker and selenium from our integration tests, simplifying things a lot.

lucas-clemente:
The benefits:

- Speed up chrome tests (from 160s to 34s on my machine).
- Allow us to remove docker, which caused quite a few headaches.
- Gets us a more current Chrome, so now we run tests against all
  supported versions.
- We don't touch the FS anymore.
- We no longer depend on sudo for travis, which means we can use
  their new container-based infrastructure. This also speeds up our
  tests :)

Tests are now implemented in javascript; the server detects when
they are complete and kills Chrome.

Fixes #417.

### DIFF
--- a/integrationtests/chrome_test.go
+++ b/integrationtests/chrome_test.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/lucas-clemente/quic-go/protocol"
@@ -14,6 +15,10 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
+)
+
+const (
+	nChromeRetries = 8
 )
 
 func getChromePath() string {
@@ -24,10 +29,20 @@ func getChromePath() string {
 }
 
 func chromeTest(version protocol.VersionNumber, url string, blockUntilDone func()) {
+	// Chrome sometimes starts but doesn't send any HTTP requests for no apparent reason.
+	// Retry starting it a couple of times.
+	for i := 0; i < nChromeRetries; i++ {
+		if chromeTestImpl(version, url, blockUntilDone) {
+			return
+		}
+	}
+	Fail("Chrome didn't hit the testing endpoints")
+}
+
+func chromeTestImpl(version protocol.VersionNumber, url string, blockUntilDone func()) bool {
 	userDataDir, err := ioutil.TempDir("", "quic-go-test-chrome-dir")
 	Expect(err).NotTo(HaveOccurred())
 	defer os.RemoveAll(userDataDir)
-
 	path := getChromePath()
 	args := []string{
 		"--disable-gpu",
@@ -46,7 +61,19 @@ func chromeTest(version protocol.VersionNumber, url string, blockUntilDone func(
 	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
 	Expect(err).NotTo(HaveOccurred())
 	defer session.Kill()
+	const pollInterval = 100 * time.Millisecond
+	const pollDuration = 10 * time.Second
+	for i := 0; i < int(pollDuration/pollInterval); i++ {
+		time.Sleep(pollInterval)
+		if testEndpointCalled {
+			break
+		}
+	}
+	if !testEndpointCalled {
+		return false
+	}
 	blockUntilDone()
+	return true
 }
 
 var _ = Describe("Chrome tests", func() {
--- a/integrationtests/chrome_test.go
+++ b/integrationtests/chrome_test.go
@@ -2,234 +2,115 @@ package integrationtests
 
 import (
 	"fmt"
-	"io"
 	"io/ioutil"
-	"net/http"
 	"os"
-	"path/filepath"
-	"strconv"
+	"os/exec"
+	"runtime"
+	"strings"
 
-	"github.com/tebeka/selenium"
-
+	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/lucas-clemente/quic-go/protocol"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
 )
 
-const nImgs = 200
-const imgSize = 40
+func getChromePath() string {
+	if runtime.GOOS == "darwin" {
+		return "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+	}
+	return "google-chrome"
+}
 
-func init() {
-	http.HandleFunc("/tile", func(w http.ResponseWriter, r *http.Request) {
-		// Small 40x40 png
-		w.Write([]byte{
-			0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 0x00, 0x00, 0x00, 0x0d,
-			0x49, 0x48, 0x44, 0x52, 0x00, 0x00, 0x00, 0x28, 0x00, 0x00, 0x00, 0x28,
-			0x01, 0x03, 0x00, 0x00, 0x00, 0xb6, 0x30, 0x2a, 0x2e, 0x00, 0x00, 0x00,
-			0x03, 0x50, 0x4c, 0x54, 0x45, 0x5a, 0xc3, 0x5a, 0xad, 0x38, 0xaa, 0xdb,
-			0x00, 0x00, 0x00, 0x0b, 0x49, 0x44, 0x41, 0x54, 0x78, 0x01, 0x63, 0x18,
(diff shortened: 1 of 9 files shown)


marten-seemann:
I haven't had the time to look at the code yet. I already noticed two things:
1. Why are the Travis builds failing? Seems to be a problem with the integration tests.
2. There shouldn't be any increase in test coverage, since Travis is configured to only measure unit test coverage.

lucas-clemente:
One of the travis failures is the known flakyness, the other one looks like a new timeout (I'm guessing it's Chrome "do you really want to shutdown" dialog). I'll look into that.

Re coverage: There should be no change. The changes to the travis.yml file definitely shouldn't affect coverage, and the rest is contained to integrationtests. Either we have a pre-existing issue with how we take the samples, or it's an issue with codecov. Would you mind looking into that?

marten-seemann:
I can't see anything wrong with the Codecov config. I'll have a closer look once you pushed the fix for the flakiness. Maybe this is just one of the weird things Codecov does sometimes, and we don't have to fix anything.

lucas-clemente:
Tests should be fixed now. Codecov also looked good on the last run.

lucas-clemente:
Still not fixed, I'll take another look 😕

lucas-clemente:
@marten-seemann PTAL. The first two commits are somewhat unrelated to the docker change, I can pull them out into separate PRs if you prefer that.

integrationtests/chrome_test.go
```diff
@@ -2,234 +2,144 @@ package integrationtests
...
+		"--origin-to-force-quic-on=quic.clemente.io:443",
+		fmt.Sprintf(`--host-resolver-rules=MAP quic.clemente.io:443 localhost:%s`, port),
+		fmt.Sprintf("--quic-version=QUIC_VERSION_%d", version),
+		url,
+	}
+	utils.Infof("Running chrome: %s '%s'", getChromePath(), strings.Join(args, "' '"))
+	command := exec.Command(path, args...)
+	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+	Expect(err).NotTo(HaveOccurred())
+	defer func() {
```
marten-seemann: `defer session.Kill()`

integrationtests/chrome_test.go
```diff
@@ -2,234 +2,144 @@ package integrationtests
...
+		session.Kill()
+	}()
+	const pollInterval = 100 * time.Millisecond
+	const pollDuration = 10 * time.Second
+	for i := 0; i < int(pollDuration/pollInterval); i++ {
```
marten-seemann: Why don't you use `Eventually` for this?

integrationtests/integrationtests_suite_test.go
```diff
@@ -210,137 +192,90 @@ func stopQuicServer() {
...
-		"lclemente/standalone-chrome:dev",
-	)
-	docker, err = gexec.Start(dockerCmd, GinkgoWriter, GinkgoWriter)
-	Expect(err).NotTo(HaveOccurred())
-	Eventually(docker.Out, 10).Should(gbytes.Say("Selenium Server is up and running"))
-}
-
-func stopSelenium() {
-	docker.Interrupt().Wait(10)
+func waitForDone() {
```
marten-seemann: Maybe move this to chrome.go.

integrationtests/integrationtests_suite_test.go
```diff
@@ -210,137 +192,90 @@ func stopQuicServer() {
...
-		"chromeOptions": map[string]interface{}{
-			"args": []string{
-				"--enable-quic",
-				"--no-proxy-server",
-				"--origin-to-force-quic-on=quic.clemente.io:443",
-				fmt.Sprintf(`--host-resolver-rules=MAP quic.clemente.io:443 %s:%s`, GetLocalIP(), port),
-				fmt.Sprintf(`--quic-version=QUIC_VERSION_%d`, version),
-			},
-		},
+func waitForNUploaded(expected int) func() {
```
marten-seemann: This as well.

integrationtests/integrationtests_suite_test.go
```diff
@@ -210,137 +192,90 @@ func stopQuicServer() {
...
-	Expect(err).NotTo(HaveOccurred())
-	Eventually(session, 5).Should(gexec.Exit())
-	if session.ExitCode() != 0 {
-		return nil
+</script>
+</body>
+</html>
+`
+
+func generatePRData(l int) []byte {
```
marten-seemann: Reference the Wikipedia article here and mention that we're generating the same sequence as the JavaScript.

integrationtests/integrationtests_suite_test.go
```diff
@@ -210,137 +192,90 @@ func stopQuicServer() {
...
-				"--enable-quic",
-				"--no-proxy-server",
-				"--origin-to-force-quic-on=quic.clemente.io:443",
-				fmt.Sprintf(`--host-resolver-rules=MAP quic.clemente.io:443 %s:%s`, GetLocalIP(), port),
-				fmt.Sprintf(`--quic-version=QUIC_VERSION_%d`, version),
-			},
-		},
+func waitForNUploaded(expected int) func() {
+	return func() {
+		Eventually(func() int32 { return nFilesUploaded }, 60).Should(BeEquivalentTo(expected))
```
marten-seemann: Since you're incrementing `nFilesUploaded` atomically, shouldn't you read it atomically as well?

integrationtests/integrationtests_suite_test.go
```diff
@@ -210,137 +192,90 @@ func stopQuicServer() {
...
+<html>
+<body>
+<script>
+	console.log("Running DL test...");
+
+  ` + commonJS + `
+	for (var i = 0; i < NUM; i++) {
+		var req = new XMLHttpRequest();
+		req.open("POST", "/uploadhandler?len=" + LENGTH, true);
+		req.send(buf);
```
marten-seemann: We're sending the same data on every stream. Wouldn't it make sense to send different files (maybe by passing the seed as another GET parameter)?

integrationtests/integrationtests_suite_test.go
```diff
@@ -210,137 +192,90 @@ func stopQuicServer() {
...
-	session, err := gexec.Start(cmd, &out, GinkgoWriter)
-	Expect(err).NotTo(HaveOccurred())
-	Eventually(session, 5).Should(gexec.Exit())
-	if session.ExitCode() != 0 {
-		return 0
+	var nOK = 0;
+	for (var i = 0; i < NUM; i++) {
+		let req = new XMLHttpRequest();
+		req.responseType = "arraybuffer";
+		req.open("POST", "/prdata?len=" + LENGTH, true);
```
marten-seemann: Same here. Maybe use different seeds.

integrationtests/chrome_test.go
lucas-clemente: Because we don't want the test to fail if it's not fulfilled, we want to retry. I don't think there's a way of doing that with Eventually?

integrationtests/integrationtests_suite_test.go
lucas-clemente: I'm gonna refactor this soon so that we can have multiple files of chrome tests in a separate package. I don't see a point in moving it right now.

integrationtests/chrome_test.go
lucas-clemente: Done. There used to be a Wait() in there, but that didn't work (for some reason).

integrationtests/integrationtests_suite_test.go
lucas-clemente: See above.

integrationtests/chrome_test.go
marten-seemann: Makes sense.

integrationtests/integrationtests_suite_test.go
lucas-clemente: Done. Strictly speaking all reads of a word are atomic, so the Load* functions are really NOPs (see https://golang.org/src/sync/atomic/asm_amd64.s#L94). Even if they weren't, it wouldn't matter here as we call this many times.

integrationtests/integrationtests_suite_test.go
lucas-clemente: I thought about this, but frankly I'm not sure it's worth the effort. I think the class of bugs that result in exchange of stream data at the same offset between streams should be really small, and I guess we'd have many other things fail.

I can add a seed if you insist, but I'd rather not over-complicate the test setup.

integrationtests/integrationtests_suite_test.go
lucas-clemente: Done.

integrationtests/integrationtests_suite_test.go
marten-seemann: Ok, let's merge this for now. We can still add this later if we feel the need to.

PR #548, opened 2017-04-18
add integration tests with the quic_server

### LINKED ISSUE #545: add integration tests using the quic_server

marten-seemann:
Closes #545.

There are certainly more tests we can (and should) add, but here are 3 very basic tests: downloading a "Hello World", downloading a small file, and downloading a larger file, all on a lossless local connection.

I think we'll need to restructure the integration tests soon (maybe after merging the new Chrome tests?). I find the current naming scheme "client tests", "server tests" etc. not very helpful, now that we have both a client and server implementation and we're testing with quic_client and quic_server.

### DIFF
--- a/integrationtests/server_test.go
+++ b/integrationtests/server_test.go
@@ -1,0 +1,213 @@
+package integrationtests
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/pem"
+	"io/ioutil"
+	"math/big"
+	mrand "math/rand"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"github.com/lucas-clemente/quic-go/h2quic"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("Server tests", func() {
+	mrand.Seed(time.Now().UnixNano())
+
+	var (
+		serverPort string
+		tmpDir     string
+		session    *Session
+		client     *http.Client
+	)
+
+	generateCA := func() (*rsa.PrivateKey, *x509.Certificate) {
+		key, err := rsa.GenerateKey(rand.Reader, 1024)
+		Expect(err).ToNot(HaveOccurred())
+
+		t := time.Now().Add(-time.Minute)
+		templateRoot := &x509.Certificate{
+			SerialNumber: big.NewInt(1),
+			NotBefore:    t,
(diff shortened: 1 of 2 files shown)


integrationtests/server_test.go
```diff
@@ -0,0 +1,201 @@
...
+		err := os.RemoveAll(tmpDir)
+		Expect(err).ToNot(HaveOccurred())
+		tmpDir = ""
+	})
+
+	It("downloads a hello", func(done Done) {
+		data := []byte("Hello world!\n")
+		createDownloadFile("hello", data)
+
+		startServer()
```
lucas-clemente: Should we move this to Before / AfterEach? We could then get rid of the startServer / stopServer functions.

integrationtests/server_test.go
```diff
@@ -0,0 +1,201 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("Server tests", func() {
+	mrand.Seed(time.Now().UnixNano())
+
+	var (
+		port    string
```
lucas-clemente: This variable is shadowing the global `port` we currently have in the tests. That won't be an issue if we restructure it out later, but for now I'd like to avoid that. Rename to `serverPort`?

integrationtests/server_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		}()
+	}
+
+	stopServer := func() {
+		session.Kill()
+	}
+
+	BeforeEach(func() {
+		port = strconv.Itoa(20000 + int(mrand.Int31n(10000)))
```
lucas-clemente: Is there a way of setting port=0 and extracting it from the output of the server?

integrationtests/pkcs8.go
```diff
@@ -0,0 +1,75 @@
...
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"errors"
+)
+
+// copied from https://github.com/AGWA-forks/golang-crypto/blob/master/pkcs12/pkcs8.go
```
lucas-clemente: No copied code.

integrationtests/server_test.go
marten-seemann: Unfortunately, this `quic_server` doesn't support that. If I omit the `--port`, it listens on 6121 by default.

integrationtests/server_test.go
marten-seemann: That's not possible. See the comment for `startServer`.

lucas-clemente: Ok, let's revisit this for the refactoring.

lucas-clemente:
LGTM, but please review the changes I just pushed.

PR #556, opened 2017-04-20
Fixing deadlock with runloop in session.go

Eichhoernchen:
If a client is in `handleSHLOMessage` in parallel to running the runloop (run) in session.go (this is usually the case) and being past the select statement there, other calls in the runloop can wait for the lock (e.g. idleTimeout()), while `h.aeadChanged <- ...` blocks until the select statement in the runloop is called, thus causing a potential deadlock.

This fix releases the lock to make progress. The channel itself acts like a lock in this situation causing the deadlock.

### DIFF
--- a/handshake/crypto_setup_client.go
+++ b/handshake/crypto_setup_client.go
@@ -247,9 +247,11 @@ func (h *cryptoSetupClient) handleSHLOMessage(cryptoData map[Tag][]byte) error {
 	if err != nil {
 		return qerr.InvalidCryptoMessageParameter
 	}
-
+	
+	h.mutex.Unlock()
 	h.aeadChanged <- protocol.EncryptionForwardSecure
-
+	h.mutex.Lock()
+	
 	return nil
 }
 


Eichhoernchen:
|I signed it!|

lucas-clemente:
Thanks a lot for the PR! 👍 I'd like @marten-seemann to properly review this (as he wrote the client), but he's on vacation until EOW.

Quick question: Did you actually see the deadlock in action? I'm wondering because the aeadChanged channel is made as a buffered channel with capacity 2 ([here](https://github.com/lucas-clemente/quic-go/blob/master/session.go#L184)), so I wouldn't expect a deadlock here. That being said, it might be that we have an issue with how the cryptoSetupClient works (maybe it tries to upgrade the crypto multiple times?), so we should look into that as well.

Eichhoernchen:
I saw the deadlock in action and it took me some time to figure out the problem, however, I did not have the head of the current master branch during testing and it seems that my local version does not have the change from two month ago that increases the channel capacity to two.

Given this change (6cb19e42a5f0ac1539ba5d6aa8b6cb1183a6f8b4), I'm unsure if the deadlock can still appear or if both processes always alternate in their synchronization. If the latter is the case, the deadlock cannot appear given the increased capacity, otherwise it may still appear.

So I guess it might be good to get a review of someone who has a better understanding of the concurrency that might occur.

marten-seemann:
Hello @Eichhoernchen, thanks for the PR.
If I see that correctly, the commit you mention (6cb19e4) prevents the deadlock. The `cryptoSetupClient` should make sure that the AEAD is changed exactly twice during a successful connection establishment, so the channel should never block.
That being said, deadlocks are notoriously hard to find, so if you ever see any problem here again, please let us know!

PR #559, opened 2017-04-22
use random port numbers in integration tests

marten-seemann:
This helps prevent “address already in use” errors that occur from time to time when running the tests with the untilItFails option.

PR #560, opened 2017-04-22
rewrite the QUIC proxy used in the integrationtests

marten-seemann:
The new QUIC proxy allows to listen on “:0”, which allows us to get rid of all “address already in use” errors. The constructor now takes an `Opts` struct, which makes configuring it easier.

Furthermore, I expect a reduction of reordering with this new proxy. I'm now using `time.AfterFunc` instead of spawning a new goroutine for every packet (for RTT simulations), which seems to perform better (although still not perfectly).

This PR superseeds #559.


integrationtests/drop_test.go
```diff
@@ -47,7 +47,8 @@ var _ = Describe("Drop Proxy", func() {
 	}
 
 	AfterEach(func() {
-		dropproxy.Stop()
+		err := proxy.Close()
+		Expect(err).NotTo(HaveOccurred())
 		time.Sleep(time.Millisecond)
```
lucas-clemente: We can probably delete this sleep now.

integrationtests/drop_test.go
```diff
@@ -64,11 +65,17 @@ var _ = Describe("Drop Proxy", func() {
 				}
 
 				It("gets a file when many outgoing packets are dropped", func() {
-					runDropTest(nil, dropper, version)
+					dropCallback := func(d quicproxy.Direction, p protocol.PacketNumber) bool {
+						return d == quicproxy.DirectionOutgoing && dropper(p)
```
lucas-clemente: I think for clarity I'd prefer to inline dropper here and below.

PR #571, opened 2017-04-27
add support for QUIC 37

marten-seemann:


integrationtests/chrome_test.go
```diff
@@ -45,7 +45,7 @@ func init() {
 var _ = Describe("Chrome tests", func() {
 	It("does not work with mismatching versions", func() {
 		versionForUs := protocol.SupportedVersions[0]
-		versionForChrome := protocol.SupportedVersions[len(protocol.SupportedVersions)-1]
+		versionForChrome := protocol.SupportedVersions[1]
 
 		// If both are equal, this test doesn't make any sense.
 		if versionForChrome == versionForUs {
```
lucas-clemente: This check can be removed with your above change.

marten-seemann: I was hoping that the above change is only temporary and we can remove it either by updating the Docker image or merging your Chrome-branch.

lucas-clemente: The change doesn't worsen the ability of this test to detect an improper testing setup, so I don't see why it should be temporary.

marten-seemann: Got it. I'll remove it.

PR #572, opened 2017-04-27
remove the closeCallback from the session

marten-seemann:
This is what I mentioned in https://github.com/lucas-clemente/quic-go/issues/564#issuecomment-296434663.

The `closeCallback` was run when a session was closed, i.e. after the run loop of the session stopped. Instead of explicitely calling this callback from the session, the caller of `session.run()` can just execute the code after `session.run()` returns.

### DIFF
--- a/server.go
+++ b/server.go
@@ -18,7 +18,7 @@ import (
 type packetHandler interface {
 	Session
 	handlePacket(*receivedPacket)
-	run()
+	run() error
 }
 
 // A Listener of QUIC
@@ -34,7 +34,7 @@ type server struct {
 	sessionsMutex             sync.RWMutex
 	deleteClosedSessionsAfter time.Duration
 
-	newSession func(conn connection, v protocol.VersionNumber, connectionID protocol.ConnectionID, sCfg *handshake.ServerConfig, closeCallback closeCallback, cryptoChangeCallback cryptoChangeCallback) (packetHandler, error)
+	newSession func(conn connection, v protocol.VersionNumber, connectionID protocol.ConnectionID, sCfg *handshake.ServerConfig, cryptoChangeCallback cryptoChangeCallback) (packetHandler, error)
 }
 
 var _ Listener = &server{}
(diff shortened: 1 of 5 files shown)


server.go
```diff
@@ -182,16 +182,22 @@ func (s *server) handlePacket(pconn net.PacketConn, remoteAddr net.Addr, packet
...
 		s.sessionsMutex.Lock()
 		s.sessions[hdr.ConnectionID] = session
 		s.sessionsMutex.Unlock()
+
+		go func(id protocol.ConnectionID) {
+			// session.run() returns as soon as the session is closed
+			_ = session.run()
+
+			s.removeConnection(id)
+		}(hdr.ConnectionID)
```
lucas-clemente: hdr is not in a loop, so it should be ok to just capture hdr.ConnectionID, right?

session.go
```diff
@@ -73,6 +69,8 @@ type session struct {
 	// closeChan is used to notify the run loop that it should terminate.
 	// If the value is not nil, the error is sent as a CONNECTION_CLOSE.
 	closeChan chan *qerr.QuicError
+	// the error this session was closed with
+	closeErr  error
```
lucas-clemente: What is the difference between closeErr and the error sent on the closeChan? Could we simply return the latter one from run()?

marten-seemann: For the error for closing a session in order to replace it with a new one (`errCloseSessionForNewVersion`), we're sending a `nil` into the channel. Same for the cases where we send a Public Reset instead of a ConnectionClose.

server.go
marten-seemann: Yeah, that should work.

session.go
lucas-clemente: I know, but I don't understand why that means we require a second way of delivering errors.

marten-seemann: Because we need to return an error from `session.run`.

lucas-clemente: Sure. Right now we have two break statements in the run loop, and at both points we just read the error from the chan. Can we not simply change the breaks to returns of the error? Then we wouldn't need this additional closeErr you're adding here.

marten-seemann: This won't work, because as I said before, we're not passing the error into the channel.

lucas-clemente: I see. Then I'd rather change the chan to a chan struct{} and always use the closeErr (and document it there). I really dislike having multiple spots for errors to propagate.

marten-seemann:
@lucas-clemente: I changed the error handling in the session. Please have another look.

session.go
```diff
@@ -503,12 +495,11 @@ func (s *session) closeImpl(e error, remoteClose bool) error {
 	if !atomic.CompareAndSwapUint32(&s.closed, 0, 1) {
 		return errSessionAlreadyClosed
 	}
-	s.closeErr = e
 
 	if e == errCloseSessionForNewVersion {
 		s.streamsMap.CloseWithError(e)
 		s.closeStreamsWithError(e)
-		s.closeChan <- nil
+		s.closeChan <- e
```
lucas-clemente: All paths leaving this function send `e` on the closeChan. I think it would be nicer to `defer s.closeChan <- e` at the top of the function. We then have to move the `e==nil` case to the top, but that shouldn't change anything.

marten-seemann: Done.

PR #577, opened 2017-04-29
add a quic.Config option for QUIC versions

### LINKED ISSUE #565: Make QUIC versions configurable
There should be a `quic.Config` options for QUIC versions, of type `[]protocol.VersionNumber`.
It determines which QUIC versions a server / client is willing to use. The default value is `protocol.SupportedVersions`.
For the client, the order of this slice determines the preference: The first version in the slice would be tried on the first dial attempt.

marten-seemann:
Fixes #565.

### DIFF
--- a/client.go
+++ b/client.go
@@ -205,20 +205,20 @@ func (c *client) handlePacketWithVersionFlag(hdr *PublicHeader) error {
 		}
 	}
 
-	ok, highestSupportedVersion := protocol.HighestSupportedVersion(c.config.Versions, hdr.SupportedVersions)
-	if !ok {
+	newVersion := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
+	if newVersion == protocol.VersionUnsupported {
 		return qerr.InvalidVersion
 	}
 
 	// switch to negotiated version
-	c.version = highestSupportedVersion
+	c.version = newVersion
 	c.connState = ConnStateVersionNegotiated
 	var err error
 	c.connectionID, err = utils.GenerateConnectionID()
 	if err != nil {
 		return err
 	}
-	utils.Infof("Switching to QUIC version %d. New connection ID: %x", highestSupportedVersion, c.connectionID)
+	utils.Infof("Switching to QUIC version %d. New connection ID: %x", newVersion, c.connectionID)
 
 	c.session.Close(errCloseSessionForNewVersion)
 	err = c.createNewSession(hdr.SupportedVersions)
--- a/client.go
+++ b/client.go
@@ -47,12 +47,13 @@ func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config
 		return nil, err
 	}
 
+	clientConfig := populateClientConfig(config)
 	c := &client{
 		conn:         &conn{pconn: pconn, currentAddr: remoteAddr},
 		connectionID: connID,
 		hostname:     hostname,
-		config:       config,
-		version:      protocol.SupportedVersions[len(protocol.SupportedVersions)-1], // use the highest supported version by default
+		config:       clientConfig,
+		version:      clientConfig.Versions[0],
 	}
 
 	c.connStateChangeOrErrCond.L = &c.mutex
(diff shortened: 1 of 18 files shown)


marten-seemann:
@lucas-clemente: What is your suggestion regarding our new changelog? Should I add an entry for the new quic.Config option in this PR?

lucas-clemente:
Yes, I'd generally say it's best to update the changelog from the same commit as the actual change is made (i.e. where the config option was added). That makes it easy to find the relevant commit by blaming the Changelog.

interface.go
```diff
@@ -63,6 +63,9 @@ type Config struct {
 	// If this field is not set, the Dial functions will return only when the connection is forward secure.
 	// Callbacks have to be thread-safe, since they might be called in separate goroutines.
 	ConnState ConnStateCallback
+	// The QUIC versions that can be negotiated.
+	// If not set, it uses all versions available.
+	Versions []protocol.VersionNumber
```
lucas-clemente: Add a comment here that this field should be considered unstable, as `protocol` is not part of our stable API (and we will need to figure out what to do here).

server.go
```diff
@@ -77,6 +77,20 @@ func Listen(conn net.PacketConn, config *Config) (Listener, error) {
 	}, nil
 }
 
+func populateConfig(config *Config) *Config {
+	versions := config.Versions
+	if len(versions) == 0 {
+		versions = make([]protocol.VersionNumber, len(protocol.SupportedVersions))
```
lucas-clemente: Why not `versions = protocol.SupportedVersions`?

server.go
```diff
@@ -77,6 +77,20 @@ func Listen(conn net.PacketConn, config *Config) (Listener, error) {
...
+		copy(versions, protocol.SupportedVersions)
+	}
+
+	return &Config{
+		TLSConfig: config.TLSConfig,
```
lucas-clemente: It's interesting that we copy the config, but not the TLS config. That means that we'll pick up runtime changes to the TLSConfig, but not to ours. I'm not sure that matters, just as a note.

client.go
```diff
@@ -67,6 +68,20 @@ func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config
 	return c.establishConnection()
 }
 
+func populateClientConfig(config *Config) *Config {
+	versions := config.Versions
+	if len(versions) == 0 {
+		versions = make([]protocol.VersionNumber, len(protocol.SupportedVersions))
```
lucas-clemente: Same thing here, why copy?

protocol/version.go
```diff
@@ -40,23 +40,21 @@ func IsSupportedVersion(supported []VersionNumber, v VersionNumber) bool {
...
 // it returns true and the version number, if there is one, otherwise false
-func HighestSupportedVersion(other []VersionNumber) (bool, VersionNumber) {
-	var otherSupported []VersionNumber
-	for _, ver := range other {
+func HighestSupportedVersion(ours, theirs []VersionNumber) (bool, VersionNumber) {
+	var theirsSupported []VersionNumber
+	for _, ver := range theirs {
 		if ver != VersionUnsupported {
-			otherSupported = append(otherSupported, ver)
+			theirsSupported = append(theirsSupported, ver)
```
lucas-clemente: This is allocation-heavy without much work from the client, we should benchmark that at some time.

server.go
marten-seemann: Well, first of all, you can't copy a `tls.Config`, because it contains a mutex.
I'd like to have a function that takes a `quic.Config` provided by the user and populates all unset fields with default values, so that later on in the code we don't have to worry about that any more.
What do you think is the nicer way to do it? Create a new struct, effectively treating the user-supplied `quic.Config` as a `const`, or doing the replacements in that struct?

protocol/version.go
marten-seemann: I think we don't need this at all. Before this PR, the `PublicHeader` would parse a Version Negotiation Packet such that unsupported versions (i.e. versions for which we didn't have a `protocol.VersionNumber` defined) were replaced by `VersionUnsupported`.
I changed that in one of the commits in the PR, since "supported" / "unsupported" now has two different meanings, if the user specifies the acceptable versions in the `quic.Config`. Thus, we don't need this check here anymore.

client.go
marten-seemann: Because I don't want to allow runtime-changes of the supported versions. Doing so will likely lead to nasty race conditions (clients won't like it if a VNP contains a certain set of versions, but the SHLO contains a different one).
The cost for copying this slice is minimal, but makes it explicit what is going on (and that runtime changes are not possible).

server.go
lucas-clemente: You can copy a tls.Config using https://golang.org/pkg/crypto/tls/#Config.Clone.

It seems net/tls copies its config when it makes changes, so I'd prefer to do that too. However it doesn't protect against races if the user changes stuff in the config after passing it (the clone is shallow), so I don't feel like we have to.

lucas-clemente: Forgot the link: https://golang.org/src/crypto/tls/tls.go#L131

marten-seemann:
@lucas-clemente: I think I addressed all your comments. Please rereview.

marten-seemann:
Apparently, `tls.Config.Clone()` was introduced in Go 1.8 :(

lucas-clemente:
LGTM

PR #579, opened 2017-04-30
implement stream deadlines

### LINKED ISSUE #514: implement deadlines for streams
We should implement a `SetDeadline`, `SetReadDeadline` and `SetWriteDeadline` for `Stream`s, analogous to the `net.Conn` (https://golang.org/pkg/net/#Conn).

marten-seemann:
Fixes #514.

I ran the benchmark tests a couple of times, and I can't see any performance impact. 
@lucas-clemente, maybe you have more ideas for benchmarking this?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@
 - Add a `quic.Config` option to configure the source address validation
 - Add a `quic.Config` option to configure the handshake timeout
 - Add a `quic.Config` option to configure keep-alive
+- Implement `net.Conn`-style deadlines for streams
 - Remove the `tls.Config` from the `quic.Config`. The `tls.Config` must now be passed to the `Dial` and `Listen` functions as a separate parameter. See the [Godoc](https://godoc.org/github.com/lucas-clemente/quic-go) for details.
 - Changed the log level environment variable to only accept strings ("DEBUG", "INFO", "ERROR"), see [the wiki](https://github.com/lucas-clemente/quic-go/wiki/Logging) for more details.
 - Rename the `h2quic.QuicRoundTripper` to `h2quic.RoundTripper`
--- a/stream.go
+++ b/stream.go
@@ -43,10 +43,10 @@ type stream struct {
 	frameQueue *streamFrameSorter
 	readChan   chan struct{}
 
-	dataForWriting       []byte
-	finSent              utils.AtomicBool
-	rstSent              utils.AtomicBool
-	doneWritingOrErrCond sync.Cond
+	dataForWriting []byte
+	finSent        utils.AtomicBool
+	rstSent        utils.AtomicBool
+	writeChan      chan struct{}
 
 	flowControlManager flowcontrol.FlowControlManager
 }
@@ -56,16 +56,15 @@ func newStream(StreamID protocol.StreamID,
 	onData func(),
 	onReset func(protocol.StreamID, protocol.ByteCount),
 	flowControlManager flowcontrol.FlowControlManager) *stream {
-	s := &stream{
+	return &stream{
 		onData:             onData,
 		onReset:            onReset,
 		streamID:           StreamID,
 		flowControlManager: flowControlManager,
 		frameQueue:         newStreamFrameSorter(),
 		readChan:           make(chan struct{}, 1),
+		writeChan:          make(chan struct{}, 1),
 	}
-	s.doneWritingOrErrCond.L = &s.mutex
-	return s
 }
 
 // Read implements io.Reader. It is not thread safe!
@@ -147,29 +146,34 @@ func (s *stream) Read(p []byte) (int, error) {
 }
 
 func (s *stream) Write(p []byte) (int, error) {
-	if s.resetLocally.Get() {
-		return 0, s.err
-	}
-
 	s.mutex.Lock()
-	defer s.mutex.Unlock()
-
-	if s.err != nil {
-		return 0, s.err
+	if s.resetLocally.Get() || s.err != nil {
+		err := s.err
+		s.mutex.Unlock()
+		return 0, err
 	}
-
 	if len(p) == 0 {
+		s.mutex.Unlock()
 		return 0, nil
 	}
 
 	s.dataForWriting = make([]byte, len(p))
 	copy(s.dataForWriting, p)
-
 	s.onData()
+	s.mutex.Unlock()
 
-	for s.dataForWriting != nil && s.err == nil {
-		s.doneWritingOrErrCond.Wait()
+	for {
+		s.mutex.Lock()
+		if s.dataForWriting == nil || s.err != nil {
+			s.mutex.Unlock()
+			break
+		}
+		s.mutex.Unlock()
+		<-s.writeChan
 	}
+
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
 
 	if s.err != nil {
 		return len(p) - len(s.dataForWriting), s.err
@@ -190,14 +194,12 @@ func (s *stream) lenOfDataForWriting() protocol.ByteCount {
 
 func (s *stream) getDataForWriting(maxBytes protocol.ByteCount) []byte {
 	s.mutex.Lock()
-	if s.err != nil {
-		s.mutex.Unlock()
+	defer s.mutex.Unlock()
+
+	if s.err != nil || s.dataForWriting == nil {
 		return nil
 	}
-	if s.dataForWriting == nil {
-		s.mutex.Unlock()
-		return nil
-	}
+
 	var ret []byte
 	if protocol.ByteCount(len(s.dataForWriting)) > maxBytes {
 		ret = s.dataForWriting[:maxBytes]
@@ -205,10 +207,9 @@ func (s *stream) getDataForWriting(maxBytes protocol.ByteCount) []byte {
 	} else {
 		ret = s.dataForWriting
 		s.dataForWriting = nil
-		s.doneWritingOrErrCond.Signal()
+		s.signalWrite()
 	}
 	s.writeOffset += protocol.ByteCount(len(ret))
-	s.mutex.Unlock()
 	return ret
 }
 
@@ -263,6 +264,14 @@ func (s *stream) signalRead() {
 	}
 }
 
+// signalRead performs a non-blocking send on the writeChan
+func (s *stream) signalWrite() {
+	select {
+	case s.writeChan <- struct{}{}:
+	default:
+	}
+}
+
 // CloseRemote makes the stream receive a "virtual" FIN stream frame at a given offset
 func (s *stream) CloseRemote(offset protocol.ByteCount) {
 	s.AddStreamFrame(&frames.StreamFrame{FinBit: true, Offset: offset})
@@ -277,7 +286,7 @@ func (s *stream) Cancel(err error) {
 	if s.err == nil {
 		s.err = err
 		s.signalRead()
-		s.doneWritingOrErrCond.Signal()
+		s.signalWrite()
 	}
 	s.mutex.Unlock()
 }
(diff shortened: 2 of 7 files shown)


marten-seemann:
I finally found the race condition (thanks to the awesome [deadlock package](https://github.com/sasha-s/go-deadlock)), and the CIs seem happy with this PR now.
@lucas-clemente: You can review now.

stream.go
```diff
@@ -96,11 +101,30 @@ func (s *stream) Read(p []byte) (int, error) {
...
 				break
 			}
-			<-s.readChan
+
+			if deadline.IsZero() {
+				<-s.readChan
+			} else {
+				select {
+				case <-s.readChan:
+				case <-time.After(deadline.Sub(time.Now())):
```
lucas-clemente: There's a race here: You should only call time.Now() once, and use the value for the comparison above and the duration calculation here. Otherwise, the condition above might be false, but here the duration is negative (which leads to the timer now firing).

stream.go
```diff
@@ -190,14 +191,36 @@ func (s *stream) Write(p []byte) (int, error) {
...
 		}
 		s.mutex.Unlock()
-		<-s.writeChan
+
+		if deadline.IsZero() {
+			<-s.writeChan
+		} else {
+			select {
+			case <-s.writeChan:
+			case <-time.After(deadline.Sub(time.Now())):
```
lucas-clemente: There's a race here: You should only call time.Now() once, and use the value for the comparison above and the duration calculation here. Otherwise, the condition above might be false, but here the duration is negative (which leads to the timer now firing).

stream.go
```diff
@@ -323,6 +323,8 @@ func (s *stream) signalWrite() {
 	}
 }
 
+// SetDeadline sets the read deadline
```
lucas-clemente: Nit: Please use the proper comment rules for godoc (i.e. fullstops at the end of each line / sentence, capitalize new lines / sentences) throughout this commit.

Changelog.md
```diff
@@ -7,6 +7,7 @@
 - Add a `quic.Config` option to request truncation of the connection ID from a server
 - Add a `quic.Config` option to configure the source address validation
 - Add a `quic.Config` option to configure the handshake timeout
+- Implement `net.Conn`-style deadlines for streams
```
lucas-clemente: This PR doesn't actually change the stream interface, so these functions aren't visible publicly unless you type-assert. Is this intentional? If so, this should be documented someplace.

lucas-clemente:
And I just noticed that one of the tests is already flaky on appveyor, please fix that before merging :)

```
Stream writing deadlines [It] doesn't unblock if the deadline is changed before the first one expires 
c:/gopath/src/github.com/lucas-clemente/quic-go/stream_test.go:811
  Expected
      <time.Time>: 2017-06-27T16:36:58.7773287Z
  to be ~
      <time.Time>: 2017-06-27T16:36:58.6913271Z
  c:/gopath/src/github.com/lucas-clemente/quic-go/stream_test.go:810
  Full Stack Trace
  	c:/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:69 +0x1f6
  github.com/onsi/gomega/internal/assertion.(*Assertion).To(0xc04656a240, 0xb66680, 0xc04656a280, 0x0, 0x0, 0x0, 0x1)
  	c:/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:35 +0xb5
  github.com/lucas-clemente/quic-go.glob..func20.7.7.3()
  	c:/gopath/src/github.com/lucas-clemente/quic-go/stream_test.go:810 +0x53b
  github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc046c3b2c0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
  	c:/gopath/src/github.com/lucas-clemente/quic-go/quic_suite_test.go:13 +0x6b
  testing.tRunner(0xc042044680, 0x969820)
  	C:/go/src/testing/testing.go:657 +0x9d
  created by testing.(*T).Run
  	C:/go/src/testing/testing.go:697 +0x2d1
```

marten-seemann:
The problem with the tests is that we need to set a deadline, do a `Read` or a `Write` and check that it returns at the right time. Apparently we need even longer durations to make that work, which slows down the whole test suite :(

lucas-clemente:
I think it should be fine to test it with a huge deadline and check that it doesn't return (Consistently), and check that with an alost-zero deadline it returns immediately. As you know, I generally dislike tests that assert a specific time…

marten-seemann:
I completely agree about that, but I don't think we can covers all cases like that. 
For example, there's one test that sets a deadline, and sets a new deadline (further in the future) before the first one expires. `Read` / `Write` is supposed to unblock after the second deadline.

lucas-clemente:
This issue is generally solved (e.g. in the congestion controller) by injecting a clock and not using time.Now directly. Then you can mock that clock in the tests and arbitrarily advance time, see https://github.com/lucas-clemente/quic-go/blob/2102e19ca4ff53d31f55f6a6de65a9a485458872/congestion/cubic_sender_test.go#L14. I just really don't think we can trust timing to be more accurate than 10s on a container based system such as travis.

Options I can think of:

- Switch to mocked clocks
- Mark the test as pending and switch to mocked clocks later
- Increase times to tens of seconds.
- Use ginkgo's flaky tests setup
- Introduce a constant "timescale" factor across tests that we set higher for travis than for local dev

I'd prefer the first or second.

marten-seemann:
I don't think a mock clock would work here (because I'm using `time.After` when setting the deadline).
I kind of like the idea with the scaling factor. I don't care if the tests take a couple of seconds longer on the CIs, but running them as fast as possible locally is essential in my workflow.

stream.go
marten-seemann: I don't really see a race condition here, just an optimization to save one syscall.
In the situation you're describing the timer immediately fires, so the loop will loop and we'll break. If I replace this by one call of `time.Now()`, the timer will fire a fraction of a millisecond later, and the loop will loop and break.

stream.go
```diff
@@ -147,30 +146,39 @@ func (s *stream) Read(p []byte) (int, error) {
...
+		s.mutex.Unlock()
+		return 0, err
 	}
-
-	s.mutex.Lock()
-	defer s.mutex.Unlock()
-
 	if s.err != nil {
+		s.mutex.Unlock()
 		return 0, s.err
```
marten-seemann: Fixed.

stream.go
lucas-clemente: Oh right, I thought a negative timer wouldn't fire. Then disregard my comment.

Changelog.md
marten-seemann: No, I just forgot to add them to the `Stream` interface.

stream.go
marten-seemann: We can still do it as a performance optimization if you want.

stream.go
```diff
@@ -97,14 +102,34 @@ func (s *stream) Read(p []byte) (int, error) {
 				err = s.err
```
lucas-clemente: Race here

stream.go
```diff
@@ -84,9 +89,9 @@ func (s *stream) Read(p []byte) (int, error) {
 	for bytesRead < len(p) {
 		s.mutex.Lock()
 		frame := s.frameQueue.Head()
+		s.mutex.Unlock()
 
 		if frame == nil && bytesRead > 0 {
-			s.mutex.Unlock()
 			return bytesRead, s.err
```
lucas-clemente: Race here

marten-seemann:
I just implemented the scaling factor for the deadline test, mainly to see how it performs. I could reduce the deadline duration a lot, such that all of the deadline tests only take around 0.4s locally in total. The CI scaling factor is currently 20, which means that the tests will take around 8s on the CIs.

stream_test.go
```diff
@@ -253,22 +265,22 @@ var _ = Describe("Stream", func() {
 			})
 
 			It("unblocks after the deadline", func() {
-				deadline := time.Now().Add(200 * time.Millisecond)
+				deadline := time.Now().Add(50 * getDeadlineScalingFactor() * time.Millisecond)
```
lucas-clemente: What do you think of making this a function, `scaleMs(50)`?

stream_test.go
```diff
@@ -27,6 +29,16 @@ var _ = Describe("Stream", func() {
 		mockFcm *mocks_fc.MockFlowControlManager
 	)
 
+	// in the tests for the stream deadlines we set a deadline
+	// and wait to make an assertion when Read / Write was unblocked
+	// on the CIs, the timing is a lot less precise, so scale every duration by this factor
+	getDeadlineScalingFactor := func() time.Duration {
+		if os.Getenv("CI") == "true" {
```
lucas-clemente: I dislike detecting a magic env and changing behavior based on that. Can you add the scaling factor as an int/float flag instead?

marten-seemann: I don't understand. At some point we will have to read the "magical" CI env var, if we want to detect that we're running on a CI.

stream_test.go
marten-seemann: I don't have any preference here, but I can change it if you want to.

stream_test.go
lucas-clemente: Add `timescaleFactor` flag to the test suite and set that from the command line on the CI. That way it's pretty clear what the differences between the two setups are.

marten-seemann: That's a pretty neat idea. We can use that for a couple of other tests as well.

stream_test.go
lucas-clemente: It seems more readable, especially if it's in a lot of places.

stream_test.go
marten-seemann: Done.

stream_test.go
marten-seemann: Done.

marten-seemann:
@lucas-clemente: How do we proceed with this?

lucas-clemente:
Last I heard you wanted to make it a flag but had issues. Why did that work out in the end?How did that go in the end?

marten-seemann:
Yes, the issues are:
* Go is very unhappy with flags not parsed (meaning it will error and stop execution), so we can't run `ginkgo -r` with this flag.
* I couldn't get Ginkgo to skip the root package as well as the integrationtests package, which would be necessary if we want to run the root package as a separate Ginkgo task with that flag.

lucas-clemente:
`ginkgo -r -skipPackage=integrationtests *` works for me.

marten-seemann:
Strictly speaking, you're not testing recursively from the root package anymore, but from each of the subpackages. But that should work as well.

lucas-clemente:
Yep, that's the idea :)

marten-seemann:
Unfortunately, it doesn't work. Windows isn't able to expand wildcards 🤦‍♂️, so all our tests fail on Appveyor.
I now already spent way more than 2 hours on these flags, and to be honest, I don't think it was worth it.

@lucas-clemente: If you insist on this flag, please have a look yourself. Otherwise, I still don't see anything wrong with setting an environment variable.

lucas-clemente:
Fine, use the env.

marten-seemann:
Done.

stream.go
```diff
@@ -40,32 +42,35 @@ type stream struct {
...
+	dataForWriting []byte
+	finSent        utils.AtomicBool
+	rstSent        utils.AtomicBool
+	writeChan      chan struct{}
+	writeDeadline  time.Time
 
 	flowControlManager flowcontrol.FlowControlManager
 }
 
+var errDeadline = errors.New("deadline exceeded")
```
AudriusButkevicius: This should implement https://golang.org/pkg/net/#Error

marten-seemann: You are right. Thanks! 👍 
I opened #741 to track this.

PR #580, opened 2017-04-30
set ginkgo timeouts for client integrationtests

marten-seemann:
If a test of our client is taking too long, ginkgo should kill it.


marten-seemann:
Unfortunately, ginkgo doesn't seem to have this option. I don't know why, it is even possible with the normal go tests (via `go test -timeout *x*s`).

PR #589, opened 2017-05-05
accept server config if one of the key exchange algorithms is Curve25519

### LINKED ISSUE #588: CryptoNoSupport error
I am running the example client against a local server and I get this error:
`Closing session with error: CryptoNoSupport: KEXS`

I don't quite understand what this error means. Looks like the server doesn't support the cipher which the client requests? If so, how can I make a request with a different cipher?

marten-seemann:
In the server config, the server lists all supported key exchange algorithms. The client should accept the config as long as the list contains C255 (Curve25519).

### DIFF
--- a/handshake/server_config_client.go
+++ b/handshake/server_config_client.go
@@ -57,8 +57,6 @@ func (s *serverConfigClient) parseValues(tagMap map[Tag][]byte) error {
 	s.ID = scfgID
 
 	// KEXS
-	// TODO: allow for P256 in the list
-	// TODO: setup Key Exchange
 	kexs, ok := tagMap[TagKEXS]
 	if !ok {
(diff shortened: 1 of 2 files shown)


lucas-clemente:
Yeah. I also don't want to support P256, see https://safecurves.cr.yp.to/

marten-seemann:
That's not the issue. I never planned to implement P256.
The issue here is, that we currently refuse to talk to a peer who offers **both** P256 and C255.

lucas-clemente:
I know, my comment was just meant as an aside.

marten-seemann:
Closing this for now. We haven't seen those kinds of servers or clients out in the wild, so I'm reluctant to spend more time on this, especially since the switch to TLS 1.3 is happening soon (hopefully).

PR #595, opened 2017-05-07
replace CryptoSetup.HandshakeComplete() by closing the aeadChanged channel

marten-seemann:
This allows us to remove `HandshakeComplete()` from the `CryptoSetup` interface. 
More importantly, it also provides a signal to the session when the handshake completes. This will probably come handy when implementing the new API.

### DIFF
--- a/session.go
+++ b/session.go
@@ -78,7 +78,7 @@ type session struct {
 
 	// this channel is passed to the CryptoSetup and receives the current encryption level
 	// it is closed as soon as the handshake is complete
-	aeadChanged       chan protocol.EncryptionLevel
+	aeadChanged       <-chan protocol.EncryptionLevel
 	handshakeComplete bool
 
 	nextAckScheduledTime time.Time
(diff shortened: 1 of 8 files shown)


session.go
```diff
@@ -226,7 +235,12 @@ runLoop:
 			// This is a bit unclean, but works properly, since the packet always
 			// begins with the public header and we never copy it.
 			putPacketBuffer(p.publicHeader.Raw)
-		case l := <-s.aeadChanged:
+		case l, ok := <-aeadChanged:
+			if !ok {
+				s.handshakeComplete = true
+				aeadChanged = nil // prevent this case from ever being selected again
+				continue
```
lucas-clemente: I don't have the full context right now, but are you sure about this continue? I think that completing the handshake unblock some things, or at the least change some timeouts, so that we should go through the full loop once?

marten-seemann: Good catch. I fixed that.

PR #599, opened 2017-05-09
fix infinite loop in session test

marten-seemann:
fixes #598.

The `mockReceivedPacketHandler` returned the same ACK frame over and over again, so that the loop in `session.sendPacket()` would send packets containing this packet indefinitely.

Note that while this fixes the flaky test, there's still a race condition. I'll create a separate issue for this.

PR #601, opened 2017-05-09
fix race condition when sending ConnectionClose frames

marten-seemann:
fixes #599

### DIFF
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/lucas-clemente/quic-go/ackhandler"
 	"github.com/lucas-clemente/quic-go/frames"
@@ -20,6 +21,8 @@ type packedPacket struct {
 }
 
 type packetPacker struct {
+	mutex sync.Mutex
+
 	connectionID protocol.ConnectionID
 	perspective  protocol.Perspective
 	version      protocol.VersionNumber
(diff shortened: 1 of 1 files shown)


packet_packer.go
```diff
@@ -20,6 +21,8 @@ type packedPacket struct {
 }
 
 type packetPacker struct {
+	mutex sync.Mutex
```
lucas-clemente: I'm not a big fan of this mutex here. We're more and more breaking up our concept of not having concurrent code apart from the session, and this makes stuff much harder to reason about. In particular, to understand this mutex, I need to know which entry paths into this class exist that may be concurrent.

So, I would propose two options:

- Either make packing the ConnectionClose sufficiently independant from normal packing that we don't need it anymore (could we e.g. pass the controlFrames as param to packPacket?).
- Fully protect everything in this class with the mutex (e.g. RetransmitNonForwardSecurePacket), so that there's no hidden knowledge required to extend this in the future.

marten-seemann: I don't like it either. I think I forgot the mutex for `RetransmitNonForwardSecurePacket`. The mutex here protects two things: first the `controlFrames` slice in the `packetPacker`, and second the `packetNumberGenerator`.
Interestingly, the **only** place where a concurrent access might occur is when sending `ConnectionClose`s. So it would be nice if we could find a non-concurrent solution for that.

lucas-clemente: SGTM. Then please add the mutex to RetransmitNonForwardSecurePacket.

marten-seemann:
Superseeded by #608.

PR #606, opened 2017-05-09
add a quic.Config option to request connection ID truncation

### LINKED ISSUE #434: ask server to truncate connection ID in client
The client should send the `TCID` tag during the crypto handshake, to allow the server to omit the connection ID.

Once we have a `quic.Config`, this should be a configurable option, but truncating should be the default in this direction (packets sent from server to client).

### LINKED ISSUE #524: Make packet number truncation configurable
Packet number truncation should be configurable via [quic.Config](https://github.com/lucas-clemente/quic-go/blob/master/interface.go#L60). I think we should add a enum there with values Default, Yes, No. Default would be no for a client and yes for a server.

marten-seemann:
Fixes #524, fixes #434.

I don't really like the name of the `quic.Config` option though, maybe @lucas-clemente has a better idea (`RequestConnectionIDTruncation` or so, but that's too long).

I added a `handshake.TransportParameters` struct. The name comes from the QUIC WG draft (https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#transport-parameter-definitions). Currently it only has one field, but it will need some more in the future. I hope this will make contributing to the other *want-help* issues easier.

The option is currently only valid for the client. In principle, we could also allow servers to allow connection ID truncation, but we'll need to add some logic to the `server` to assign packets to the right sessions then.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -67,6 +67,10 @@ type Config struct {
 	// If not set, it uses all versions available.
 	// Warning: This API should not be considered stable and will change soon.
 	Versions []protocol.VersionNumber
+	// Ask the server to truncate the connection ID sent in the Public Header.
+	// This saves 8 bytes in the Public Header in every packet. However, if the IP address of the server changes, the connection cannot be migrated.
+	// Currently only valid for the client.
+	RequestConnectionIDTruncation bool
 }
 
 // A Listener for incoming QUIC connections
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,4 +3,5 @@
 ## v0.6.0 (unreleased)
 
 - Add a `quic.Config` option for QUIC versions
+- Add a `quic.Config` option to request truncation of the connection ID from a server
 - Various bugfixes
--- a/client.go
+++ b/client.go
@@ -75,9 +75,10 @@ func populateClientConfig(config *Config) *Config {
 	}
 
 	return &Config{
-		TLSConfig: config.TLSConfig,
-		ConnState: config.ConnState,
-		Versions:  versions,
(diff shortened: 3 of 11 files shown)


interface.go
```diff
@@ -67,6 +67,9 @@ type Config struct {
 	// If not set, it uses all versions available.
 	// Warning: This API should not be considered stable and will change soon.
 	Versions []protocol.VersionNumber
+	// Ask the server to truncate the connection ID sent in the Public Header.
+	// Only valid for the client.
```
lucas-clemente: s/Only/Currently only/

interface.go
```diff
@@ -67,6 +67,9 @@ type Config struct {
 	// If not set, it uses all versions available.
 	// Warning: This API should not be considered stable and will change soon.
 	Versions []protocol.VersionNumber
+	// Ask the server to truncate the connection ID sent in the Public Header.
```
lucas-clemente: Please explain why a user would enable this (space savings), and the requirements for this option to be safe.

interface.go
```diff
@@ -67,6 +67,9 @@ type Config struct {
...
+	// Only valid for the client.
+	TruncateConnectionID bool
```
lucas-clemente: I also don't like this name. I think either "Request" or "Accept" should be in there, to make it clear this option applies to the other peer.

marten-seemann: Yeah, that's what I wrote in my PR. Would you be ok with `RequestConnectionIDTruncation`? It's a bit too long for my taste, but tbh, I don't have any better idea.

lucas-clemente: SGTM

PR #610, opened 2017-05-11
fix a race condition in the client crypto setup

marten-seemann:
This is a pretty large PR for fixing a simple race condition.

When doing the handshake for a the server, there's only one thing that can trigger an escalation of the crypto level: a new CHLO arriving on the crypto stream. That's why it's a good solution to just have one go routine block on reading the crypto stream (`HandleCryptoStream`).
For the client however, there are two things that can trigger an escalation of the crypto level: a new message (REJ or SHLO) on the crypto stream, and a diversification nonce (which is needed to create the `secureAEAD`). This PR introduces a `select` in `HandleCryptoStream`, which accepts the diversification nonce as well as handshake messages, which are passed there from a separate go routine parsing the crypto channel.

### DIFF
--- a/handshake/crypto_setup_client.go
+++ b/handshake/crypto_setup_client.go
@@ -37,7 +37,7 @@ type cryptoSetupClient struct {
 	lastSentCHLO     []byte
 	certManager      crypto.CertManager
 
-	divNonceChan         chan struct{}
+	divNonceChan         chan []byte
 	diversificationNonce []byte
 
 	clientHelloCounter int
@@ -85,7 +85,7 @@ func NewCryptoSetupClient(
 		nullAEAD:             crypto.NewNullAEAD(protocol.PerspectiveClient, version),
 		aeadChanged:          aeadChanged,
 		negotiatedVersions:   negotiatedVersions,
-		divNonceChan:         make(chan struct{}),
+		divNonceChan:         make(chan []byte),
 	}, nil
 }
 
@@ -123,7 +123,11 @@ func (h *cryptoSetupClient) HandleCryptoStream() error {
 
 		var message HandshakeMessage
 		select {
-		case <-h.divNonceChan:
+		case divNonce := <-h.divNonceChan:
+			if len(h.diversificationNonce) != 0 && !bytes.Equal(h.diversificationNonce, divNonce) {
+				return errConflictingDiversificationNonces
+			}
+			h.diversificationNonce = divNonce
 			// there's no message to process, but we should try upgrading the crypto again
 			continue
 		case message = <-messageChan:
@@ -372,19 +376,8 @@ func (h *cryptoSetupClient) DiversificationNonce() []byte {
 	panic("not needed for cryptoSetupClient")
 }
 
-func (h *cryptoSetupClient) SetDiversificationNonce(data []byte) error {
-	h.mutex.Lock()
-	defer h.mutex.Unlock()
-
-	if len(h.diversificationNonce) == 0 {
-		h.diversificationNonce = data
-		h.divNonceChan <- struct{}{}
-		return nil
-	}
-	if !bytes.Equal(h.diversificationNonce, data) {
-		return errConflictingDiversificationNonces
-	}
-	return nil
+func (h *cryptoSetupClient) SetDiversificationNonce(data []byte) {
+	h.divNonceChan <- data
 }
 
 func (h *cryptoSetupClient) sendCHLO() error {
--- a/handshake/crypto_setup_client.go
+++ b/handshake/crypto_setup_client.go
@@ -29,14 +29,16 @@ type cryptoSetupClient struct {
 
 	serverConfig *serverConfigClient
 
-	stk                  []byte
-	sno                  []byte
-	nonc                 []byte
-	proof                []byte
+	stk              []byte
+	sno              []byte
+	nonc             []byte
+	proof            []byte
+	chloForSignature []byte
+	lastSentCHLO     []byte
+	certManager      crypto.CertManager
+
+	divNonceChan         chan struct{}
 	diversificationNonce []byte
-	chloForSignature     []byte
-	lastSentCHLO         []byte
-	certManager          crypto.CertManager
 
 	clientHelloCounter int
 	serverVerified     bool // has the certificate chain and the proof already been verified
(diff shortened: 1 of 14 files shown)


marten-seemann:
I added some more tests for the `HandshakeMessage`, to make Codecov happy.

handshake/crypto_setup_client.go
```diff
@@ -104,35 +104,30 @@ func (h *cryptoSetupClient) HandleCryptoStream() error {
...
-		}
-
-		if messageTag == TagSHLO {
-			utils.Debugf("Got SHLO:\n%s", printHandshakeMessage(cryptoData))
-			err = h.handleSHLOMessage(cryptoData)
+		utils.Debugf("Got %s", message)
+		switch message.Tag {
+		case TagREJ:
+			err = h.handleREJMessage(message.Data)
 			if err != nil {
```
lucas-clemente: You could merge the two err != nil cases into one after the switch statement.

handshake/handshake_message.go
```diff
@@ -12,27 +12,36 @@ import (
...
+type HandshakeMessage struct {
+	Tag  Tag
+	Data map[Tag][]byte
+}
+
+var _ fmt.Stringer = &HandshakeMessage{}
+
 // ParseHandshakeMessage reads a crypto message
-func ParseHandshakeMessage(r io.Reader) (Tag, map[Tag][]byte, error) {
+func ParseHandshakeMessage(r io.Reader) (hm HandshakeMessage, err error) {
```
lucas-clemente: I'm not a fan of having named return values for longer functions (and, in particular, mixing naked and normal returns). Unless you have a very specific reason for using them here, I'd strongly prefer using normal returns.

handshake/crypto_setup_client.go
```diff
@@ -83,30 +85,51 @@ func NewCryptoSetupClient(
...
 
 func (h *cryptoSetupClient) HandleCryptoStream() error {
+	messageChan := make(chan HandshakeMessage)
+
+	go func() {
+		for {
+			message, err := ParseHandshakeMessage(h.cryptoStream)
+			if err != nil {
+				close(messageChan)
+				return
```
lucas-clemente: This silently discards the error here. Either use this error as the return value of the function (which is notoriously tricky with go channels 🙄), or log the error.

handshake/crypto_setup_client.go
```diff
@@ -83,30 +85,51 @@ func NewCryptoSetupClient(
...
 			err = h.sendCHLO()
 			if err != nil {
 				return err
 			}
 		}
 
-		message, err := ParseHandshakeMessage(h.cryptoStream)
-		if err != nil {
-			return qerr.HandshakeFailed
+		var message HandshakeMessage
```
lucas-clemente: I'd prefer doing `var ok bool` here instead.

handshake/crypto_setup_client.go
```diff
@@ -354,9 +377,13 @@ func (h *cryptoSetupClient) DiversificationNonce() []byte {
 }
 
 func (h *cryptoSetupClient) SetDiversificationNonce(data []byte) error {
+	h.mutex.Lock()
+	defer h.mutex.Unlock()
+
 	if len(h.diversificationNonce) == 0 {
 		h.diversificationNonce = data
-		return h.maybeUpgradeCrypto()
+		h.divNonceChan <- struct{}{}
```
lucas-clemente: This can deadlock if the chan is not being read (e.g. if we received an invalid handshake message and the loop terminated).

handshake/crypto_setup_server_test.go
```diff
@@ -100,7 +101,11 @@ type mockStream struct {
 }
 
 func (s *mockStream) Read(p []byte) (int, error) {
-	return s.dataToRead.Read(p)
+	n, _ := s.dataToRead.Read(p)
+	if n == 0 { // block if there's no data
+		time.Sleep(time.Hour)
```
lucas-clemente: Nonononononono! This makes clean test shutdown impossible, don't do this.

handshake/crypto_setup_client_test.go
```diff
@@ -640,13 +640,15 @@ var _ = Describe("Client Crypto Setup", func() {
 		})
 
 		It("tries to escalate the crypto after receiving a diversification nonce", func(done Done) {
-			go cs.HandleCryptoStream()
-			time.Sleep(50 * time.Millisecond) // wait for the first maybeUpgradeCrypto to finish
+			go func() {
+				defer GinkgoRecover()
+				cs.HandleCryptoStream()
+				Fail("HandleCryptoStream should not have returned")
```
lucas-clemente: This makes clean test shutdown impossible, don't do this.

handshake/crypto_setup_client.go
lucas-clemente: Just saw that this was resolved in a later commit, disregard.

handshake/crypto_setup_server_test.go
marten-seemann: Any other suggestion? I don't want to get an `io.EOF` here. This immediately kills the `HandleCryptoStream` in the crypto setups.

handshake/crypto_setup_client_test.go
marten-seemann: Well, if there's no error, then it doesn't return. We have that at a bunch of paces, where we just call `cs.HandleCryptoStream()`. Just here, we're explicit about that.

handshake/crypto_setup_server_test.go
lucas-clemente: If you need to block, block on something that you can unblock during test shutdown (e.g. a chan).

handshake/crypto_setup_client_test.go
lucas-clemente: Then we should fix all of those places at some point… And the first step to that is not allowing this for new tests ;)

Having goroutines running after a test finishes causes all kinds of issues, just have a look at the hoops I had to jump through for the tests in the main package. I really want to avoid that.

marten-seemann: I'll do the chan solution for unblocking the read. That should take care of all leaking `go HandleCryptoStream` routines.

marten-seemann: Should be fixed now.

handshake/crypto_setup_client.go
marten-seemann: Not tricky at all. Just make another chan.

marten-seemann:
@lucas-clemente: I think I addressed all your points. Please review again.

handshake/crypto_setup_client.go
```diff
@@ -83,58 +85,73 @@ func NewCryptoSetupClient(
...
+		var message HandshakeMessage
+		select {
+		case divNonce := <-h.divNonceChan:
+			if len(h.diversificationNonce) != 0 && !bytes.Equal(h.diversificationNonce, divNonce) {
+				return errConflictingDiversificationNonces
+			}
+			h.diversificationNonce = divNonce
+			// there's no message to process, but we should try upgrading the crypto again
+			continue
+		case msg := <-messageChan:
```
lucas-clemente: This variable is unnecessary, you should be able to read directly into `message`.

PR #612, opened 2017-05-11
Added parsing of PUBS tag and filtering according to KEXS

Eichhoernchen:
This now allows quic-go to contact servers announcing multiple KEXS methods, e.g. like Akamai.

### DIFF
--- a/handshake/server_config_client.go
+++ b/handshake/server_config_client.go
@@ -73,8 +73,8 @@ func (s *serverConfigClient) parseValues(tagMap map[Tag][]byte) error {
 			break
 		}
 	}
-	if (c255Foundat < 0) {
-		return qerr.Error(qerr.CryptoNoSupport, "KEXS")
+	if c255Foundat < 0 {
+		return qerr.Error(qerr.CryptoNoSupport, "KEXS: Could not find C255, other key exchanges are not supported")
 	}
 
 	// AEAD
--- a/handshake/server_config_client.go
+++ b/handshake/server_config_client.go
@@ -110,13 +110,21 @@ func (s *serverConfigClient) parseValues(tagMap map[Tag][]byte) error {
 
 		err := binary.Read(bytes.NewReader([]byte{pubs[i], pubs[i+1], pubs[i+2], 0x00}), binary.LittleEndian, &last_len);
 		if err != nil {
+			return qerr.Error(qerr.CryptoInvalidValueLength, "PUBS not decodable")
+		}
+		if last_len == 0 {
 			return qerr.Error(qerr.CryptoInvalidValueLength, "PUBS")
 		}
+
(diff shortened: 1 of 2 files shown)


handshake/server_config_client.go
```diff
@@ -66,7 +65,15 @@ func (s *serverConfigClient) parseValues(tagMap map[Tag][]byte) error {
...
+	c255Foundat := -1
+
+	for i := 0; i < len(kexs)/4; i++ {
+		if bytes.Equal(kexs[4*i:4*i+4], []byte("C255")) {
+			c255Foundat = i
+			break
+		}
+	}
+	if (c255Foundat < 0) {
 		return qerr.Error(qerr.CryptoNoSupport, "KEXS")
```
lucas-clemente: I'd like this error to be more specific, e.g. "KEXS: Could not find C255, other key exchanges are not supported", as this is something that we might actually encounter in the wild :)

handshake/server_config_client_test.go
```diff
@@ -184,6 +184,12 @@ var _ = Describe("Server Config", func() {
 				Expect(err).To(MatchError("CryptoInvalidValueLength: PUBS"))
 			})
 
+			It("rejects PUBS values that have a zero length", func() {
+				tagMap[TagPUBS] = bytes.Repeat([]byte{0}, 100) // completely wrong length
+				err := scfg.parseValues(tagMap)
+				Expect(err).To(MatchError("CryptoInvalidValueLength: PUBS"))
+			})
+
```
lucas-clemente: This test is good, but I'd also like to have one where the C255 pubs is not at index 0. I'm fine with having a dummy value before that (e.g. 0x01, 0x00, 0x00, 0x00).

handshake/server_config_client.go
```diff
@@ -66,7 +65,15 @@ func (s *serverConfigClient) parseValues(tagMap map[Tag][]byte) error {
...
-	if !bytes.Equal(kexs, []byte("C255")) {
+	c255Foundat := -1
+
+	for i := 0; i < len(kexs)/4; i++ {
+		if bytes.Equal(kexs[4*i:4*i+4], []byte("C255")) {
+			c255Foundat = i
+			break
+		}
+	}
+	if (c255Foundat < 0) {
```
lucas-clemente: Nit: parens not necessary in Go :)

Eichhoernchen:
Hi, thanks for the inline comments (I haven't coded in go before ;)).
I'll include your feedback and then everything should be fine.

marten-seemann:
Thanks @Eichhoernchen for this PR!

Is it a public Akamai server that already speaks QUIC? If so, can you give me the URL? I'd love to point the QUIC client to their server and play around a bit.

PR #613, opened 2017-05-12
implement a new API

### LINKED ISSUE #506: Re-enable skipped client test
As of #503, the "doesn't return after version negotiation is established if no ConnState is defined" test is disabled.

### LINKED ISSUE #563: handle handshake errors in h2quic.Client
If the crypto handshake fails, the `h2quic.Client.Do` should return immediately.
At the moment, it is stuck waiting at the `cryptoChangedCond` `sync.Cond`, and will never return.

### LINKED ISSUE #564: return an error in Dial() if the handshake fails
Related to #563.

### LINKED ISSUE #566: think about replacing the ConnStateCallback by a channel
This came up in the discussion of #564. I don't understand how it helps there, but it might be an idea worth exploring on its own.
At the moment, users of our API have to deal with the fact that the `ConnStateCallback` is called in a separate goroutine, i.e. a common practice will be to protect the `ConnStateCallback` by a mutex (see for example [client.go#L69](https://github.com/lucas-clemente/quic-go/blob/master/h2quic/client.go#L69) . Using a channel, this could probably be avoided.

### LINKED ISSUE #568: Rethink API
Current new API proposal after discussion with @marten-seemann 

```go
// Server

listener, err := quic.ListenAddr(addr, cfg)
go listener.Serve() // <-- problematic, could be in ListenAddr()

for {
	session, err := listener.Accept()
	go handleSession(session)
}

// For non-fw cases

type NonFwSession interface {
	Session
	WaitUntilForwardSecure() error
}

for {
	session, err := listener.AcceptNonForwardSecure()
	// session is of type NonFwSession, which is
	go func() {
		doNonFwWork(session)
		err := session.WaitUntilForwardSecure()
		doWork(session)
	}()
}

// Client

session, err := quic.DialAddr(addr, cfg)
// session is fw-secure here

// For non-fw cases

fwSecureSession, err := quic.DialAddr(addr, &quic.Config{
	OnNonForwardSecure: func(nonFwSecureSession Session) {
		/* go */ doNonFwWork(nonFwSecureSession)
	},
})
// we block, so fwSecureSession here is fw secure

// Client #2

session, err := quic.DialAddr(addr, cfg)
// session is fw-secure here

// For non-fw cases
session, err := quic.DialAddrNonForwardSecure(addr, cfg)
// session is non-fw secure here
session.WaitUntilForwardSecure()
// session is fw-secure here
```

marten-seemann:
Fixes #568, fixes #564, fixes #563.
Closes #566, closes #506 (obsolete now).

### DIFF
--- a/client.go
+++ b/client.go
@@ -3,6 +3,7 @@ package quic
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"net"
 	"strings"
 	"sync"
@@ -17,11 +18,11 @@ type client struct {
 	mutex     sync.Mutex
 	listenErr error
 
-	conn      connection
-	hostname  string
-	errorChan chan struct{}
+	conn     connection
+	hostname string
 
-	handshakeChan chan struct{} // is closed as soon as the handshake completes
+	errorChan     chan struct{}
+	handshakeChan <-chan handshakeEvent
 
 	config            *Config
 	versionNegotiated bool // has version negotiation completed yet
@@ -50,9 +51,9 @@ func DialAddr(addr string, config *Config) (Session, error) {
 	return Dial(udpConn, udpAddr, addr, config)
 }
 
-// Dial establishes a new QUIC connection to a server using a net.PacketConn.
+// DialNonFWSecure establishes a new non-forward-secure QUIC connection to a server using a net.PacketConn.
 // The host parameter is used for SNI.
-func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config) (Session, error) {
+func DialNonFWSecure(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config) (NonFWSession, error) {
 	connID, err := utils.GenerateConnectionID()
 	if err != nil {
 		return nil, err
@@ -65,13 +66,12 @@ func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config
 
 	clientConfig := populateClientConfig(config)
 	c := &client{
-		conn:          &conn{pconn: pconn, currentAddr: remoteAddr},
-		connectionID:  connID,
-		hostname:      hostname,
-		config:        clientConfig,
-		version:       clientConfig.Versions[0],
-		errorChan:     make(chan struct{}),
-		handshakeChan: make(chan struct{}),
+		conn:         &conn{pconn: pconn, currentAddr: remoteAddr},
+		connectionID: connID,
+		hostname:     hostname,
+		config:       clientConfig,
+		version:      clientConfig.Versions[0],
+		errorChan:    make(chan struct{}),
 	}
 
 	err = c.createNewSession(nil)
@@ -81,7 +81,21 @@ func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config
 
 	utils.Infof("Starting new connection to %s (%s), connectionID %x, version %d", hostname, c.conn.RemoteAddr().String(), c.connectionID, c.version)
 
-	return c.establishConnection()
+	return c.session.(NonFWSession), c.establishSecureConnection()
+}
+
+// Dial establishes a new QUIC connection to a server using a net.PacketConn.
+// The host parameter is used for SNI.
+func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config) (Session, error) {
+	sess, err := DialNonFWSecure(pconn, remoteAddr, host, config)
+	if err != nil {
+		return nil, err
+	}
+	err = sess.WaitUntilHandshakeComplete()
+	if err != nil {
+		return nil, err
+	}
+	return sess, nil
 }
 
 func populateClientConfig(config *Config) *Config {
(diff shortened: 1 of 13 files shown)


h2quic/client.go
```diff
@@ -169,16 +138,14 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 
 	hasBody := (req.Body != nil)
 
-	c.mutex.Lock()
-	for c.encryptionLevel != protocol.EncryptionForwardSecure {
-		c.cryptoChangedCond.Wait()
-	}
+	<-c.handshakeChan // wait until the handshake has completed
```
lucas-clemente: We should probably have a way to unblock this if an error occurs between dialing and closing the chan.

session.go
```diff
@@ -83,10 +83,11 @@ type session struct {
...
 	// this channel is passed to the CryptoSetup and receives the current encryption level
 	// it is closed as soon as the handshake is complete
-	aeadChanged        <-chan protocol.EncryptionLevel
-	handshakeComplete  bool
-	handshakeChan      chan struct{} // will be closed as soon as the handshake completes
-	handshakeErrorChan chan error
+	aeadChanged       <-chan protocol.EncryptionLevel
+	handshakeComplete bool
+	// will be closed as soon as the handshake completes, and receive any error that might occur until then
+	// it to block WaitUntilHandshakeComplete()
```
lucas-clemente: Grammar is weird here.

client.go
```diff
@@ -81,7 +81,21 @@ func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config
...
+func Dial(pconn net.PacketConn, remoteAddr net.Addr, host string, config *Config) (Session, error) {
+	sess, err := DialNonFWSecure(pconn, remoteAddr, host, config)
+	if err != nil {
+		return nil, err
+	}
+	err = sess.WaitUntilHandshakeComplete()
+	if err != nil {
+		return nil, err
+	}
+	return sess, nil
```
lucas-clemente: Can be simplified to `return sess, sess.WaitUntilHandshakeComplete()`

session.go
```diff
@@ -159,17 +175,19 @@ func newClientSession(conn connection, hostname string, v protocol.VersionNumber
 
 	aeadChanged := make(chan protocol.EncryptionLevel, 2)
 	s.aeadChanged = aeadChanged
+	handshakeChan := make(chan handshakeEvent, 1)
```
lucas-clemente: Nit: don't really need the var here, can assign to s.handshakeChan directly.

session.go
```diff
@@ -287,6 +311,8 @@ runLoop:
 	// otherwise this chan will already be closed
 	if !s.handshakeComplete {
 		s.handshakeCompleteChan <- closeErr.err
+		s.handshakeChan <- handshakeEvent{err: closeErr.err}
+
```
lucas-clemente: Nit: remove empty line

server.go
```diff
@@ -229,9 +229,21 @@ func (s *server) handlePacket(pconn net.PacketConn, remoteAddr net.Addr, packet
...
 			_ = session.run()
-
 			s.removeConnection(hdr.ConnectionID)
 		}()
+
+		go func() {
+			for {
+				ev := <-handshakeChan
+				if ev.err != nil {
+					return
```
lucas-clemente: Should we log the err here, or is that done someplace else?

session.go
```diff
@@ -95,6 +89,7 @@ type session struct {
 	handshakeCompleteChan chan error
 	// handshakeChan receives handshake events and is closed as soon the handshake completes
 	// the receiving end of this channel is passed to the creator of the session
+	// it receives at most 3 handshake event: 2 when the encryption level changes, and one error
```
lucas-clemente: nit: s/event/events/

session.go
marten-seemann: We do. `session.handshakeChan` is a directed channel.

client.go
marten-seemann: Only if you're willing to hand out the `sess`, even if `WaitUntilHandshakeComplete()` fails. I'd prefer to stick to the pattern that all return values are the respective null values if `err != nil`.

server.go
marten-seemann: This error should get already get logged by session, when it closes.

marten-seemann:
@lucas-clemente: Done.

h2quic/client.go
```diff
@@ -44,57 +46,34 @@ var _ h2quicClient = &Client{}
...
 }
 
 // Dial dials the connection
-func (c *Client) Dial() error {
-	_, err := quic.DialAddr(c.hostname, c.config)
-	return err
-}
+func (c *Client) Dial() (err error) {
+	defer func() {
+		c.mutex.Lock()
```
lucas-clemente: I don't think you actually need the mutex here if you always set the error before closing the chan.

h2quic/client.go
```diff
@@ -170,17 +149,23 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 
 	hasBody := (req.Body != nil)
 
-	c.mutex.Lock()
-	for c.encryptionLevel != protocol.EncryptionForwardSecure {
-		c.cryptoChangedCond.Wait()
+	// wait until the handshake is complete
+	<-c.handshakeChan
+	c.mutex.RLock()
```
lucas-clemente: Same here, I think this Lock is unnecessary.

client.go
lucas-clemente: I wouldn't say it's necessary to return a null value if err != nil, that's what the err is for – but I'm fine with keeping it as-is if you insist.

h2quic/client.go
```diff
@@ -44,57 +46,34 @@ var _ h2quicClient = &Client{}
...
 		t:               t,
+		dialAddr:        quic.DialAddr,
 		hostname:        authorityAddr("https", hostname),
 		responses:       make(map[protocol.StreamID]chan *http.Response),
 		encryptionLevel: protocol.EncryptionUnencrypted,
+		config: &quic.Config{
+			TLSConfig:                     tlsConfig,
+			RequestConnectionIDTruncation: true,
+		},
+		handshakeChan: make(chan struct{}),
```
lucas-clemente: I think the name collision with the handshakeChan in client.go is unfortunate, could you maybe rename this one here?

lucas-clemente:
LGTM modulo my mutex & renaming comments.

Please also update quic-conn at the same time as merging this.

marten-seemann:
Merged. I just noticed we forgot to update the CHANGELOG.
We should probably also have another look at our godoc.

PR #616, opened 2017-05-13
properly handle errors that occur while handling packets in the session

### LINKED ISSUE #614: panic in session
```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x7041fb]
goroutine 7631 [running]:
panic(0x847240, 0xc420012190)
	/home/travis/.gimme/versions/go1.7.5.linux.amd64/src/runtime/panic.go:500 +0x1a1
github.com/lucas-clemente/quic-go/qerr.ToQuicError(0x0, 0x0, 0xc428b0fd10)
	/home/travis/gopath/src/github.com/lucas-clemente/quic-go/qerr/quic_error.go:44 +0xeb
github.com/lucas-clemente/quic-go.(*session).handleCloseError(0xc43b0adba0, 0x0, 0x0, 0x0, 0x0, 0xbc7040)
	/home/travis/gopath/src/github.com/lucas-clemente/quic-go/session.go:562 +0x337
github.com/lucas-clemente/quic-go.(*session).run(0xc43b0adba0, 0xc430316fa0, 0x5dbeb5)
	/home/travis/gopath/src/github.com/lucas-clemente/quic-go/session.go:308 +0x7fd
github.com/lucas-clemente/quic-go.(*client).createNewSession.func1(0xc431f28f00)
	/home/travis/gopath/src/github.com/lucas-clemente/quic-go/client.go:271 +0x3e
created by github.com/lucas-clemente/quic-go.(*client).createNewSession
	/home/travis/gopath/src/github.com/lucas-clemente/quic-go/client.go:283 +0x14b
```

marten-seemann:
fixes #614

### DIFF
--- a/session.go
+++ b/session.go
@@ -265,7 +265,7 @@ runLoop:
 					continue
 				}
 				s.close(err)
-				break runLoop
+				continue
 			}
 			// This is a bit unclean, but works properly, since the packet always
 			// begins with the public header and we never copy it.
--- a/session_test.go
+++ b/session_test.go
@@ -708,6 +708,20 @@ var _ = Describe("Session", func() {
 			Expect(sess.largestRcvdPacketNumber).To(Equal(protocol.PacketNumber(5)))
 		})
 
+		It("closes when handling a packet fails", func(done Done) {
+			testErr := errors.New("unpack error")
+			hdr.PacketNumber = 5
+			var runErr error
+			go func() {
+				runErr = sess.run()
+			}()
+			sess.unpacker.(*mockUnpacker).unpackErr = testErr
+			sess.handlePacket(&receivedPacket{publicHeader: hdr})
+			Eventually(func() error { return runErr }).Should(MatchError(testErr))
+			Expect(sess.runClosed).To(BeClosed())
+			close(done)
+		})
+
 		It("sets the {last,largest}RcvdPacketNumber, for an out-of-order packet", func() {
 			hdr.PacketNumber = 5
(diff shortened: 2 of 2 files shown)


session.go
```diff
@@ -265,7 +265,6 @@ runLoop:
 					continue
 				}
 				s.close(err)
-				break runLoop
```
lucas-clemente: I'm not sure this is the right fix – I'd be more happy with a `continue`. That way, we don't enter the code at the bottom of the run loop (e.g. sending packets). Keep in mind that this error here may be triggered by a peer doing something security-relevant, so I don't think we should do much more work other than sending the close frame.

session_test.go
```diff
@@ -697,7 +697,10 @@ var _ = Describe("Session", func() {
 		BeforeEach(func() {
 			sess.unpacker = &mockUnpacker{}
 			clientSess.unpacker = &mockUnpacker{}
-			hdr = &PublicHeader{PacketNumberLen: protocol.PacketNumberLen6}
+			hdr = &PublicHeader{
+				PacketNumberLen: protocol.PacketNumberLen6,
+				Raw:             getPacketBuffer(),
```
lucas-clemente: If you adopt the continue above, this change is no longer necessary.

session.go
marten-seemann: Good point. Fixed.

PR #619, opened 2017-05-15
log timestamps

### LINKED ISSUE #444: Improve logging
We could think about wrapping os.Stdout in the default logger, or using a log.Logger.

marten-seemann:
Print timestamps in log files.

Fixes #444.

### DIFF
--- a/integrationtests/integrationtests_suite_test.go
+++ b/integrationtests/integrationtests_suite_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"mime/multipart"
 	"net"
 	"net/http"
@@ -78,6 +79,9 @@ func init() {
 }
 
 var _ = BeforeEach(func() {
+	// set custom time format for logs
+	utils.SetLogTimeFormat("15:04:05.000")
+
 	// create a new uploadDir for every test
 	var err error
 	uploadDir, err = ioutil.TempDir("", "quic-upload-dest")
@@ -95,7 +99,7 @@ var _ = BeforeEach(func() {
 	if len(logFileName) > 0 {
 		logFile, err = os.Create("./log.txt")
 		Expect(err).ToNot(HaveOccurred())
-		utils.SetLogWriter(logFile)
+		log.SetOutput(logFile)
 		utils.SetLogLevel(utils.LogLevelDebug)
 	}
 })
--- a/utils/log.go
+++ b/utils/log.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strconv"
 	"sync"
+	"time"
 )
 
 // LogLevel of quic-go
@@ -25,8 +26,9 @@ const (
 )
 
 var (
-	logLevel           = LogLevelNothing
-	out      io.Writer = os.Stdout
+	logLevel             = LogLevelNothing
+	out        io.Writer = os.Stdout
+	timeFormat           = "15:04:05.000"
 
 	mutex sync.Mutex
 )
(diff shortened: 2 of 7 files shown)


marten-seemann:
I'd like to keep a way to set the log level manually. I'm not really worried about the API, since utils is not part of our public API anyway (to make that even clearer, we can think about making it an internal package some time).

lucas-clemente:
I'm fine with keeping utils.SetLogLevel. What do you think about using Go's log package?

marten-seemann:
tbh, I haven't used it so far.
From the docs I'd say that we'd only use `Logger.SetPrefix` (for setting the timestamp prefix), `Logger.SetOutput` (to set the writer) and `Logger.Printf`, right?

lucas-clemente:
Yes. Wrt the prefix or output, it actually has a proper defaults, which we definitely shouldn't change in non-test code. For testing code we can decide about the format, but I'd argue for keeping the default there as well.

marten-seemann:
The `log` package is pretty unflexible. I just tried to use it.
Disabling timestamp works with the self-explanatory command `log.SetFlags(0)`. `log.SetPrefix` is pretty much useless, since it only takes a string, not a callback. And logging with millisecond resolution is impossible. I'd prefer to not use this package.

lucas-clemente:
I don't think it's the job of a QUIC implementation to solve the logging problem. Especially if we ever want to get in there, we don't have a choice here. If this package works for the rest of the world, it should work for us.

Regarding the resolution, have you seen `log.Lmicroseconds`?

marten-seemann:
Well, there is a reason there is a whole bunch of logging libraries around. Logging is definitely our smallest problem regarding standard library compatibility, so I don't see any need to move to an inconvenient solution because of that, at this point.
Since we'll use logging a lot for debugging, I prefer to use our own solution that provides an API that I don't have to look up in the godoc every single time.

Regarding the resolution: This is just a minor point, but yet another example of the bad API of `log`. I'm aware of `log.Lmicrosecond`, but there's no `log.Lmillisecond`, and no way to specify our own format.

marten-seemann:
What we can do is keep the `SetLogTimeFormat` function. Then we can still use the `log` package, but we're completely free to do what we want for our tests.

lucas-clemente:
I think the additional code is not worth saving 3 chars per log line. I'd strongly prefer to use Lmicroseconds and not have any custom code for the format / prefix.

marten-seemann:
I pushed a commit that now uses the standard library log package. I think this is a good compromise fulfilling the logging needs of quic-go users on the one hand, while at the same time providing a more convenient API for internal purposes.

utils/log.go
```diff
@@ -26,16 +25,13 @@ const (
...
-	out        io.Writer = os.Stdout
-	timeFormat           = "15:04:05.000"
-
-	mutex sync.Mutex
+	logLevel   = LogLevelNothing
+	timeFormat = ""
 )
 
 // SetLogWriter sets the log writer.
 func SetLogWriter(w io.Writer) {
```
lucas-clemente: Remove this function, now that it doesn't do anything.

integrationtests/integrationtests_suite_test.go
```diff
@@ -78,6 +78,9 @@ func init() {
 }
 
 var _ = BeforeEach(func() {
+	// set custom time format for logs
+	utils.SetLogTimeFormat("15:04:05.000")
```
lucas-clemente: I still think it's kind of ridiculous have code in a QUIC library just for this, but if you really must…

utils/log.go
```diff
@@ -46,42 +42,36 @@ func SetLogLevel(level LogLevel) {
...
-		mutex.Unlock()
+		logMessage(format, args...)
 	}
 }
 
-func logTimestamp() {
+func logMessage(format string, args ...interface{}) {
 	if len(timeFormat) > 0 {
-		fmt.Fprintf(out, time.Now().Format(timeFormat)+" ")
+		log.Printf(time.Now().Format(timeFormat)+" "+format, args...)
```
lucas-clemente: This is a red flag from a security point of view (never have dynamic format strings). It doesn't really make a difference right now, but please do this in two prints.

marten-seemann: That's not possible. `log.Printf` adds a newline.

marten-seemann:
@lucas-clemente: lgty?

PR #623, opened 2017-05-16
make sure that the SHLO is always sent with secure encryption level

### LINKED ISSUE #544: unencrypted SHLO in random RTT tests
After 60 attempts:
screen shot 2017-04-16 at 18 17 07

marten-seemann:
Fixes #544.

### DIFF
--- a/handshake/crypto_setup_server.go
+++ b/handshake/crypto_setup_server.go
@@ -22,6 +22,8 @@ type KeyExchangeFunction func() crypto.KeyExchange
 
 // The CryptoSetupServer handles all things crypto for the Session
 type cryptoSetupServer struct {
+	mutex sync.RWMutex
+
 	connID               protocol.ConnectionID
 	sourceAddr           []byte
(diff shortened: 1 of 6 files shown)


lucas-clemente:
Note that this makes #611 much more likely and pressing.

handshake/crypto_setup_server.go
```diff
@@ -152,6 +153,9 @@ func (h *cryptoSetupServer) handleMessage(chloData []byte, cryptoData map[Tag][]
 		if err != nil {
 			return false, err
 		}
+		h.mutex.Lock()
+		h.wroteSHLO = true
```
lucas-clemente: Isn't there a race condition here, as the frame may already be packed (and Write() returned) before the sealer is actually requested?

marten-seemann: I don't see how to fix this without putting the `Write` into the mutex as well. And this will deadlock for sure.

PR #626, opened 2017-05-18
add tests measuring how many RTTs a handshake takes

marten-seemann:


marten-seemann:
I don't the idea of taking an average here. We might run into the situation where a handshake that's supposed to take 2 RTTs takes 3 RTTs in 10% of the cases (for whatever reason). This test is designed to catch that kind of bug, and by averaging we'd obscure this.

The better way is to increase the RTT. I'll play around a bit with the parameters, to make sure there's no (additional) flakiness on Travis.

lucas-clemente:
SGTM

PR #627, opened 2017-05-20
add a quic.Config option to verify source address tokens

### LINKED ISSUE #303: add a quic.Config option for STK checks
We currently require a STK on all connections, which requires 2 RTTs for the initial connection. We should think about relaxing this, or making it configurable.

marten-seemann:
Fixes #303.

### DIFF
--- a/crypto/source_address_token.go
+++ b/crypto/source_address_token.go
@@ -58,7 +58,11 @@ const stkKeySize = 16
 const stkNonceSize = 16
 
 // NewStkSource creates a source for source address tokens
-func NewStkSource(secret []byte) (StkSource, error) {
+func NewStkSource() (StkSource, error) {
+	secret := make([]byte, 32)
+	if _, err := rand.Read(secret); err != nil {
+		return nil, err
+	}
 	key, err := deriveKey(secret)
 	if err != nil {
 		return nil, err
--- a/crypto/source_address_token.go
+++ b/crypto/source_address_token.go
@@ -5,7 +5,6 @@ import (
 	"crypto/cipher"
 	"crypto/rand"
 	"crypto/sha256"
-	"crypto/subtle"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -20,20 +19,20 @@ import (
 type StkSource interface {
 	// NewToken creates a new token for a given IP address
 	NewToken(sourceAddress []byte) ([]byte, error)
-	// VerifyToken verifies if a token matches a given IP address
-	VerifyToken(sourceAddress []byte, data []byte) (time.Time, error)
+	// DecodeToken decodes a token and returns the source address and the timestamp
+	DecodeToken(data []byte) ([]byte, time.Time, error)
 }
 
 type sourceAddressToken struct {
-	sourceAddr []byte
+	data []byte
 	// unix timestamp in seconds
 	timestamp uint64
 }
 
 func (t *sourceAddressToken) serialize() []byte {
-	res := make([]byte, 8+len(t.sourceAddr))
+	res := make([]byte, 8+len(t.data))
 	binary.LittleEndian.PutUint64(res, t.timestamp)
-	copy(res[8:], t.sourceAddr)
+	copy(res[8:], t.data)
 	return res
 }
 
@@ -42,8 +41,8 @@ func parseToken(data []byte) (*sourceAddressToken, error) {
 		return nil, fmt.Errorf("invalid STK length: %d", len(data))
 	}
 	return &sourceAddressToken{
-		sourceAddr: data[8:],
-		timestamp:  binary.LittleEndian.Uint64(data),
+		data:      data[8:],
+		timestamp: binary.LittleEndian.Uint64(data),
 	}, nil
 }
 
@@ -78,38 +77,31 @@ func NewStkSource() (StkSource, error) {
 	return &stkSource{aead: aead}, nil
 }
 
-func (s *stkSource) NewToken(sourceAddr []byte) ([]byte, error) {
+func (s *stkSource) NewToken(data []byte) ([]byte, error) {
 	return encryptToken(s.aead, &sourceAddressToken{
-		sourceAddr: sourceAddr,
-		timestamp:  uint64(time.Now().Unix()),
+		data:      data,
+		timestamp: uint64(time.Now().Unix()),
 	})
 }
 
-func (s *stkSource) VerifyToken(sourceAddr []byte, data []byte) (time.Time, error) {
+func (s *stkSource) DecodeToken(data []byte) ([]byte, time.Time, error) {
 	if len(data) < stkNonceSize {
-		return time.Time{}, errors.New("STK too short")
+		return nil, time.Time{}, errors.New("STK too short")
 	}
 	nonce := data[:stkNonceSize]
-
 	res, err := s.aead.Open(nil, nonce, data[stkNonceSize:], nil)
 	if err != nil {
-		return time.Time{}, err
+		return nil, time.Time{}, err
 	}
 
 	token, err := parseToken(res)
 	if err != nil {
-		return time.Time{}, err
+		return nil, time.Time{}, err
 	}
-
-	if subtle.ConstantTimeCompare(token.sourceAddr, sourceAddr) != 1 {
-		return time.Time{}, errors.New("invalid source address in STK")
-	}
-
 	if token.timestamp > math.MaxInt64 {
-		return time.Time{}, errors.New("invalid timestamp")
+		return nil, time.Time{}, errors.New("invalid timestamp")
 	}
-
-	return time.Unix(int64(token.timestamp), 0), nil
+	return token.data, time.Unix(int64(token.timestamp), 0), nil
 }
 
 func deriveKey(secret []byte) ([]byte, error) {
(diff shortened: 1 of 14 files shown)


handshake/stk_generator.go
```diff
@@ -0,0 +1,72 @@
...
+// it ensures that we're binary compatible with Google's implementation of STKs
+func encodeRemoteAddr(remoteAddr net.Addr) []byte {
+	// if the address is a UDP address, just use the byte representation of the IP address
+	// the length of an IP address is 4 bytes (for IPv4) or 16 bytes (for IPv6)
+	if udpAddr, ok := remoteAddr.(*net.UDPAddr); ok {
+		return udpAddr.IP
+	}
+	// if the address is not a UDP address, prepend 16 bytes
+	// that way it can be distinguished from an IP address
+	return append(bytes.Repeat([]byte{0}, 16), []byte(remoteAddr.String())...)
```
lucas-clemente: This seems weird, why not have a different prefix for each option? That should also simplify the decoding.

interface.go
```diff
@@ -45,6 +46,18 @@ type NonFWSession interface {
 	WaitUntilHandshakeComplete() error
 }
 
+// An STK is a Source Address token.
+// It is issued by the server and sent to the client. For the client, it is an opaque blob.
+// The client can send the STK in subsequent handshakes to prove ownership of his IP address.
```
lucas-clemente: s/his/{their,its}/

interface.go
```diff
@@ -54,9 +67,14 @@ type Config struct {
 	// Warning: This API should not be considered stable and will change soon.
 	Versions []protocol.VersionNumber
 	// Ask the server to truncate the connection ID sent in the Public Header.
+	// If not set, the default checks if
 	// This saves 8 bytes in the Public Header in every packet. However, if the IP address of the server changes, the connection cannot be migrated.
 	// Currently only valid for the client.
 	RequestConnectionIDTruncation bool
+	// If true is returned, a source address validation is performed
```
lucas-clemente: It's uncommon to have a verify function return true if stuff is wrong (verbs are usually imperative in function names). Maybe rename to AcceptSTK, IsValidSTK, CheckSTK and invert the return value?

server.go
```diff
@@ -85,15 +85,36 @@ func Listen(conn net.PacketConn, config *Config) (Listener, error) {
...
+	if time.Now().After(stk.sentTime.Add(protocol.STKExpiryTime)) {
+		return true
+	}
+	var sourceAddr string
+	if udpAddr, ok := clientAddr.(*net.UDPAddr); ok {
+		sourceAddr = udpAddr.IP.String()
+	} else {
+		sourceAddr = clientAddr.String()
+	}
+	return sourceAddr != stk.remoteAddr
```
lucas-clemente: This might be pedantic, but I'd prefer a constant time comparison here.

session.go
```diff
@@ -137,25 +137,29 @@ func newSession(
...
 	s.aeadChanged = aeadChanged
 	handshakeChan := make(chan handshakeEvent, 3)
 	s.handshakeChan = handshakeChan
+	verifySourceAddr := func(clientAddr net.Addr, hstk *handshake.STK) bool {
+		if hstk == nil {
+			return config.VerifySourceAddress(clientAddr, nil)
+		}
+		return config.VerifySourceAddress(
+			clientAddr,
+			&STK{remoteAddr: hstk.RemoteAddr, sentTime: hstk.SentTime},
```
lucas-clemente: This should become unnecessary with 1.9's type aliases :)

crypto/source_address_token.go
```diff
@@ -58,7 +58,11 @@ const stkKeySize = 16
 const stkNonceSize = 16
 
 // NewStkSource creates a source for source address tokens
-func NewStkSource(secret []byte) (StkSource, error) {
+func NewStkSource() (StkSource, error) {
```
marten-seemann: The STK doesn't have to do anything with the server config. In fact, even when the server config changes, we still want to be able to read STKs issues before that change.

handshake/stk_generator.go
marten-seemann: This way, we're compatible with Google's STK format.

interface.go
marten-seemann: Yes! I like that.

server.go
marten-seemann: Which attack do we mitigate by doing so?

session.go
marten-seemann: Oh, finally. So we can remove this when dropping support for 1.8, which means in Feb 2018...

handshake/stk_generator.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// An STK is a source address token
+type STK struct {
+	RemoteAddr string
+	SentTime   time.Time
+}
+
+// An STKGenerator generates STKs
+type STKGenerator struct {
```
marten-seemann: The one in the handshake package deals with the net.Addr of the client and decides how to encode this. The one in the crypto package encodes arbitrary byte slices into an STK.

crypto/source_address_token.go
lucas-clemente: True, but I still think it makes sense to keep (and probably even refresh) them together. Both are secrets that will have to live across server restarts, and be shared by multiple servers in a load-balancing setup. FWIW, gQUIC also keeps them together.

handshake/stk_generator.go
lucas-clemente: Well, the crypto struct decodes the timestamp. Maybe move that to here, so that the struct in the crypto package only does (un)boxing?

handshake/stk_generator.go
lucas-clemente: We're not, as gQUIC uses protobufs for the STKs. I also recall that we decided to drop that idea. I really don't like having a zero prefix, that just seems needlessly specific and inextensible.

server.go
lucas-clemente: I didn't have anything specific in mind, but it's a good idea to always use constant time comparisons for secret values, as there's practically no disadvantage.

marten-seemann: Don't we have to do that for the timestamp as well?

lucas-clemente: Just realized that this data is authenticated, so it doesn't matter.

handshake/stk_generator.go
marten-seemann: I don't really like the idea of returning the timestamp as a byte-slice from the `crypto` package.

lucas-clemente: That's not what I meant. I think the stuff in the crypto package should just treat the token as a byte slice that is boxed / unboxed. The timestamp can then be (de)serialized in the handshake packet. This is how the our crypto/handshake package split usually works.

For simplicity we could then even use Go's binary encoding of the struct, so we don't even need parsing logic ourselves (which is always a win).

marten-seemann: I don't think we can use Go's binary encoding, because the data in the source address has variable length. I tried it in the playground: https://play.golang.org/p/3j_VzfKxa-.

lucas-clemente: encoding/binary is only for numbers, I meant encoding/gob.

https://play.golang.org/p/ZDWvBLS6pw

marten-seemann: We could use [asn1 encoding](https://golang.org/pkg/encoding/asn1/) here: https://play.golang.org/p/ToqhV7dBEC

lucas-clemente: Though looking at gob in more detail that looks extremely wasteful, as it always encodes the type. I was hoping it was more like protobufs…

crypto/source_address_token.go
marten-seemann: We agreed to keep them separate OOB, especially sind the server configs are not going to stay when we switch to TLS 1.3.

interface.go
marten-seemann: Renamed to `AcceptSTK`.

marten-seemann:
@lucas-clemente I think this can be approved now.

handshake/stk_generator.go
```diff
@@ -52,21 +56,17 @@ func (g *STKGenerator) DecodeToken(data []byte) (*STK, error) {
...
-	// that way it can be distinguished from an IP address
-	return append(bytes.Repeat([]byte{0}, 16), []byte(remoteAddr.String())...)
+	return append([]byte{stkPrefixString}, []byte(remoteAddr.String())...)
 }
 
+// decodeRemoteAddr decodes the remote address saved in the STK
 func decodeRemoteAddr(data []byte) string {
-	if len(data) <= 16 {
-		return net.IP(data).String()
+	if data[0] == stkPrefixIP {
```
lucas-clemente: I think we should check the length here to be on the safe side. Returning "" should be fine in that case.

handshake/stk_generator.go
```diff
@@ -0,0 +1,96 @@
...
+func encodeRemoteAddr(remoteAddr net.Addr) []byte {
+	if udpAddr, ok := remoteAddr.(*net.UDPAddr); ok {
+		return append([]byte{stkPrefixIP}, udpAddr.IP...)
+	}
+	return append([]byte{stkPrefixString}, []byte(remoteAddr.String())...)
+}
+
+// decodeRemoteAddr decodes the remote address saved in the STK
+func decodeRemoteAddr(data []byte) string {
+	if data[0] == stkPrefixIP {
```
lucas-clemente: Commenting again because GH: Check the length here before accessing the slice.

PR #631, opened 2017-06-03
simplify the dialing logic in the h2quic client, fix mutex in roundtripper held to long

### LINKED ISSUE #620: mutex held to long in h2quic.QuicRoundTripper
in `QuicRoundTripper.getClient` the mutex is locked while a connection is dialed.

marten-seemann:
Fixes #620.

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -20,8 +20,8 @@ import (
 	"github.com/lucas-clemente/quic-go/utils"
 )
 
-// Client is a HTTP2 client doing QUIC requests
-type Client struct {
+// client is a HTTP2 client doing QUIC requests
+type client struct {
 	mutex sync.RWMutex
 
 	dialAddr func(hostname string, config *quic.Config) (quic.Session, error)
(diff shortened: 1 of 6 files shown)


h2quic/client.go
```diff
@@ -137,8 +132,8 @@ func (c *client) handleHeaderStream() {
 	c.mutex.Unlock()
 }
 
-// Do executes a request and returns a response
-func (c *client) Do(req *http.Request) (*http.Response, error) {
+// Roundtrip executes a request and returns a response
+func (c *client) RoundTrip(req *http.Request) (*http.Response, error) {
 	// TODO: add port to address, if it doesn't have one
```
lucas-clemente: Is this TODO still relevant? If so, do we have an issue?

h2quic/client.go
```diff
@@ -147,14 +142,16 @@ func (c *client) Do(req *http.Request) (*http.Response, error) {
 		return nil, fmt.Errorf("h2quic Client BUG: RoundTrip called for the wrong client (expected %s, got %s)", c.hostname, req.Host)
 	}
 
-	hasBody := (req.Body != nil)
+	c.dialOnce.Do(func() {
+		c.handshakeErr = c.dial()
```
lucas-clemente: Can we make handshakeErr a local variable here?

marten-seemann: No, because the Do is only executed once, but the error has to be returned from all calls to Roundtrip.

lucas-clemente: Right – Go doesn't have static vars. Sorry :)

Issue #555, opened 2017-04-19
Drop test flaky


lucas-clemente:
```
[BeforeEach] [Top Level]
  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/integrationtests_suite_test.go:83
[BeforeEach] Drop Proxy
  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/drop_test.go:22
[It] gets a file when many outgoing packets are dropped
  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/drop_test.go:68
Connected to 127.0.0.1:12345
Request:
headers:
{
  :method GET
  :scheme https
  :authority quic.clemente.io
  :path /data
}
body: 
Response:
headers: 
{
  :status 200
}
body: 
trailers: {}
Request succeeded (200).
[AfterEach] Drop Proxy
  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/drop_test.go:52
[AfterEach] [Top Level]
  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/integrationtests_suite_test.go:94
• Failure [4.412 seconds]
Drop Proxy with quic version 35 dropping every 4th packet after the crypto handshake [It] gets a file when many outgoing packets are dropped 
/home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/drop_test.go:68
  Expected error:
      <*qerr.QuicError | 0x20c22a90>: {
          ErrorCode: 25,
          ErrorMessage: "No recent network activity.",
      }
      NetworkIdleTimeout: No recent network activity.
  not to have occurred
  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/integrationtests_suite_test.go:110
  Full Stack Trace
  	/home/travis/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:69 +0x164
  github.com/onsi/gomega/internal/assertion.(*Assertion).NotTo(0x20c08500, 0x85f33c0, 0x865fa24, 0x0, 0x0, 0x0, 0x20c08500)
  	/home/travis/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:43 +0x85
  github.com/lucas-clemente/quic-go/integrationtests.setupHTTPHandlers.func2(0x85f31e0, 0x20a6ffb0, 0x18968880)
[...]

PR #637, opened 2017-06-06
Always send retransmissions, irrespective of congestion state

lucas-clemente:
This seems to work around part of #555, but I'm still seeing a similar failure. I propose to merge this (temporary) workaround now anyway, so that can collect logs for the other issue(s).

### DIFF
--- a/ackhandler/sent_packet_handler.go
+++ b/ackhandler/sent_packet_handler.go
@@ -333,7 +333,11 @@ func (h *sentPacketHandler) SendingAllowed() bool {
 			h.bytesInFlight,
 			h.congestion.GetCongestionWindow())
 	}
-	return !(congestionLimited || maxTrackedLimited)
+	// Workaround for #555:
+	// Always allow sending of retransmissions. This should probably be limited
+	// to RTOs, but we currently don't have a nice way of distinguishing them.
+	haveRetransmissions := len(h.retransmissionQueue) > 0
+	return !maxTrackedLimited && (!congestionLimited || haveRetransmissions)
 }
 
 func (h *sentPacketHandler) retransmitOldestTwoPackets() {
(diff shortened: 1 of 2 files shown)


ackhandler/sent_packet_handler.go
```diff
@@ -333,7 +333,11 @@ func (h *sentPacketHandler) SendingAllowed() bool {
 			h.bytesInFlight,
 			h.congestion.GetCongestionWindow())
 	}
-	return !(congestionLimited || maxTrackedLimited)
+	// Workaround for #555:
+	// Always allow sending of retransmissions. This should probably be limited
+	// to RTOs, but we currently don't have a nice way of distinguishing them.
```
marten-seemann: That shouldn't be that hard to build, right?

lucas-clemente: Right now I don't know the correct, full solution here. I think this is an appropriate workaround without many downsides for part of the issue so that we can continue working on finding other causes.

I don't want to spend time adding more complexity that we probably won't need once we kill the workaround.

marten-seemann: Why is this only a workaround? I thought this is (part of) the solution, to always send out RTOs, no matter what the congestion controller says.

lucas-clemente: We can allow retransmission of (1) the oldest two outstanding packets (per RTO), (2) of all outstanding packets, (3) of the oldest two outstanding packets due to RTO (per RTO), or (4) of all outstanding packets due to RTO. This CL implements (2), but I don't think it's the right solution.

If you don't like merging this I'm fine with waiting, but I thought it'd be nice to have a partial fix out asap, as people are waiting for it (#555).

marten-seemann: Doesn't this completely defeat the purpose of congestion control? We detect congestion by observing loss events. If we now just blast out all retransmissions for lost packets, congestion control won't have any effect at all.

lucas-clemente: Well, it does in a way. RTXs still count towards the bytesInFlight, so subsequent "normal" TXs will be delayed accordingly. But: I'm trying to fix incorrect behavior here, and I'd be happy taking the perf hit for now.

But again, I'm fine just killing this PR until we have a proper fix (which might take a few weeks).

PR #645, opened 2017-06-07
fix stalls when retransmitting connection-level Window Update Frames

qdeconinck:
When a Window Update Frame with streamID 0 is lost, we want to know
the receive window when retransmitting the frame. However, the connection
flow controller is not in streamFlowController, leading to map access errors
and then connections stalls since the sender is stuck by the not-updated
receive window it previously received.


qdeconinck:
I signed it!

qdeconinck:
Added a unit test to get receive window at connection level :-)

Issue #633, opened 2017-06-05
Wiki logging example not up to date


twdkeule:
The example on the wiki page does not use the standard library logging module, but the old one from quic-go/utils.

 ```
@@ -20,7 +20,7 @@
 	panic(err)
 }
 
-log.SetLogWriter(f)
+utils.SetLogWriter(f)
 utils.SetLogLevel(utils.LogLevelDebug)

@@ -33,6 +33,6 @@
 }
 writer := bufio.NewWriter(f)
 
-log.SetLogWriter(writer)
+utils.SetLogWriter(writer)
 utils.SetLogLevel(utils.LogLevelDebug)
 ```

lucas-clemente:
Fixed. Also remove utils.SetLogLevel and replaced it with a hint to the env.

lucas-clemente:
Also updated https://github.com/mholt/caddy/wiki/QUIC. Thanks for the report! :)

twdkeule:
I can't get the logging to work with Caddy.
I used
```
log.SetOutput(f)
utils.SetLogLevel(utils.LogLevelDebug)
```
in caddy/caddy/main.go on both the file and a Buffered Writer, but nothing happens.
I also tried setting the environment variable, but to no avail.

lucas-clemente:
Can you check that you have the latest version of quic-go? I think Caddy now vendors deps.

twdkeule:
Ok, I removed lucas-clemente/ from caddy/vendor/github.com (and form vendor/manifest) and rebased my branch of quic-go on origin/master but this changes nothing. The logfile is still empty.

`go install -v github.com/mholt/caddy/caddy` lists `github.com/lucas-clemente/quic-go` so I assume it is using the correct repo.

Btw, is removing the files from caddy/vendor the correct method to unvendor?

lucas-clemente:
That is unexpected. Are you seeing logs on stdout if you remove the SetOutput() call?

I haven't used vendoring so far, but that seems like a safe way to unvendor ;)

twdkeule:
Nothing.
However if I add `utils.Debugf()` inside the caddy/caddy/main.go it does print debug messages:
```
Log level is debug: true
2017/06/06 10:56:59 DEBUG
2017/06/06 10:56:59 INFO
```
But nothing from quic-go

lucas-clemente:
[...]

PR #647, opened 2017-06-07
better log levels

marten-seemann:
Make the log level numbers more intuitive. Logging nothing is now log level 0, debug logging is level 3.

We'll need to change the wiki again ;)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,4 +6,5 @@
 - Add a `quic.Config` option to request truncation of the connection ID from a server
 - Add a `quic.Config` option to configure the source address validation
 - Add a `quic.Config` option to configure the handshake timeout
+- Changed the log level environment variable to only accept strings ("DEBUG", "INFO", "ERROR"), see [the wiki](https://github.com/lucas-clemente/quic-go/wiki/Logging) for more details.
 - Various bugfixes
--- a/internal/utils/log.go
+++ b/internal/utils/log.go
@@ -13,14 +13,14 @@ type LogLevel uint8
 const (
 	logEnv = "QUIC_GO_LOG_LEVEL"
 
-	// LogLevelDebug enables debug logs (e.g. packet contents)
-	LogLevelDebug LogLevel = iota
-	// LogLevelInfo enables info logs (e.g. packets)
-	LogLevelInfo
-	// LogLevelError enables err logs
-	LogLevelError
 	// LogLevelNothing disables
-	LogLevelNothing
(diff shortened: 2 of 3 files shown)


lucas-clemente:
This is a breaking change, so I'd like to get a bit more reasoning before we do this. I agree that probably our initial choice was weird, but is there a clear advantage of reversing the ordering at this time?

Second, maybe exposing numeric log levels as part of the API are a bad idea in general? What do you think about `QUIC_GO_LOG_LEVEL=DEBUG` instead?

Also, this should be in the changelog.

marten-seemann:
@lucas-clemente: Please have another look.

utils/log.go
```diff
@@ -79,13 +78,12 @@ func init() {
 }
 
 func readLoggingEnv() {
-	env := os.Getenv(logEnv)
-	if env == "" {
-		return
+	switch os.Getenv(logEnv) {
```
lucas-clemente: I'd like to be backwards compatible for the time being (and emit a noticable warning to switch to the new strings). So, please add parsing with the old numeric log levels.

utils/log.go
```diff
@@ -14,13 +13,13 @@ const (
 	logEnv = "QUIC_GO_LOG_LEVEL"
```
lucas-clemente: I'd prefer to have nothing = 0, so move this to a separate const block.

marten-seemann: Fixed.

utils/log.go
marten-seemann: Backward compatibility shouldn't be necessary after we merge #648.

marten-seemann:
@lucas-clemente: I just added a Changelog entry. Please approve this PR, then I'll merge it once #648 is merged.

lucas-clemente:
Thanks for adding the line, but I don't think it's particularly helpful like this (what is the change?!).  What about:

```
Changed the log level environment variable to only accept strings ("DEBUG", "INFO", "ERROR"), see [the wiki](https://github.com/lucas-clemente/quic-go/wiki/Logging) for more details.
```

lucas-clemente:
Otherwise LGTM.

lucas-clemente:
Oh, and please update our and caddy's wiki as soon as this is merged, and close #633 :)

twdkeule:
This broke my method to add a prefix to the log 
I will write a PR to add a `QUIC_GO_LOG_PREFIX` ENV if you don't have any objections

lucas-clemente:
Sounds good to me.

PR #648, opened 2017-06-07
make utils an internal package

marten-seemann:
This intentionally breaks the build for everyone who is using the old logging API. This makes sense because we force people to stop using a deprecated API.


marten-seemann:
Apparently, @codecov-io doesn't include the internal package.

PR #454, opened 2017-02-27
accept overlapping stream data

marten-seemann:
Fixes #408 

Since we're still using the linked list, we should reopen #300.

PR #652, opened 2017-06-08
accept duplicate and delayed packets

marten-seemann:
We used to reject duplicate and packets with packet numbers lower than the LeastUnacked we received in a STOP_WAITING frame, because we didn't accept overlapping stream data. For all other frames, duplicates never were an issue. Now that we accept overlapping stream data (#454), there's no need to reject those packets, in fact, processing a delayed packet will be beneficial for performance.

This will also be really nice for implementing QUIC 38, since we lose the explicit lower boundary on which packets to accept, that was provided by the STOP_WAITING.

### DIFF
--- a/ackhandler/received_packet_handler.go
+++ b/ackhandler/received_packet_handler.go
@@ -8,13 +8,6 @@ import (
 	"github.com/lucas-clemente/quic-go/protocol"
 )
 
-var (
-	// ErrDuplicatePacket occurres when a duplicate packet is received
-	ErrDuplicatePacket = errors.New("ReceivedPacketHandler: Duplicate Packet")
-	// ErrPacketSmallerThanLastStopWaiting occurs when a packet arrives with a packet number smaller than the largest LeastUnacked of a StopWaitingFrame. If this error occurs, the packet should be ignored
-	ErrPacketSmallerThanLastStopWaiting = errors.New("ReceivedPacketHandler: Packet number smaller than highest StopWaiting")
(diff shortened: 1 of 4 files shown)


ackhandler/received_packet_handler.go
```diff
@@ -52,19 +45,10 @@ func (h *receivedPacketHandler) ReceivedPacket(packetNumber protocol.PacketNumbe
...
-	}
-
-	if h.packetHistory.IsDuplicate(packetNumber) {
-		return ErrDuplicatePacket
-	}
-
-	err := h.packetHistory.ReceivedPacket(packetNumber)
-	if err != nil {
-		return err
+	if packetNumber > h.ignorePacketsBelow {
```
lucas-clemente: Do we still need this, or should we already kill this check?

marten-seemann: Yeah, otherwise we'll insert it again into the packet history and generate another ACK for this packet.

lucas-clemente: Then this will be an issue for v38, won't it?

marten-seemann: I don't think so. In v37, this value is set explicitly from the STOP_WAITING, in v38 we derive it by tracking the ACK for an ACK. The important thing is that there actually is a lower bound for the packet history, which both mechanisms ensure.

lucas-clemente:
sgtm

PR #655, opened 2017-06-08
Generate some mocks using gomock

lucas-clemente:
For now the flowcontrol manager and the connections parameter manager.


marten-seemann:
In the future, we can also generate some public mocks for the public interfaces.

PR #658, opened 2017-06-09
Fix unencrypted SHLO

lucas-clemente:


session.go
```diff
@@ -219,7 +219,8 @@ func (s *session) setup(
 		return nil, nil, err
 	}
 
-	s.packer = newPacketPacker(s.connectionID, s.cryptoSetup, s.connectionParameters, s.streamFramer, s.perspective, s.version)
+	s.packer = newPacketPacker(s.connectionID, s.cryptoSetup, s.connectionParameters, s.streamFramer,
+		s.perspective, s.version)
```
marten-seemann: Put this on 1 line or on 6 lines.

packet_packer.go
```diff
@@ -187,27 +173,24 @@ func (p *packetPacker) packPacket(stopWaitingFrame *frames.StopWaitingFrame, lea
...
-		number:          currentPacketNumber,
+		number:          publicHeader.PacketNumber,
 		raw:             raw,
 		frames:          payloadFrames,
 		encryptionLevel: encLevel,
 	}, nil
 }
 
-func (p *packetPacker) composeNextPacket(stopWaitingFrame *frames.StopWaitingFrame, maxFrameSize protocol.ByteCount) ([]frames.Frame, error) {
+func (p *packetPacker) composeNextPacket(stopWaitingFrame *frames.StopWaitingFrame,
```
marten-seemann: 1 line or 3 lines.

lucas-clemente: See above.

session.go
lucas-clemente: I'd like to keep to about 100 chars per line, so a single line is out. I find this more readable than putting each parameter on its own line (as long as it fits into two lines with this format). If you don't feel strongly about it, I'd prefer to keep it this way.

FWIW, I see this format often used in other languages (C++ / Java).

marten-seemann: "Don't worry about overflowing a punched card."

I strongly prefer to keep a consistent style across the project, which means either 1 line or 6 lines. I don't think C++ is a good reference point when talking about readability of code anyway.

lucas-clemente: Done.

PR #662, opened 2017-06-09
Fix flaky session tests

lucas-clemente:


session_test.go
```diff
@@ -1179,7 +1179,7 @@ var _ = Describe("Session", func() {
 			sess.ackAlarmChanged(time.Now().Add(10 * time.Millisecond))
 			time.Sleep(10 * time.Millisecond)
 			Eventually(func() int { return len(mconn.written) }).ShouldNot(BeZero())
-			Expect(mconn.written[0]).To(ContainSubstring(string([]byte{0x37, 0x13})))
+			Eventually(func() []byte { return mconn.written[0] }).Should(ContainSubstring("foobar"))
```
marten-seemann: I don't really understand what we're testing here, and why it is valid to not test for the ACK anymore. Can we find a more meaningful name for the test?

session_test.go
```diff
@@ -1198,8 +1198,8 @@ var _ = Describe("Session", func() {
 				defer sess.Close(nil)
 
 				Eventually(func() [][]byte { return mconn.written }).Should(HaveLen(1))
-				Expect(mconn.written[0]).To(ContainSubstring("foobar1"))
-				Expect(mconn.written[0]).To(ContainSubstring("foobar2"))
+				Eventually(func() []byte { return mconn.written[0] }).Should(ContainSubstring("foobar"))
+				Eventually(func() []byte { return mconn.written[0] }).Should(ContainSubstring("foobar"))
```
marten-seemann: You should test for foobar1 and foobar2 here.

session_test.go
lucas-clemente: Copy & paste error, fix pushed.

session_test.go
lucas-clemente: Same as above.

lucas-clemente:
The two commits fix issues with different things (a sessino member vs the mock conn), so I'd like to keep them separate.

marten-seemann:
Oh, I didn't see that. Why does Github hide the second line of the commit message here?

lucas-clemente:
It's github. They don't do UX.

lucas-clemente:
This seems to make matters worse the more races I fix. I don't think this approach here will get us anywhere soon, and I don't have the time right now to massively rewrite the tests.

PR #664, opened 2017-06-12
[WIP] Make max incoming streams configurable

### LINKED ISSUE #523: Make max incoming streams configurable
The max number of incoming streams is currently fixed in the server parameters, we should make it configurable via [quic.Config](https://github.com/lucas-clemente/quic-go/blob/master/interface.go#L148).

twdkeule:
Fix #523


twdkeule:
I ran tests with `go test` (in the root folder), but that does not run all tests?
Should tests always be run with `$GOPATH/bin/ginkgo -r --cover --randomizeAllSpecs --randomizeSuites --trace --progress --skipPackage integrationtests --skipMeasurements` or `ginkgo -r --randomizeAllSpecs --randomizeSuites --trace --progress integrationtests`?

twdkeule:
I still have 1 test failing:
```
• Failure [0.001 seconds]
ConnectionsParameterManager max streams per connection incoming connections [It] always uses the constant value, no matter what the client sent 
/home/thomas/Programming/golang/src/github.com/lucas-clemente/quic-go/handshake/connection_parameters_manager_test.go:279
  Expected
      <uint32>: 84
  to be >
      <int>: 100
```

However I don't see why.
the result of `GetMaxIncomingStreams` used to be based on `protocol.MaxIncomingDynamicStreamsPerConnection` so I changed it to the configured value (`h.configuredMaxIncomingDynamicStreamsPerConnection`), but the test compares the result to `protocol.MaxStreamsPerConnection`

lucas-clemente:
First of all, thanks for the PR! You can run the tests with `ginkgo -r -skipMeasurements -skipPackage integrationtests .` or watch with `ginkgo watch -r -skipMeasurements -skipPackage integrationtests .`.

I'll try to review and look into the failing test tomorrow (don't have the time right now, sorry).

handshake/connection_parameters_manager.go
```diff
@@ -61,14 +62,19 @@ var (
 )
 
 // NewConnectionParamatersManager creates a new connection parameters manager
-func NewConnectionParamatersManager(pers protocol.Perspective, v protocol.VersionNumber) ConnectionParametersManager {
+func NewConnectionParamatersManager(pers protocol.Perspective, v protocol.VersionNumber, maxIncomingDynamicStreamsPerConnection uint32) ConnectionParametersManager {
+	if maxIncomingDynamicStreamsPerConnection == 0 {
```
lucas-clemente: I'd prefer to move this logic to the populate config functions (similar to your other PR), so that the config has valid values.

handshake/crypto_setup_server.go
```diff
@@ -47,6 +47,7 @@ type cryptoSetupServer struct {
 
 	cryptoStream io.ReadWriter
 
+	params               *TransportParameters
```
lucas-clemente: Remove this line.

handshake/connection_parameters_manager_test.go
```diff
@@ -290,7 +290,7 @@ var _ = Describe("ConnectionsParameterManager", func() {
 					TagMIDS: {3, 0, 0, 0},
 				})
 				Expect(err).ToNot(HaveOccurred())
-				Expect(cpm.GetMaxIncomingStreams()).To(BeNumerically(">", protocol.MaxStreamsPerConnection))
+				Expect(cpm.GetMaxIncomingStreams()).To(BeNumerically(">", maxStreamsInTest))
```
twdkeule: I fixed this based on the code it is supposed to check. If that was incorrect, this might be too

twdkeule:
This should not be that hard to merge right?

lucas-clemente:
I don't think so – do you want to rebase it or should I take a look?

twdkeule:
No, I meant Git should be able to auto-merge this.
I'm already rebasing :)

Changelog.md
```diff
@@ -3,6 +3,7 @@
 ## v0.6.0 (unreleased)
 
 - Added `quic.Config` options for maximal flow control windows
+- Add a `quic.Config` option for the maximal amount of incoming dynamic streams
```
marten-seemann: Nit: maximum number

handshake/connection_parameters_manager.go
```diff
@@ -242,7 +246,7 @@ func (h *connectionParametersManager) GetMaxIncomingStreams() uint32 {
 	h.mutex.RLock()
 	defer h.mutex.RUnlock()
 
-	maxStreams := protocol.MaxIncomingDynamicStreamsPerConnection
+	maxStreams := h.configuredMaxIncomingDynamicStreamsPerConnection
```
marten-seemann: Don't we have to use the negotiated value here?

twdkeule: That does seem logical.
I'm also wondering whether [this](https://github.com/lucas-clemente/quic-go/blob/master/handshake/connection_parameters_manager.go#L84) and [this](https://github.com/lucas-clemente/quic-go/blob/master/handshake/connection_parameters_manager.go#L88) should not be based on protocol.MaxStreamsPerConnection but on configuredMaxIncomingDynamicStreamsPerConnection

lucas-clemente:
Something is weird with the commits here – could you please rebase onto master and check that only your own commits show up here? :)

twdkeule:
I'm slightly confused now. Can one of you explain the difference between `maxStreamsPerConnection` and `maxIncomingDynamicStreamsPerConnection`?
Because sometimes 'protocol.maxStreamsPerConnection' is used to set `h.maxIncomingDynamicStreamsPerConnection` and both `GetMaxOutgoingStreams()` and `GetMaxIncomingStreams()` are based on `maxIncomingDynamicStreamsPerConnection`, making me wondering what `maxStreamsPerConnection` is actually used for.

lucas-clemente:
I'm not totally up-to-date with the current state of our implementation in that regard, @marten-seemann worked on that last. Stream negotiation used to happen for both endpoints at once in an earlier QUIC version, that's where MSPC (maxStreams) comes from. As we dropped support for Q034, I don't think that's needed anymore. Maybe we can just kill it?

marten-seemann:
That was a long time ago. Actually, I don't really remember how that works, and unfortunately there's basically no documentation around about that. So what one needs to do here is look in the Chromium code, or run some tests with Chromium, and see which tags it sends in the CHLO, and if it abides by the negotiated values.

twdkeule:
I already tried to remove maxStreams completely but it broke the integrations test suite, so the solution is  definitively more complicated :(
I'll look into it when have some time again

Issue #526, opened 2017-04-11
Make max flow control windows configurable


lucas-clemente:
The flow control windows are currently fixed in https://github.com/lucas-clemente/quic-go/blob/master/protocol/server_parameters.go#L42 and below, we should make them configurable via [quic.Config](https://github.com/lucas-clemente/quic-go/blob/master/interface.go#L60).

See also https://github.com/golang/net/commit/906cda9512f77671ab44f8c8563b13a8e707b230.

marten-seemann:
Fixed by #666.

PR #666, opened 2017-06-13
Make max flow control windows configurable

twdkeule:
#526 

I'm not entirely satisfied with the new signature of NewConnectionParamatersManager. It's too long, certainly if `maxIncomingDynamicStreamsPerConnection` gets added too.
```
func NewConnectionParamatersManager(
	pers protocol.Perspective, v protocol.VersionNumber,
	maxReceiveStreamFlowControlWindowServer protocol.ByteCount, maxReceiveConnectionFlowControlWindowServer protocol.ByteCount,
	maxReceiveStreamFlowControlWindowClient protocol.ByteCount, maxReceiveConnectionFlowControlWindowClient protocol.ByteCount,
) ConnectionParametersManager
```

### DIFF
--- a/handshake/connection_parameters_manager.go
+++ b/handshake/connection_parameters_manager.go
@@ -5,9 +5,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/lucas-clemente/quic-go/protocol"
 	"github.com/lucas-clemente/quic-go/qerr"
-	"github.com/lucas-clemente/quic-go/internal/utils"
 )
 
 // ConnectionParametersManager negotiates and stores the connection parameters
@@ -50,6 +50,11 @@ type connectionParametersManager struct {
 	sendConnectionFlowControlWindow        protocol.ByteCount
 	receiveStreamFlowControlWindow         protocol.ByteCount
 	receiveConnectionFlowControlWindow     protocol.ByteCount
+
+	maxReceiveStreamFlowControlWindowServer     protocol.ByteCount
+	maxReceiveConnectionFlowControlWindowServer protocol.ByteCount
+	maxReceiveStreamFlowControlWindowClient     protocol.ByteCount
+	maxReceiveConnectionFlowControlWindowClient protocol.ByteCount
 }
 
 var _ ConnectionParametersManager = &connectionParametersManager{}
@@ -61,7 +66,24 @@ var (
 )
 
 // NewConnectionParamatersManager creates a new connection parameters manager
-func NewConnectionParamatersManager(pers protocol.Perspective, v protocol.VersionNumber) ConnectionParametersManager {
+func NewConnectionParamatersManager(
+	pers protocol.Perspective, v protocol.VersionNumber,
+	maxReceiveStreamFlowControlWindowServer protocol.ByteCount, maxReceiveConnectionFlowControlWindowServer protocol.ByteCount,
+	maxReceiveStreamFlowControlWindowClient protocol.ByteCount, maxReceiveConnectionFlowControlWindowClient protocol.ByteCount,
+) ConnectionParametersManager {
+	if maxReceiveStreamFlowControlWindowServer == 0 {
+		maxReceiveStreamFlowControlWindowServer = protocol.DefaultMaxReceiveStreamFlowControlWindowServer
+	}
+	if maxReceiveConnectionFlowControlWindowServer == 0 {
+		maxReceiveConnectionFlowControlWindowServer = protocol.DefaultMaxReceiveConnectionFlowControlWindowServer
+	}
+	if maxReceiveStreamFlowControlWindowClient == 0 {
+		maxReceiveStreamFlowControlWindowClient = protocol.DefaultMaxReceiveStreamFlowControlWindowClient
+	}
+	if maxReceiveConnectionFlowControlWindowClient == 0 {
+		maxReceiveConnectionFlowControlWindowClient = protocol.DefaultMaxReceiveConnectionFlowControlWindowClient
+	}
+
 	h := &connectionParametersManager{
 		perspective:                        pers,
 		version:                            v,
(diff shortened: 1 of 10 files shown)


interface.go
```diff
@@ -80,6 +80,18 @@ type Config struct {
...
+	// MaxReceiveConnectionFlowControlWindowServer is the connection-level flow control window for receiving data, for the server
+	// If this value is zero, the timeout is set to protocol.DefaultMaxReceiveConnectionFlowControlWindowServer
+	MaxReceiveConnectionFlowControlWindowServer protocol.ByteCount
+	// MaxReceiveStreamFlowControlWindowClient is the maximum stream-level flow control window for receiving data, for the client
+	// If this value is zero, the timeout is set to protocol.DefaultMaxReceiveStreamFlowControlWindowClient
+	MaxReceiveStreamFlowControlWindowClient protocol.ByteCount
+	// MaxReceiveConnectionFlowControlWindowClient is the connection-level flow control window for receiving data, for the client
+	// If this value is zero, the timeout is set to protocol.DefaultMaxReceiveConnectionFlowControlWindowClient
+	MaxReceiveConnectionFlowControlWindowClient protocol.ByteCount
 }
```
marten-seemann: We only need two values here: `MaxReceiveStreamFlowControlWindow` and `MaxReceiveConnectionFlowControlWindow`. Then fill in the appropriate values in `populateServerConfig` and `populateClientConfig`.

In the comment, please don't refer to the protocol package. Instead, do something like this:
`MaxReceiveStreamFlowControlWindow is the maximum stream-level flow control window for receiving data. If not set, it will be x MB (as a client) or y MB (as a server)`

handshake/connection_parameters_manager.go
```diff
@@ -61,14 +66,36 @@ var (
...
+	if maxReceiveConnectionFlowControlWindowServer == 0 {
+		maxReceiveConnectionFlowControlWindowServer = protocol.DefaultMaxReceiveConnectionFlowControlWindowServer
+	}
+	if maxReceiveStreamFlowControlWindowClient == 0 {
+		maxReceiveStreamFlowControlWindowClient = protocol.DefaultMaxReceiveStreamFlowControlWindowClient
+	}
+	if maxReceiveConnectionFlowControlWindowClient == 0 {
+		maxReceiveConnectionFlowControlWindowClient = protocol.DefaultMaxReceiveConnectionFlowControlWindowClient
+	}
+
```
marten-seemann: Use the `populateServerConfig` and `populateClientConfig` to set default values. The idea here is that once the `quic.Config` has passed through one of these functions (which happens right after the `quic.Config` is passed to quic-go), all values in the config are valid.
This will also allow you to just pass 2 function parameters here.

twdkeule: Indeed, then it everything will be cleaner. Thanks!

handshake/connection_parameters_manager.go
```diff
@@ -208,9 +235,9 @@ func (h *connectionParametersManager) GetReceiveStreamFlowControlWindow() protoc
 // GetMaxReceiveStreamFlowControlWindow gets the maximum size of the stream-level flow control window for sending data
 func (h *connectionParametersManager) GetMaxReceiveStreamFlowControlWindow() protocol.ByteCount {
 	if h.perspective == protocol.PerspectiveServer {
```
marten-seemann: We probably can get rid off this if.

handshake/connection_parameters_manager.go
```diff
@@ -223,9 +250,9 @@ func (h *connectionParametersManager) GetReceiveConnectionFlowControlWindow() pr
 // GetMaxReceiveConnectionFlowControlWindow gets the maximum size of the stream-level flow control window for sending data
 func (h *connectionParametersManager) GetMaxReceiveConnectionFlowControlWindow() protocol.ByteCount {
 	if h.perspective == protocol.PerspectiveServer {
```
marten-seemann: And of this one as well.

handshake/connection_parameters_manager.go
```diff
@@ -61,14 +63,19 @@ var (
 )
 
 // NewConnectionParamatersManager creates a new connection parameters manager
-func NewConnectionParamatersManager(pers protocol.Perspective, v protocol.VersionNumber) ConnectionParametersManager {
+func NewConnectionParamatersManager(
+	pers protocol.Perspective, v protocol.VersionNumber,
+	maxReceiveStreamFlowControlWindow protocol.ByteCount, maxReceiveConnectionFlowControlWindow protocol.ByteCount,
```
lucas-clemente: As a comment (no change needed now): Maybe we can pass in the config here at some point (I know about the import issues, but those should be fixable). @marten-seemann

interface.go
```diff
@@ -80,6 +80,12 @@ type Config struct {
 	// If not set, it verifies that the address matches, and that the STK was issued within the last 24 hours
 	// This option is only valid for the server.
 	AcceptSTK func(clientAddr net.Addr, stk *STK) bool
+	// MaxReceiveStreamFlowControlWindowServer is the maximum stream-level flow control window for receiving data
```
lucas-clemente: Nit: Remove the "Server" from the comment, and add a dot at the end of the lines (godoc will put them in one paragraph).

interface.go
```diff
@@ -80,6 +80,12 @@ type Config struct {
...
+	// If this value is zero, it will default to 1 MB for the server and 6 MB for the client
+	MaxReceiveStreamFlowControlWindow protocol.ByteCount
+	// MaxReceiveConnectionFlowControlWindowServer is the connection-level flow control window for receiving data
```
lucas-clemente: Nit: Remove the "Server" from the comment, and add a dot at the end of the lines (godoc will put them in one paragraph).

handshake/connection_parameters_manager.go
marten-seemann: That will create an import loop, and I don't see how you can resolve that.
We could use the `TransportParameters` struct for that https://github.com/lucas-clemente/quic-go/blob/master/handshake/interface.go#L22-L24.

lucas-clemente: Go 1.9 should fix that. On the other hand, I don't think the split between the quic package and the handshake package is justified right now anyway, but that's not an issue we need to resolve right now.

twdkeule:
I noticed an incomplete sentence at https://github.com/lucas-clemente/quic-go/blob/master/interface.go#L70

lucas-clemente:
Nice catch, fixed :)

PR #667, opened 2017-06-13
run gofmt to fix ordering of imports

marten-seemann:
I forgot to do so in #648.


marten-seemann:
There shouldn't be any merge conflict since those PRs are (hopefully) also gofmted.

twdkeule:
Not intentionally, but I use [go-plus](https://atom.io/packages/go-plus)

marten-seemann:
That runs gofmt for you, among others.

Issue #624, opened 2017-05-17
Connection seems dead for 30s, then closes and starts a new one.


twdkeule:
I have an issue where a QUIC video stream seems dead for about 30 seconds before restarting a new connection (new CHLO etc.), this happens about 8 times in the 10min video.

Setup:
- Server: Caddy using QUIC
- Client: Chromium 58.0.3029.96 with force-quic-on server
- node in between to limit throughput with tcconfig (using netem/tc) and tcpdump capture

Might be related to #576 because:

1. The last packet sent by the server is at 03:31:24.277
2. everything is congestion limited
3. The client keeps sending Ack frames (I suppose for earlier video segments)
4. Everythin seems normal up until (and including) 03:31:35.415
5. Then no traffic for 20 seconds after which the client sends something strange:
```
frames.RstStreamFrame{StreamID:0x7, ErrorCode:0x6, ByteOffset:0x0}
Ignoring error in session: RST_STREAM received for unknown stream
```
6. This is ignored, but another 20 seconds later, the client closes the connection: `ConnectionCloseFrame{ErrorCode:0x19, ReasonPhrase:"No recent network activity."}`\
7. After this the client starts a new connection an everything is fine for a while.

I have a Wireshark capture, but it is 140MB because of the video. 
 [edit: removed log as it cluttered the conversation]

lucas-clemente:
Thanks a lot for the report! This looks like a duplicate of #555, but I'll keep it around since your log is super helpful :)

In particular, this looks interesting:

```
Congestion limited: bytes in flight 65170, window 64240
[...]
Queueing packet 0x131 for retransmission (RTO), 49 outstanding
Queueing packet 0x132 for retransmission (RTO), 48 outstanding
Congestion limited: bytes in flight 62968, window 2920
```

The congestion window collapses much faster than I would expect here, there may be a bug in the congestion controller.

twdkeule:
Here is the whole log: https://www.dropbox.com/s/kjmwtgh3qqrq4uw/quic.txt?dl=0 (25MB)
It happens 8 times in the whole log, just look for the CHLOs and work back from there.

I wanted to add the Wireshark capture, but then I remembered it can't decrypt the QUIC traffic anyway. Do you know of any way to decrypt the traffic?

lucas-clemente:
[...]

PR #679, opened 2017-06-15
Added test with large RTT

twdkeule:
I managed adding a test with a large RTT (75ms) and while downloading a large file. But iI am not yet sure whether it triggers issue #624, as it is one large file and not different (<1MB files at a 2s interval (as in the issue).

However, the filesizes of the logs are quite large and the tests quite slow (for localhost)

### DIFF
--- a/integrationtests/rtt_test.go
+++ b/integrationtests/rtt_test.go
@@ -20,7 +20,7 @@ import (
 	. "github.com/onsi/gomega/gexec"
 )
 
-var _ = FDescribe("non-zero RTT", func() {
+var _ = Describe("non-zero RTT", func() {
 	BeforeEach(func() {
 		dataMan.GenerateData(dataLen)
 	})
@@ -67,14 +67,15 @@ var _ = FDescribe("non-zero RTT", func() {
 				runRTTTest(10*time.Millisecond, version, 20)
 			})
 		})
-
-		Context(fmt.Sprintf("with quic version %d", version), func() {
-			It("gets a large file with 75ms RTT", func() {
-				dataMan.GenerateData(dataLongLen)
-				utils.SetLogLevel(utils.LogLevelDebug)
-
+	}
+	version := protocol.Version35
+	Context(fmt.Sprintf("with quic version %d", version), func() {
+		fileSizes := [...]int{dataLen, 5 * 1024 * 1024, dataLongLen}
+		for _, fileSize := range fileSizes {
+			fileSizeMB := fmt.Sprintf("%.2f", float64(fileSize)/1024/1024)
+			It("gets a "+fileSizeMB+"MB file with 75ms RTT", func() {
 				rtt := 75 * time.Millisecond
-				testname := "./large_rtt_" + rtt.String() + "_Q" + fmt.Sprint(version) + ".log"
+				testname := "./large_rtt_" + rtt.String() + "_Q" + fmt.Sprint(version) + "_" + fileSizeMB + "MB.log"
 
 				f, err := os.Create(testname)
 				defer f.Close()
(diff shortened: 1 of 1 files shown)


integrationtests/rtt_test.go
```diff
@@ -59,8 +64,32 @@ var _ = Describe("non-zero RTT", func() {
...
+	version := protocol.Version35
+	Context(fmt.Sprintf("with quic version %d", version), func() {
+		fileSizes := [...]int{dataLen, 5 * 1024 * 1024, dataLongLen}
+		for _, fileSize := range fileSizes {
+			fileSizeMB := fmt.Sprintf("%.2f", float64(fileSize)/1024/1024)
+			It("gets a "+fileSizeMB+"MB file with 75ms RTT", func() {
+				rtt := 75 * time.Millisecond
+				testname := "./large_rtt_" + rtt.String() + "_Q" + fmt.Sprint(version) + "_" + fileSizeMB + "MB.log"
+
+				f, err := os.Create(testname)
```
marten-seemann: Remove the logging here.

FYI, when running integrationtests, you can run them with `QUIC_GO_LOG_LEVEL=DEBUG gingko -- -logfile=log.txt`, so you don't have to modify the test code when you want a log.

integrationtests/rtt_test.go
```diff
@@ -59,8 +64,32 @@ var _ = Describe("non-zero RTT", func() {
 
 		Context(fmt.Sprintf("with quic version %d", version), func() {
 			It("gets a file with 10ms RTT", func() {
-				runRTTTest(10*time.Millisecond, version)
+				runRTTTest(10*time.Millisecond, version, 20)
 			})
 		})
 	}
+	version := protocol.Version35
```
marten-seemann: You probably forgot to remove this line.

twdkeule:
Ah yes, I forgot about the environment variable. I was testing too specific for the issue

integrationtests/rtt_test.go
lucas-clemente: Nit: you don't actually need the env when you set the flag.

marten-seemann: Oh right. I guess I tried to be too clever when implementing that flag ;)

integrationtests/rtt_test.go
```diff
@@ -59,8 +64,22 @@ var _ = Describe("non-zero RTT", func() {
...
-				runRTTTest(10*time.Millisecond, version)
+				runRTTTest(10*time.Millisecond, version, 20)
 			})
+
+			fileSizes := [...]int{dataLen, 5 * 1024 * 1024, dataLongLen}
+			for _, fileSize := range fileSizes {
+				fileSizeMB := fmt.Sprintf("%.2f", float64(fileSize)/1024/1024)
+				It("gets a "+fileSizeMB+"MB file with 75ms RTT", func() {
+					flags := log.Flags()
+					log.SetFlags(log.Ldate | log.Ltime | log.Lmicroseconds)
```
lucas-clemente: Remove the 5 logging relate lines here and below – we don't always need the logs and they may skew the tests by slowing them down. We also already globally set the timestamp format for logs.

integrationtests/rtt_test.go
```diff
@@ -1,14 +1,18 @@
 package integrationtests
 
 import (
+	"bufio"
```
lucas-clemente: Run goimports to remove the unused ones.

integrationtests/rtt_test.go
```diff
@@ -59,8 +64,22 @@ var _ = Describe("non-zero RTT", func() {
 
 		Context(fmt.Sprintf("with quic version %d", version), func() {
 			It("gets a file with 10ms RTT", func() {
-				runRTTTest(10*time.Millisecond, version)
+				runRTTTest(10*time.Millisecond, version, 20)
 			})
+
+			fileSizes := [...]int{dataLen, 5 * 1024 * 1024, dataLongLen}
+			for _, fileSize := range fileSizes {
+				fileSizeMB := fmt.Sprintf("%.2f", float64(fileSize)/1024/1024)
```
marten-seemann: I don't think we gain anything from running the test with various file sizes.
I'd rather run the test with different RTT values here, not just 10ms and 75ms.

PR #680, opened 2017-06-15
add a method to the session that blocks until it is closed

### LINKED ISSUE #632: implement some way to tell if a session is closed
An application might want to know if a session is closed (e.g. in order to garbage-collect it). Our current `Session` API doesn't support that.

marten-seemann:
Fixes #632.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -37,6 +37,9 @@ type Session interface {
 	RemoteAddr() net.Addr
 	// Close closes the connection. The error will be sent to the remote peer in a CONNECTION_CLOSE frame. An error value of nil is allowed and will cause a normal PeerGoingAway to be sent.
 	Close(error) error
+	// WaitUntilClosed() blocks until the session is closed.
+	// Warning: This API should not be considered stable and might change soon.
+	WaitUntilClosed()
 }
 
 // A NonFWSession is a QUIC connection between two peers half-way through the handshake.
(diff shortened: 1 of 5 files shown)


interface.go
```diff
@@ -37,6 +37,8 @@ type Session interface {
 	RemoteAddr() net.Addr
 	// Close closes the connection. The error will be sent to the remote peer in a CONNECTION_CLOSE frame. An error value of nil is allowed and will cause a normal PeerGoingAway to be sent.
 	Close(error) error
+	// WaitUntilClosed() blocks until the session is closed.
```
lucas-clemente: Please add a comment that this signature might change (e.g. we might want to add an error down the road).

session_test.go
```diff
@@ -772,6 +772,17 @@ var _ = Describe("Session", func() {
 			Expect(mconn.written[0][0] & 0x02).ToNot(BeZero()) // Public Reset
 			Expect(sess.runClosed).To(BeClosed())
 		})
+
+		It("unblocks WaitUntilClosed when the run loop exists", func() {
```
lucas-clemente: Please use an atomic bool.

lucas-clemente:
Also, you have to update the mocks.

session_test.go
marten-seemann: We're making our tests race-free? That'll be a long way to go and a lot of bools to change.

PR #681, opened 2017-06-16
add some more logging to the session

marten-seemann:


marten-seemann:
@lucas-clemente: Sorry for sending the PR too early. I now fixed all the tests, so you can review now.

session.go
```diff
@@ -684,7 +684,7 @@ func (s *session) logPacket(packet *packedPacket) {
 		return
 	}
 	if utils.Debug() {
-		utils.Debugf("-> Sending packet 0x%x (%d bytes), %s", packet.number, len(packet.raw), packet.encryptionLevel)
+		utils.Debugf("-> Sending packet 0x%x (%d bytes) for connection %x to %s, %s", packet.number, len(packet.raw), s.connectionID, s.conn.RemoteAddr().Network()+" "+s.conn.RemoteAddr().String(), packet.encryptionLevel)
```
lucas-clemente: Can we use the same format `(s -> r)` here as for the "Starting new connection" line above?

session.go
```diff
@@ -769,6 +769,7 @@ func (s *session) scheduleSending() {
 
 func (s *session) tryQueueingUndecryptablePacket(p *receivedPacket) {
 	if s.handshakeComplete {
+		utils.Debugf("Received undecryptable packet from %s after the handshake: %#v, %d bytes data\n", p.remoteAddr.String(), p.publicHeader, len(p.data))
```
lucas-clemente: No need for newlines in log messages.

session.go
marten-seemann: I don't think we should log the remote address here at all. I just forgot to remove this. Fixed now.

session.go
marten-seemann: Fixed.

PR #686, opened 2017-06-17
use TLSConfig.ServerName as the client hostname if provided

phuslu:
Currently quic client always parse `hostname` from the addr. It prevent us to specific a different `hostname` to the client.

This PR is to enable this by the `TLSConfig.ServerName` field. Thanks.

Signed-off-by: Phus Lu <phuslu@hotmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -73,9 +73,16 @@ func DialNonFWSecure(pconn net.PacketConn, remoteAddr net.Addr, host string, con
 		return nil, err
 	}
 
-	hostname, _, err := net.SplitHostPort(host)
-	if err != nil {
-		return nil, err
+	var hostname string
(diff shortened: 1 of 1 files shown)


lucas-clemente:
Thanks for the PR! :) This is definitely something we want to support, but I think it would be better to add this code to client.go, in particular around [here](https://github.com/lucas-clemente/quic-go/blob/master/client.go#L76) (and don't try to extract the hostname from the addr if it is set in the tls.Config). Would you mind moving it there?

phuslu:
@lucas-clemente Thanks, I done the modification. :)

marten-seemann:
Superseeded by #693.

PR #688, opened 2017-06-19
Fix non-retransmittable packet handling in SPH

lucas-clemente:


ackhandler/sent_packet_handler.go
```diff
@@ -310,10 +311,11 @@ func (h *sentPacketHandler) DequeuePacketForRetransmission() *Packet {
...
 	}
-	queueLen := len(h.retransmissionQueue)
-	// packets are usually NACKed in descending order. So use the slice as a stack
-	packet := h.retransmissionQueue[queueLen-1]
-	h.retransmissionQueue = h.retransmissionQueue[:queueLen-1]
+	packet := h.retransmissionQueue[0]
+	// Shift the slice and don't retain anything that isn't needed.
+	copy(h.retransmissionQueue, h.retransmissionQueue[1:])
+	h.retransmissionQueue[len(h.retransmissionQueue)-1] = nil
+	h.retransmissionQueue = h.retransmissionQueue[:len(h.retransmissionQueue)-1]
```
marten-seemann: What exactly are you doing here?

lucas-clemente: This is the standard pattern for deleting an element from a slice in such a way that it can be GCd. If you just do `s = s[1:]`, the first element will stay in memory until the slice is re-allocated.

See https://github.com/golang/go/wiki/SliceTricks

marten-seemann: 🤢 Okay.

PR #692, opened 2017-06-20
Implement Timeout()/Temproary() for QuicError

phuslu:
By adding Timeout()/Temparay() to QuicError, we could distinguish QuicError to different types.
Like idioms in golang net, context, http packages.

I'm not familar with quic very well, so I'm not sure current PR code is suitable or not.

Please give comments, thanks!

### DIFF
--- a/qerr/quic_error.go
+++ b/qerr/quic_error.go
@@ -31,6 +31,16 @@ func (e *QuicError) Error() string {
 	return fmt.Sprintf("%s: %s", e.ErrorCode.String(), e.ErrorMessage)
 }
 
+func (e *QuicError) Timeout() bool {
+	switch e.ErrorCode {
+	case NetworkIdleTimeout,
+		HandshakeTimeout,
+		TimeoutsWithOpenStreams:
+		return true
+	}
+	return false
+}
+
 // ToQuicError converts an arbitrary error to a QuicError. It leaves QuicErrors
 // unchanged, and properly handles `ErrorCode`s.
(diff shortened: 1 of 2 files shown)


lucas-clemente:
Thanks for the PR! This is an interesting suggestion that we hadn't thought of so far.

First, let me provide a bit more context (apologies if you already know some of these):

On the IETF side, error codes are currently undergoing some changes, see https://github.com/quicwg/base-drafts/pull/467. Basically, many are going to be removed, so we will probably change that soon.

Second, the qerr package is not meant to be publicly consumed (and will almost certainly move to internal/ soon). However, your PR still makes sense: People can check whether the error returned implements a simple interface with just the Timeout() function (see https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully, the section named "Assert errors for behaviour, not type").

Now, there's probably two options here. We can either wait what the IETF error space converges to and have another look at whether it makes sense to expose the error type then, or we can introduce Timeout() now. I'm not such a big fan of Temporary(), as currently no error fulfills that category.

@marten-seemann, WDYT?

phuslu:
@lucas-clemente I total agree you points. :)

In fact, I'm managing to integrate my project with quic-go. And I wish that I could assert quic timeout errors here
https://github.com/phuslu/goproxy/blob/b755a53388fed9613da407a34051677ca6d07517/httpproxy/filters/gae/gae.go#L538-L547
That's why I send this PR 😄 

And I removed Temporary() now.

PR #693, opened 2017-06-20
use TLSConfig.ServerName as the client hostname if provided

marten-seemann:
For some reason, Github didn't let me push into PR #686, so here's a new PR.

### DIFF
--- a/crypto/cert_manager.go
+++ b/crypto/cert_manager.go
@@ -107,15 +107,14 @@ func (c *certManager) Verify(hostname string) error {
 	var opts x509.VerifyOptions
 	if c.config != nil {
 		opts.Roots = c.config.RootCAs
-		opts.DNSName = c.config.ServerName
 		if c.config.Time == nil {
 			opts.CurrentTime = time.Now()
 		} else {
 			opts.CurrentTime = c.config.Time()
 		}
-	} else {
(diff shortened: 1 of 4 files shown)


marten-seemann:
Hi @phuslu,

thanks for your PR! I added some tests and removed some now obsolete code in the crypto package, and we can now merge these changes.
Apparently, @googlebot is not happy that I created a new branch for that. Can you please confirm here that you're ok with us merging your contribution here in this PR?

crypto/cert_manager.go
```diff
@@ -106,8 +106,8 @@ func (c *certManager) Verify(hostname string) error {
 
 	var opts x509.VerifyOptions
 	if c.config != nil {
+		// we don't need to care about the tls.Config.ServerName here, since hostname has already been set to that value in the session setup
```
lucas-clemente: I still don't really understand why you are doing this change… If the config is not nil, you don't set DNSName at all, not even to hostname, right? How is the verification supposed to work then?

phuslu:
Hi @marten-seemann and others, I'm OK for that.

crypto/cert_manager.go
marten-seemann: Oh, right. Must have missed that. Please review again.

PR #690, opened 2017-06-19
Packer refactoring

lucas-clemente:

PR #694, opened 2017-06-20
Send ACKs and SWFs even if we are congestion limited

lucas-clemente:
Only the last two commits differ from #690.


marten-seemann:
Please rebase after merging #690, and have a look at Codecov. Right now, it says merging would decrease our coverage quite a bit, but I don't really trust it...

PR #695, opened 2017-06-20
Session.Send refactor

lucas-clemente:
To be merged after #694.

### DIFF
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -253,7 +253,7 @@ func (p *packetPacker) composeNextPacket(
 	return payloadFrames, nil
 }
 
-func (p *packetPacker) QueueControlFrameForNextPacket(frame frames.Frame) {
+func (p *packetPacker) QueueControlFrame(frame frames.Frame) {
 	switch f := frame.(type) {
 	case *frames.StopWaitingFrame:
 		p.stopWaiting = f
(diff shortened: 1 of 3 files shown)


packet_packer.go
```diff
@@ -185,9 +191,24 @@ func (p *packetPacker) composeNextPacket(
 	var payloadLength protocol.ByteCount
 	var payloadFrames []frames.Frame
 
+	// STOP_WAITING and ACK will always fit
```
marten-seemann: Until we implement timestamps in ACKs.

packet_packer.go
```diff
@@ -185,9 +191,24 @@ func (p *packetPacker) composeNextPacket(
...
+		payloadFrames = append(payloadFrames, p.stopWaiting)
+		l, err := p.stopWaiting.MinLength(p.version)
+		if err != nil {
+			return nil, err
+		}
+		payloadLength += l
 	}
+	if p.ackFrame != nil {
+		payloadFrames = append(payloadFrames, p.ackFrame)
+		l, err := p.ackFrame.MinLength(p.version)
```
marten-seemann: Strictly speaking, this never errors for an ACK frame. I don't mind checking for the error here though for now, and removing the error from `MinLength` as soon as we finally drop the STOP_WAITING.

lucas-clemente: Yeah, I just don't like ignoring error values, as there's no compile-time check prohibiting you from e.g. changing AckFrame.MinLength to actually return an error.

packet_packer.go
lucas-clemente: Good point, but then ACK frame writing will become much harder anyway (probably with the need to pass in a max len to Write()).

Issue #583, opened 2017-05-02
Implement CloseNotifier interface


mholt:
Hi! :smile:

Is it trivial to implement a CloseNotify method on h2quic.ResponseWriter? I'm asking in reference to this issue: https://github.com/mholt/caddy/issues/1643

If so, I could potentially submit a pull request, I just need to know if this would be a no-op like Flush... or if it's more complicated maybe I should leave it to you too; I'm not super well versed in quic's operational requirements.

lucas-clemente:
Thanks for raising this! I think a "real" implementation might be a bit tricky, but definitely not impossible. But, as you're saying, I think it should be fine to just throw it in as a NOP for now (returning a new chan each time). We can then figure out the proper implementation later.

mholt:
Sounds good to me, PR is currently finishing up tests. Kinda sorry to add to your list of things to do. 😇 Thanks for all the great work you do though!

PR #698, opened 2017-06-21
implement the context for the server http.Request

### LINKED ISSUE #585: Actually implement http.CloseNotifier
As part of #584 we now have a NOP implementation, but we should do the real thing there.

twdkeule:
Fix #585 

I wanted to modify `session.runClosed` to use signal true when a the session was closed, but kept running into unclosed channels.

As a side note, my integration tests keep failing on my machine but travis/appveyor do not complain.
[integration_test.txt](https://github.com/lucas-clemente/quic-go/files/1091161/integration_test.txt)

### DIFF
--- a/h2quic/client_test.go
+++ b/h2quic/client_test.go
@@ -3,6 +3,7 @@ package h2quic
 import (
 	"bytes"
 	"compress/gzip"
+	"context"
 	"crypto/tls"
 	"errors"
 	"io"
@@ -36,6 +37,7 @@ var _ = Describe("Client", func() {
 		client = newClient(hostname, nil, &roundTripperOpts{}, nil)
 		Expect(client.hostname).To(Equal(hostname))
 		session = &mockSession{}
+		session.ctx, session.ctxCancel = context.WithCancel(context.Background())
 		client.session = session
 
 		headerStream = newMockStream(3)
--- a/h2quic/response_writer.go
+++ b/h2quic/response_writer.go
@@ -83,7 +83,7 @@ func (w *responseWriter) Write(p []byte) (int, error) {
 
 func (w *responseWriter) Flush() {}
 
-// TODO: Implement a functional CloseNotify method.
+// This is a NOP. Use http.Request.Context
 func (w *responseWriter) CloseNotify() <-chan bool { return make(<-chan bool) }
 
 // test that we implement http.Flusher
--- a/h2quic/response_writer_test.go
+++ b/h2quic/response_writer_test.go
@@ -25,20 +25,24 @@ type mockStream struct {
 	remoteClosed bool
 
 	unblockRead chan struct{}
+	ctx         context.Context
+	ctxCancel   context.CancelFunc
 }
 
 func newMockStream(id protocol.StreamID) *mockStream {
-	return &mockStream{
+	s := &mockStream{
 		id:          id,
 		unblockRead: make(chan struct{}),
 	}
+	s.ctx, s.ctxCancel = context.WithCancel(context.Background())
+	return s
 }
 
-func (s *mockStream) Close() error                          { s.closed = true; return nil }
+func (s *mockStream) Close() error                          { s.closed = true; s.ctxCancel(); return nil }
 func (s *mockStream) Reset(error)                           { s.reset = true }
-func (s *mockStream) CloseRemote(offset protocol.ByteCount) { s.remoteClosed = true }
+func (s *mockStream) CloseRemote(offset protocol.ByteCount) { s.remoteClosed = true; s.ctxCancel() }
 func (s mockStream) StreamID() protocol.StreamID            { return s.id }
-func (s *mockStream) Context() context.Context              { return context.Background() }
+func (s *mockStream) Context() context.Context              { return s.ctx }
 func (s *mockStream) SetDeadline(time.Time) error           { panic("not implemented") }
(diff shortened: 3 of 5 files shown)


marten-seemann:
Hello @twdkeule, thanks for this PR. I just had a call with @lucas-clemente and we talked about this topic as well.

As you already suggested in the comment, we agree that the `WaitUntilClosed`, which we recently introduced (#680), is probably not the best solution for telling if a session is closed. 

Probably the best way is to use the `runClosed` channel, i.e. add this function to the `Session`: `CloseNotify() <-chan struct{}`. I like this more than the `CloseNotify() <-chan bool` because it allows a more flexible use: If you close a channel, you can check the condition multiple times, whereas the `chan bool` receives at most one value. Furthermore, to me it feels a bit odd to use a `bool`, but then specify that the only possible value is `true`.

To summarize, what I'd suggest doing here is the following:

1. replace `CloseNotify() <-chan bool` by `CloseNotify() <-chan struct{}` in your PR.
2. remove the `WaitUntilClosed`.

What do you think?

twdkeule:
I completely agree that using runClosed would be the better option. However the CloseNotify() interface requires returning a `chan bool`. My knowledge of Go is very young, so I don't really know how to go from the `chan struct {}` to `chan bool`. i was thinking about starting a goroutine to block on the `chan struct{}` and  send a `chan bool`, but it seemed quite wasteful to start this goroutine for every stream.

lucas-clemente:
Let me chime in here :) I also think that the right API for the session is a chan that is just closed.

You are right that "transforming" that from a `chan struct{}` to a `chan bool` requires a go routine, but it should only need a single one per session. One per request would not be super costly, but it's nice if we can avoid it.

So, in principle we can just have a slice of response writers, with each their own `chan bool`(of size 1), and a single go routine that first blocks on the `chan struct{}`, and then sends a bool to all of the chans in the slice. However, we now have memory that grows with the number of streams that ever existed, which is a DOS issue… So, I would actually use a `map[*responseWriter] chan bool`, and remove the entries for the response writers as soon as they are done. Note that this map has to be mutex protected; I'd probably throw the mutex together with the map into a small struct `closeNotifier` or something.

NB: There's actually a bug with your current implementation, in that only one bool will be sent per session, so as soon as the first responseWriter reads it, it's gone, and the others will block again.

lucas-clemente:
Another thing I just found: https://stackoverflow.com/questions/44107868/is-http-closenotify-entirely-replaced-by-http-request-context-done

Maybe we don't need to bother with the chan of bool at all and should rather cancel the context instead…

twdkeule:
Looks a lot simpler to implement.
But the CloseNotify() interface was introduced because Go required it on http.responseWriters, no? Would that not reintroduce #583?

lucas-clemente:
I think it should be fine to leave it with its current NOP implementation and add a note somewhere that we won't fix this and people should use contexts.

twdkeule:
The unit tests for session.go fail, but I'm not sure why. I replaced `session.runClosed` by `context.Context.Done()` and `context.CancelFunc`.
`Done()` also returns a `<-chan struct{}` and calling `CancelFunc()` should do the same as `close(chan)`, so the behaviour should be the same.

marten-seemann:
Hi @twdkeule,

sorry for getting back so late on this! The reason is that I'm not really sure what the correct solution here is.

The easy part of this is the client side: Looking at the spec for `http.Request.Context` (https://golang.org/pkg/net/http/#Request.Context), it says that for the client the context controls cancellation. My interpretation of this is that the client is supposed to cancel the context if it is no longer interested in receiving a response. For QUIC this would mean resetting the corresponding (data?) stream.
The problem is that on the server side, a stream reset doesn't generate any immediate signal for the `h2quic.Server`. The first time the server learns about the canceled stream is when it calls `Write` on the stream and gets an error for that. At that point we could cancel the context in the server. However, the `Write` call probably occurs *after* completing the potentially expensive operation, so that doesn't solve the problem.

One possible solution would be to implement something somewhat analogous to `Session.WaitUntilClosed()` for the stream, i.e. a function (or channel) that blocks until the stream is closed by the other peer. I'm not sure if this a good way to move forward though, and I'd really like to hear what you and @lucas-clemente think about this.

marten-seemann:
@twdkeule: I don't know if you got a notification about it, I merged #774 today. The stream now has a `context.Context`, which should make implementing this possible.

twdkeule:
Thanks for the heads up, but it'll have to wait a month or so until I have more time :-)

twdkeule:
Should the server use the context from session and pass it to the request, like I did? Or should it be the streams context.
And why would we need a `WaitUntilClosed` function? The session has a context

marten-seemann:
The `WaitUntilClosed` was removed from the session in favor of the context in #774.

The requests must use the stream contexts, because there can be multiple requests multiplexed on one QUIC connection, and then it's possible to cancel them separately. Note that when the session itself goes away, all streams are closed, which will also cancel their contexts, so we'll have that special case covered for free.

twdkeule:
So it really is that simple. I was expecting something more complicated.
I'll add a test

twdkeule:
I think this is ready to merge.

twdkeule:
@marten-seemann test is added :smile:

lucas-clemente:
Thanks again :)

PR #702, opened 2017-06-22
fix INVALID_STREAM error when a retransmission for a server-side stream arrives after the stream is closed

### LINKED ISSUE #691: Stream ID sequencing issue
I'm seeing errors like:

 - InvalidStreamID: attempted to open stream 4 from client-side

I've read the QUIC docs but am still left with the questions:

 - Are we able to arbitrarily open streams on the client and server, or does it have to be Client, Server, Client, alternating?
 - If we are able to open streams in any order, then why is this error in place?

Thanks!

marten-seemann:
Fixes #691.

### DIFF
--- a/streams_map.go
+++ b/streams_map.go
@@ -83,15 +83,27 @@ func (m *streamsMap) GetOrOpenStream(id protocol.StreamID) (*stream, error) {
 		return s, nil
 	}
 
-	if id <= m.highestStreamOpenedByPeer {
-		return nil, nil
+	if m.perspective == protocol.PerspectiveServer {
+		if id%2 == 0 {
+			if id <= m.nextStream { // this is a server-side stream that we already opened. Must have been closed already
+				return nil, nil
+			}
+			return nil, qerr.Error(qerr.InvalidStreamID, fmt.Sprintf("attempted to open stream %d from client-side", id))
+		}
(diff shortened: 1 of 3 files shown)


streams_map.go
```diff
@@ -83,15 +83,27 @@ func (m *streamsMap) GetOrOpenStream(id protocol.StreamID) (*stream, error) {
...
+	if m.perspective == protocol.PerspectiveClient {
+		if id%2 == 1 {
+			if id <= m.nextStream { // this is a client-side stream that we already opened.
+				return nil, nil
+			}
+			return nil, qerr.Error(qerr.InvalidStreamID, fmt.Sprintf("attempted to open stream %d from server-side", id))
+		}
+		if id <= m.highestStreamOpenedByPeer { // this is a server-side stream that doesn't exist anymore. Must have been closed already
+			return nil, nil
+		}
```
lucas-clemente: I'm fine with merging this as is, but we could also factor this out into a separate function (which would be pretty simple to unit test). You could also remove the duplicate code by making the `0, 1` and the error string a var that you set depending on the perspective, but I don't think that will make it nicer.

marten-seemann: I agree that this code isn't particularly nice. I don't have the time to refactor it now, so let's merge it now and do the refactoring later, so we don't block anyone with this.

Issue #677, opened 2017-06-15
Expose quic.Config in h2quic.Server and h2quic.RoundTripper


lucas-clemente:

marten-seemann:
We need to think about what we do with the `tls.Config` then, which is currently a member of the `QuicRoundTripper`.

PR #703, opened 2017-06-23
Add HandshakeTimeout to h2quic.QuicRoundTripper

phuslu:
Currently h2quic.QuicRoundTripper always use default handshake timeout(10s).
This PR is to expose HandshakeTimeout for QuicRoundTripper.

For tests parts, I'm not sure the added unittest is a right one, please comment that.
Thanks!

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"time"
 
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/hpack"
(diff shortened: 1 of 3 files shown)


lucas-clemente:
Thanks for the PR! We definitely want to support configuring quic settings in the h2quic package, but I think that the nicer solution will be to allow users to specify the quic.Config from there; #677 tracks this. As Marten mentioned there, this is unfortunately not straight-forward, as we currently set the tls.Config in the quic.Config as well as in the h2quic.QuicRoundTripper and the h2quic.Server (and it would then be duplicate). I'm tending towards removing the tls.Config from the quic.Config, but I have to think about it a bit more :/

So, I'm gonna close this PR, sorry for the time you put into it! If you want to work on #677 instead please ping the issue!

PR #713, opened 2017-06-30
remove the tls.Config from the quic.Config

marten-seemann:
The `tls.Config` now is a separate parameter to all Listen and Dial functions in the `quic` package.


marten-seemann:
The problem was a missing re-initializiation of the `quic.Config` between the handshake RTT tests. Northing worth reviewing, I'll merge directly if the CIs pass.

PR #708, opened 2017-06-26
Add Go 1.9beta2 to travis config

lucas-clemente:

PR #714, opened 2017-07-03
Fix cookie test for Go1.9

lucas-clemente:
Sending this as a separate PR as #708 will probably take more time and I'd like 1.9 to be usable for dev.


marten-seemann:
I don't see the point merging this as long as there is other stuff that doesn't work properly with Go 1.9.

lucas-clemente:
It's a valid commit, I don't see why we should delay it. I generally use betas locally and this breaks my workflow; that should be reason enough.

lucas-clemente:
Also, this simplifies local debugging ;)

marten-seemann:
Fair enough. What's the problem with 1.9 anyway?

marten-seemann:
Please check what's going on with the CIs before merging.

PR #720, opened 2017-07-06
Add qtrace - a Tracer for events/packets in quic

konnykonny50:
I implemented a new module called qtrace. A Tracer can be attached to a server or client session to get some insights about events and messages being handled by quic. You can provide own methods to receive these events.
Let me know what you think about it in the comments.


konnykonny50:
I signed it!

konnykonny50:
I signed it!

handshake/crypto_setup_server.go
```diff
@@ -28,6 +29,7 @@ type cryptoSetupServer struct {
 	scfg                 *ServerConfig
 	stkGenerator         *STKGenerator
 	diversificationNonce []byte
+	QuicTracer           qtrace.Tracer
```
Eichhoernchen: I think this should be a pointer as in scfg above, to not copy the tracer over and over

handshake/crypto_setup_server.go
```diff
@@ -70,6 +72,7 @@ func NewCryptoSetup(
 	supportedVersions []protocol.VersionNumber,
 	acceptSTK func(net.Addr, *STK) bool,
 	aeadChanged chan<- protocol.EncryptionLevel,
+	QuicTracer qtrace.Tracer,
```
Eichhoernchen: The same here, I think a pointer makes more sense

handshake/crypto_setup_server.go
konnykonny50: Yes, that might be a better idea. I've changed it. Thanks.

lucas-clemente:
Thanks for the PR! We certainly need more possibilities for our users to get to detailed information about what is happening, and I think this is a very promising start!

I have a few high level comments that I think we should discuss before I do a proper code review. I'd like very much to hear your input, especially which requirements motivated you to implement this :)

- I'm a bit surprised at the choice of struct + lambdas for the tracing. Naively, I'd have taken an interface with a single method for this, and essentially an enum (with the event type) and a map or struct with additional information as parameters. One thing we certainly don't want to do is an interface with many methods (b/c we'll break compatibility when adding new events), but is there a downside you see with the interface approach?
- We are planning to make everything but the root and the h2quic package internal at some point – we don't want it to be part of our public API (which we try to not break). So, passing `interfaces {}` values won't work. An alternative would be structs, but I'm wondering whether we really need this information to be strictly typed. Maybe a `map[string]interface{}` would work…
- I'm a bit surprised you are tracing very low level events, rather than details of packet loss, RTT, or the state of the congestion controller. I'd really like to hear your use-case for information about the underlying QUIC frames. I'm a bit worried whether that exposes too much of our internal structure (which will certainly change a lot when we're moving towards IETF QUIC).
- This is all QUIC, not HTTP/2. Do you think information about H2 is interesting as well?
- Finally, I'm wondering whether maybe our logging should be implemented using tracers ^^

I'd also like @marten-seemann to chime in here, and maybe @twdkeule who might want to use similar info for doing measurements about QUIC performance?

twdkeule:
To be honest, my research is about video streaming so I'm not focussing on the low level events of QUIC. (Though I needed it to implement the push promise.)
My interests lie more towards the RTT, packet loss and congestion. And I think anyone using HTTP/2 over QUICK would indeed be more interested in those reports.

Eichhoernchen:
@konnykonny50 is one of my students. We are researching how QUIC is being used in the Internet and we believe that quic-go is a really good implementation to do research on and build tools from. So this tracer allows us to write tools that give us the opportunity to investigate how QUIC is used in the Internet.

The idea to use the `interfaces{}` came because we got into include cycles. Yet, if the goal is to make all types private this will break anyways. 
The tracer is inspired by the "net/http/httptrace" package which defines several methods for different tracing events, as we did not want to have a copy of some types we opted for the `interface{}` approach, yet we can easily introduce new tracing types to have a stable interface.

We started from the bottom and at some point we are definitively interested in connection level parameters such as RTT and packet loss. Yet, currently I feel that quic-go is not something I would use for performance measurements it still misses things like packet pacing if I'm not mistaken.

Honestly, we started out by capturing the debug log yet when introducing parallelism this becomes a real mess, so maybe moving from a debug log to a tracer might actually be a good idea when things get more involved.

Regarding H2, I'm not quite sure if there should be anything H2-specific in quic. Maybe just something like which stream maps to which quic stream, but anything else I would trace in H2.

konnykonny50:
I hope @Eichhoernchen was able to answer some of your questions. In the latest commit I've changed the interface of the tracer to be compatible when switching to private types at some time.

marten-seemann:
Hello @konnykonny50, thanks for this PR!
We believe that the approach described https://github.com/lucas-clemente/quic-go/issues/1258#issuecomment-376844078 is the way want to take in quic-go. I'm therefore closing this PR.
Please also have a look at the debug log. We added a lot more logging there recently, and also prefixed each line with {server, client}, so that should make reading quic-go logs a lot easier.

PR #724, opened 2017-07-10
add Ginkgo timeout for unit tests that do stream reads and writes

marten-seemann:
`Read` and `Write` are calls that are supposed to block under certain circumstances. Adding a timeout to the tests ensures that test actually fail when `Read` or `Write` doesn't unblock properly.

I wish there was a way to set a global timeout for each test with Ginkgo. We don't have any unit tests that is supposed to run longer than 1s.

### DIFF
--- a/session_test.go
+++ b/session_test.go
@@ -246,7 +246,7 @@ var _ = Describe("Session", func() {
 	})
 
 	Context("when handling stream frames", func() {
-		It("makes new streams", func() {
+		It("makes new streams", func(done Done) {
 			sess.handleStreamFrame(&frames.StreamFrame{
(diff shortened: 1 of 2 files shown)


lucas-clemente:
I've opened https://github.com/onsi/ginkgo/pull/362 which I think is a nicer fix for this issue.

lucas-clemente:
Following the discussion at ginkgo, I think we should do this without async tests. We can introduce a small helper function for io.Closer and io.Writer, and probably just use gbytes for the Reader. Thoughts?

lucas-clemente:
Superseded by #742.

PR #728, opened 2017-07-11
various session improvements

marten-seemann:


session.go
```diff
@@ -716,11 +710,9 @@ func (s *session) logPacket(packet *packedPacket) {
 		// We don't need to allocate the slices for calling the format functions
 		return
 	}
-	if utils.Debug() {
-		utils.Debugf("-> Sending packet 0x%x (%d bytes) for connection %x, %s", packet.number, len(packet.raw), s.connectionID, packet.encryptionLevel)
-		for _, frame := range packet.frames {
-			frames.LogFrame(frame, true)
-		}
+	utils.Debugf("-> Sending packet 0x%x (%d bytes) for connection %x, %s", packet.number, len(packet.raw), s.connectionID, packet.encryptionLevel)
```
lucas-clemente: This change will cause a lot of unnecessary slice allocations if debug logging is not enabled.

marten-seemann: It won't, because there's a `return` 2 lines above.

lucas-clemente: Huh, okay. Why did we have this if then at all…

lucas-clemente: Also, GH should really show more in the diff view 🙄

PR #729, opened 2017-07-11
implement receiving of Public Resets for the client

### LINKED ISSUE #34: Implement public reset receiving

marten-seemann:
When a Public Reset is received, the client validates if it was sent from the correct remote address and if the connection ID matches. When a valid Public Reset is received, the connection is closed immediately.

This fixes #34. For the server-side, there's no need to receive Public Resets.

### DIFF
--- a/client.go
+++ b/client.go
@@ -233,6 +233,24 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
+	if hdr.ResetFlag {
+		cr := c.conn.RemoteAddr()
+		// check if the remote address and the connection ID match
+		// otherwise this might be an attacker trying to inject a PUBLIC_RESET to kill the connection
+		if cr.Network() != remoteAddr.Network() || cr.String() != remoteAddr.String() || hdr.ConnectionID != c.connectionID {
+			utils.Infof("Received a spoofed Public Reset. Ignoring.")
+			return nil
+		}
+		pr, err := parsePublicReset(r)
+		if err != nil {
+			utils.Infof("Received a Public Reset for connection %x. An error occurred parsing the packet.")
+			return nil
+		}
+		utils.Infof("Received Public Reset, rejected packet number: %#x.", pr.rejectedPacketNumber)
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -233,6 +233,23 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
+	if hdr.ResetFlag {
+		cr := c.conn.RemoteAddr()
+		// check if the remote address PUBLIC_RESET matches the address we were sending to before and that the connection ID matches
```
lucas-clemente: Nit: weird grammar here.

client.go
```diff
@@ -233,6 +233,23 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
...
+		// otherwise this might be an attacker trying to inject a PUBLIC_RESET to kill the connection
+		if (cr.Network() != remoteAddr.Network() || cr.String() != remoteAddr.String()) || hdr.ConnectionID != c.connectionID {
```
lucas-clemente: parens unnecessary

client.go
```diff
@@ -233,6 +233,23 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
...
+			utils.Infof("Received a spoofed Public Reset. Ignoring.")
+			return nil
+		}
+		pr, err := parsePublicReset(r)
+		if err != nil {
+			utils.Infof("Received a Public Reset for connection %x. An error occurred parsing the packet.")
+			return nil
+		}
+		utils.Infof("Received Public Reset, rejected packet number: 0x%x.", pr.rejectedPacketNumber)
```
lucas-clemente: Nit: s/0x%x/%#x/

client.go
```diff
@@ -233,6 +233,23 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
...
+		c.session.closeRemote(qerr.Error(qerr.PublicReset, fmt.Sprintf("Received a Public Reset for packet number 0x%x", pr.rejectedPacketNumber)))
```
lucas-clemente: Nit: s/0x%x/%#x/

client.go
```diff
@@ -233,6 +233,23 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
...
+	}
```
lucas-clemente: We should probably return here?

PR #731, opened 2017-07-11
don't close the client connection when the Public Header can't be parsed

### LINKED ISSUE #730: don't close the connection when receiving a packet with an un-parseable Public Header in the client

marten-seemann:
Fixes #730.

When receiving an unparseable packet with a spoofed remote address, we should not close the connection.

### DIFF
--- a/client.go
+++ b/client.go
@@ -211,22 +211,19 @@ func (c *client) listen() {
 		}
 		data = data[:n]
 
-		err = c.handlePacket(addr, data)
-		if err != nil {
-			utils.Errorf("error handling packet: %s", err.Error())
-			c.session.Close(err)
-			break
-		}
+		c.handlePacket(addr, data)
 	}
 }
 
-func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
+func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) {
 	rcvTime := time.Now()
 
 	r := bytes.NewReader(packet)
 	hdr, err := ParsePublicHeader(r, protocol.PerspectiveServer)
 	if err != nil {
-		return qerr.Error(qerr.InvalidPacketHeader, err.Error())
+		utils.Errorf("error parsing packet from %s: %s", remoteAddr.String(), err.Error())
+		// drop this packet if we can't parse the Public Header
+		return
 	}
 	hdr.Raw = packet[:len(packet)-r.Len()]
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -211,22 +211,20 @@ func (c *client) listen() {
...
 	rcvTime := time.Now()
 
 	r := bytes.NewReader(packet)
 	hdr, err := ParsePublicHeader(r, protocol.PerspectiveServer)
 	if err != nil {
-		return qerr.Error(qerr.InvalidPacketHeader, err.Error())
+		utils.Errorf("error handling packet: %s", err.Error())
+		// drop this packet if we can't read parse the Public Header
+		// this packet could be malformed, or an injected packet with a spoofed remote address
+		return
```
marten-seemann: @lucas-clemente: Do we need to put back the packet buffer here?

lucas-clemente: We can put it back on every code path that doesn't enter the session. Putting it back is never required though, just a (small) matter of performance.

client.go
```diff
@@ -211,22 +211,20 @@ func (c *client) listen() {
...
+func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) {
 	rcvTime := time.Now()
 
 	r := bytes.NewReader(packet)
 	hdr, err := ParsePublicHeader(r, protocol.PerspectiveServer)
 	if err != nil {
-		return qerr.Error(qerr.InvalidPacketHeader, err.Error())
+		utils.Errorf("error handling packet: %s", err.Error())
+		// drop this packet if we can't read parse the Public Header
+		// this packet could be malformed, or an injected packet with a spoofed remote address
```
lucas-clemente: I don't understand this comment. How would ParsePublicHeader know it was a spoofed packet?

client.go
```diff
@@ -211,22 +211,20 @@ func (c *client) listen() {
...
-func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
+func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) {
 	rcvTime := time.Now()
 
 	r := bytes.NewReader(packet)
 	hdr, err := ParsePublicHeader(r, protocol.PerspectiveServer)
 	if err != nil {
-		return qerr.Error(qerr.InvalidPacketHeader, err.Error())
+		utils.Errorf("error handling packet: %s", err.Error())
+		// drop this packet if we can't read parse the Public Header
```
lucas-clemente: duplicate verb

client.go
```diff
@@ -211,22 +211,20 @@ func (c *client) listen() {
...
 
-func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
+func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) {
 	rcvTime := time.Now()
 
 	r := bytes.NewReader(packet)
 	hdr, err := ParsePublicHeader(r, protocol.PerspectiveServer)
 	if err != nil {
-		return qerr.Error(qerr.InvalidPacketHeader, err.Error())
+		utils.Errorf("error handling packet: %s", err.Error())
```
lucas-clemente: s/handling/parsing/

I'd include the remote addr in this message as well.

client.go
```diff
@@ -239,21 +237,21 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) error {
 		// otherwise this might be an attacker trying to inject a PUBLIC_RESET to kill the connection
 		if cr.Network() != remoteAddr.Network() || cr.String() != remoteAddr.String() || hdr.ConnectionID != c.connectionID {
```
lucas-clemente: Remark: it's trivial to just spoof the sender address, so the only really relevant thing here is the connection id, right?

client_test.go
```diff
@@ -265,62 +257,53 @@ var _ = Describe("Client", func() {
...
 			})
 
-			It("errors if the version is supported by quic-go, but disabled by the quic.Config", func() {
-				v := protocol.SupportedVersions[1]
-				Expect(v).ToNot(Equal(cl.version))
-				Expect(config.Versions).ToNot(ContainElement(v))
-				err := cl.handlePacket(nil, composeVersionNegotiation(0x1337, []protocol.VersionNumber{v}))
-				Expect(err).To(MatchError(qerr.InvalidVersion))
-			})
+			// It("errors if the version is supported by quic-go, but disabled by the quic.Config", func() {
```
lucas-clemente: No commented code please. Either kill or set pending with an explanation.

client.go
marten-seemann: It doesn't really. It could just be random junk arriving on our UDP socket.

client.go
marten-seemann: Correct.

client_test.go
marten-seemann: Oh, I forgot to reenable this test.

client.go
lucas-clemente: Then I'd remove the second line of the comment.

PR #739, opened 2017-07-15
fix timestamp test such that it works in all timezones

### LINKED ISSUE #738: log test fails in ICT time zone
```
• Failure [0.000 seconds]
Log [It] adds a timestamp
/Users/marten/src/go/src/github.com/lucas-clemente/quic-go/internal/utils/log_test.go:92

  Expected error:
      <*time.ParseError | 0xc420011d60>: {
          Layout: "Jan 2, 2006 at 3:04:05pm (MST)",
          Value: "Jul 14, 2017 at 6:07:17pm (+07)",
          LayoutElem: "MST",
          ValueElem: "+07)",
          Message: "",
      }
      parsing time "Jul 14, 2017 at 6:07:17pm (+07)" as "Jan 2, 2006 at 3:04:05pm (MST)": cannot parse "+07)" as "MST"
  not to have occurred

  /Users/marten/src/go/src/github.com/lucas-clemente/quic-go/internal/utils/log_test.go:90
```

marten-seemann:
Fixes #738.

### DIFF
--- a/internal/utils/log_test.go
+++ b/internal/utils/log_test.go
@@ -82,13 +82,13 @@ var _ = Describe("Log", func() {
 	})
 
 	It("adds a timestamp", func() {
-		format := "Jan 2, 2006 at 3:04:05pm (MST)"
+		format := "Jan 2, 2006"
 		SetLogTimeFormat(format)
 		SetLogLevel(LogLevelInfo)
 		Infof("info")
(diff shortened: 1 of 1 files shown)


marten-seemann:
I think it's a Go bug. Apparently the problem is that *(MST)* and *(+07)*, which both indicate the time zone, are apparently not recognised as the same time format.

PR #742, opened 2017-07-17
use the new gbytes.TimeoutReader and TimeoutWriter in the stream tests

marten-seemann:
This makes sure that our tests fail, even if `stream.Read` or `stream.Write` block.

Supersedes #724.

### DIFF
--- a/stream_test.go
+++ b/stream_test.go
@@ -14,14 +14,16 @@ import (
 	"github.com/lucas-clemente/quic-go/protocol"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 )
 
 var _ = Describe("Stream", func() {
 	const streamID protocol.StreamID = 1337
 
 	var (
-		str          *stream
-		onDataCalled bool
+		str            *stream
+		strWithTimeout io.ReadWriter // str wrapped with gbytes.Timeout{Reader,Writer}
+		onDataCalled   bool
 
 		resetCalled          bool
 		resetCalledForStream protocol.StreamID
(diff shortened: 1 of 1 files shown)


stream_test.go
```diff
@@ -52,6 +53,18 @@ var _ = Describe("Stream", func() {
 		resetCalledAtOffset = offset
 	}
 
+	// streamWithTimeout creates an io.ReadWriter that makes sure that Read and Write time out
+	streamWithTimeout := func(str io.ReadWriter) io.ReadWriter {
+		timeout := scaleDuration(250 * time.Millisecond)
```
lucas-clemente: Do we really need to do this everywhere? Why not just pick 1s, it shouldn't fail that often locally, right? I though this was only an issue for the deadlines…

stream_test.go
```diff
@@ -52,6 +53,18 @@ var _ = Describe("Stream", func() {
 		resetCalledAtOffset = offset
 	}
 
+	// streamWithTimeout creates an io.ReadWriter that makes sure that Read and Write time out
+	streamWithTimeout := func(str io.ReadWriter) io.ReadWriter {
```
lucas-clemente: The stream is stored in `str` anyway, so this could be another local variable. Then we don't need the function call everywhere.

stream_test.go
marten-seemann: The deadline tests will fail on the CIs, because the deadline will be after the timeout.

lucas-clemente: Damn it. I really hate the scaling thing, but fine…

stream_test.go
marten-seemann: Is that more readable?

PR #744, opened 2017-07-18
Removal of Alternate-Protocol header

Injust:
Google no longer uses the Alternate-Protocol header to advertise support for QUIC, instead using the Alt-Svc header, and the latest IETF draft has also dropped the Alternate-Protocol header.
Would doing the same thing here be good to stay in sync?


Injust:
I signed it!

lucas-clemente:
Cool, thanks! :)

PR #762, opened 2017-07-24
print commands on Travis

marten-seemann:
See https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps for documentation.

### DIFF
--- a/.travis/after_success.sh
+++ b/.travis/after_success.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
 
 if [ ${TESTMODE} == "unit" ]; then
   cat `find . -name "*.coverprofile"` > coverage.txt
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
 
 go get -t ./...
 if [ ${TESTMODE} == "unit" ]; then


.travis/after_success.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ev
```
lucas-clemente: Use -x, that expands the variables before printing. Also, this could be part of the shebang line.

.travis/script.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ev
```
lucas-clemente: As above.

marten-seemann:
Looks a bit better with the `-x` option, but still not perfect.

lucas-clemente:
What don't you like about it?

marten-seemann:
Have a look how the commands are printed when they're not in a bash-script, but specified in .travis.yml directly: https://travis-ci.org/marten-seemann/quic-conn/jobs/257168386

lucas-clemente:
Oh right, travis does that. Tbh I don't like it ;)

marten-seemann:
We're running the benchmark-tests twice (with and without race detector). I find it very useful to have the command printed above the results, to see which test output I'm looking at.

lucas-clemente:
We can always split them up into separate jobs…

PR #766, opened 2017-07-25
move benchmarks to separate package, speed up CI builds

marten-seemann:
Compared to #764, which already reduced the Travis build time by 100s, this PR achieves a further speedup by reducing the number of samples and the data size transferred (for the test run with the race detector), thus saving another:
* 25s on the benchmark tests run without the race detector (from 50s down to 25s)
* 160s on the benchmark tests run without the race detector (from ~170s down to less than 10s)

In total, we can reduce the Travis build time by around 3 minutes by merging this PR.
On AppVeyor, we save about 10 - 15s on every build, which is nice, however, that was never our bottleneck.

### DIFF
--- a/benchmark/benchmark_test.go
+++ b/benchmark/benchmark_test.go
@@ -1,4 +1,4 @@
-package quic
+package benchmark
 
 import (
 	"bytes"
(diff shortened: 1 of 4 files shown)


benchmark/benchmark_test.go
```diff
@@ -69,7 +67,8 @@ var _ = Describe("Benchmarks", func() {
 
 				ln.Close()
 				sess.Close(nil)
-			}, 6)
+				}, samples)
```
lucas-clemente: Please gofmt this file ;)

marten-seemann: The file is properly gofmted. Please note that all `Describe` code was moved into the `init`. This is the only way to use the flag values.

PR #769, opened 2017-07-26
Add Connect support to h2quic

phuslu:
Currently, I'm working on using h2quic as a CONNECT tunnel.
This pull request makes h2quic.Server allow `path` is empty if request method is `CONNECT`

Signed-off-by: Phus Lu <phuslu@hotmail.com>

### DIFF
--- a/h2quic/request.go
+++ b/h2quic/request.go
@@ -37,7 +37,7 @@ func requestFromHeaders(headers []hpack.HeaderField) (*http.Request, error) {
 		httpHeaders.Set("Cookie", strings.Join(httpHeaders["Cookie"], "; "))
 	}
 
-	if len(path) == 0 || len(authority) == 0 || len(method) == 0 {
+	if (len(path) == 0 && method != http.MethodConnect) || len(authority) == 0 || len(method) == 0 {
 		return nil, errors.New(":path, :authority and :method must not be empty")
 	}
 


phuslu:
If the changes is OK, I'm happy to add tests for it. 😄

h2quic/client.go
```diff
@@ -178,7 +178,7 @@ func (c *client) RoundTrip(req *http.Request) (*http.Response, error) {
 		requestedGzip = true
 	}
 	// TODO: add support for trailers
-	endStream := !hasBody
+	endStream := !hasBody && req.Method != http.MethodConnect
```
marten-seemann: According to the documentation (e.g. here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) it looks like the request should have a body. The http2 package in the standard library doesn't do any checks for the request method here as well.
Can you explain why you think we need this addition here?

phuslu: Emm, my understanding is below,

The statement `req.Method != http.MethodConnect` here does not mean there's a body with request.

It just enforce 'enStream = true' during a CONNECT request. 
So h2quic.Server here https://github.com/lucas-clemente/quic-go/blob/master/h2quic/server.go#L190-L195 will say, "OK, I will not close datastream for reading, maybe the client will do a tls handshake through me".

Please correct me any misunderstanding.

marten-seemann: I don't think you need to enforce this, since a CONNECT request already contains a body.

phuslu: In some cases, (e.g. my POC codes), I cannot set req.Body filed before RoundTrip it.
Doing a preRead should solve this, but it maybe bring other side effects.

marten-seemann: I don't see how a CONNECT with a `request.Body = nil` even makes sense. You won't be able to proxy any data after all.

Please have a look at the standard library H2 implementation here: https://github.com/golang/net/blob/master/http2/transport.go#L769. They don't need any special cases for CONNECT, and I'm not convinced we need one in QUIC.

phuslu: Oh, Thanks! Agree that. I will remove that.

phuslu: I removed this change and changed PR's description.

For the left `path` checking, does it make sense?

h2quic/request.go
```diff
@@ -37,7 +37,7 @@ func requestFromHeaders(headers []hpack.HeaderField) (*http.Request, error) {
 		httpHeaders.Set("Cookie", strings.Join(httpHeaders["Cookie"], "; "))
 	}
 
-	if len(path) == 0 || len(authority) == 0 || len(method) == 0 {
+	if (len(path) == 0 && method != http.MethodConnect) || len(authority) == 0 || len(method) == 0 {
```
marten-seemann: This looks fine, this is what the H2 standard specifies (https://http2.github.io/http2-spec/index.html#CONNECT).

There's one thing I'd like to check before approving this PR. We're parsing the path into an URL a couple of lines later here. Parsing an empty string works fine with `url.Parse`, however, I'm wondering if the returned `http.Request` should have a `nil` value for `http.Request.URL`, or this empty URL. Could you setup an example with a Go H2 server to check how the standard library handles that (please paste the code here or a link to a Go playground file)?

Issue #770, opened 2017-07-26
Add client support for server push


twdkeule:
I tried a setup where a Caddy proxy and a Caddy upstream server were using QUIC,
This fails as soon as the server pushes a resource to the client:
```
Sending PUSH_PROMISE for target '/dash.all.debug.js' on stream 3, promised on stream 2
2017/07/26 13:39:32.775762 -> Sending packet 0x9 (175 bytes) for connection 4c0eabfc5dac0c83, forward-secure
2017/07/26 13:39:32.775782 	-> &frames.StreamFrame{StreamID: 3, FinBit: false, Offset: 0x0, Data length: 0x9e, Offset + Data length: 0x9e}
2017/07/26 13:39:32.779200 Responding with 200
2017/07/26 13:39:32.779292 -> Sending packet 0xa (141 bytes) for connection 4c0eabfc5dac0c83, forward-secure
2017/07/26 13:39:32.779313 	-> &frames.StreamFrame{StreamID: 3, FinBit: false, Offset: 0x9e, Data length: 0x7a, Offset + Data length: 0x118}
2017/07/26 13:39:32.779454 -> Sending packet 0xb (1350 bytes) for connection 4c0eabfc5dac0c83, forward-secure
2017/07/26 13:39:32.779475 	-> &frames.StreamFrame{StreamID: 2, FinBit: false, Offset: 0x0, Data length: 0x535, Offset + Data length: 0x535}
2017/07/26 13:39:32.779513 -> Sending packet 0xc (1350 bytes) for connection 4c0eabfc5dac0c83, forward-secure
2017/07/26 13:39:32.779537 	-> &frames.StreamFrame{StreamID: 2, FinBit: false, Offset: 0x535, Data length: 0x533, Offset + Data length: 0xa68}
2017/07/26 13:39:32.779568 -> Sending packet 0xd (1350 bytes) for connection 4c0eabfc5dac0c83, forward-secure
2017/07/26 13:39:32.779611 	-> &frames.StreamFrame{StreamID: 2, FinBit: false, Offset: 0xa68, Data length: 0x533, Offset + Data length: 0xf9b}
2017/07/26 13:39:32.779642 -> Sending packet 0xe (1350 bytes) for connection 4c0eabfc5dac0c83, forward-secure
[...]

PR #771, opened 2017-07-26
Initial client push support [WIP]

twdkeule:
Client does not handle receiving pushed data
Includes logging

Attempt at fixing #770


twdkeule:
It works for 1 small file (1,6 kB) but not for a large one (1,8MB).
It might be a deadlock but I'm not sure. It might also have something to do with the ReadCloser interface for the datastream of the pushed resource.
Any help would be much appreciated.

marten-seemann:
Do you have a debug log of the failing transfer (the one with the 1.8 MB)? 

I could imagine a couple of reasons why this fails: the stream could be reset, or it could be a flow control problem, or maybe we actually have a deadlock in our stream implementation. I'd be surprised to see the last two though, since that would probably show up at other places as well, but you never know, often a new use case brings up bugs that were hidden before.

twdkeule:
This is the client's log (Caddy proxy acting as client): [client.txt](https://github.com/lucas-clemente/quic-go/files/1179305/client.txt)
And the server's log: [server.txt](https://github.com/lucas-clemente/quic-go/files/1179306/server.txt)

Some general remarks/to-do's:
- Roundtrip() should check if a push_promise has been received for the request and if it has, not request the response but for the data. Right now it will only check if a response (the 200 header) has already been receive.
- Looking at the  Chromium source code for how the client handles push promises might be interesting.

marten-seemann:
Thanks for the logs. Turns out one of my guesses was right :) The client probably isn't reading the resource pushed by the server, is it? That means that at some point flow control kicks in and prevents new data from being sent. You can see at the end of the server log, the server is getting flow control limited and send a `frames.BlockedFrame{StreamID:0x2}`. That also explains why you only succeed pushing small files: They fit into the initial flow control window (32 kb).

I'm not really sure how to correctly handle Pushes in the client (I haven't thought about it in detail yet). Currently, the `RoundTripper` doesn't store the path of the requests it's executing, which we probably would need to do if we want to speed up `RoundTrip()` by returning pushed resources, right?
Do you know what the Push implementation in the standard library is doing in those cases? I think it makes the most sense that `h2quic` mirros this behavior as closely as possible.

twdkeule:
Ah, yes. The `Roundtripper` leaves the body open and does not read from it. 
So it should probably read it and store it in a new body.

The modified `Roundtripper` stores the request paths in the new `pushedResponses` map and for every new request first checks if it isn't in there.

twdkeule:
So I quickly used a `bytes.Buffer` wrapper with a `Close()` function to act as the response body so it can read the dataStream and it works in my case!
However I still don't know when to throw away the pushed resources if they are not requested.

marten-seemann:
I'm not sure about that. How does the standard library handle this case?

twdkeule:
Poorly:
```
func (rl *clientConnReadLoop) processPushPromise(f *PushPromiseFrame) error {
	// We told the peer we don't want them.
	...
	return ConnectionError(ErrCodeProtocol)
}
```

marten-seemann:
Maybe that's how we should handle it as well? We just need to make sure that the client actually sends the correct setting (i.e. disabling push), then we don't have to deal with it.
By the way, does your server push PR handle this setting correctly?

twdkeule:
But this is our chance to get ahead of the standard library! ;-)

I think we should by default disable receiving push (like the std lib), but allow a config setting to enable it. We can look at the Chromium implementation for inspiration on how to continue. 
It would be a waste to throw a working (experimental) implementation away.

I will make a new PR for the disabled receive and keep this one open for further development.

twdkeule:
Strike that, it would require implementing the http2 SETTINGS frame. 
For now I will leave this PR as is (concerning the enabling/disabling).

marten-seemann:
There's currently an open issue in the Go issue tracker to support H2 Pushes in the client: https://github.com/golang/go/issues/18594. Apparently this is planned for Go 1.10 (which will be released in early 2018).

Considering that they're still discussing the right API, I think we'll be best off waiting what they decide for and then follow suit with our implementation.

twdkeule:
Yes, that's probably be the best approach. 
I'm closing this PR then.

PR #773, opened 2017-07-28
return an error when Write is called on a closed stream

marten-seemann:


stream.go
```diff
@@ -180,6 +181,9 @@ func (s *stream) Write(p []byte) (int, error) {
 	if s.resetLocally.Get() || s.err != nil {
 		return 0, s.err
 	}
+	if s.finishedWriting.Get() {
+		return 0, errors.New("write on closed stream")
```
lucas-clemente: `fmt.Errorf("write on closed stream %d", s.streamID)`

PR #774, opened 2017-07-29
implement a context for streams and for the session

marten-seemann:
This came up in #698 and is a prerequisite for implementing the `http.CloseNotifier` in h2quic.

### DIFF
--- a/session.go
+++ b/session.go
@@ -1,6 +1,7 @@
 package quic
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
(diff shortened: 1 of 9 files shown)


session.go
```diff
@@ -333,12 +334,12 @@ runLoop:
 		s.handshakeChan <- handshakeEvent{err: closeErr.err}
 	}
 	s.handleCloseError(closeErr)
-	close(s.runClosed)
+	s.ctxCancel()
```
lucas-clemente: I know we shouldn't return earlier, but I'd feel safer if this was deferred at the beginning of the func.

session.go
```diff
@@ -543,7 +544,7 @@ func (s *session) closeRemote(e error) {
 // It waits until the run loop has stopped before returning
 func (s *session) Close(e error) error {
 	s.closeLocal(e)
-	<-s.runClosed
+	<-s.ctx.Done()
```
lucas-clemente: Nice :)

session.go
marten-seemann: Well, it's literally one line above the `return`, but I'm fine with this if it makes you feel safer.

PR #778, opened 2017-08-02
disable verbose logging for integration tests on Travis

marten-seemann:
@lucas-clemente: Do you know why Chrome is dumping so much debug information into the Travis log? I disabled both `outWriter` and `errWriter` in `gexec.Start`, but that doesn't seem to change anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
   - export GOARCH=$TRAVIS_GOARCH
   - go env # for debugging
   - "export DISPLAY=:99.0"
-  - "Xvfb $DISPLAY &"
+  - "Xvfb $DISPLAY &> /dev/null &"
 
 script:
   - .travis/script.sh
--- a/integrationtests/chrome/chrome_suite_test.go
+++ b/integrationtests/chrome/chrome_suite_test.go
@@ -142,7 +142,7 @@ func chromeTestImpl(version protocol.VersionNumber, url string, blockUntilDone f
 	}
 	utils.Infof("Running chrome: %s '%s'", getChromePath(), strings.Join(args, "' '"))
 	command := exec.Command(path, args...)
-	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+	session, err := gexec.Start(command, nil, nil)
 	Expect(err).NotTo(HaveOccurred())
 	defer session.Kill()
 	const pollInterval = 100 * time.Millisecond
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -4,17 +4,17 @@ set -e
 
 go get -t ./...
 if [ ${TESTMODE} == "unit" ]; then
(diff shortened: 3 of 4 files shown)


integrationtests/chrome_test.go
```diff
@@ -58,7 +58,7 @@ func chromeTestImpl(version protocol.VersionNumber, url string, blockUntilDone f
 	}
 	utils.Infof("Running chrome: %s '%s'", getChromePath(), strings.Join(args, "' '"))
 	command := exec.Command(path, args...)
-	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+	session, err := gexec.Start(command, nil, nil)
```
lucas-clemente: I don't like this, I'd like to have the Chrome log if it fails. Why did you change this?

lucas-clemente:
It's probably because Chrome starts subprocesses without forking. I don't think there's anything we can do about it, and tbh I don't really care. I only look at logs if something fails, and then it might be useful.

integrationtests/chrome_test.go
marten-seemann: A desperate attempt to make Chrome shut up.

marten-seemann:
It's a lot of spam in our logs. It's annoying to find what is failing in all the Chrome debug output.

lucas-clemente:
I don't think there's a nice solution for that.

marten-seemann:
@lucas-clemente: There is. The output was not coming from Chrome, but from the Xvfb. I fixed that now.

marten-seemann:
@lucas-clemente: This is ready for merging now. Please review.

.travis.yml
```diff
@@ -31,7 +31,7 @@ before_install:
   - export GOARCH=$TRAVIS_GOARCH
   - go env # for debugging
   - "export DISPLAY=:99.0"
-  - "Xvfb $DISPLAY &"
+  - "Xvfb $DISPLAY > /dev/null 2>&1 &"
```
lucas-clemente: You can use &> to redirect both stdout and stderr at the same time.

integrationtests/chrome/chrome_suite_test.go
```diff
@@ -142,7 +142,7 @@ func chromeTestImpl(version protocol.VersionNumber, url string, blockUntilDone f
 	}
 	utils.Infof("Running chrome: %s '%s'", getChromePath(), strings.Join(args, "' '"))
 	command := exec.Command(path, args...)
-	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+	session, err := gexec.Start(command, nil, nil)
```
lucas-clemente: I'm still not convinced dropping Chrome logs is the right thing to do, but if you insist…

marten-seemann: The problem is that Ginkgo dumps the output of the gexec [errWriter](https://godoc.org/github.com/onsi/gomega/gexec#Start) if we run the tests with -v.

PR #781, opened 2017-08-03
Refactor inttests

### LINKED ISSUE #743: Refactor integration tests
- [x] Split up into packages by type (quic-go, quic_{client,server}, chrome)
- [x] Split up the test files
- [x] Remove datamanager, replace with PRNG
- [ ] Make tests race free

lucas-clemente:
Fixes #743.

The datamanager is still in there, will pull that out in a future PR.


marten-seemann:
LGTM 👍🏻

I don't think that `gquic` is a good package name, since we already started moving towards IETF QUIC.
What do you think about moving the `benchmark` package here as well?

2 nits regarding the commit history:
- The creation of the `testserver` package and the usage of that package should be in the same commit.
- Creating the Go 1.8 files should be done when you create the new packages.

lucas-clemente:
The idea is to have a split between tests with Google QUIC (quic_{server,client}) and our self-tests, e.g. the handshake rtt test. I picked the name gquic as it seems like the easiest term that sums up quic_server and quic_client, feel free to suggest another name.

We can move the benchmark suite here, but I'd like to keep it a separate package.

Regarding the testserver, I don't know what you mean. I created the package in the same commit as factoring out the Chrome tests (and they use it).

I rebased the Go 1.8 files.

marten-seemann:
What about h2quic or so?

lucas-clemente:
If we have self-tests for h2quic we should add them to self/, so I dislike the name h2quic here. I think the issue here is that Google QUIC can mean either the spec or the implementation. As you say, the spec is going away, so gQUIC in the future will mean Google's implementation 🙃

marten-seemann:
Does "self" mean quic-go talking to quic-go? Or quic-go speaking plain QUIC (without the H2 layer)?

lucas-clemente:
The former, quic-go 2 quic-go. I don't like the name but I didn't find a better one.

marten-seemann:
Then I think you forgot to move some tests there. We already have a bunch of tests where the quic-go server talks to the quic-go client.

lucas-clemente:
Done and removed dataManager, PTAL.

PR #782, opened 2017-08-04
add support for QUIC 39

### LINKED ISSUE #500: Add support for big-endian encoding
We can do some work on this (testing with our client & server) even before Chromium supports it.

### LINKED ISSUE #530: add support for QUIC 39
- [x] implement big-endian Public Header parsing / writing
- [x] implement big-endian frame parsing / writing
- [x] don't ack ACKs, send retransmittable packets at least every 20 packets

see https://codereview.chromium.org/2809773002/

marten-seemann:
Fixes #530, fixes #500.

We should support QUIC 38 (#784) before merging this.

The way to force the sending of retransmittable packets is not ideal:

1. Chromium sends a connection-level WINDOW_UPDATE instead of a PING frame. Obviously, this is more efficient.
2. We immediately send the 20th packet containing the PING, and don't wait for the next ACK to be queued.


marten-seemann:
@lucas-clemente: I rebased the branch and activated version 39.
There seems to be a problem with the Chrome tests on Travis though (probably the Chrome we're using on Travis doesn't support 39 yet). Can you have a look at that please?

public_header.go
```diff
@@ -110,6 +110,35 @@ func (h *PublicHeader) Write(b *bytes.Buffer, version protocol.VersionNumber, pe
 	return nil
 }
 
+// ParseConnectionID parses the connection ID from a QUIC packet's public header.
+// It no error occurs, it restores the read position in the bytes.Reader.
```
lucas-clemente: s/It/If/

public_header.go
```diff
@@ -110,6 +110,35 @@ func (h *PublicHeader) Write(b *bytes.Buffer, version protocol.VersionNumber, pe
...
+func ParseConnectionID(b *bytes.Reader, packetSentBy protocol.Perspective) (protocol.ConnectionID, error) {
```
lucas-clemente: Maybe rename this to `PeekConnectionID`?

protocol/version.go
```diff
@@ -9,8 +9,10 @@ const (
 	Version36
 	Version37
 	Version38
+	Version39
 	VersionWhatever    VersionNumber = 0 // for when the version doesn't matter
 	VersionUnsupported VersionNumber = -1
+	VersionUnknown     VersionNumber = -2
```
lucas-clemente: I'm not a big fan of having 3 sentinel values in this enum. I get what you use them for, but can we perhaps pass the information out of line (as a separate bool in the ParseHeader func), or just merge Unknown, Unsupported and Whatever?

session_test.go
```diff
@@ -655,6 +655,11 @@ var _ = Describe("Session", func() {
 		close(done)
 	})
 
+	It("tells it's versions", func() {
```
lucas-clemente: s/it's/its/

ackhandler/sent_packet_handler.go
```diff
@@ -89,6 +91,10 @@ func (h *sentPacketHandler) largestInOrderAcked() protocol.PacketNumber {
 	return h.LargestAcked
 }
 
+func (h *sentPacketHandler) ShouldSendRetransmittablePacket() bool {
+	return h.numNonRetransmittablePackets >= 19
```
lucas-clemente: No magic numbers, please make this a const.

public_header.go
marten-seemann: That sounds much better. Thanks!

protocol/version.go
marten-seemann: `VersionWhatever` is only intended for tests (although it's also used for writing the Version Negotiation Packet). We could consider moving that to the test code, so that it's not available in the production code anymore (if Go allows that).
I don't think that we should get rid of `VersionUnsupported`, since this is used for doing the version negotiation on the client side.

I'm not sure what the right solution here is. Maybe keep `VersionWhatever` and `VersionUnsupported`, and use the bool instead of `VersionUnknown`, as you suggested. What do you think?

marten-seemann: @lucas-clemente: What do you want to do about this? I'd like to merge this PR soon.

lucas-clemente: You can leave it as-is for now, but we should think about merging the three special enum values at some point.

PR #783, opened 2017-08-04
use io.ReadFull instead of Read to read into known length slices

marten-seemann:
No functional changes expected.

### DIFF
--- a/frames/stream_frame.go
+++ b/frames/stream_frame.go
@@ -3,6 +3,7 @@ package frames
 import (
 	"bytes"
 	"errors"
+	"io"
 
 	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/lucas-clemente/quic-go/protocol"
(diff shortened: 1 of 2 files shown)


frames/stream_frame.go
```diff
@@ -70,11 +71,7 @@ func ParseStreamFrame(r *bytes.Reader) (*StreamFrame, error) {
 	}
 	if dataLen != 0 {
 		frame.Data = make([]byte, dataLen)
-		n, err := r.Read(frame.Data)
-		if n != int(dataLen) {
-			return nil, errors.New("BUG: StreamFrame could not read dataLen bytes")
-		}
-		if err != nil {
+		if _, err := io.ReadFull(r, frame.Data); err != nil {
```
lucas-clemente: I don't like this particular change as it gets rid of a helpful error message in favor of a generic EOF, but we don't have helpful messages in many places where we should…

marten-seemann: Well, this error message was misleading anyway. It's not a bug (on our side) if the data was not long enough.

PR #791, opened 2017-08-21
Run gofmt -s and gosimple

lucas-clemente:


stream.go
```diff
@@ -134,7 +134,7 @@ func (s *stream) Read(p []byte) (int, error) {
 			} else {
 				select {
 				case <-s.readChan:
-				case <-time.After(deadline.Sub(time.Now())):
+				case <-time.After(time.Until(deadline)):
```
marten-seemann: This isn't really easy to read.

lucas-clemente: True, but I think it's still easier than the old version.

marten-seemann: The alternative would be `time.NewTimer(deadline).C`. I'm fine with either.

lucas-clemente: That would cause more garbage, so I'd keep it as is.

PR #792, opened 2017-08-22
add a quic.Config option to configure the idle timeout

### LINKED ISSUE #441: Make idle state timeout configurable
I am using QUIC in a non-HTTP setup, and I need to alter the idle state timeout, but I cannot from what is exposed. Can this be made configurable? I hesitate to do it myself as it appears a lot of things just got upheaved in #420.

Also, is there a place for discussion besides these issues?

marten-seemann:
Fixes #441.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -100,6 +100,12 @@ type Config struct {
 	// If the timeout is exceeded, the connection is closed.
 	// If this value is zero, the timeout is set to 10 seconds.
 	HandshakeTimeout time.Duration
+	// IdleTimeout is the maximum duration that may pass without any incoming network activity.
+	// This value only applies after the handshake has completed.
+	// Before that, the idle timeout is set to half the duration of the HandshakeTimeout.
+	// If the timeout is exceeded, the connection is closed.
+	// If this value is zero, the timeout is set to 30 seconds.
+	IdleTimeout time.Duration
 	// AcceptSTK determines if an STK is accepted.
 	// It is called with stk = nil if the client didn't send an STK.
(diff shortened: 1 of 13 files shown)


session.go
```diff
@@ -371,7 +371,7 @@ func (s *session) idleTimeout() time.Duration {
 	if s.handshakeComplete {
 		return s.connectionParameters.GetIdleConnectionStateLifetime()
 	}
-	return protocol.InitialIdleTimeout
+	return s.config.HandshakeTimeout / 2
```
lucas-clemente: No magic numbers… Either move the 2 to the top of the file (or server_parameters.go) or add a comment.

interface.go
```diff
@@ -100,6 +100,12 @@ type Config struct {
 	// If the timeout is exceeded, the connection is closed.
 	// If this value is zero, the timeout is set to 10 seconds.
 	HandshakeTimeout time.Duration
+	// IdleTimeout is the maximum duration that may pass without any network activity.
```
lucas-clemente: any *incoming* network activity

interface.go
```diff
@@ -100,6 +100,12 @@ type Config struct {
...
+	// This value only applies after the handshake is completed.
```
lucas-clemente: Nit: is complete or has completed.

session.go
marten-seemann: This also breaks the handshake RTT test, where we test the handshake timeout. I'm not sure what to do about this, it seems like the only way to solve this is making the initial idle timeout configurable as well.

marten-seemann:
I removed the idle timeout before the handshake has completed.
@lucas-clemente: PHAL.

PR #794, opened 2017-08-23
make the integration tests race free

### LINKED ISSUE #558: Make integration tests race free
The current setup with the start and stop functions is race-prone and pollutes Go's race detection output. We should figure out a way to run the tests without races.

marten-seemann:
Fixes #558.

This PR fixes a bunch of race conditions in the integration tests. We're now at a point where can run the whole test suite with the Go race detector.

However, I'm not what the right way to do so is. Running the tests with the race detector takes considerably longer (see the Travis build log of the last commit), slowing down our Travis build even more.

### DIFF
--- a/integrationtests/chrome/chrome_suite_test.go
+++ b/integrationtests/chrome/chrome_suite_test.go
@@ -38,6 +38,7 @@ var (
 	nFilesUploaded     int32 // should be used atomically
 	testEndpointCalled utils.AtomicBool
 	doneCalled         utils.AtomicBool
+	version            protocol.VersionNumber
 )
 
 func TestChrome(t *testing.T) {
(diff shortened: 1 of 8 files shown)


lucas-clemente:
> Running the tests with the race detector takes considerably longer (see the Travis build log of the last commit), slowing down our Travis build even more.

I can't think of a good way to speed up travis. We could maybe restrict our usage of multi-version tests… Does your workflow contain waiting for travis to build – maybe we can improve tooling?

marten-seemann:
Only for merging PRs. I'd like to have all 4 CI builds succeed before merging a PR.
Sometimes, when we rebase and merge a bunch of stuff, our Travis queue becomes rather long, so that a build might have to wait for an hour or so before being built. Adding slow Travis jobs doesn't help solve that problem.

I'm not really sure what to do about it. Restricting the QUIC versions tested will require a lot of extra code in the integration tests, so maybe we should just leave it as is.

marten-seemann:
The CI build fails due to #803. We should fix that one before merging.

marten-seemann:
@lucas-clemente: I removed the commit that ran the integration tests with the race detector on Travis.
As we decided in the morning, we can now merge this PR. The integration tests should now be race free. We can then activate the race detector on the CIs in a separate PR.

lucas-clemente:
sgtm

PR #798, opened 2017-08-25
use Go 1.9 on CIs

marten-seemann:


.travis.yml
```diff
@@ -8,7 +8,7 @@ language: go
 
 go:
   - 1.8.x
-  - 1.9rc2
+  - 1.9
```
lucas-clemente: This should be 1.9.x, see https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use.

marten-seemann: I tried that, but unfortunately, 1.9.x doesn't match 1.9, see https://travis-ci.org/lucas-clemente/quic-go/builds/268253712.

lucas-clemente: Okay, LGTM then. For the future, this would have been a helpful addition to the commit msg ;)

PR #801, opened 2017-08-27
make sure that the SHLO was sent before returning Listener.Accept

### LINKED ISSUE #800: undecryptable packets in quic-conn integration tests
[log.txt](https://github.com/lucas-clemente/quic-go/files/1254340/log.txt)

marten-seemann:
Fixes #800.

The test occasionally failed due to a race condition: In the server crypto setup, we would send `protocol.EncryptionForwardSecure` on the `aeadChan` as soon as the SHLO was composed, but before it was written to the crypto stream. This lead to `Listener.Accept` returning the session already. If the server was ready to write a lot of data then, this data could be sent before the crypto setup would write on the crypto stream, therefore sending a lot of undecryptable packets to the client, which would eventually lead to a connection error (too many undecryptable packets).

Note that this doesn't fully solve the problem. If the packet containing the SHLO is lost, the client might still receive a lot of undecryptable packets before receiving the  retransmission of the SHLO.  The only way to solve this on the server-side is to not return `Listener.Accept` until it received the first forward-secure packet, but this would add another RTT to the connection establishment.


handshake/crypto_setup_server.go
```diff
@@ -193,7 +194,6 @@ func (h *cryptoSetupServer) Open(dst, src []byte, packetNumber protocol.PacketNu
 		if err == nil {
 			if !h.receivedForwardSecurePacket { // this is the first forward secure packet we receive from the client
 				h.receivedForwardSecurePacket = true
-				close(h.aeadChanged)
```
lucas-clemente: Just to be clear, this change removes the additional RTT (waiting for the client to send its first FW packet) you describe in the PR msg? If we didn't change this, we would be correct, but with an additional RTT?

marten-seemann: No. We already returned the session as soon as the encryption level was forward-secure in https://github.com/lucas-clemente/quic-go/blob/master/server.go#L307. We never waited for the handshake to complete (which, according to the old definition, was when we received the first forward-secure packet from the client).

lucas-clemente: Ok, I thought we were waiting for the chan to be closed.

lucas-clemente: But you are right that this changes our definition of handshake completeness. This also affects the timeout. Do we want that?

marten-seemann: I don't have a strong opinion about the timeout.

Once we move to TLS 1.3, we will have to wait for the TLS Finished message (which would be roughly equivalent to receiving the forward-secure packet now). The question is, when do we want to return a connection in `Listener.Accept`?

marten-seemann:
@lucas-clemente: I just changed back the behaviour of the server crypto setup such that the `aeadChan` is only closed when we receive a forward-secure packet. This way, this PR doesn't change the definition of a completed handshake (and the associated timeout) anymore.
However, it was necessary to introduce additional synchronisation in the crypto setup. Because `Open` and `handleMessage` are executed concurrently, we must make sure, that `protocol.EncryptionForwardSecure` is sent on the `aeadChan` before `Open` closes the channel.

PR #804, opened 2017-08-28
delete empty files in the integration tests folder

marten-seemann:
These empty files were necessary in Go 1.7, but are obsolete now.

### DIFF
--- a/integrationtests/chrome/chrome.go
+++ b/integrationtests/chrome/chrome.go
@@ -1,1 +1,0 @@
-package chrome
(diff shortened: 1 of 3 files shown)


marten-seemann:
Apparently this is not true. It's still needed in Go 1.8, but obsolete for 1.9.
So we need to keep those empty files around until dropping support for 1.8.

lucas-clemente:
Yes. I tried that during my PR but it broke travis.

marten-seemann:
@lucas-clemente: This can be merged now. Please approve.

PR #807, opened 2017-08-29
use type aliases to clean up the API

marten-seemann:


interface.go
```diff
@@ -6,9 +6,18 @@ import (
...
 
-	"github.com/lucas-clemente/quic-go/protocol"
+	"github.com/lucas-clemente/quic-go/internal/protocol"
 )
 
+// The StreamID is the ID of a QUIC stream.
+type StreamID = protocol.StreamID
+
+// A ByteCount is a number of bytes.
+type ByteCount = protocol.ByteCount
```
marten-seemann: Do we really want to export a `ByteCount` type?
The standard library usually uses a built-in type there, for example an `int` in the [http.Server](https://golang.org/pkg/net/http/#Server) or a `int64` in the [http.Request](https://golang.org/pkg/net/http/#Request).

marten-seemann: @lucas-clemente: What do you think?

README.md
```diff
@@ -16,7 +16,7 @@ As Google's QUIC versions are expected to converge towards the [IETF QUIC draft]
 
 ## Guides
 
-We currently support Go 1.8+.
+We currently support Go 1.9+.
```
lucas-clemente: We should probably also add this to the changelog.

interface.go
lucas-clemente: Using a signed type looks weird, but I guess we can do that here too. We just need to check the sign everywhere :/

I'm fine with using uint{32,64} though.

marten-seemann: I just tried that. It doesn't seem to be that easy:
When I change the type in the `quic.Config` to `uint64`, we'd have to use `uint64`s all across the quic package. We could declare `type uint64 = protocol.ByteCount`, but that is displayed slightly differently on GoDoc (built in types are usually linked to the GoDoc of the builtin package, but this uint64 would not).

PR #810, opened 2017-09-05
implement a very basic TLS 1.3 handshake

marten-seemann:


crypto/aesgcm_aead.go
```diff
@@ -17,31 +16,34 @@ type aeadAESGCM struct {
...
+// NewAEADAESGCM creates a AEAD using AES-GCM
 func NewAEADAESGCM(otherKey []byte, myKey []byte, otherIV []byte, myIV []byte) (AEAD, error) {
-	if len(myKey) != 16 || len(otherKey) != 16 || len(myIV) != 4 || len(otherIV) != 4 {
-		return nil, errors.New("AES-GCM: expected 16-byte keys and 4-byte IVs")
+	// the IVs need to be at least 8 bytes long, otherwise we can't compute the nonce
+	if len(otherIV) != ivLen || len(myIV) != ivLen {
+		return nil, errors.New("AES-GCM: expected 12 byte IVs")
 	}
-	encrypterCipher, err := aes12.NewCipher(myKey)
+
```
lucas-clemente: Is there a reason why you're not checking the key size here (anymore)?

crypto/key_derivation.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/bifurcation/mint"
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+)
+
+const (
+	clientExporterLabel = "EXPORTER-QUIC client 1-RTT Secret"
+	serverExporterLabel = "EXPORTER-QUIC server 1-RTT Secret"
+)
+
+// MintState contains
```
lucas-clemente: Nit

crypto/key_derivation.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	otherKey, otherIV, err := computeKeyAndIV(ms, otherLabel)
+	if err != nil {
+		return nil, err
+	}
+	return NewAEADAESGCM(otherKey, myKey, otherIV, myIV)
+}
+
+func computeKeyAndIV(ms MintState, label string) (key, iv []byte, err error) {
+	cs := ms.GetCipherSuite()
```
lucas-clemente: Should we check here that this is actually AES-GCM?

handshake/crypto_setup_tls.go
```diff
@@ -0,0 +1,140 @@
...
+	if perspective == protocol.PerspectiveServer {
+		conn = mint.Server(&fakeConn{cryptoStream}, mintConf)
+	} else {
+		conn = mint.Client(&fakeConn{cryptoStream}, mintConf)
+	}
+	return &cryptoSetupServerTLS{
+		perspective:   perspective,
+		mintConf:      mintConf,
+		conn:          conn,
+		nullAEAD:      crypto.NewNullAEAD(perspective, protocol.Version39),
```
lucas-clemente: Why is there a version const here?

handshake/crypto_setup_tls.go
```diff
@@ -0,0 +1,140 @@
...
+		keyDerivation: crypto.DeriveAESKeys,
+		aeadChanged:   aeadChanged,
+	}, nil
+}
+
+func (h *cryptoSetupServerTLS) HandleCryptoStream() error {
+	alert := h.conn.Handshake()
+	utils.Debugf("ALERT: %#v", alert)
```
lucas-clemente: Remove this line or move it into the if block below.

crypto/aesgcm_aead.go
marten-seemann: Yes, `aes.NewCipher` can be called with different key sizes, so it performs the check if the key has one of the accepted sizes. See https://golang.org/pkg/crypto/aes/#NewCipher

lucas-clemente: Okay, and mint will check for us that we don't accept weak keys?

marten-seemann: TLS 1.3 doesn't have any weak cipher suites. These are the ones supported by mint: https://github.com/bifurcation/mint/blob/master/common.go#L43-L52.

crypto/key_derivation.go
marten-seemann: We probably should do that. That will mean checking the cipher suite against the values defined in https://github.com/bifurcation/mint/blob/master/common.go#L43-L52 though, I don't see a general way to check for AES.

handshake/crypto_setup_tls.go
marten-seemann: I think I forgot to write the correct null AEAD constructor. I will add this later today.

crypto/key_derivation.go
marten-seemann: Actually, what we should do is set a list of supported ciphers in the `mint.Config`, and only use AES ciphers there. We can then trust mint to negotiate one of them.

handshake/crypto_setup_tls.go
marten-seemann: Fixed.

crypto/aesgcm_aead.go
marten-seemann: I added tests with 16 and 32 byte keys, and another test that we error on invalid key sizes.

marten-seemann:
@lucas-clemente: I think I addressed all your comments. Furthermore, I moved all mint-related methods into one interface, which should make testing easier. PHAL.

PR #817, opened 2017-09-11
improve the QUIC proxy

marten-seemann:


integrationtests/tools/proxy/proxy.go
```diff
@@ -200,21 +200,19 @@ func (p *QuicProxy) runConnection(conn *connection) error {
...
+		if err != nil {
+			return err
+		}
 
-		v := atomic.AddUint64(&conn.outgoingPacketCounter, 1)
+		atomic.AddUint64(&conn.outgoingPacketCounter, 1)
 
-		packetNumber := protocol.PacketNumber(v)
-		if p.dropPacket(DirectionOutgoing, packetNumber) {
+		if p.dropPacket(DirectionOutgoing, hdr.PacketNumber) {
```
lucas-clemente: I'd like to know more about the motivation behind this change. Why is it not sufficient to use the counter for the drop function? Especially with the packet number inferral, I feel like the proxy is becoming quite complicated – and we're going to have to drop all of this anyway when packet number randomization is introduced.

integrationtests/tools/proxy/proxy_test.go
```diff
@@ -34,13 +32,9 @@ var _ = Describe("QUIC Proxy", func() {
...
 	}
 
-	BeforeEach(func() {
-		serverAddr = "localhost:7331"
-	})
-
 	Context("Proxy setup and teardown", func() {
 		It("sets up the UDPProxy", func() {
-			proxy, err := NewQuicProxy("localhost:0", protocol.VersionWhatever, Opts{RemoteAddr: serverAddr})
+			proxy, err := NewQuicProxy("localhost:0", protocol.VersionWhatever, Opts{})
```
lucas-clemente: s/Opts{}/nil and a default value if opts == nil.

integrationtests/tools/proxy/proxy.go
marten-seemann: Because the interval drop test doesn't work as intended when we don't have the correct packet number. Inferring should be easy here, because there won't be packet loss on localhost.

How do you think packet number randomization will affect this? We can still read the 32 bit packet number on the first packet, and start inferring from there. We will have to do some changes in the drop test, but that's independent.

marten-seemann: Thinking further about that, we could get rid of the Public Header parsing completely and just pass a "packet count" to the delay and drop callbacks. We won't need any changes to the drop tests then.

lucas-clemente: Yes, that was my intention. I don't see a difference between counts (what we use now), or the actual pnum for the purpose of our tests.

marten-seemann:
@lucas-clemente: PHAL.

PR #820, opened 2017-09-12
add an integration test with dropped packets during the handshake

### LINKED ISSUE #392: add integration tests with dropped packets during the crypto handshake
This should work already, but will be an excellent test for #391.

marten-seemann:
Fixes #392.

Should be merged after #819.

### DIFF
--- a/integrationtests/gquic/drop_test.go
+++ b/integrationtests/gquic/drop_test.go
@@ -16,6 +16,8 @@ import (
 	. "github.com/onsi/gomega/gexec"
 )
 
+var directions = []quicproxy.Direction{quicproxy.DirectionIncoming, quicproxy.DirectionOutgoing, quicproxy.DirectionBoth}
+
 var _ = Describe("Drop tests", func() {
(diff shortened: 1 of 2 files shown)


integrationtests/gquic/drop_test.go
```diff
@@ -22,74 +23,115 @@ var directions = []quicproxy.Direction{quicproxy.DirectionIncoming, quicproxy.Di
...
-			Context(fmt.Sprintf("with QUIC version %d", version), func() {
+	for _, v := range protocol.SupportedVersions {
+		version := v
+
+		Context(fmt.Sprintf("with QUIC version %d", version), func() {
+			Context("during the crypto handshake", func() {
+				for _, d := range directions {
+					direction := d
+
+					It("establishes a connection when the first packet is lost", func() {
```
lucas-clemente: The direction should be in the It string. Also in the next 2 cases.

marten-seemann: Done.

PR #824, opened 2017-09-15
fix version negotiation

### LINKED ISSUE #823: version negotiation broken
```
GET https://10.1.1.164:6121/index.html
Starting new connection to 10.1.1.164 ([::]:52815 -> 10.1.1.164:6121), connectionID 75b565747fd19055, version 39
Switching to QUIC version 38. New connection ID: 48edd5af3829c7d1
Internal error: closing session in order to recreate it with a new version
Closing session with error: closing session in order to recreate it with a new version
```

marten-seemann:
Fixes #823.

This was probably broken for a long time. I fixed the client code and added an integration test.

### DIFF
--- a/client.go
+++ b/client.go
@@ -24,8 +24,9 @@ type client struct {
 
 	handshakeChan <-chan handshakeEvent
 
-	versionNegotiationChan chan struct{} // the versionNegotiationChan is closed as soon as the server accepted the suggested version
-	versionNegotiated      bool          // has version negotiation completed yet
+	versionNegotiationChan           chan struct{} // the versionNegotiationChan is closed as soon as the server accepted the suggested version
+	versionNegotiated                bool          // has version negotiation completed yet
+	receivedVersionNegotiationPacket bool
 
 	tlsConf *tls.Config
 	config  *Config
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -181,11 +178,37 @@ func populateClientConfig(config *Config) *Config {
...
+	if err := c.createNewSession(nil); err != nil {
+		return err
+	}
 	go c.listen()
 
+	var runErr error
+	errorChan := make(chan struct{})
+	go func() {
+		// session.run() returns as soon as the session is closed
+		for {
```
lucas-clemente: Should we have some logic here to stop retrying after some number of tries?

client_test.go
```diff
@@ -268,25 +321,45 @@ var _ = Describe("Client", func() {
...
+				Eventually(func() bool { return firstSession.closed }).Should(BeTrue())
+				Expect(firstSession.closeReason).To(Equal(errCloseSessionForNewVersion))
+				Consistently(func() bool { return secondSession.closed }).Should(BeFalse())
+				Expect(cl.connectionID).ToNot(BeEquivalentTo(0x1337))
 				Expect(negotiatedVersions).To(Equal([]protocol.VersionNumber{newVersion}))
+
+				handshakeChan <- handshakeEvent{encLevel: protocol.EncryptionSecure}
+				Eventually(established).Should(BeClosed())
+
+				// cl.handlePacket(nil)
```
lucas-clemente: No commented code please

client.go
marten-seemann: There must not be more than one try. I'll fix that.

marten-seemann:
@lucas-clemente : I made sure that we only accept a single version negotiation packet (the last commit). PHAL.

client.go
lucas-clemente: Can we just get rid of the for loop? I find it very non-intuitive if it just runs twice.

marten-seemann: Done. This code actually looks a lot nicer now.

PR #831, opened 2017-09-20
reject packets with the wrong connection ID in the client

### LINKED ISSUE #825: reject packets with the wrong connection ID in the client

marten-seemann:
Fixes #825.

### DIFF
--- a/client.go
+++ b/client.go
@@ -38,6 +38,8 @@ type client struct {
 }
 
 var (
+	// make it possible to mock connection ID generation in the tests
+	generateConnectionID         = utils.GenerateConnectionID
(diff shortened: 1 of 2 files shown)


marten-seemann:
A Public Reset is not supposed to be sent by the client, see https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#rfc.section.7.7.5.

marten-seemann:
@lucas-clemente: PHAL. I had to dependency-inject the connection ID generation in the client, such that we can mock it in the tests, because all the tests for dialing new connections failed because we now drop packets with the wrong connection ID.

PR #832, opened 2017-09-20
Small fixes in session, h2quic

jbenoist:
- populateServerConfig() doesn't honor the keepalive setting of the caller
- h2quic seems to pass as TCP connection instead of TLS to the http2 stack in the non-quic case


lucas-clemente:
LGTM, but we need you to sign the CLA before merging. Let me know if you have any questions.

jbenoist:
> On Sep 22, 2017, at 3:40 AM, Lucas Clemente <notifications@github.com> wrote:
> 
> LGTM, but we need you to sign the CLA before merging. Let me know if you have any questions.

I signed it actually, but it turns out that the email address of my google account is different (e.g: langton@benoist.name).
If this is problematic, my employer has a corporate CLA that might be able to use. Just let me know.

—
Julien Benoist

jbenoist:
the address should match now, I just updated it.

server.go
```diff
@@ -149,6 +149,7 @@ func populateServerConfig(config *Config) *Config {
 		HandshakeTimeout:                      handshakeTimeout,
 		IdleTimeout:                           idleTimeout,
 		AcceptSTK:                             vsa,
+		KeepAlive:                             config.KeepAlive,
```
marten-seemann: Can you please add a test for this value in the two tests here: https://github.com/lucas-clemente/quic-go/blob/master/server_test.go#L344-L373.

jbenoist:
If the second function you mentioned happens to be the one that checks whether the defaults are correctly used when options are not set, there isn't one for keepalive since type default suffices.

marten-seemann:
@jbenoist: Don't worry about the merge conflict. I'll take care of that.

marten-seemann:
Closing in favor of #842.

PR #843, opened 2017-09-23
reduce bytes_in_flight before reporting packets to congestion

marten-seemann:
According the Loss Recovery Draft (https://quicwg.github.io/base-drafts/draft-ietf-quic-recovery.html#rfc.section.4.7.6), we should reduce the `bytes_in_flight` value before reporting a packet as lost to the congestion controller.

Note that due to this change, we don't reduce the `bytes_in_flight` value for RTOs. This is consistent with the Draft, I suspect this is a bug in the Draft though (https://github.com/quicwg/base-drafts/issues/790). We should probably delay merging this PR until we get feedback on that issue.

### DIFF
--- a/ackhandler/sent_packet_handler.go
+++ b/ackhandler/sent_packet_handler.go
@@ -282,11 +282,12 @@ func (h *sentPacketHandler) detectLostPackets() {
 		}
 	}
 
-	if len(lostPackets) > 0 {
-		for _, p := range lostPackets {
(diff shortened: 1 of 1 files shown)


marten-seemann:
I talked with Jana at the Seattle interim, and I think I now know how to fix our retransmission logic. It will take a lot more than the change here, so I'm closing this PR.

PR #849, opened 2017-09-27
implement a string representation of the version number

marten-seemann:
This will be useful as soon as we start supporting non-gQUIC versions.


marten-seemann:
The string representation for the Alt-Svc isn't particularly readable, especially for the gQUIC versions: https://quicwg.github.io/base-drafts/draft-ietf-quic-http.html#alt-svc-version-hint.

PR #853, opened 2017-09-29
implement the TLS Transport Parameters

marten-seemann:


internal/handshake/params_negotiator.go
```diff
@@ -0,0 +1,104 @@
...
+	var foundInitialMaxData bool
+	var foundInitialMaxStreamID bool
+	var foundIdleTimeout bool
+
+	for _, p := range params {
+		switch p.Parameter {
+		case initialMaxStreamDataParameterID:
+			foundInitialMaxStreamData = true
+			if len(p.Value) != 4 {
+				return fmt.Errorf("wrong length for initial_max_stream_data: %d", len(p.Value))
```
lucas-clemente: Should we include the expected len in the error message?

internal/handshake/params_negotiator.go
```diff
@@ -0,0 +1,104 @@
...
+// newParamsNegotiator creates a new connection parameters manager
+func newParamsNegotiator(pers protocol.Perspective, v protocol.VersionNumber, params *TransportParameters) *paramsNegotiator {
+	h := &paramsNegotiator{}
+	h.perspective = pers
+	h.version = v
+	h.init(params)
+	return h
+}
+
+func (h *paramsNegotiator) SetFromTransportParameters(params []transportParameter) error {
```
lucas-clemente: Do we have to lock the mutex here?

marten-seemann: Yeah, I missed that. Same for `GetTransportParameters`.

internal/handshake/params_negotiator.go
marten-seemann: We can do that.

PR #858, opened 2017-09-29
rename connection ID truncation to connection ID omission

marten-seemann:
The IETF draft uses omit instead of truncate, e.g. here: https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#rfc.section.7.3.

No functional change.


lucas-clemente:
Please fix the proxy before merging.

PR #859, opened 2017-09-29
Update echo example

byronwilliams:
I noticed that the echo example did not function on `master`. This PR provides a working version

### DIFF
--- a/example/echo/echo.go
+++ b/example/echo/echo.go
@@ -31,25 +31,41 @@ func main() {
 
 // Start a server that echos all data on the first stream opened by the client
 func echoServer() error {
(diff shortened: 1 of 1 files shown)


byronwilliams:
I signed it!

marten-seemann:
Hi @byronwilliams, thanks for taking the time for putting together this PR!

However, I don't think it's correct. We removed the `ConnStateCallback` about half a year ago, and replaced it by a (hopefully) easier public API. I suspect you might be running an old version of quic-go, can you please check?

I'm closing this PR now, but feel free to comment or open a new issue.

PR #861, opened 2017-10-01
add an integration test for invalid certificates

marten-seemann:


internal/handshake/crypto_setup_client.go
```diff
@@ -198,15 +198,15 @@ func (h *cryptoSetupClient) handleREJMessage(cryptoData map[Tag][]byte) error {
 		err = h.certManager.Verify(h.hostname)
 		if err != nil {
 			utils.Infof("Certificate validation failed: %s", err.Error())
-			return qerr.ProofInvalid
+			return qerr.Error(qerr.ProofInvalid, err.Error())
```
marten-seemann: @lucas-clemente: Do you see any risk with sending the error message in the CONNECTION_CLOSE?

lucas-clemente: No, but given that most TLS implementations are very silent about errors I'd rather not include it.

PR #862, opened 2017-10-01
implement loss detection for handshake packets

### LINKED ISSUE #496: Handle handshake packets separately in loss recovery
See the wg draft.

marten-seemann:
Fixes #496.

### DIFF
--- a/ackhandler/interfaces.go
+++ b/ackhandler/interfaces.go
@@ -12,6 +12,7 @@ type SentPacketHandler interface {
 	// SentPacket may modify the packet
 	SentPacket(packet *Packet) error
 	ReceivedAck(ackFrame *wire.AckFrame, withPacketNumber protocol.PacketNumber, recvTime time.Time) error
+	SetHandshakeComplete()
 
 	SendingAllowed() bool
 	GetStopWaitingFrame(force bool) *wire.StopWaitingFrame
--- a/ackhandler/sent_packet_handler.go
+++ b/ackhandler/sent_packet_handler.go
@@ -16,8 +16,13 @@ const (
 	// Maximum reordering in time space before time based loss detection considers a packet lost.
 	// In fraction of an RTT.
 	timeReorderingFraction = 1.0 / 8
+	// The default RTT used before an RTT sample is taken.
+	// Note: This constant is also defined in the congestion package.
+	defaultInitialRTT = 100 * time.Millisecond
 	// defaultRTOTimeout is the RTO time on new connections
 	defaultRTOTimeout = 500 * time.Millisecond
+	// Minimum time in the future a tail loss probe alarm may be set for.
+	minTPLTimeout = 10 * time.Millisecond
 	// Minimum time in the future an RTO alarm may be set for.
 	minRTOTimeout = 200 * time.Millisecond
 	// maxRTOTimeout is the maximum RTO time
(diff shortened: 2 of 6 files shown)


ackhandler/interfaces.go
```diff
@@ -12,6 +12,7 @@ type SentPacketHandler interface {
 	// SentPacket may modify the packet
 	SentPacket(packet *Packet) error
 	ReceivedAck(ackFrame *wire.AckFrame, withPacketNumber protocol.PacketNumber, recvTime time.Time) error
+	HandshakeComplete()
```
lucas-clemente: Nit: SetHandshakeComplete, as non-"const" functions should start with a verb.

ackhandler/sent_packet_handler.go
```diff
@@ -16,8 +16,12 @@ const (
 	// Maximum reordering in time space before time based loss detection considers a packet lost.
 	// In fraction of an RTT.
 	timeReorderingFraction = 1.0 / 8
+	// The default RTT used before an RTT sample is taken.
+	defaultInitialRTT = 100 * time.Millisecond
```
lucas-clemente: This is duplicate with the const in the rttstats. Can we change SmoothedRTT to return that there if the value is zero?

ackhandler/sent_packet_handler.go
```diff
@@ -380,6 +407,16 @@ func (h *sentPacketHandler) queuePacketForRetransmission(packetElement *PacketEl
 	h.stopWaitingManager.QueuedRetransmissionForPacketNumber(packet.PacketNumber)
 }
 
+func (h *sentPacketHandler) computeHandshakeTimeout() time.Duration {
+	duration := 2 * h.rttStats.SmoothedRTT()
+	if duration == 0 {
+		duration = 2 * defaultInitialRTT
+	}
+	duration = utils.MaxDuration(duration, minTPLTimeout)
```
lucas-clemente: See above, I'd rather have SmoothedRTT be initialized with the 100ms. Also, why the factor of 2?

ackhandler/sent_packet_handler.go
```diff
@@ -380,6 +407,16 @@ func (h *sentPacketHandler) queuePacketForRetransmission(packetElement *PacketEl
...
+	// exponential backoff
+	return duration << h.handshakeCount
```
lucas-clemente: Add a comment that there's an implicit max for the handshakeCount due to the timeout.

ackhandler/sent_packet_handler.go
marten-seemann: https://quicwg.github.io/base-drafts/draft-ietf-quic-recovery.html#pseudocode

ackhandler/sent_packet_handler.go
marten-seemann: `SmoothedRTT` is used a lot all across quic-go, including the congestion, and compared for 0 at a few places. Changing the return value will cause all those checks to fail, so I don't think this is something we can do.

ackhandler/sent_packet_handler.go
marten-seemann: What do you mean? I don't see an implicit max anywhere.

ackhandler/interfaces.go
marten-seemann: Done.

ackhandler/sent_packet_handler.go
lucas-clemente: It's not actually used a lot. I only see the use in the flowcontroller (where a default initial value would make sense), BandwidthEstimate() (which isn't called from anywhere and could be removed), and RetransmissionDelay() (where a too-low initial default doesn't matter). I'll leave it up to you, but I find it very strange to have the same const in two places.

ackhandler/sent_packet_handler.go
lucas-clemente: The handshake timeout results in an implicit max value. Otherwise, we'd have to worry about handshakeCount > 64.

marten-seemann:
@lucas-clemente: I added the comments. PHAL.

PR #864, opened 2017-10-01
reject ACKs that ack packets with a higher encryption level

### LINKED ISSUE #863: check that an unprotected ACK doesn't ack protected packets
We should close the connection when receiving such an ACK:

The TLS draft says:

> Endpoints MUST NOT use an ACK frame in an unprotected packet to acknowledge packets that were protected by 0-RTT or 1-RTT keys. An endpoint MUST treat receipt of an ACK frame in an unprotected packet that claims to acknowledge protected packets as a connection error of type OPTIMISTIC_ACK.

Waiting for confirmation of my PR for the Loss Recovery draft (https://github.com/quicwg/base-drafts/pull/810).

marten-seemann:
Fixes #863.

### DIFF
--- a/ackhandler/sent_packet_handler.go
+++ b/ackhandler/sent_packet_handler.go
@@ -152,7 +152,7 @@ func (h *sentPacketHandler) SentPacket(packet *Packet) error {
 	return nil
 }
 
-func (h *sentPacketHandler) ReceivedAck(ackFrame *wire.AckFrame, withPacketNumber protocol.PacketNumber, rcvTime time.Time) error {
+func (h *sentPacketHandler) ReceivedAck(ackFrame *wire.AckFrame, withPacketNumber protocol.PacketNumber, encLevel protocol.EncryptionLevel, rcvTime time.Time) error {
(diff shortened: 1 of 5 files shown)


ackhandler/sent_packet_handler.go
```diff
@@ -173,6 +173,9 @@ func (h *sentPacketHandler) ReceivedAck(ackFrame *wire.AckFrame, withPacketNumbe
 
 	if len(ackedPackets) > 0 {
 		for _, p := range ackedPackets {
+			if encLevel < p.Value.EncryptionLevel {
+				return fmt.Errorf("Received ACK with encryption level %s that acks a packet %d (encryption level %s)", encLevel, p.Value.PacketNumber, p.Value.EncryptionLevel)
```
lucas-clemente: I think we should use a quic error here – otherwise it'll result in an internal error sent to the peer.

marten-seemann:
TODO (@marten-seemann): Check how Chrome handles this before merging.

marten-seemann:
I talked to Jana about this in Singapore. He said that Google is already complying with this rule in gQUIC, so this PR should be safe to merge.

PR #867, opened 2017-10-02
improve garbage collection of stream

### LINKED ISSUE #828: find a better way to garbage collect streams
`streamsMap.Iterate` is only used to delete closed streams. We call it once per run of the run-loop in the session, and it makes a copy of the slice containing the open streams every time. This is wasteful.

We could think about getting rid of `streamsMap.Iterate`, and implement a `RemoveClosedStreams` on the `streamsMap` directly. We'd need to pass a callback to the `streamsMap` then (which would take care of deleting the stream from the flow controller.

marten-seemann:
Fixes #828.

This removes the Iterate method of the `streamsMap`, and moves the garbage collection of closed streams from the session to the `streamsMap`.

### DIFF
--- a/streams_map.go
+++ b/streams_map.go
@@ -29,26 +29,27 @@ type streamsMap struct {
 	closeErr           error
 	nextStreamToAccept protocol.StreamID
 
-	newStream newStreamLambda
+	newStream            newStreamLambda
+	removeStreamCallback removeStreamCallback
 
 	numOutgoingStreams uint32
 	numIncomingStreams uint32
 }
 
 type streamLambda func(*stream) (bool, error)
+type removeStreamCallback func(protocol.StreamID)
 type newStreamLambda func(protocol.StreamID) *stream
 
-var (
-	errMapAccess = errors.New("streamsMap: Error accessing the streams map")
-)
+var errMapAccess = errors.New("streamsMap: Error accessing the streams map")
 
-func newStreamsMap(newStream newStreamLambda, pers protocol.Perspective, connParams handshake.ParamsNegotiator) *streamsMap {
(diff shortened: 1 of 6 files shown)


streams_map.go
```diff
@@ -29,26 +29,27 @@ type streamsMap struct {
 	closeErr           error
 	nextStreamToAccept protocol.StreamID
 
-	newStream newStreamLambda
+	newStream            newStreamLambda
+	removeStreamCallback removeStreamCallback
```
lucas-clemente: It might be nicer to define a streamMapDelegate interface with these two methods, and pass the session in. Your call.

streams_map.go
```diff
@@ -29,26 +29,27 @@ type streamsMap struct {
...
 
 	numOutgoingStreams uint32
 	numIncomingStreams uint32
 }
 
 type streamLambda func(*stream) (bool, error)
```
lucas-clemente: This type can be removed.

streams_map.go
```diff
@@ -216,21 +217,50 @@ func (m *streamsMap) AcceptStream() (*stream, error) {
...
+	var j int
+	for i, id := range m.openStreams {
+		if i != j {
+			m.openStreams[j] = m.openStreams[i]
 		}
-		if !cont {
-			break
+		if id != 0 {
+			j++
+		} else if uint32(j) < m.roundRobinIndex {
```
lucas-clemente: Should we change the roundRobinIndex to an int?

marten-seemann: Yes.

streams_map.go
marten-seemann: It's still need for `RoundRobinIterate`.

PR #870, opened 2017-10-03
don't negotiate the idle timeout

marten-seemann:
The idle timeout doesn't need to be negotiated, simply announcing it to the peer is sufficient. We now close the session if there's no incoming network activity for the duration of the local idle timeout, and we send a PING frame after half the remote idle timeout (if keep alives are enabled).

This is a small step towards eliminating the `ParamsNegotiator`.


internal/handshake/params_negotiator.go
```diff
@@ -62,8 +62,7 @@ func (h *paramsNegotiator) SetFromTransportParameters(params []transportParamete
 			if len(p.Value) != 2 {
 				return fmt.Errorf("wrong length for idle_timeout: %d (expected 2)", len(p.Value))
 			}
-			val := time.Duration(binary.BigEndian.Uint16(p.Value)) * time.Second
-			h.idleConnectionStateLifetime = h.negotiateIdleConnectionStateLifetime(val)
+			h.remoteIdleTimeout = time.Duration(binary.BigEndian.Uint16(p.Value)) * time.Second
```
lucas-clemente: We should have a implementation-defined minimum value here (5-10s?), to avoid a PING DOS if the peer announces 0.

internal/handshake/params_negotiator_base.go
```diff
@@ -19,6 +19,7 @@ type ParamsNegotiator interface {
 	GetMaxReceiveConnectionFlowControlWindow() protocol.ByteCount
 	GetMaxOutgoingStreams() uint32
 	GetMaxIncomingStreams() uint32
+	// get the idle timeout that was sent by the peer
 	GetIdleConnectionStateLifetime() time.Duration
```
lucas-clemente: GetRemoteIdleTimeout()?

internal/handshake/params_negotiator.go
marten-seemann: Good catch! I'll add that.

PR #875, opened 2017-10-05
update CIs to Go 1.9.1

marten-seemann:
We should delay merging until Travis supports 1.9.1.


marten-seemann:
Superseeded by #903.

PR #879, opened 2017-10-16
remove the params negotiator

marten-seemann:
Transport parameters are only announced to the peer, not negotiated anymore. We don't need the `paramsNegotiator` anymore.

When initializing the crypto setup, the session passes in a `TransportParameters` struct. This struct contains our values, which then get announced to the peer. As soon as we receive the peer's transport parameters, they are passed back to the session on a newly introduced channel. The session then makes sure to update the values everywhere, e.g. in the flow controller and in the streams map.


internal/flowcontrol/flow_control_manager.go
```diff
@@ -13,32 +13,31 @@ import (
...
 	connFlowController   *flowController
 	mutex                sync.RWMutex
+
+	initialStreamSendWindow protocol.ByteCount
 }
 
 var _ FlowControlManager = &flowControlManager{}
 
-var errMapAccess = errors.New("Error accessing the flowController map.")
+var errMapAccess = errors.New("Error accessing the flowController map")
```
lucas-clemente: Nit: (because you're touching it): error strings shouldn't be capitalized.

internal/flowcontrol/flow_control_manager.go
```diff
@@ -61,6 +60,17 @@ func (f *flowControlManager) RemoveStream(streamID protocol.StreamID) {
 	f.mutex.Unlock()
 }
 
+func (f *flowControlManager) UpdateTransportParameters(params *handshake.TransportParameters) {
```
lucas-clemente: I think it would be nicer to remove the handshake pkg dep here and just have the relevant ints as parameters.

internal/handshake/crypto_setup_client.go
```diff
@@ -49,10 +49,11 @@ type cryptoSetupClient struct {
 	nullAEAD             crypto.AEAD
 	secureAEAD           crypto.AEAD
 	forwardSecureAEAD    crypto.AEAD
-	aeadChanged          chan<- protocol.EncryptionLevel
 
-	requestConnIDOmission bool
-	params                *paramsNegotiatorGQUIC
+	paramsReceived chan<- *TransportParameters
```
lucas-clemente: Nit: For small structs I would say it's preferable to send a value instead of a pointer, as that makes it explicit that the value is immutable and no synchronization is needed.

lucas-clemente: Second nit: I don't like the name of the chan, it sounds like one of the close-only chans. Why not just call `transportParams` or `paramsChan`?

internal/handshake/crypto_setup_client.go
```diff
@@ -141,15 +142,21 @@ func (h *cryptoSetupClient) HandleCryptoStream(stream io.ReadWriter) error {
...
+			if err := h.handleREJMessage(message.Data); err != nil {
+				return err
+			}
 		case TagSHLO:
-			err = h.handleSHLOMessage(message.Data)
+			params, err := h.handleSHLOMessage(message.Data)
+			if err != nil {
+				return err
+			}
+			// blocks until the session has received the parameters
```
lucas-clemente: Nit: The API could be changed so that it guarantees this, which is currently not the case, as the chan is passed in from the outside. I'll leave it up to you.

internal/handshake/crypto_setup_server.go
```diff
@@ -40,14 +40,17 @@ type cryptoSetupServer struct {
 	receivedForwardSecurePacket bool
 	receivedSecurePacket        bool
 	sentSHLO                    chan struct{} // this channel is closed as soon as the SHLO has been written
-	aeadChanged                 chan<- protocol.EncryptionLevel
+
+	receivedParams bool
+	paramsReceived chan<- *TransportParameters
```
lucas-clemente: Nits from above apply.

internal/handshake/crypto_setup_tls.go
```diff
@@ -38,52 +38,52 @@ var newMintController = func(conn *mint.Conn) crypto.MintController {
 // NewCryptoSetupTLSServer creates a new TLS CryptoSetup instance for a server
 func NewCryptoSetupTLSServer(
 	tlsConfig *tls.Config,
-	transportParams *TransportParameters,
+	params *TransportParameters,
+	paramsReceived chan<- *TransportParameters,
```
lucas-clemente: Nits from above.

internal/handshake/crypto_setup_tls.go
```diff
@@ -38,52 +38,52 @@ var newMintController = func(conn *mint.Conn) crypto.MintController {
...
+	}, nil
 }
 
 // NewCryptoSetupTLSClient creates a new TLS CryptoSetup instance for a client
 func NewCryptoSetupTLSClient(
 	hostname string, // only needed for the client
 	tlsConfig *tls.Config,
-	transportParams *TransportParameters,
+	params *TransportParameters,
+	paramsReceived chan<- *TransportParameters,
```
lucas-clemente: Nits from above.

internal/handshake/tls_extension_handler_client.go
```diff
@@ -3,16 +3,19 @@ package handshake
...
 	"github.com/bifurcation/mint"
 	"github.com/bifurcation/mint/syntax"
 	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/utils"
 )
 
 type extensionHandlerClient struct {
-	params *paramsNegotiator
+	params         *TransportParameters
+	paramsReceived chan<- *TransportParameters
```
lucas-clemente: And finally, nits from above :)

internal/handshake/tls_extension_handler_client.go
```diff
@@ -47,6 +57,7 @@ func (h *extensionHandlerClient) Send(hType mint.HandshakeType, el *mint.Extensi
 }
 
 func (h *extensionHandlerClient) Receive(hType mint.HandshakeType, el *mint.ExtensionList) error {
+	utils.Debugf("Received %#v", hType)
```
lucas-clemente: Aside: We should probably note somewhere that DEBUG should not be enabled in prod :)

internal/handshake/tls_extension_handler_server.go
```diff
@@ -13,17 +14,24 @@ import (
 )
 
 type extensionHandlerServer struct {
-	params *paramsNegotiator
+	paramsReceived chan<- *TransportParameters
```
lucas-clemente: Nits from above.

internal/handshake/transport_parameters.go
```diff
@@ -0,0 +1,165 @@
+package handshake
```
lucas-clemente: The functions in this file could be package private, right?

session.go
```diff
@@ -88,6 +88,8 @@ type session struct {
 	undecryptablePackets                   []*receivedPacket
 	receivedTooManyUndecrytablePacketsTime time.Time
 
+	// this channel is passed to the CryptoSetup and receives the transport parameters, as soon as the peer sends them
+	paramsReceived <-chan *handshake.TransportParameters
```
lucas-clemente: Nit from above.

streams_map.go
```diff
@@ -340,3 +341,7 @@ func (m *streamsMap) CloseWithError(err error) {
 		m.streams[s].Cancel(err)
 	}
 }
+
+func (m *streamsMap) UpdateTransportParameters(params *handshake.TransportParameters) {
+	m.maxOutgoingStreams = params.MaxStreams
```
lucas-clemente: I think this function has to acquire the lock.

internal/flowcontrol/flow_control_manager.go
marten-seemann: You're right, but I won't fix this, since the `FlowControlManager` will be gone as soon as I merge #881 anyway.

internal/flowcontrol/flow_control_manager.go
marten-seemann: You're right, but I won't fix this, since the `FlowControlManager` will be gone as soon as I merge #881 anyway.

internal/handshake/tls_extension_handler_client.go
marten-seemann: Isn't that the definition of DEBUG? ;)

Regarding that log statement here, I just forgot to remove that.

internal/handshake/transport_parameters.go
marten-seemann: Yes. Fixed.

PR #881, opened 2017-10-18
remove the flow control manager

marten-seemann:
Every stream now has its own stream-level flow controller. Every stream-level flow-controller now has a reference to the connection-level flow controller, and updates it accordingly (if the stream contributes to connection-level flow control).
This way, we can get rid of the `FlowControlManager`, and we only keep a single map of `StreamID` => some state.

This PR also removes the optimization that we update WINDOW_UPDATEs for retransmissions. We can reintroduce that later, I want to get this big change merged first.

This PR also changes a lot in the tests in the `quic` package. We now have a stream interface in that package (which has a lot more methods than the exported one), and that is mocked in the tests. This allowed me to remove a bunch of tests in the session, and make the `streamFramer` and `streamsMap` tests easier.


internal/flowcontrol/interface.go
```diff
@@ -24,5 +24,6 @@ type FlowControlManager interface {
 	AddBytesSent(streamID protocol.StreamID, n protocol.ByteCount) error
 	SendWindowSize(streamID protocol.StreamID) (protocol.ByteCount, error)
 	RemainingConnectionWindowSize() protocol.ByteCount
-	UpdateWindow(streamID protocol.StreamID, offset protocol.ByteCount) (bool, error)
+	UpdateStreamWindow(streamID protocol.StreamID, offset protocol.ByteCount) (bool, error)
```
lucas-clemente: Do we actually still need the bool return value? Seems to be ignored everywhere.

internal/mocks/mockgen_internal.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Mockgen refuses to generate mocks for internal packages.
+# This script copies the internal directory and renames it to internalpackage.
+# That way, mockgen can generate the mock.
+# Afterwards, it corrects the import paths (replaces internalpackage back to internal).
+
+TEMP_DIR=$(mktemp -d)
+mkdir -p $TEMP_DIR/src/github.com/lucas-clemente/quic-go/internalpackage
+
+cp -r $GOPATH/src/github.com/lucas-clemente/quic-go/internal/* $TEMP_DIR/src/github.com/lucas-clemente/quic-go/internalpackage
```
lucas-clemente: It's not required to have GOPATH set anymore, the script should work in that case :)

internal/mocks/mockgen_internal.sh
```diff
@@ -0,0 +1,21 @@
...
+find $TEMP_DIR -type f -name "*.go" -exec sed -i '' 's/internal/internalpackage/g' {} \;
```
lucas-clemente: This doesn't work on linux, sed behaves non-standard on macOS iirc.

internal/mocks/mockgen_internal.sh
```diff
@@ -0,0 +1,21 @@
...
+
+export GOPATH="$TEMP_DIR:$GOPATH"
+PACKAGE_PATH=${3/internal/internalpackage}
+
+
+mockgen -package $1 -self_package $1 -destination $2 $PACKAGE_PATH $4
+sed -i '' 's/internalpackage/internal/g' $2
+
+rm -rf "$TEMP_DIR"
```
lucas-clemente: To make this a bit safer you could remove the -f.

internal/mocks/mockgen_internal.sh
```diff
@@ -0,0 +1,21 @@
...
+
+cp -r $GOPATH/src/github.com/lucas-clemente/quic-go/internal/* $TEMP_DIR/src/github.com/lucas-clemente/quic-go/internalpackage
+find $TEMP_DIR -type f -name "*.go" -exec sed -i '' 's/internal/internalpackage/g' {} \;
+
+export GOPATH="$TEMP_DIR:$GOPATH"
+PACKAGE_PATH=${3/internal/internalpackage}
+
+
+mockgen -package $1 -self_package $1 -destination $2 $PACKAGE_PATH $4
+sed -i '' 's/internalpackage/internal/g' $2
```
lucas-clemente: Same as above.

internal/flowcontrol/interface.go
```diff
@@ -1,29 +1,33 @@
...
-	// methods needed for sending data
-	AddBytesSent(streamID protocol.StreamID, n protocol.ByteCount) error
-	SendWindowSize(streamID protocol.StreamID) (protocol.ByteCount, error)
-	RemainingConnectionWindowSize() protocol.ByteCount
-	UpdateStreamWindow(streamID protocol.StreamID, offset protocol.ByteCount) (bool, error)
-	UpdateConnectionWindow(offset protocol.ByteCount) bool
+type StreamFlowController interface {
+	flowController
+	// for receiving
+	UpdateHighestReceived(offset protocol.ByteCount, final bool) error
```
lucas-clemente: The `final` param is not self-explanatory here, please document what it means in the interface.

internal/mocks/mockgen_stream.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
lucas-clemente: Same comments as in the previous script.

session.go
```diff
@@ -109,7 +109,8 @@ type session struct {
 
 	sessionCreationTime     time.Time
 	lastNetworkActivityTime time.Time
-	remoteIdleTimeout       time.Duration
+
+	peerParams *handshake.TransportParameters
```
lucas-clemente: Can you explain this change? I'd prefer having the params as normal members instead of in the struct: Lesser coupling and simpler to reason about.

session.go
```diff
@@ -625,12 +624,15 @@ func (s *session) handleCloseError(closeErr closeError) error {
 }
 
 func (s *session) processTransportParameters(params *handshake.TransportParameters) {
-	s.remoteIdleTimeout = params.IdleTimeout
-	s.flowControlManager.UpdateTransportParameters(params)
+	s.peerParams = params
 	s.streamsMap.UpdateTransportParameters(params)
```
lucas-clemente: I must have missed this earlier, but I would suggest not passing the params struct here too.

stream.go
```diff
@@ -14,6 +14,24 @@ import (
 	"github.com/lucas-clemente/quic-go/internal/wire"
 )
 
+type streamI interface {
```
lucas-clemente: I'm strongly opposed to making the internal stream methods public (even if invisible). This way they can easily (without pointer magic) be called from external code – and might even break if the stdlib introduces more interfaces in the future.

Also, that change should really be its own commit if at all possible.

internal/mocks/mockgen_internal.sh
marten-seemann: You can even have multiple GOPATHs. In principle, this script should handle that as well, but I don't want to handle all those cases here. mockgen is a pain!

stream.go
marten-seemann: Making those methods public is the only way to mock them. As it was before, we were just inconsistent (for example, `AddStreamFrame` was public already).
I'm not sure I understand your comment about the stdlib.

Separating this out into a separate commit is challenging, because this will involve changing the tests as well (which used to be tightly coupled to the flow control stuff).

session.go
marten-seemann: Fixed in #879.

internal/mocks/mockgen_internal.sh
marten-seemann: I'm not planning to fix this. We're both working on OSX, so we can regenerate the mocks. This is the dirty work that mockgen is supposed to do.

stream.go
marten-seemann: What we can do is create a separate `Stream` struct on which we **only** define the methods defined on the interface, e.g.

```go
type Stream struct {
  stream streamI
}

func (s *Stream) Read(b []byte) (int, error) {
  return s.stream.Read(b)
}

// ... all other Stream interface methods
```

That would allow us to use a mock for `streamI` while at the same time preventing users to type-assert the public methods on the `quic.Stream`.

session.go
marten-seemann: `newStream` is checking if the `peerParams` are set, and using the value to initialize the flow controller.

internal/flowcontrol/interface.go
marten-seemann: I believe this was already fixed by a later commit. The method is now `flowController.UpdateSendWindow(protocol.ByteCount)` (no return value).

session.go
lucas-clemente: Couldn’t you do the same check using the params themselves?

stream.go
marten-seemann: This is not a mockgen problem here. mockgen generates the right mock calls here, but Go doesn't allow private methods on an implementation that lives in another package.

session.go
marten-seemann: I don't understand where you're going with this. Of course we could create 3 or 4 new members of the `session` here, but I don't see any benefit in doing so (since they're all set at the same time). I'll merge this PR now, we can still change this later if you want to.

PR #885, opened 2017-10-18
drop packets for new connections that don't have the required min size

### LINKED ISSUE #884: drop Client Hellos that are too small
At the moment, we're sending a CONNECTION_CLOSE. We should instead just drop a too small CHLO.

marten-seemann:
Fixes #884.

### DIFF
--- a/server.go
+++ b/server.go
@@ -223,6 +223,12 @@ func (s *server) handlePacket(pconn net.PacketConn, remoteAddr net.Addr, packet 
 	session, ok := s.sessions[connID]
 	s.sessionsMutex.RUnlock()
 
+	// This is (potentially) a Client Hello.
+	// Make sure it has the minimum required size before spending any more ressources on it.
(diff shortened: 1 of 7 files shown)


marten-seemann:
Merging this needs to wait until #885 is fixed.

server.go
```diff
@@ -223,6 +223,12 @@ func (s *server) handlePacket(pconn net.PacketConn, remoteAddr net.Addr, packet
 	session, ok := s.sessions[connID]
 	s.sessionsMutex.RUnlock()
 
+	// This is (potentially) a Client Hello.
```
lucas-clemente: Just to be clear, this means that all packets sent in the first RTT need to fulfill this requirement? Do we do that in the client?

marten-seemann: We do, at least for gQUIC. #886 is the issue to track that we also do that for TLS.

Issue #874, opened 2017-10-05
0-RTT not working


marten-seemann:
Currently, we're creating a new `CookieGenerator` in the `cryptoSetupServer`. That means that the STK won't be valid across sessions. We should only have a single `CookieGenerator` for the whole server.

marten-seemann:
Fixed by #890.

PR #890, opened 2017-10-20
Fix 0-RTT by moving cookie generation to server config

jbenoist:


marten-seemann:
Hello @jbenoist, thanks for this PR!

This is a partial fix for #874 (it only works for the gQUIC handshake, not the TLS handshake). I think I can merge it like this, since it will improve performance in production. Can you please fix the tests?

I've been thinking about the right API for this after our discussion in Seattle, but not come to a conclusion yet. For a large-scale deployment there needs to be a mechanism to rotate Server Configs, and probably the Cookie Generators as well, which suggests to move this to the `quic.Config` in form or the other (probably as separate entries, so it can be used by gQUIC and TLS as well). This will be a more complicated change, so let's first move forward with this PR.

jbenoist:
Sure thing, I've cleaned up the server_config/crypto_setup_server  tests and seems to be good now. However, in my environment, I'm having a hard time repro the timeout as seen here

https://travis-ci.org/lucas-clemente/quic-go/jobs/291687171#L843

Attempting a re-run

marten-seemann:
Don't worry about the Travis build. It's probably one of those loss recovery issues we talked about.

PR #891, opened 2017-10-21
handle the crypto stream separately

### LINKED ISSUE #668: Don't count the crypto stream as incoming on the server
We should fix this in the streams map.

marten-seemann:
This will be very useful when moving forward with the IETF draft implementation, especially when (if) the crypto stream gets stream ID 0.

This also fixes #668.
Should be merged after #881.


lucas-clemente:
For some reason github doesn’t show the approval dialog here. LGTM.

PR #892, opened 2017-10-21
remove the prioritisation of the H2 header stream

marten-seemann:
This was an optimization for H2/QUIC, which always violated the QUIC layering. The H2 header stream will (probably) be removed in IETF QUIC.

This will simplify things a lot when the crypto stream is moved to stream ID 0.

This PR depends on #891.


lucas-clemente:
LGTM

Issue #894, opened 2017-10-23
send IETF draft style Version Negotiation Packets


marten-seemann:
The format for VNPs is different, they now have the 0x80 bit set in the Flag Byte (because it's a Long Header), and they echo the packet number.

For parsing this shouldn't cause any problem. In the server, we should probably sends two VNPs (one in gQUIC format and one in IETF format), if the server supports both versions.

PR #896, opened 2017-10-24
make version numbers consistent with the IETF draft

marten-seemann:
Depends on #895. Prerequisite for #894 and #901.

### DIFF
--- a/internal/protocol/version.go
+++ b/internal/protocol/version.go
@@ -1,13 +1,19 @@
 package protocol
 
-import "fmt"
+import (
+	"fmt"
+)
 
 // VersionNumber is a version number as int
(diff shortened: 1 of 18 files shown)


internal/protocol/version.go
```diff
@@ -40,19 +42,19 @@ func (vn VersionNumber) String() string {
 	case VersionTLS:
 		return "TLS dev version (WIP)"
 	default:
+		if vn > 0x51303300 && vn <= 0x51303399 {
```
lucas-clemente: Define 0x51303300 as const and use it for the other magic numbers.

marten-seemann: Done

PR #897, opened 2017-10-25
make flow control stricter when receiving final offsets

### LINKED ISSUE #382: check for StreamDataAfterTermination errors
After half-closing (remote) a stream, either by receiving a `RstStreamFrames` or by receiving a FIN, an incorrect implementation of the peer could still send `StreamFrames` with a higher offset. At the moment, we're simply ignoring all `StreamFrames` on closed streams.
We should send a `StreamDataAfterTermination` error instead.

marten-seemann:
1. Reject STREAM frames with a higher offset than a previously received highest offset.
2. Reject inconsistent final offsets.

Fixes #382.

### DIFF
--- a/internal/flowcontrol/stream_flow_controller.go
+++ b/internal/flowcontrol/stream_flow_controller.go
@@ -12,10 +12,12 @@ import (
 type streamFlowController struct {
 	baseFlowController
 
-	connection connectionFlowControllerI
+	streamID protocol.StreamID
 
-	streamID                protocol.StreamID
+	connection              connectionFlowControllerI
 	contributesToConnection bool // does the stream contribute to connection level flow control
+
+	receivedFinalOffset bool
 }
 
 var _ StreamFlowController = &streamFlowController{}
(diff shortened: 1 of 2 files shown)


internal/flowcontrol/stream_flow_controller.go
```diff
@@ -52,6 +52,11 @@ func (c *streamFlowController) UpdateHighestReceived(byteOffset protocol.ByteCou
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
+	// when receiving a final offset, check that this final offset is consistent with a final offset we might have received earlier
+	if final && c.receivedFinalOffset && byteOffset != c.highestReceived {
+		return qerr.Error(qerr.StreamDataAfterTermination, fmt.Sprintf("Received inconsistent final offset for stream %d (%d vs %d bytes)", c.streamID, byteOffset, c.highestReceived))
```
lucas-clemente: Nit: "old: 42, new: 23" instead of "42 vs 23".

Issue #898, opened 2017-10-25
implement the new AEAD for TLS handshake packets


marten-seemann:
The FNV checksum was recently replaced with AEAD (see https://quicwg.github.io/base-drafts/draft-ietf-quic-tls.html#rfc.section.5.2.1).

PR #902, opened 2017-10-27
set the Long Header type

marten-seemann:
Depends on #898.


internal/handshake/mint_utils.go
```diff
@@ -69,11 +75,30 @@ func (mc *mintController) Handshake() mint.Alert {
 // mint expects a net.Conn, but we're doing the handshake on a stream
 // so we wrap a stream such that implements a net.Conn
 type fakeConn struct {
-	io.ReadWriter
+	stream io.ReadWriter
+	pers   protocol.Perspective
+
+	blockRead bool
```
lucas-clemente: Are you sure this doesn't need to be atomic?

marten-seemann: Why should it be atomic? There's only one `Read` at a time, and it is reset before the next `Read` starts.

lucas-clemente: If you can guarantee that Read and UnblockRead will never be called from separate goroutines, then yes, it should be fine. It just seems counter-intuitive that blocking doesn't involve coroutines.

marten-seemann: Yeah, the name is kind of unintuitive. It comes from mint, where it's called non-blocking mode, when `Handshake()` (which normally runs the whole TLS handshake from beginning to end) returns when it thinks `Read` would "block" (which apparently is met by the condition that it returns `0, nil`).

Issue #893, opened 2017-10-22
implement Stateless Reset receiving


marten-seemann:
Stateless Resets work completely different than in gQUIC, so we can't just read a flag in the packet header.

PR #907, opened 2017-10-29
pad the Client Initial packet to the required minimum size

### LINKED ISSUE #886: implement the minimum packet size requirements for the IETF draft
This will require an inspection of the Header Type byte, to make sure that the padding is applied to all Client Initials (and their retransmissions).

marten-seemann:
Fixes #886.

### DIFF
--- a/internal/protocol/protocol.go
+++ b/internal/protocol/protocol.go
@@ -61,8 +61,11 @@ const MaxReceivePacketSize ByteCount = 1452
 // Used in QUIC for congestion window computations in bytes.
 const DefaultTCPMSS ByteCount = 1460
 
-// ClientHelloMinimumSize is the minimum size the server expects an inchoate CHLO to have.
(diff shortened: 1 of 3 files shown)


marten-seemann:
This PR is currently wrong. We can't just add padding at the end of the packet, since the STREAM frame carrying the ClientHello doesn't have the data length set.

marten-seemann:
Fixed that. Just can't come up with a good unit test for this, but this is guaranteed to fail the integration test immediately, so it should be safe to merge.

marten-seemann:
This currently fails because of a problem with the TLS crypto setup. We're sending packets as Client Initials that are actually not Client Initials. We would have to transition away from packet type ClientInitial immediately after sending a ClientInitial, however, this requires that we properly handle Server Stateless Resets (because we need to send the regenerated Client Hello then).
Merging this probably has to wait until #893 is fixed.

marten-seemann:
Superseeded by #1001.

PR #910, opened 2017-10-29
send IETF draft style version negotiation packets

### LINKED ISSUE #894: send IETF draft style Version Negotiation Packets

marten-seemann:
Fixes #894.

If we don't understand the version the client is offering, we don't know if this is an unsupported gQUIC version or an unsupported IETF QUIC version. We therefore send two Version Negotiation Packets, and let the client pick the one it is able to parse.

### DIFF
--- a/internal/wire/version_negotiation.go
+++ b/internal/wire/version_negotiation.go
@@ -7,18 +7,36 @@ import (
 	"github.com/lucas-clemente/quic-go/internal/utils"
 )
 
-// ComposeVersionNegotiation composes a Version Negotiation Packet
-// TODO(894): implement the IETF draft format of Version Negotiation Packets
-func ComposeVersionNegotiation(connectionID protocol.ConnectionID, versions []protocol.VersionNumber) []byte {
+// ComposeGQUICVersionNegotiation composes a Version Negotiation Packet for gQUIC
+func ComposeGQUICVersionNegotiation(connID protocol.ConnectionID, versions []protocol.VersionNumber) []byte {
 	fullReply := &bytes.Buffer{}
 	ph := Header{
-		ConnectionID: connectionID,
+		ConnectionID: connID,
 		PacketNumber: 1,
 		VersionFlag:  true,
 	}
-	err := ph.writePublicHeader(fullReply, protocol.PerspectiveServer, protocol.VersionWhatever)
-	if err != nil {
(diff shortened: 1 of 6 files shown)


marten-seemann:
I just realized that my previous commit was wrong (and amended).
There'll never be a situation where we need to send two VNPs. We can tell by the packet header if the client speaks gQUIC or IETF QUIC, and we can just send the VNP accordingly.

internal/wire/version_negotiation.go
```diff
@@ -7,17 +7,33 @@ import (
...
 	ph := Header{
-		ConnectionID: connectionID,
+		ConnectionID: connID,
 		PacketNumber: 1,
 		VersionFlag:  true,
 	}
-	err := ph.writePublicHeader(fullReply, protocol.PerspectiveServer, protocol.VersionWhatever)
-	if err != nil {
+	if err := ph.writePublicHeader(fullReply, protocol.PerspectiveServer, protocol.VersionWhatever); err != nil {
+		utils.Errorf("error composing version negotiation packet: %s", err.Error())
```
lucas-clemente: Is a `return nil` missing here?

internal/wire/version_negotiation.go
```diff
@@ -7,17 +7,33 @@ import (
...
+func ComposeVersionNegotiation(connID protocol.ConnectionID, pn protocol.PacketNumber, versions []protocol.VersionNumber) []byte {
+	fullReply := &bytes.Buffer{}
+	ph := Header{
+		IsLongHeader: true,
+		Type:         protocol.PacketTypeVersionNegotiation,
+		ConnectionID: connID,
+		PacketNumber: pn,
+	}
+	if err := ph.writeHeader(fullReply); err != nil {
 		utils.Errorf("error composing version negotiation packet: %s", err.Error())
```
lucas-clemente: Same here?

internal/wire/version_negotiation.go
marten-seemann: Yes, but this was an existing issue. `writePublicHeader` shouldn't fail when writing to a `bytes.Buffer`, unless given the wrong input values.

PR #912, opened 2017-10-30
fix gQUIC version numbers

### LINKED ISSUE #911: fix gQUIC version numbers
Deriving the gQUIC version from the version number only works for versions 30-39.

marten-seemann:
Fixes #911.

I got the math wrong in #896, it only worked for gQUIC version numbers Q030 - Q039.
I now added tests with unsupported version numbers from the QUIC WG wiki.

### DIFF
--- a/internal/protocol/version.go
+++ b/internal/protocol/version.go
@@ -7,13 +7,15 @@ import (
 // VersionNumber is a version number as int
 type VersionNumber int
 
-// gquicVersion0 is the "base" for gQUIC versions
-// e.g. version 39 is gquicVersion + 0x39
-const gquicVersion0 = 0x51303300
+// gQUIC version range as defined in the wiki: https://github.com/quicwg/base-drafts/wiki/QUIC-Versions
+const (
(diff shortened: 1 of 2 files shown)


internal/protocol/version.go
```diff
@@ -7,13 +7,15 @@ import (
 // VersionNumber is a version number as int
 type VersionNumber int
 
-// gquicVersion0 is the "base" for gQUIC versions
-// e.g. version 39 is gquicVersion + 0x39
-const gquicVersion0 = 0x51303300
+// gQUIC version range as defined in the wiki: https://github.com/quicwg/base-drafts/wiki/QUIC-Versions
+const (
+	gquicVersion0   = 0x51303030
+	maxGquicVersion = 0x51303439
```
lucas-clemente: That seems quite low – why not just 0x5130ffff?

marten-seemann: Just following the WG wiki here: https://github.com/quicwg/base-drafts/wiki/QUIC-Versions. We can extrapolate, if you want to.

PR #916, opened 2017-10-30
fix the exponential backoff when sending handshake retransmissions

marten-seemann:


ackhandler/sent_packet_handler_test.go
```diff
@@ -857,6 +857,7 @@ var _ = Describe("SentPacketHandler", func() {
 			Expect(p.PacketNumber).To(Equal(protocol.PacketNumber(4)))
 			Expect(handler.packetHistory.Len()).To(Equal(1))
 			Expect(handler.packetHistory.Front().Value.PacketNumber).To(Equal(protocol.PacketNumber(3)))
+			Expect(handler.handshakeCount).To(BeEquivalentTo(1))
```
lucas-clemente: Can we have a test that checks whether the var's value is actually used?

marten-seemann: Done.

PR #928, opened 2017-11-02
use stream 0 for the crypto stream when using TLS

### LINKED ISSUE #880: use stream 0 for the crypto stream when using TLS

### LINKED ISSUE #923: find a solution for streams not contributing to connection-level flow control
In gQUIC stream 1 and 3 don't contribute, in IETF QUIC only stream 0 doesn't.

marten-seemann:
Fixes #880. Fixes #923.

This also requires the usage of MAX_DATA, MAX_STREAM_DATA as well as BLOCKED and STREAM_BLOCKED, since stream ID 0 cannot be used for in band signalling any more.


stream.go
```diff
@@ -274,7 +274,7 @@ func (s *stream) GetDataForWriting(maxBytes protocol.ByteCount) []byte {
 	}
 
 	// TODO(#657): Flow control for the crypto stream
-	if s.streamID != 1 {
+	if s.streamID >= 2 {
```
lucas-clemente: Same here?

internal/wire/max_data_frame.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, err
+	}
+	frame.ByteOffset = protocol.ByteCount(byteOffset)
+	return frame, nil
+}
+
+//Write writes a MAX_STREAM_DATA frame
+func (f *MaxDataFrame) Write(b *bytes.Buffer, version protocol.VersionNumber) error {
+	if version.CryptoStreamID() != 0 {
```
lucas-clemente: Are we sure that MAX_DATA will appear in the same version as stream 0 for the crypto stream?

If yes, then I'd still add a bool-valued function to the version, so that we don't need this comparison everywhere.

packet_unpacker.go
```diff
@@ -63,43 +63,50 @@ func (u *packetUnpacker) Unpack(headerBinary []byte, hdr *wire.Header, data []by
...
-			case 0x06:
-				frame, err = wire.ParseStopWaitingFrame(r, hdr.PacketNumber, hdr.PacketNumberLen, u.version)
-				if err != nil {
-					err = qerr.Error(qerr.InvalidStopWaitingData, err.Error())
-				}
-			case 0x07:
-				frame, err = wire.ParsePingFrame(r, u.version)
-			default:
-				err = qerr.Error(qerr.InvalidFrameData, fmt.Sprintf("unknown type byte 0x%x", typeByte))
+		} else if typeByte == 0x01 {
```
lucas-clemente: I think it would be nicer here to duplicate code and have two separate switches for the two cases instead of this if-else mess. WDYT?

lucas-clemente: Also, this could probably be made shorter by combining the err check with the assignments.

session.go
```diff
@@ -820,7 +820,7 @@ func (s *session) queueResetStreamFrame(id protocol.StreamID, offset protocol.By
 func (s *session) newStream(id protocol.StreamID) streamI {
 	// TODO: find a better solution for determining which streams contribute to connection level flow control
 	var contributesToConnection bool
-	if id != 1 && id != 3 {
+	if id != 0 && id != 1 && id != 3 {
```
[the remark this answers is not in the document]
lucas-clemente: This is outdated with a later commit, the other comment still applies.

stream.go
lucas-clemente: (see the hidden comment)

internal/wire/max_data_frame.go
marten-seemann: I think there's no other way. How else would you distinguish a WINDOW_UPDATE for stream 0 and the connection?

packet_unpacker.go
marten-seemann: But there are more than two cases. Not all IETF frames will be introduced at the same time.

stream.go
marten-seemann: Right, we need to check the version here.

Issue #931, opened 2017-11-03
h2quic.RoundTripper reusing dead Session / Connection


heri16:
`h2quic.RoundTripper` keeps a cached session to every host in a map.

However, if i understand correctly from the source code, if there is any connectivity problem to a host, `qerr.NetworkIdleTimeout` is returned as an error from the runLoop, causing the client connection to be closed. Even when the client connection has been closed, `h2quic.RoundTripper` still keeps the dead client with the stale connection it has in its map, and tries to reuse the dead client/session on the next RoundTrip() to the same host. This causes third-party libraries that uses `h2quic.RoundTripper` like Caddy to not be able to recover on failed connectivity.

I thought a better idea would be to automatically detect and flush stale sessions/clients from `h2quic.RoundTripper`. This could be done via `client.session.Context` or some sort of client.isConnClosed()?

Is this bug valid? If so I would like to attempt a fix on this and submit a PR.

marten-seemann:
Duplicate of #765.

Issue #765, opened 2017-07-25
handle closed clients in http3.RoundTripper


marten-seemann:
When a QUIC connection is closed (e.g. due to an idle timeout), subsequent `RoundTrip`calls in the  `h2quic.RoundTripper` will fail.

marten-seemann:
Since this came up a couple of times, but I don't have the time to fix this issue myself at the moment:

There two cases for closed connections here: The connection could be closed regularly, or time out, or receive a Public Reset. In these cases, `RoundTrip` should re-dial the connection. If the connection was closed as a result of a failed version negotiation, failed crypto negotiation or a general protocol violation, there's no need to try again.

heri16:
@marten-seemann could you layout a high-level design that makes sense since the approaches of both PR #933 and PR #1095 has been rejected. With a design specification, it would be easier to write a PR that would be acceptable.

marten-seemann:
I'll take care of it.

fbzhong:
@marten-seemann can you write down high level design of this? I encounter this issue and want to fix.

marten-seemann:
There should be similar code in the standard library HTTP/2 package.

gabrielperezs:
Hi. Here is a proposal to detect "done" clients and force it to create a new one: https://github.com/lucas-clemente/quic-go/compare/master...gabrielperezs:reuse-roundTripper-if-client-fails

I didn't send a PR because I'm not sure about one of the unit tests that returns an error:

```
• Failure [0.001 seconds]
RoundTripper
/go/src/github.com/lucas-clemente/quic-go/http3/roundtrip_test.go:60
  dialing hosts
  /go/src/github.com/lucas-clemente/quic-go/http3/roundtrip_test.go:79
    reuses existing clients [It]
    /go/src/github.com/lucas-clemente/quic-go/http3/roundtrip_test.go:136

    Unexpected call to *mockquic.MockEarlySession.Context([]) at /go/src/github.com/lucas-clemente/quic-go/internal/mocks/quic/early_session.go:101 because: there are no expected calls of the method "Context" for that receiver

    /go/src/github.com/lucas-clemente/quic-go/internal/mocks/quic/early_session.go:101
```

I suppose that if the changes linked before are ok, then this test can be updated because it is not an issue any more?

Let me know so I can send the PR with the mentioned changes.

spacewander:
Is there any updates?

kaiyuhou:
[...]

PR #933, opened 2017-11-03
Fix Flushing of dead RoundTripper clients

heri16:
Auto-delete clients from RoundTripper with sessions that are dead/stale/done.
Issue #931

### DIFF
--- a/h2quic/roundtrip.go
+++ b/h2quic/roundtrip.go
@@ -1,6 +1,7 @@
 package h2quic
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


heri16:
Tests are failing on Travis as [](https://quic.clemente.io/foobar.html) and [](https://quic.clemente.io/file1.html) no longer exists. What should be done?

heri16:
I signed it!

marten-seemann:
Hello @heri16, sorry for the late reply, I must have missed your PR somehow.

Deleting closed clients in the roundtripper is only the correct thing to do under one condition: when the connection was closed properly. There are a lot of cases where you don't want to delete the client: when the connection timed out because the peer didn't reply at all (i.e. it doesn't support QUIC), or when a protocol violation occurred.

marten-seemann:
Closing this PR due to age. Feel free to open a new one.

heri16:
Connection Migration does not work properly for me without an ability for the RoundTripper to delete clients with sessions that are dead. For example, when there is a NAT rebinding, using an existing client causes the connection to not be able to reestablish after a connection timeout.

marten-seemann:
That's a known issue: #765.

PR #935, opened 2017-11-03
don't inform the congestion controller about retransmission timeouts

marten-seemann:


marten-seemann:
I was hoping that this might fix our flaky Travis tests. Doesn't seem to work out that way.
We probably should wait until the rewrite of the loss recovery before making this change.

marten-seemann:
I'm dropping this PR. This will have to wait.

PR #944, opened 2017-11-09
parse the packet header before getting the session from the session map

marten-seemann:
After dropping support for the QUIC versions that encoded values in little endian, we can now parse the packet header without looking up the session in the `sessionsMap`.

For the server, the form of the header flag byte allows us to distinguish between gQUIC and IETF QUIC packets without any additional information. For the client, the version is known before parsing (except for Version Negotiation Packets, which in turn are distinguishable by their flag byte). See https://docs.google.com/document/d/162BtfR84c6xTvAG6AaPK0dXSQXKahMrtJOXxNnpcwDw/edit?usp=sharing for details.

This PR is an important prerequisite for stateless connection establishment: Now that we can parse the header by just looking at the bytes sent on the wire, we will be able to statelessly handle the initial packet.


internal/wire/header.go
```diff
@@ -32,59 +31,56 @@ type Header struct {
...
-	var typeByte uint8
-	if version == protocol.VersionUnknown {
-		var err error
-		typeByte, err = b.ReadByte()
-		if err != nil {
-			return nil, err
-		}
-		_ = b.UnreadByte() // unread the type byte
+// ParseHeaderFromServer parses the header for a packet that was sent by the server.
+func ParseHeaderFromServer(b *bytes.Reader, version protocol.VersionNumber) (*Header, error) {
```
lucas-clemente: The name is ambiguous (and I got it wrong on my first reading). Either ParseHeaderSentByServer or ParseHeaderForClient.

internal/wire/header.go
```diff
@@ -32,59 +31,56 @@ type Header struct {
...
+		// For all packets that are not Version Negotiation Packets, the client knows the version that this packet was sent with
+		isPublicHeader = !version.UsesTLS()
 	}
+	return parsePacketHeader(b, protocol.PerspectiveServer, isPublicHeader)
+}
 
-	// This is a gQUIC Public Header.
-	hdr, err := parsePublicHeader(b, sentBy, version)
+// ParseHeaderFromClient parses the header for a packet that was sent by the client.
+func ParseHeaderFromClient(b *bytes.Reader) (*Header, error) {
```
lucas-clemente: Same here.

server.go
```diff
@@ -219,10 +219,12 @@ func (s *server) handlePacket(pconn net.PacketConn, remoteAddr net.Addr, packet
...
-	connID, err := wire.PeekConnectionID(r)
+	hdr, err := wire.ParseHeaderFromClient(r)
 	if err != nil {
 		return qerr.Error(qerr.InvalidPacketHeader, err.Error())
 	}
+	hdr.Raw = packet[:len(packet)-r.Len()]
+	connID := hdr.ConnectionID
 
 	s.sessionsMutex.RLock()
 	session, ok := s.sessions[connID]
```
lucas-clemente: I think the rest of the function would be clearer if we do s/ok/sessionKnown/g

PR #947, opened 2017-11-12
fix parsing of the Header type byte

marten-seemann:
In order to determine if a packet is a Version Negotiation Packet, it is not sufficient to just look at bit 0x1. Other packet types also have that bit set, e.g. the Retry packet (packet type 0x3). Instead, we have to look at the last 3 bits. This fix will work as long as IETF QUIC doesn't define more than 8 long header types.


marten-seemann:
Things seem to be more complicated than expected. And all just for the weird special case that we receive a gQUIC VNP in response to an IETF QUIC packet.

PR #949, opened 2017-11-12
echo the offered version in IETF Version Negotiation Packets

### LINKED ISSUE #948: echo the client's version number in IETF style Version Negotiation Packets

marten-seemann:
Fixes #948.


lucas-clemente:
We should probably also check this in the client (or add a TODO), right?

marten-seemann:
This code path with change completely with my upcoming PR for stateless connection establishment. I’ll add the check there. I believe this check is not really necessary, since echoing the packet number and the connection ID should be sufficient proof to verify that the sender of the VNP is on path.

PR #955, opened 2017-11-22
Make error chan not block for a single write

### LINKED ISSUE #953: HandleCryptoStream errorChan not always read
Hi,

quic-go is leaking goroutines in `HandleCryptoStream` (https://github.com/lucas-clemente/quic-go/blob/master/internal/handshake/crypto_setup_client.go#L111).

I haven't quite figured out what chain of events trigger this leak but I started to notice that my quic-go enabled application consumes more and more memory.

I used pprof and delve and found that I have many outstanding goroutines:
`github.com/lucas-clemente/quic-go/internal/handshake(*cryptoSetupClient).HandleCryptoStream.func1+0x122`
This is the goroutine that parses handshake messages.

Upon debugging with delve I found that these routines hang on sending on the error channel in `HandleCryptoStream`. It seems that the `cryptoSetupClient` has already been freed (so the session is closed and `HandleCryptoStream` has exited). So at this point and there is no one left to read the error from the error channel causing the block in the goroutine to write to the channel effectively leaking the routine.

I think `HandleCryptoStream` should make sure that the goroutine has finished before exiting.
For example, if `maybeUpgradeCrypto` or `sendCHLO` fail, the goroutine could still process data and fail (also on success the `messageChan` would block) and cause leaks.

Best
 Jan

Eichhoernchen:
This fixes #953 

It allows the ParseHandshakeMessage goroutine to gracefully exit if the cryptoSetupClient has already exited.

### DIFF
--- a/internal/handshake/crypto_setup_client.go
+++ b/internal/handshake/crypto_setup_client.go
@@ -102,7 +102,7 @@ func NewCryptoSetupClient(
 
 func (h *cryptoSetupClient) HandleCryptoStream() error {
 	messageChan := make(chan HandshakeMessage)
-	errorChan := make(chan error)
+	errorChan := make(chan error, 1)
 
 	go func() {
 		for {


Eichhoernchen:
googlebot do your thing

PR #962, opened 2017-11-30
send reserved version numbers

### LINKED ISSUE #871: send reserved versions in version negotiation packet
As described in https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#using-reserved-versions.

marten-seemann:
Fixes #871.

This PR implements sending reserved version numbers for:

- IETF Version Negotiation Packets
- gQUIC Version Negotiation Packets
- EncryptedExtensions message (for the TLS handshake)
- SHLO (for the gQUIC handshake)

I'm not really happy how I'm reading the random numbers here (using `crypto.Rand`), but there doesn't seem to be a good way to initialise a `math.Rand` as a library, if you want to be able to use concurrent reads.

### DIFF
--- a/internal/handshake/crypto_setup_server.go
+++ b/internal/handshake/crypto_setup_server.go
@@ -431,7 +431,7 @@ func (h *cryptoSetupServer) handleCHLO(sni string, data []byte, cryptoData map[T
 	replyMap := h.params.getHelloMap()
 	// add crypto parameters
 	verTag := &bytes.Buffer{}
-	for _, v := range h.supportedVersions {
+	for _, v := range protocol.GetGreasedVersions(h.supportedVersions) {
 		utils.BigEndian.WriteUint32(verTag, uint32(v))
 	}
 	replyMap[TagPUBS] = ephermalKex.PublicKey()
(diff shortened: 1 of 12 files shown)


internal/handshake/crypto_setup_server.go
```diff
@@ -449,6 +445,25 @@ func (h *cryptoSetupServer) handleCHLO(sni string, data []byte, cryptoData map[T
 	return reply.Bytes(), nil
 }
 
+func (h *cryptoSetupServer) getSupportedVersions() []byte {
```
lucas-clemente: s/getSupportedVersions/getGreasedVersions/g

internal/handshake/tls_extension_handler_server.go
```diff
@@ -61,6 +59,29 @@ func (h *extensionHandlerServer) Send(hType mint.HandshakeType, el *mint.Extensi
...
+			j++
+		}
+		supportedVersions[j] = uint32(v)
+		j++
+	}
+	if randPos == len(h.supportedVersions) {
+		supportedVersions[randPos] = uint32(reservedVersion)
+	}
+	return supportedVersions
+}
```
lucas-clemente: This code is fragile enough that I'd like to have it in a single place. WDYT about a single getGreasedVersions in version.go that returns a greased slice?

marten-seemann: Will be a few more allocations, but probably worth the clarity we gain by doing so.

marten-seemann:
@lucas-clemente: I updated the PR. PHAL.

PR #963, opened 2017-11-30
disable connection migration

marten-seemann:
Connection migration is complicated: We need to verify the new path, and reset the congestion controller. In the IETF draft, connection migration is still work in progress.


julienschmidt:
Are there any major problems with the implementation this PR removed?
Or would it work at least for demo purposes / experiments?

marten-seemann:
Yes and yes. The implementation didn't verify the peer's new address, and didn't reset the congestion controller, so this could have been used in a reflection DoS.

PR #967, opened 2017-12-01
implement the new varint frame types

marten-seemann:
This PR implements the variable integer length frame types for the IETF draft, for all frame types that have so far been implemented.

There's still one TODO for the ACK frame: Since the ACK ranges are also varint encoded, and their number is not limited any more, it might overflow the maximum packet size. I'm not a fan of the current code for the ACK frame, so maybe @lucas-clemente has an idea for an elegant solution here.
I think we can merge this PR without resolving this TODO for now, but I'd still be happy to know what you think about this.

### DIFF
--- a/internal/utils/varint.go
+++ b/internal/utils/varint.go
@@ -1,0 +1,101 @@
+package utils
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+)
+
+// taken from the QUIC draft
+const (
+	maxVarInt1 = 63
+	maxVarInt2 = 16383
+	maxVarInt4 = 1073741823
+	maxVarInt8 = 4611686018427387903
+)
+
+// ReadVarInt reads a number in the QUIC varint format
+func ReadVarInt(b io.ByteReader) (uint64, error) {
+	firstByte, err := b.ReadByte()
+	if err != nil {
+		return 0, err
+	}
+	// the first two bits of the first byte encode the length
+	len := 1 << ((firstByte & 0xc0) >> 6)
+	b1 := firstByte & (0xff - 0xc0)
+	if len == 1 {
+		return uint64(b1), nil
+	}
+	b2, err := b.ReadByte()
+	if err != nil {
+		return 0, err
+	}
+	if len == 2 {
+		return uint64(b2) + uint64(b1)<<8, nil
+	}
+	b3, err := b.ReadByte()
+	if err != nil {
+		return 0, err
+	}
+	b4, err := b.ReadByte()
+	if err != nil {
+		return 0, err
+	}
+	if len == 4 {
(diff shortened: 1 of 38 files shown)


internal/utils/varint.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	b8, err := b.ReadByte()
+	if err != nil {
+		return 0, err
+	}
+	return uint64(b8) + uint64(b7)<<8 + uint64(b6)<<16 + uint64(b5)<<24 + uint64(b4)<<32 + uint64(b3)<<40 + uint64(b2)<<48 + uint64(b1)<<56, nil
+}
+
+// WriteVarInt writes a number in the QUIC varint format
+func WriteVarInt(b *bytes.Buffer, i uint64) {
```
lucas-clemente: Nit: I think this could be nicer using else if and w/o the returns. Potentially also easier on the compiler.

internal/wire/ack_frame.go
```diff
@@ -9,340 +9,166 @@ import (
...
-				ackRange.First = ackRange.Last - length + 1
-				frame.AckRanges = append(frame.AckRanges, ackRange)
-			}
-
-			if length > 0 {
-				lastRangeComplete = true
-			}
-
-			inLongBlock = (ackBlockLength == 0)
+		gap := protocol.PacketNumber(g)
```
lucas-clemente: Maybe introduce a utils.ReadVarPacketNumber?

internal/wire/ack_frame.go
```diff
@@ -9,340 +9,166 @@ import (
...
 
-	missingSequenceNumberDeltaLen := f.getMissingSequenceNumberDeltaLen()
-	if missingSequenceNumberDeltaLen != protocol.PacketNumberLen1 {
-		typeByte ^= (uint8(missingSequenceNumberDeltaLen / 2))
-	}
+	b.WriteByte(0xe)
+	utils.WriteVarInt(b, uint64(f.LargestAcked))
+	utils.WriteVarInt(b, encodeAckDelay(f.DelayTime))
 
+	// TODO: limit the number of ACK ranges, such that the frame doesn't grow larger than an upper bound
```
lucas-clemente: I'm still fine with introducing an arbitrary cutoff. Maybe we should raise this in the IETF though.

marten-seemann: What exactly do you want to raise there?

internal/wire/ack_frame.go
marten-seemann: Then we'd need the same for stream IDs and byte counts. Not sure if it's worth it.

internal/wire/ack_frame.go
lucas-clemente: Whether the spec should define a max number of ACK ranges.

internal/utils/varint.go
marten-seemann: Done.

Issue #964, opened 2017-11-30
ignore STOP_WAITING frames


marten-seemann:
We still need to use them for gQUIC for now, but we should remove them for IETF QUIC (and derive the lower bound for acking packets ourselves).

PR #969, opened 2017-12-01
bundle PING frame with other frames

marten-seemann:
Should be merged after #968.

We're sending a retransmittable packet every 20 packets (if there are no other frames to send). To make a packet retransmittable, we add a PING frame. We should bundle this PING with an ACK.

This will be especially important when we remove STOP_WAITING frames (#964). Then we keep track of the largest acked of every retransmittable packet sent to increase the lower bound of packets we include in the ACK frame. That means we need to make sure that a packet containing an ACK actually is a retransmittable packet from time to time.

### DIFF
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -27,12 +27,13 @@ type packetPacker struct {
 	packetNumberGenerator *packetNumberGenerator
 	streamFramer          *streamFramer
 
-	controlFrames    []wire.Frame
-	stopWaiting      *wire.StopWaitingFrame
-	ackFrame         *wire.AckFrame
-	leastUnacked     protocol.PacketNumber
-	omitConnectionID bool
-	hasSentPacket    bool // has the packetPacker already sent a packet
+	controlFrames                 []wire.Frame
+	stopWaiting                   *wire.StopWaitingFrame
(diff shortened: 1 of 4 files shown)


packet_packer.go
```diff
@@ -146,6 +147,14 @@ func (p *packetPacker) PackPacket() (*packedPacket, error) {
 	if len(payloadFrames) == 1 && p.stopWaiting != nil {
 		return nil, nil
 	}
+	if (len(payloadFrames) == 1 && p.ackFrame != nil) || (len(payloadFrames) == 2 && p.ackFrame != nil && p.stopWaiting != nil) { // check if this packet only contains an ACK and / or STOP_WAITING
```
lucas-clemente: This looks very fragile. We already have code to determine whether a frame is retransmittable, right? Can we introduce a `hasRetransmittableFrames([]frames.Frame) bool` function that iterates the slice here?

marten-seemann: Good idea. Actually, that function already exists: `ackhandler.HasRetransmittableFrames([]wire.Frame) bool`.

marten-seemann:
@lucas-clemente: I updated the PR. PHAL.

PR #974, opened 2017-12-05
implement the recent changes to the version negotiation packet

### LINKED ISSUE #972: implement recent changes to the version negotiation packet
The header format was changed - again...: https://github.com/quicwg/base-drafts/pull/968

The version negotiation packet now looks different, so we have to make changes both to the writing and the parsing logic.

marten-seemann:
Fixes #972. Should be merged after #961.

Our header parsing is not perfect. Technically, nothing but the connection ID and the version field are an invariant, so technically, we should read the version and stop parsing the header if it's an unsupported version. Practically, the current parsing should work for both gQUIC and IETF QUIC.

### DIFF
--- a/client.go
+++ b/client.go
@@ -286,10 +286,8 @@ func (c *client) handlePacket(remoteAddr net.Addr, packet []byte) {
 		return
 	}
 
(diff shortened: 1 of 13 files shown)


marten-seemann:
Yeah, the header is changing quite often currently. And the next change is already on the horizon: varint packet number encoding.

PR #980, opened 2017-12-05
implement stateless handling of Initial packets for the TLS server

marten-seemann:
Depends on https://github.com/bifurcation/mint/pull/145 and https://github.com/bifurcation/mint/pull/151. It can be reviewed right now, the interaction with mint is not the interesting part of this PR.

This is a pretty big change, which unfortunately cannot be split up into smaller parts. Here's the high-level description: When the server receives an Initial packet from the client, it doesn't create a session yet. It only creates what is needed to handle that packet (e.g. the nullAEAD and the mint connection). When a stateless retry is performed, the server sends out the Retry packet and forgets all of this state. Only once the Initial packet is accepted, a new session is created.
The client on the other hand creates its session immediately. By doing so, we're making use of the loss recovery logic of QUIC right away (which will be even more important when we support 0-RTT). If the server does a retry, the session is closed and replaced by a new session. Among others, this procedure makes sure that and no retransmissions are sent for the Initial packet after receiving the Retry, and that the stream offset is properly reset.


internal/handshake/crypto_stream_conn.go
```diff
@@ -0,0 +1,99 @@
...
+	if c.stream != nil {
+		return c.stream.Write(p)
+	}
+	return c.writeBuf.Write(p)
+}
+
+// GetDataForWriting returns all data currently in the write buffer, and resets this buffer.
+func (c *CryptoStreamConn) GetDataForWriting() []byte {
+	defer c.writeBuf.Reset()
+	return c.writeBuf.Bytes()
```
lucas-clemente: You have to copy the slice here I think:

> The slice is valid for use only until the next buffer modification (that is, only until the next call to a method like Read, Write, Reset, or Truncate). 

https://golang.org/pkg/bytes/#Buffer.Bytes

mint_utils.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if err := mconf.Init(pers == protocol.PerspectiveClient); err != nil {
+		return nil, err
+	}
+	return mconf, nil
+}
+
+// unpackInitialOrRetryPacket unpacks packets Initial and Retry packets
+// These packets must contain a STREAM_FRAME for the crypto stream, starting at offset 0.
+func unpackInitialPacket(aead crypto.AEAD, hdr *wire.Header, data []byte, version protocol.VersionNumber) (*wire.StreamFrame, error) {
```
lucas-clemente: Where do we check the initial packet length?

stream.go
```diff
@@ -481,3 +487,18 @@ func (s *stream) IsFlowControlBlocked() bool {
...
+// It must not be called concurrently with any other stream methods, especially Read and Write.
+func (s *stream) SetReadOffset(offset protocol.ByteCount) {
+	s.readOffset = offset
+	s.frameQueue.readPosition = offset
+}
+
+// SetWriteOffset sets the write offset.
+// It is only needed for the crypto stream.
+// It must not be called concurrently with any other stream methods, especially Read and Write.
+func (s *stream) SetWriteOffset(offset protocol.ByteCount) {
```
lucas-clemente: This function isn't used anywhere right now, is it?

mint_utils.go
marten-seemann: Not yet, since we're not sending the right length yet: #886 will fix this once this PR is merged.

internal/handshake/crypto_stream_conn.go
marten-seemann: Good point, I missed that.

PR #983, opened 2017-12-05
fix counting of incoming and outgoing streams for the client

### LINKED ISSUE #982: incoming / outgoing stream count wrong for the client

marten-seemann:
Fixes #982.


marten-seemann:
I just discovered that this wasn't actually a bug, but just a super-weird naming of variables: for a client outgoing streams would be counted using the variable numIncomingStreams and incoming streams using numOutgoingStreams. This naming is so confusing that it might justify the bug label anyway ;)

PR #985, opened 2017-12-05
add an integration tests concurrently opening and closing many streams

### LINKED ISSUE #984: add integration tests opening and closing a lot of streams
In order to test the max stream limits. Those tests should have been able to catch #982.

marten-seemann:
Fixes #984.

The current max stream limit (without the MAX_STREAM_ID frame) is inherently racy, so I'm using a small timeout here (5 ms per stream), which should be plenty to prevent the race condition. This integration test is valuable since it makes sure that it's possible to use a large number of streams (>> number of concurrent streams) in one session.

### DIFF
--- a/integrationtests/self/stream_test.go
+++ b/integrationtests/self/stream_test.go
@@ -1,0 +1,133 @@
+package self_test
+
+import (
+	"crypto/tls"
+	"fmt"
+	"io"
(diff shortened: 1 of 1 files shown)


lucas-clemente:
Does this test actually work if it it succeeds without being based off of the fix #983?

PR #987, opened 2017-12-06
remove superfluous goroutine

julienschmidt:
`handleHeaderStream` is already called in a new goroutine.
The loop, which is currently run in a new goroutine, can be run in the same goroutine as `handleHeaderStream`.


julienschmidt:
One of the tests is failing. Is it a somehow flaky test or did I miss something?

PR #997, opened 2017-12-07
refactor sending of stream data

marten-seemann:
This PR removes two methods, `ShouldSendFin` and `SentFin` from the `stream`.

This will make splitting the stream up into two unidirectional stream halves (#996) easier.


marten-seemann:
@lucas-clemente: This is currently the most important PR to review.

PR #998, opened 2017-12-07
adapt TLS state handling to mint changes

julienschmidt:
bifurcation/mint#151 was merged an currently breaks quic-go.
I think it will be eventually fixed by #980, but that seems to be a massive change. For now, it would be nice to just make the minimal changes to let this package compile again.
This PR does exactly that changes and nothing else.


marten-seemann:
Thanks for the PR.
I just merged #980, so this won't be needed any more.

PR #1005, opened 2017-12-08
shorten blocking call of Server.handleRequest

julienschmidt:
The h2quic server calls handleRequest in a loop. Thus, to handle requests as fast as possible, the blocking call of handleRequest should be as short as possible, to minimize head-of-line blocking of other requests. 
This PR moves some more code in the goroutine started from handleRequest, including some I/O (the EOF read), allocations and function calls that can be made from the other goroutine.

### DIFF
--- a/h2quic/server.go
+++ b/h2quic/server.go
@@ -164,8 +164,6 @@ func (s *Server) handleRequest(session streamCreator, headerStream quic.Stream, 
 		return err
 	}
 
-	req.RemoteAddr = session.RemoteAddr().String()
-
 	if utils.Debug() {
 		utils.Infof("%s %s%s, on data stream %d", req.Method, req.Host, req.RequestURI, h2headersFrame.StreamID)
 	} else {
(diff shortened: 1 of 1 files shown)


marten-seemann:
Hi @julienschmidt, thanks a lot for this PR! This makes a lot sense. 

Is it correct that currently, if someone sends us a H2 frame with the End Flag, but doesn't close the QUIC stream, will effectively DoS his own connection, by making the loop in `handleHeaderStream` block on the `Read` call, so that no new requests can be handled?

julienschmidt:
I guess it might indeed be possible to construct cases where that Read is blocking all other requests.  Maybe a test for that would be good. (But I don't have time to write it right now)

marten-seemann:
I don't think we need a case for that with your fix.
If a client really wants to keep the QUIC stream open after closing the H2 stream, it should feel free to do so. QUIC limits the number of open streams, so a client can't exhaust the server's resources by opening an unlimited number of streams. Such a client would just run into this limit really fast, and then not be able to perform any more requests.

Thanks for your debugging effort and for submitting this patch!

PR #1006, opened 2017-12-08
decrease the threshold to send MAX_{STREAM}_DATA frames

marten-seemann:
This will increase the performance high BDP connections.

The optimization to send every WINDOW_UPDATE twice, which we took from Chrome, was apparently an unintended optimisation there (aka bug), so now that we're sending the updates more often, it should be even safer to remove this.

### DIFF
--- a/internal/flowcontrol/base_flow_controller.go
+++ b/internal/flowcontrol/base_flow_controller.go
@@ -67,9 +67,9 @@ func (c *baseFlowController) AddBytesRead(n protocol.ByteCount) {
 // getWindowUpdate updates the receive window, if necessary
 // it returns the new offset
 func (c *baseFlowController) getWindowUpdate() protocol.ByteCount {
-	diff := c.receiveWindow - c.bytesRead
-	// update the window when more than half of it was already consumed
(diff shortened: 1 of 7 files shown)


internal/flowcontrol/base_flow_controller.go
```diff
@@ -99,7 +99,7 @@ func (c *baseFlowController) maybeAdjustWindowIncrement() {
 
 	timeSinceLastWindowUpdate := time.Since(c.lastWindowUpdateTime)
 	// interval between the window updates is sufficiently large, no need to increase the increment
-	if timeSinceLastWindowUpdate >= 2*rtt {
+	if timeSinceLastWindowUpdate >= 4*protocol.WindowUpdateThreshold*rtt {
```
lucas-clemente: Nit: Magic number here ;)

marten-seemann: Yes, and this magic number comes from here: https://docs.google.com/document/d/1F2YfdDXKpy20WVKJueEf4abn_LVZHhMUMS5gX6Pgjl4/edit#heading=h.o1yqsbo737tc.

What is your suggestion?

PR #1007, opened 2017-12-10
don't add more STREAM frames to a packet if remaining size is too small

marten-seemann:
This is an optimization, which will become important for a later refactoring of the `stream`. 
Since it's a useful change in itself, I'd like to merge it as a separate PRö

### DIFF
--- a/internal/protocol/server_parameters.go
+++ b/internal/protocol/server_parameters.go
@@ -122,3 +122,9 @@ const ClosedSessionDeleteTimeout = time.Minute
 
 // NumCachedCertificates is the number of cached compressed certificate chains, each taking ~1K space
 const NumCachedCertificates = 128
+
+// MinStreamFrameSize is the minimum size that has to be left in a packet, so that we add another STREAM frame.
+// This avoids splitting up STREAM frames into small pieces, which has 2 advantages:
+// 1. it reduces the framing overhead
+// 2. it reduces the head-of-line blocking, when a packet is lost
+const MinStreamFrameSize ByteCount = 128
--- a/stream_framer.go
+++ b/stream_framer.go
@@ -70,32 +70,32 @@ func (f *streamFramer) PopCryptoStreamFrame(maxLen protocol.ByteCount) *wire.Str
 	return frame
 }
 
-func (f *streamFramer) maybePopFramesForRetransmission(maxLen protocol.ByteCount) (res []*wire.StreamFrame, currentLen protocol.ByteCount) {
+func (f *streamFramer) maybePopFramesForRetransmission(maxTotalLen protocol.ByteCount) (res []*wire.StreamFrame, currentLen protocol.ByteCount) {
 	for len(f.retransmissionQueue) > 0 {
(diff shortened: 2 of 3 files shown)


internal/protocol/server_parameters.go
```diff
@@ -122,3 +122,9 @@ const ClosedSessionDeleteTimeout = time.Minute
 
 // NumCachedCertificates is the number of cached compressed certificate chains, each taking ~1K space
 const NumCachedCertificates = 128
+
+// MinStreamFrameSize is the minimum size that has to be left in a packet, so that we add another STREAM frame
```
lucas-clemente: Nit: Punctuation.

stream_framer.go
```diff
@@ -71,33 +71,32 @@ func (f *streamFramer) PopCryptoStreamFrame(maxLen protocol.ByteCount) *wire.Str
...
-func (f *streamFramer) maybePopFramesForRetransmission(maxLen protocol.ByteCount) (res []*wire.StreamFrame, currentLen protocol.ByteCount) {
+func (f *streamFramer) maybePopFramesForRetransmission(maxTotalLen protocol.ByteCount) (res []*wire.StreamFrame, currentLen protocol.ByteCount) {
 	for len(f.retransmissionQueue) > 0 {
 		frame := f.retransmissionQueue[0]
 		frame.DataLenPresent = true
 
 		frameHeaderLen, _ := frame.MinLength(f.version) // can never error
-		if currentLen+frameHeaderLen >= maxLen {
+		maxLen := maxTotalLen - currentLen
+		if maxLen < protocol.MinStreamFrameSize {
```
lucas-clemente: At least for the retransmissions, we should probably change this to

`if frameHeaderLen + frame.DataLen() > maxLen && maxLen < protocol.MinStreamFrameSize`

stream_framer.go
```diff
@@ -110,13 +109,16 @@ func (f *streamFramer) maybePopNormalFrames(maxBytes protocol.ByteCount) (res []
 		frame.Offset = s.GetWriteOffset()
 		// not perfect, but thread-safe since writeOffset is only written when getting data
 		frameHeaderBytes, _ := frame.MinLength(f.version) // can never error
-		if currentLen+frameHeaderBytes > maxBytes {
+		if currentLen+frameHeaderBytes > maxTotalLen {
 			return false, nil // theoretically, we could find another stream that fits, but this is quite unlikely, so we stop here
 		}
-		maxLen := maxBytes - currentLen - frameHeaderBytes
+		maxLen := maxTotalLen - currentLen
+		if maxLen < protocol.MinStreamFrameSize { // don't try to add new STREAM frames, if only little space is left in the packet
```
lucas-clemente: Maybe same as above, but I'll leave that up to you.

marten-seemann: The idea here is to exit from the callback early, since ranging over the streams is expensive.
As long as the majority of `Write`s are larger than the packet size, we won't often pack a packet that is smaller than the maximum packet size.

PR #1012, opened 2017-12-10
remove dependencies on internal packages in h2quic and examples

julienschmidt:
Examples should be code that can be copy-pasted and used as a starting point for own development using the package.
That is currently impossible, since the examples have dependencies on internal packages and can not be compiled without replacing all these dependencies.

Likewise should `h2quic` be just one of many possible protocols building upon QUIC.
One should thus be able to develop a package with the same functionality as `h2quic`.
Or in reverse: Everything which is required in `h2quic` should also be accessible to external code.
Like with the examples, the internal packages prevent currently exactly that.

This PR adjusts the code so that no dependencies on internal code are required in the examples. In `h2quic` the `utils` package remains as the only internal import (the functionality can be easily replicated).

I did not want to export an alias of the `protocol.ByteCount` type, since its just syntactic sugar. 
Instead, I made `protocol.ByteCount` just a type alias of `uint64`, so that other packages can define an interface using `CloseRemote(uint64)`, without needing to import the protocol package.

### DIFF
--- a/example/client/main.go
+++ b/example/client/main.go
@@ -12,9 +12,6 @@ import (
 	"github.com/lucas-clemente/quic-go/h2quic"
 )
 
-// For additional logging of quic-go internals, the QUIC_GO_LOG_LEVEL env var
-// can be set. See https://github.com/lucas-clemente/quic-go/wiki/Logging
-
 func main() {
 	tls := flag.Bool("tls", false, "activate support for IETF QUIC (work in progress)")
 	flag.Parse()
--- a/example/client/main.go
+++ b/example/client/main.go
@@ -12,6 +12,9 @@ import (
 	"github.com/lucas-clemente/quic-go/h2quic"
 )
 
+// For additional logging of quic-go internals, the QUIC_GO_LOG_LEVEL env var
+// can be set. See https://github.com/lucas-clemente/quic-go/wiki/Logging
+
 func main() {
 	tls := flag.Bool("tls", false, "activate support for IETF QUIC (work in progress)")
 	flag.Parse()
--- a/example/client/main.go
+++ b/example/client/main.go
@@ -3,32 +3,23 @@ package main
 import (
 	"bytes"
 	"flag"
+	"fmt"
 	"io"
 	"net/http"
 	"sync"
 
 	quic "github.com/lucas-clemente/quic-go"
 	"github.com/lucas-clemente/quic-go/h2quic"
-	"github.com/lucas-clemente/quic-go/internal/protocol"
-	"github.com/lucas-clemente/quic-go/internal/utils"
 )
 
 func main() {
-	verbose := flag.Bool("v", false, "verbose")
 	tls := flag.Bool("tls", false, "activate support for IETF QUIC (work in progress)")
 	flag.Parse()
 	urls := flag.Args()
 
-	if *verbose {
-		utils.SetLogLevel(utils.LogLevelDebug)
-	} else {
-		utils.SetLogLevel(utils.LogLevelInfo)
-	}
-	utils.SetLogTimeFormat("")
-
-	versions := protocol.SupportedVersions
+	versions := quic.SupportedVersions
 	if *tls {
-		versions = append([]protocol.VersionNumber{protocol.VersionTLS}, versions...)
+		versions = append([]quic.VersionNumber{quic.VersionTLS}, versions...)
 	}
 
 	hclient := &http.Client{
@@ -40,21 +31,21 @@ func main() {
 	var wg sync.WaitGroup
 	wg.Add(len(urls))
 	for _, addr := range urls {
-		utils.Infof("GET %s", addr)
+		fmt.Printf("GET %s\n", addr)
 		go func(addr string) {
 			rsp, err := hclient.Get(addr)
 			if err != nil {
 				panic(err)
 			}
-			utils.Infof("Got response for %s: %#v", addr, rsp)
+			fmt.Printf("Got response for %s: %#v\n", addr, rsp)
 
 			body := &bytes.Buffer{}
 			_, err = io.Copy(body, rsp.Body)
 			if err != nil {
 				panic(err)
 			}
-			utils.Infof("Request Body:")
-			utils.Infof("%s", body.Bytes())
+			fmt.Println("Request Body:")
+			fmt.Printf("%s\n", body.Bytes())
 			wg.Done()
 		}(addr)
 	}
--- a/example/main.go
+++ b/example/main.go
@@ -21,6 +21,9 @@ import (
 	"github.com/lucas-clemente/quic-go/h2quic"
 )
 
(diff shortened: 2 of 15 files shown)


example/main.go
```diff
@@ -39,6 +37,8 @@ type Size interface {
 	Size() int64
 }
 
+var verbose bool
```
lucas-clemente: We can probably remove the verbose flag altogether and add a reference to the QUIC_GO_LOG_LEVEL env.

interface.go
```diff
@@ -16,6 +16,16 @@ type StreamID = protocol.StreamID
 // A VersionNumber is a QUIC version number.
 type VersionNumber = protocol.VersionNumber
 
+// Some known QUIC versions.
+// Warning: This API should not be considered stable and might change soon.
+const (
+	Version39  VersionNumber = protocol.Version39
+	VersionTLS VersionNumber = protocol.VersionTLS
```
lucas-clemente: I'm not a fan of exposing the VersionTLS here as it's super experimental.

internal/protocol/protocol.go
```diff
@@ -59,7 +59,7 @@ type ConnectionID uint64
 type StreamID uint64
 
 // A ByteCount in QUIC
-type ByteCount uint64
+type ByteCount = uint64
```
lucas-clemente: We introduced the ByteCount type so that we get compiler errors on invalid casts & conversions. This change defeats that purpose, so I'm against it. I don't think it's necessary either, we can just add the cast wherever CloseRemote is called.

julienschmidt: The problem is that if one (without this change) defines an interface with `CloseRemote(uint64)`, it does NOT match any defined function with `CloseRemote(ByteCount)`. E.g. if one tries to copy h2quic (e.g. to adjust it to a similar protocol) one has a problem, since one can neither import `protocol` nor can one define an interface which would match `CloseRemote(ByteCount)`.

julienschmidt: Reading you comment again, I think I understood it wrong. Did you mean to just change the function signature to `CloseRemote(uint64)`, cast the param and then keep using `ByteCount` everywhere else?

lucas-clemente: Yes, change it to CloseRemote(uint64). CloseRemote is a layer violation by gQUIC (which is why it's not part of the public interface), it will go away with the new IETF HTTP mapping.

interface.go
julienschmidt: Should I remove the `VersionNumber` constants and the `SupportedVersions` altogether?

example/main.go
julienschmidt: Since this not only applies to those examples but to all programs using this package, I added a comment to the README instead.

interface.go
marten-seemann: It seems like the only way to completely disentangle the examples from internal packages would be to declare `quic.VersionTLS` (in order to support the tls command line flag). I definitely don't want to do that, `VersionTLS` permanently changes and should under no circumstances be used.

While I generally agree that the example should be copy-pastable, I'm using -tls (as well as -v) **very** heavily in my development workflow, so I'm not really happy to merge anything that makes my life harder.

example/client/main.go
```diff
@@ -3,32 +3,23 @@ package main
...
 	tls := flag.Bool("tls", false, "activate support for IETF QUIC (work in progress)")
 	flag.Parse()
 	urls := flag.Args()
 
-	if *verbose {
-		utils.SetLogLevel(utils.LogLevelDebug)
-	} else {
-		utils.SetLogLevel(utils.LogLevelInfo)
-	}
-	utils.SetLogTimeFormat("")
```
marten-seemann: While I agree with disentangling the example from the internal packages in general, I'm using -v very much when developing, so at this point, I'm not a fan of merging anything that makes my life harder.

stream.go
```diff
@@ -386,8 +386,8 @@ func (s *stream) SetDeadline(t time.Time) error {
 }
 
 // CloseRemote makes the stream receive a "virtual" FIN stream frame at a given offset
-func (s *stream) CloseRemote(offset protocol.ByteCount) {
```
marten-seemann: The reason a `ByteCount` is used here, is that the H2 mapping in gQUIC requires this layer violation, which is why `CloseRemote` is not part of the public API.
This layer violation will be resolved in IETF QUIC

example/client/main.go
julienschmidt: Isn't it equivalent to just setting the env var `QUIC_GO_LOG_LEVEL=DEBUG`?

interface.go
julienschmidt: If we disentangle it, you could easily copy the example and adjust it to your needs 😃
How about making the TLS Version (is this really just _TLS_ or should it be the _IETF_ Version?) also available via an environment var?

example/client/main.go
marten-seemann: No, it's not, because the time format will be different. I don't think it's too hard to remove this code when copy-pasting the example, so I'd prefer not to make any change to the example client and server at this point.

interface.go
marten-seemann: This will at some point be the IETF version, at the moment it's a constantly changing mix between gQUIC and IETF QUIC.
There's no need for this change, if we don't make using the example server and client more complicated than it is today.

example/client/main.go
lucas-clemente: We could introduce another ENV for the format if you need that.

I think making the example useful for people is quite important, and having to edit the code makes the barrier of entry much harder.

Alternatively, we could split it into several examples, one set of "simple" ones that work right away, and more involved ones we can use for debugging. WDYT?

interface.go
lucas-clemente: I'm not sure I understand the issue here. You can always use an int in the versions field in the config, and still have the tls flag, even without the const being exported?

marten-seemann: I thought the whole point was to make the code better. Using a plain int achieves quite clearly the opposite, especially since the number I often change TLS version number during interop testing.

example/client/main.go
marten-seemann: Then you'd replace `go run example/main.go -v` by `QUIC_GO_LOG_LEVEL=debug DONT_SHOW_TIMESTAMPS=true go run example/main.go` to have a decent output. I don't think this provides any usability benefit to new users.

Furthermore, I don't think this is hard. We had a lot of people play with quic-go, and open all kinds of trivial issues, but this never caused any problems.
I'm don't want to merge any changes that make further development of quic-go harder for me.

lucas-clemente: Okay, then I vote for splitting up the examples into a simple client & server and a debug client & server with more features.

lucas-clemente: We can do that in a separate PR though.

marten-seemann: That sounds like a good idea.

marten-seemann:
Closing due to inactivity and age.

PR #1013, opened 2017-12-11
simplify the stream by directly popping STREAM frames from it

marten-seemann:
Depends on #1007 and #1011.

@lucas-clemente: Just the last commit needs to be reviewed. The other commits belong to the PRs cited above.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -19,9 +19,7 @@ type streamI interface {
 
 	AddStreamFrame(*wire.StreamFrame) error
 	RegisterRemoteError(error, protocol.ByteCount) error
-	HasDataForWriting() bool
-	GetDataForWriting(maxBytes protocol.ByteCount) (data []byte, shouldSendFin bool)
-	GetWriteOffset() protocol.ByteCount
+	PopStreamFrame(maxBytes protocol.ByteCount) *wire.StreamFrame
 	Finished() bool
 	Cancel(error)
 	// methods needed for flow control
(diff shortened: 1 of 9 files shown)


stream.go
```diff
@@ -256,29 +254,37 @@ func (s *stream) GetWriteOffset() protocol.ByteCount {
...
+	frame := &wire.StreamFrame{
+		StreamID:       s.streamID,
+		Offset:         s.writeOffset,
+		DataLenPresent: true,
+	}
+	frameLen, _ := frame.MinLength(s.version) // can never error
+	if frameLen >= maxBytes {                 // a STREAM frame must have at least one byte of data
+		return nil
+	}
+	frame.Data, frame.FinBit = s.getDataForWriting(maxBytes - frameLen)
```
lucas-clemente: We expect this function to be called quite often with no data for writing, so can we move this call and the checks below up above the stream frame allocation? The frame allocation in the packer itself was already visible in the profile.

stream.go
```diff
@@ -256,29 +254,37 @@ func (s *stream) GetWriteOffset() protocol.ByteCount {
 	return s.writeOffset
 }
 
-// HasDataForWriting says if there's stream available to be dequeued for writing
-func (s *stream) HasDataForWriting() bool {
+// PopStreamFrame returns the next STREAM frame that is supposed to be sent on this stream
```
lucas-clemente: Nit: punctuation

stream.go
marten-seemann: Github hid this comment from me, so here's my late reply: We need the frame to determine its length (`frameLen`), so we can't move this call. This was the same before this PR, the allocation just happened in the `streamFramer`.

PR #1015, opened 2017-12-12
pass RST_STREAM and MAX_STREAM_DATA directly to the stream

marten-seemann:
Depends on #1011 (and probably more, but I lost track of that).

@lucas-clemente: Just the last 3 commits need to be reviewed, the rest belongs to other PRs.

### DIFF
--- a/crypto_stream.go
+++ b/crypto_stream.go
@@ -11,7 +11,7 @@ import (
 type cryptoStreamI interface {
 	io.Reader
 	io.Writer
-	AddStreamFrame(*wire.StreamFrame) error
+	HandleStreamFrame(*wire.StreamFrame) error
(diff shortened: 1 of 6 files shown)


marten-seemann:
We have coupling between stream and wire already (in `AddStreamFrame`), and another PR will add `PopStreamFrame`. I found myself looking up session when `RegisterRemoteError` is actually called, so I thought it's easier to have the stream handle all frames that effect its state machine.

lucas-clemente:
I'm not convinced this improves things, but feel free to merge anyway.

marten-seemann:
I'm not 100% convinced as well, but it will make the further refactoring easier for now, so I'll merge it. It's internal API, so we can easily change it back later.

PR #1023, opened 2017-12-13
rename stream.Cancel to stream.CloseForShutdown

marten-seemann:
No functional changes expected.

`stream.CloseAbruptly` is **only** called when the session is closed.


lucas-clemente:
If it's only called by the session, can we put that in the name? E.g. OnSessionClosed()?

marten-seemann:
@lucas-clemente: Done. PHAL.

PR #1025, opened 2017-12-13
ignore STOP_WAITINGs, derive lower bound for packets to include in ACK

### LINKED ISSUE #964: ignore STOP_WAITING frames

marten-seemann:
Fixes #964.

This PR implements the logic to derive the lower bound of which packets to include in an ACK frame. It does so by keeping track of ACKs for ACKs. This will be needed when STOP_WAITING frames are removed (which will happen in IETF QUIC, and a future PR will already make that change for `VersionTLS`).
We're now still sending STOP_WAITING frames, since Chrome will need them. However, we're only parsing and then ignoring any STOP_WAITING that we receive.


ackhandler/interfaces.go
```diff
@@ -16,6 +16,7 @@ type SentPacketHandler interface {
 
 	SendingAllowed() bool
 	GetStopWaitingFrame(force bool) *wire.StopWaitingFrame
+	GetLowestPacketNotConfirmedAcked() protocol.PacketNumber
```
marten-seemann: @lucas-clemente: Any idea for a better name?

Changelog.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.6.1 (unreleased)
+- The lower boundary for packets included in ACKs is now derived, and the value sent in STOP_WAITING frames is ignored.
```
marten-seemann: @lucas-clemente: We probably want to reword this. Suggestions?

marten-seemann:
I must have closed this by mistake. Reopening.

Changelog.md
lucas-clemente: It's debatable whether it's worth a changelog entry, but SGTM.

ackhandler/interfaces.go
lucas-clemente: SGTM.

PR #1027, opened 2017-12-14
don't send STOP_WAITING frames (for IETF QUIC)

### LINKED ISSUE #1026: don't send STOP_WAITINGs (in IETF QUIC)

marten-seemann:
Fixes #1026. Depends on #1025.

### DIFF
--- a/packet_unpacker.go
+++ b/packet_unpacker.go
@@ -106,13 +106,7 @@ func (u *packetUnpacker) parseIETFFrame(r *bytes.Reader, typeByte byte, hdr *wir
 		if err != nil {
 			err = qerr.Error(qerr.InvalidWindowUpdateData, err.Error())
 		}
-	case 0x6:
-		// TODO(#964): remove STOP_WAITING frames
-		// TODO(#878): implement the MAX_STREAM_ID frame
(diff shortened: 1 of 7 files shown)


packet_unpacker.go
```diff
@@ -107,12 +107,8 @@ func (u *packetUnpacker) parseIETFFrame(r *bytes.Reader, typeByte byte, hdr *wir
 			err = qerr.Error(qerr.InvalidWindowUpdateData, err.Error())
 		}
 	case 0x6:
-		// TODO(#964): remove STOP_WAITING frames
 		// TODO(#878): implement the MAX_STREAM_ID frame
-		frame, err = wire.ParseStopWaitingFrame(r, hdr.PacketNumber, hdr.PacketNumberLen, u.version)
-		if err != nil {
-			err = qerr.Error(qerr.InvalidStopWaitingData, err.Error())
-		}
+		fallthrough
```
lucas-clemente: How is falling through correct here? Should we error instead?

marten-seemann: I'll just remove this case (which is what we do for all other not yet implemented frame types).

PR #1028, opened 2017-12-14
add the offset to the BLOCKED and STREAM_BLOCKED frames

### LINKED ISSUE #986: implement recent changes to the debugging frames
https://github.com/quicwg/base-drafts/pull/928

marten-seemann:
Fixes #986.


internal/wire/blocked_frame.go
```diff
@@ -23,13 +32,14 @@ func (f *BlockedFrame) Write(b *bytes.Buffer, version protocol.VersionNumber) er
...
 	typeByte := uint8(0x08)
 	b.WriteByte(typeByte)
+	utils.WriteVarInt(b, uint64(f.Offset))
 	return nil
 }
 
 // MinLength of a written frame
 func (f *BlockedFrame) MinLength(version protocol.VersionNumber) protocol.ByteCount {
-	if !version.UsesIETFFrameFormat() { // writing this frame would result in a legacy BLOCKED being written, which is longer
+	if !version.UsesIETFFrameFormat() {
```
lucas-clemente: Is this necessary here, given that this is the non-gQUIC BLOCKED frame? If it is, then we should also check the version when parsing / writing.

marten-seemann: gQUIC BLOCKED frames are parsed into `wire.BlockedFrame`s and `wire.StreamBlockedFrame`s by https://github.com/lucas-clemente/quic-go/blob/master/internal/wire/blocked_frame_legacy.go#L18. That way, the only place we need to care about the wire representation of these frames is in the `wire` package (and when unpacking).

PR #1030, opened 2017-12-14
when the stream write deadline expires, return the data length written

### LINKED ISSUE #1029: stream.Write should return the number of bytes written when the write deadline expires

marten-seemann:
Fixes #1029.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -243,13 +243,10 @@ func (s *stream) Write(p []byte) (int, error) {
 		s.mutex.Lock()
 	}
 
-	if err != nil {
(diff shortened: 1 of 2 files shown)


stream.go
```diff
@@ -243,11 +243,11 @@ func (s *stream) Write(p []byte) (int, error) {
 		s.mutex.Lock()
 	}
 
-	if err != nil {
```
lucas-clemente: Everything below here can be replaced with `return s.err, len(p) - len(s.dataForWriting)`.

PR #1032, opened 2017-12-15
make STOP_SENDING error code a uint16

marten-seemann:
It's an application error code, so it shouldn't be a qerr.


marten-seemann:
Introducing an `ErrorCode` is the better way to go. I'll prepare a PR.

PR #1035, opened 2017-12-15
refactor stream to support canceling Read and Write

### LINKED ISSUE #748: properly document what Stream.Close and Stream.Reset do
Close: closing the write-side.
Reset: closing both sides.

marten-seemann:
Closes #748.
Depends on #1033.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -14,6 +14,19 @@ import (
 	"github.com/lucas-clemente/quic-go/internal/wire"
 )
 
+type streamCanceledError struct {
+	error
+	errorCode protocol.ApplicationErrorCode
+}
+
+func (streamCanceledError) Canceled() bool                             { return true }
+func (e streamCanceledError) ErrorCode() protocol.ApplicationErrorCode { return e.errorCode }
+
+var _ StreamError = &streamCanceledError{}
+var _ error = &streamCanceledError{}
+
+const errorCodeStoppingGQUIC protocol.ApplicationErrorCode = 7
+
 type streamI interface {
 	Stream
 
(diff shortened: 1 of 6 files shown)


stream.go
```diff
@@ -48,27 +61,24 @@ type stream struct {
...
-	resetLocally utils.AtomicBool
-	// resetRemotely is set if HandleRstStreamFrame() is called
-	resetRemotely utils.AtomicBool
+	closedForShutdown bool // set when CloseForShutdown() is called
+	finRead           bool // set once we read a frame with a FinBit
+	finishedWriting   bool // set once Close() is called
+	canceledWrite     bool // set when CancelWrite() is called
+	canceledRead      bool // set when CancelRead() is called
+	finSent           bool // set when a STREAM_FRAME with FIN bit has b
+	resetRemotely     bool // set when HandleRstStreamFrame() is called
```
lucas-clemente: This is non-blocking for this PR, but I'd really like to get rid of all these bools and replace them with a state machine enum and a single var in the stream. Let's discuss this during one of the next weeklies.

stream.go
```diff
@@ -227,7 +238,7 @@ func (s *stream) Write(p []byte) (int, error) {
 			err = errDeadline
 			break
 		}
-		if s.dataForWriting == nil || s.err != nil {
+		if s.dataForWriting == nil || s.canceledWrite || s.closeForShutdownErr != nil {
```
lucas-clemente: I don't think we need the canceledWrite here.

stream.go
```diff
@@ -48,27 +61,24 @@ type stream struct {
...
-	// finisedWriting is set once Close() is called
-	finishedWriting utils.AtomicBool
-	// resetLocally is set if Reset() is called
-	resetLocally utils.AtomicBool
-	// resetRemotely is set if HandleRstStreamFrame() is called
-	resetRemotely utils.AtomicBool
+	closedForShutdown bool // set when CloseForShutdown() is called
+	finRead           bool // set once we read a frame with a FinBit
+	finishedWriting   bool // set once Close() is called
+	canceledWrite     bool // set when CancelWrite() is called
```
lucas-clemente: … or a STOP_SENDING frame is received.

stream.go
```diff
@@ -161,14 +177,8 @@ func (s *stream) Read(p []byte) (int, error) {
...
-
-		if err != nil {
-			return bytesRead, err
-		}
-
-		m := utils.Min(len(p)-bytesRead, int(frame.DataLen())-s.readPosInFrame)
 
+		s.mutex.Unlock()
 		if bytesRead > len(p) {
 			return bytesRead, fmt.Errorf("BUG: bytesRead (%d) > len(p) (%d) in stream.Read", bytesRead, len(p))
```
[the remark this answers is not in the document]
marten-seemann: Good catch. I'll move the `mutex.Unlock()` down after the bug checks. Those checks are basically for free, but we probably want to unlock the mutex for the `copy`.

stream.go
marten-seemann: Why not? According to the `CancelWrite` doc string, it may be called concurrently with `Write`, and is supposed to make `Write` return immediately. I'm not sure how much sense it make to call those concurrently, but if we want to change it, we should be consistent with the documentation.

stream.go
```diff
@@ -395,79 +408,114 @@ func (s *stream) CloseRemote(offset protocol.ByteCount) {
...
+		errorCode: frame.ErrorCode,
+		error:     fmt.Errorf("Stream %d was reset with error code %d", s.streamID, frame.ErrorCode),
+	}
+	s.signalRead()
 	return nil
 }
 
-func (s *stream) finishedWriteAndSentFin() bool {
-	return s.finishedWriting.Get() && s.finSent.Get()
+func (s *stream) HandleStopSendingFrame(frame *wire.StopSendingFrame) {
```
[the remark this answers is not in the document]
marten-seemann: In this PR, `HandleStopSendingFrame` is only called internally from `HandleRstStreamFrame`, which locks the mutex.
The race will be introduced by #1036, where `HandleStopSendingFrame` is added to the `streamI` interface, and called from the session.

stream.go
marten-seemann: This also belongs in #1036.

PR #1040, opened 2017-12-18
split the stream into a receive and a send stream

marten-seemann:
This PR splits the `stream` into two parts: one `sendStream` and one `receiveStream`. This is the most important step towards unidirectional stream (#996).

I'm already exporting the `ReceiveStream` and the `SendStream` interfaces in our public API. We will need them in the h2quic package once we actually deal with unidirectional streams in the streamsMap and the session.

### DIFF
--- a/receive_stream.go
+++ b/receive_stream.go
@@ -1,0 +1,284 @@
+package quic
+
+import (
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/lucas-clemente/quic-go/internal/flowcontrol"
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/utils"
+	"github.com/lucas-clemente/quic-go/internal/wire"
+)
+
+type receiveStream struct {
+	mutex sync.Mutex
+
+	streamID protocol.StreamID
+
+	// onData tells the session that there's stuff to pack into a new packet
+	onData func()
+	// queueControlFrame queues a new control frame for sending
+	// it does not call onData
+	queueControlFrame func(wire.Frame)
+
+	frameQueue     *streamFrameSorter
+	readPosInFrame int
+	readOffset     protocol.ByteCount
+
+	closeForShutdownErr error
+	cancelReadErr       error
+	resetRemotelyErr    StreamError
+
+	closedForShutdown bool // set when CloseForShutdown() is called
+	finRead           bool // set once we read a frame with a FinBit
+	canceledRead      bool // set when CancelRead() is called
+	resetRemotely     bool // set when HandleRstStreamFrame() is called
+
+	readChan     chan struct{}
+	readDeadline time.Time
+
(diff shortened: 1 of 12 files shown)


marten-seemann:
We probably want to fix our godoc before merging this PR. Currently, it looks like this:
screen shot 2017-12-18 at 09 57 57
The reason is that godoc is **too stupid** to recognize methods on combined interfaces.
On the other hand, it is also not possible to write
```go
type Stream interface {
  ReceiveStream
  SendStream
}
```
since both unidirectional streams have the `StreamID()` method, and Go itself is **too stupid** to handle that (they still didn't resolve that in Go 1.10).
The only option I see here is to duplicate the method (including documentation). Will be a lot of duplicate code, but I don't see any other way.

receive_stream.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/lucas-clemente/quic-go/internal/flowcontrol"
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/utils"
+	"github.com/lucas-clemente/quic-go/internal/wire"
+)
+
+type receiveStream struct {
+	mutex sync.Mutex
+
+	streamID protocol.StreamID
```
lucas-clemente: Would it make sense to encapsulate the mutex and streamID into a baseStream?

session_test.go
```diff
@@ -1112,25 +1112,25 @@ var _ = Describe("Session", func() {
...
+
+			// 	sess.scheduleSending()
+			// 	go sess.run()
+			// 	defer sess.Close(nil)
+
+			// 	Eventually(mconn.written).Should(HaveLen(1))
+			// 	packet := <-mconn.written
+			// 	Expect(packet).To(ContainSubstring("foobar1"))
+			// 	Expect(packet).To(ContainSubstring("foobar2"))
+			// })
```
lucas-clemente: No commented code.

session_test.go
```diff
@@ -358,17 +358,17 @@ var _ = Describe("Session", func() {
...
+			// 	fc := mocks.NewMockStreamFlowController(mockCtrl)
+			// 	offset := protocol.ByteCount(0x4321)
+			// 	fc.EXPECT().UpdateSendWindow(offset)
+			// 	sess.cryptoStream.(*cryptoStream).flowController = fc
+			// 	err := sess.handleMaxStreamDataFrame(&wire.MaxStreamDataFrame{
+			// 		StreamID:   sess.version.CryptoStreamID(),
+			// 		ByteOffset: offset,
+			// 	})
+			// 	Expect(err).ToNot(HaveOccurred())
+			// })
```
lucas-clemente: No commented code.

receive_stream.go
marten-seemann: If you  share the `baseStream` between the `sendStream` and the `receiveStream`, you'd only have a single mutex instead of two. Two is obviously better here.

PR #1048, opened 2017-12-18
 pass an interface containing the callbacks to the stream

### LINKED ISSUE #1041: pass queueControlFrame to crypto stream
This is currently nil, so if the stream tries to queue a control frame, this would panic. Currently, there's no way we can make the crypto stream queue a control frame, but it feels kind of fragile, and a source for future panics, if we don't keep this in mind all the time. It would be easy to just pass the `queueControlFrame` callback to the crypto stream as well.

marten-seemann:
Will be need for future optimisations. Also fixes #1041.


stream.go
```diff
@@ -14,6 +14,12 @@ const (
 	errorCodeStoppingGQUIC protocol.ApplicationErrorCode = 7
 )
 
+// The sender is notified by the stream about various events.
+type sender interface {
```
lucas-clemente: Nit: streamSender would make it more clear what the purpose is. Just `sender` is very generic.

PR #1057, opened 2017-12-20
make h2quic.Server.ListenAndServe error if it was closed before

marten-seemann:
This is (part of) what caused the Travis test to be flaky: There are some integration tests that are very fast (https://github.com/lucas-clemente/quic-go/blob/master/integrationtests/gquic/random_rtt_test.go#L22-L51, which we might want to move to the test tools anyway). In those cases, we would start and stop the QUIC test server so quickly, that the `Close` call happens before the `h2quic.Server.Serve` goroutine even spun up.

Note that this PR doesn't solve the problem yet.

### DIFF
--- a/h2quic/server.go
+++ b/h2quic/server.go
@@ -50,6 +50,7 @@ type Server struct {
 
 	listenerMutex sync.Mutex
 	listener      quic.Listener
+	closed        bool
 
 	supportedVersionsAsString string
 }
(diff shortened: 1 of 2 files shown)


h2quic/server.go
```diff
@@ -88,6 +89,10 @@ func (s *Server) serveImpl(tlsConfig *tls.Config, conn net.PacketConn) error {
 		return errors.New("use of h2quic.Server without http.Server")
 	}
 	s.listenerMutex.Lock()
+	if s.closed {
```
lucas-clemente: Can we defer unlock here?

marten-seemann: No, because the loop at the end of this function will block until the server is closed.

lucas-clemente: Sorry, GH didn't show the whole file 🙄

Issue #1055, opened 2017-12-20
race condition in rttStats


marten-seemann:
```
==================

WARNING: DATA RACE

Write at 0x00c42009a860 by goroutine 17:

  github.com/lucas-clemente/quic-go/congestion.(*RTTStats).UpdateRTT()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/congestion/rtt_stats.go:114 +0x35b

  github.com/lucas-clemente/quic-go/ackhandler.(*sentPacketHandler).maybeUpdateRTT()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/ackhandler/sent_packet_handler.go:256 +0x263

  github.com/lucas-clemente/quic-go/ackhandler.(*sentPacketHandler).ReceivedAck()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/ackhandler/sent_packet_handler.go:175 +0x3f2

  github.com/lucas-clemente/quic-go.(*session).handleAckFrame()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/session.go:637 +0xdc

  github.com/lucas-clemente/quic-go.(*session).handleFrames()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/session.go:522 +0x22c

  github.com/lucas-clemente/quic-go.(*session).handlePacketImpl()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/session.go:511 +0x803

  github.com/lucas-clemente/quic-go.(*session).run()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/session.go:366 +0x10fb

  github.com/lucas-clemente/quic-go.(*client).establishSecureConnection.func1()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/client.go:252 +0x63


Previous read at 0x00c42009a860 by goroutine 10:

  github.com/lucas-clemente/quic-go/internal/flowcontrol.(*baseFlowController).maybeAdjustWindowIncrement()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/congestion/rtt_stats.go:68 +0x105

  github.com/lucas-clemente/quic-go/internal/flowcontrol.(*baseFlowController).getWindowUpdate()

      /home/travis/gopath/src/github.com/lucas-clemente/quic-go/internal/flowcontrol/base_flow_controller.go:70 +0xe1

[...]

PR #1059, opened 2017-12-20
temporarily disable running the benchmark tests with race detector

marten-seemann:
The test keeps failing because of #1055.


lucas-clemente:
What is the ETA for the fix? I'd rather rollback the PR than disable the tests if this is a real-world issue in the current master.

marten-seemann:
The issue was fixed by #1065.

PR #1060, opened 2017-12-20
minor change to enable separate listen / main accept&process loop

varyukhichev:
Idea behind this minor change is following. If we have a server that should accept multiple quic/http2 connections at the same time, then we can have a tcp connection by which we would negotiate udp port for quic, and then start quic server on that port. Client, in its turn, gets port and starts quic/http2 communication over that port. However, if listen & accept are in the same method, then there is a race condition, since udp is stateless. In order to avoid that, first we need to create listener, then communicate port back to client and after that start main accept & process loop for quic/http2 server.

### DIFF
--- a/h2quic/server.go
+++ b/h2quic/server.go
@@ -107,6 +107,10 @@ func (s *Server) serveImpl(tlsConfig *tls.Config, conn net.PacketConn) error {
 	s.listener = ln
 	s.listenerMutex.Unlock()
 
+	return s.ServeLoop(ln)
+}
+
+func (s *Server) ServeLoop(ln quic.Listener) error {
 	for {
 		sess, err := ln.Accept()
 		if err != nil {


varyukhichev:
I signed CLA.

marten-seemann:
I don't think I fully understand this change. You can already run multiple h2/quic connections at the same time today, and the QUIC per is advertised via the Alt-Svc header (in fact, this is how Chrome does QUIC discovery). Can you please help me see the use case where this mechanism doesn't work?

varyukhichev:
I have tried to server many clients on the same port at the same time and each client transfers somewhat big amount of data, and very often I get "tls handshake timeout", which is set up to be 30 seconds (I tried even 120 seconds). Amount of data is 128k+ (both client request & server response), amount of clients is 50+. Network latency is 150-200ms, packet loss ratio around 10%.

marten-seemann:
To me, this sounds more like a loss recovery problem. 10% packet loss is an awful lot, and quic-go is known to not perform optimally under those conditions. Rewriting the loss recovery has been on my todo list for a long time, but I didn't have the time for that yet.

varyukhichev:
Well, as a quick solution this works. I have tried to span a new port (quic server) per each 10 clients that are accessing server at the same time and it works.

marten-seemann:
I'm not a fan of hacks, especially not if they change the public API.

PR #1063, opened 2017-12-21
rewrite flow control auto-tuning

### LINKED ISSUE #1056: flow control autotuning can be exploited
A peer can cause us to increase the flow control window size, no matter what bandwidth it has available by:

1. sending us the whole flow control window of data, except for byte 0
2. when all data arrived, sending us the missing byte 0

When we `Read` all the data available, this will trigger 1/threshold window updates at basically the same time, which will cause us to double the window.

This cannot be exploited for a DoS, since the window size has an upper limit.

marten-seemann:
Fixes #1056. Should be merged after #1062. 

Design doc: https://docs.google.com/document/d/1SExkMmGiz8VYzV3s9E35JQlJ73vhzCekKkDi85F1qCE/edit?usp=sharing.

### DIFF
--- a/internal/flowcontrol/connection_flow_controller_test.go
+++ b/internal/flowcontrol/connection_flow_controller_test.go
@@ -32,12 +32,12 @@ var _ = Describe("Connection Flow controller", func() {
 
 			fc := NewConnectionFlowController(receiveWindow, maxReceiveWindow, rttStats).(*connectionFlowController)
 			Expect(fc.receiveWindow).To(Equal(receiveWindow))
-			Expect(fc.maxReceiveWindowIncrement).To(Equal(maxReceiveWindow))
+			Expect(fc.maxReceiveWindowSize).To(Equal(maxReceiveWindow))
 		})
 	})
 
 	Context("receive flow control", func() {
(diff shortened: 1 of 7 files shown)


internal/flowcontrol/connection_flow_controller_test.go
```diff
@@ -97,14 +107,14 @@ var _ = Describe("Connection Flow controller", func() {
...
-		})
+		// It("doesn't auto-tune the window after the window size was increased", func() {
+		// 	setRtt(20 * time.Millisecond)
+		// 	controller.bytesRead = 9900 // receive window is 10000
+		// 	controller.lastWindowUpdateTime = time.Now().Add(-20 * time.Millisecond)
+		// 	controller.EnsureMinimumWindowSize(912)
+		// 	offset := controller.getWindowUpdate()
+		// 	Expect(controller.receiveWindowSize).To(Equal(protocol.ByteCount(912))) // no auto-tuning
+		// 	Expect(offset).To(Equal(protocol.ByteCount(9900 + 912)))
+		// })
```
lucas-clemente: No commented code.

Issue #1066, opened 2017-12-21
flaky client integration test


marten-seemann:
```
Client tests with QUIC version TLS dev version (WIP) 

  uploads a file

  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/self/client_test.go:83


• Failure [30.090 seconds]

Client tests

/home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/self/client_test.go:21

  with QUIC version TLS dev version (WIP)

  /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/self/client_test.go:45

    uploads a file [It]

    /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/self/client_test.go:83


    Expected error:

        <*url.Error | 0xc42b4a8120>: {

            Op: "Post",

            URL: "https://quic.clemente.io:59843/echo",

            Err: {

                ErrorCode: 97,

                ErrorMessage: "cannot read frame",

            },

        }

        Post https://quic.clemente.io:59843/echo: HeadersStreamDataDecompressFailure: cannot read frame

    not to have occurred


    /home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/self/client_test.go:89


    Full Stack Trace

    	/home/travis/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:69 +0x1ef

    github.com/onsi/gomega/internal/assertion.(*Assertion).ToNot(0xc42b415a40, 0xbbda00, 0xce5330, 0x0, 0x0, 0x0, 0xc42b415a40)

    	/home/travis/gopath/src/github.com/onsi/gomega/internal/assertion/assertion.go:39 +0xae

    github.com/lucas-clemente/quic-go/integrationtests/self_test.glob..func1.3.5()

    	/home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/self/client_test.go:89 +0x244

    github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc423503260, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)

    	/home/travis/gopath/src/github.com/lucas-clemente/quic-go/integrationtests/self/self_suite_test.go:14 +0x64

    testing.tRunner(0xc42021a0f0, 0x95f360)

[...]

PR #1068, opened 2017-12-21
call the scheduleSending callback when stream.Close is called

### LINKED ISSUE #1067: call scheduleSending callback when stream.Close is called
After calling stream.Close, the stream has data to send: The STREAM_FRAME containing the FIN bit.

marten-seemann:
Fixes #1067.

The stream needs to send the STREAM frame containing the FIN bit.


marten-seemann:
I ran the integration tests in my VM a couple of hundred times, and I'm pretty sure this also fixes #1066.

PR #1069, opened 2017-12-21
 introduce a queue of active streams for sending data

marten-seemann:
Should be merged after #1068.

The queue holds all streams that have data to send. When `stream.Write` or `stream.Close` are called, a stream is added to this queue. It is removed from the queue when all available (at that moment) data was sent.
This way, we don't need the round robin scheduling (which, for every packet sent, asked every single open stream if it had data) any more.


stream_framer.go
```diff
@@ -1,34 +1,53 @@
...
 type streamFramer struct {
-	streamsMap   *streamsMap
+	streamGetter streamGetter
 	cryptoStream cryptoStreamI
 	version      protocol.VersionNumber
 
 	retransmissionQueue []*wire.StreamFrame
+
+	streamQueueMutex sync.Mutex
+	activeStreams    map[protocol.StreamID]bool // map used as a set, bool is always true
```
lucas-clemente: Nit: map[StreamID]struct{} would be clearer.

PR #1076, opened 2017-12-22
 also use the onHasStreamData callback for the crypto stream

marten-seemann:
Depends on #1069.


stream_framer.go
```diff
@@ -57,14 +61,20 @@ func (f *streamFramer) HasFramesForRetransmission() bool {
...
+	hasCryptoStreamData := f.hasCryptoStreamData
+	f.streamQueueMutex.Unlock()
+	return hasCryptoStreamData
 }
 
-// TODO(lclemente): This is somewhat duplicate with the normal path for generating frames.
 func (f *streamFramer) PopCryptoStreamFrame(maxLen protocol.ByteCount) *wire.StreamFrame {
-	frame, _ := f.cryptoStream.popStreamFrame(maxLen)
+	frame, hasMoreData := f.cryptoStream.popStreamFrame(maxLen)
+	if !hasMoreData {
```
lucas-clemente: I think there's a race here and you have to lock the mutex before popping. What happens if a Write() and an AddActiveStream() happen while in this line?

This may also apply to the non-crypto streams path – GH doesn't show that here.

marten-seemann: Good catch. The non-crypto stream code path should be fine, the mutex is locked in `maybePopNormalFrames` from beginning to end.

marten-seemann:
@lucas-clemente: This PR is ready for review.

PR #1088, opened 2017-12-29
send CONNECTION_CLOSEs from the stateless server

### LINKED ISSUE #1002: report handshake errors when doing stateless packet handling
mint returns an internal error, so we never send out a response to an Initial packet, so the client will eventually run into a connection timeout. In fact, this is what happens in the corresponding integration test: https://github.com/lucas-clemente/quic-go/blob/master/integrationtests/self/handshake_test.go#L98-L102

marten-seemann:
Fixes #1002.

If an error occurs after a packet was successfully unpacked, we should send a CONNECTION_CLOSE.


marten-seemann:
@lucas-clemente: Please review this.

PR #1090, opened 2017-12-30
improve error handling in the h2quic client for header stream handling

marten-seemann:
When the underlying QUIC stream is closed, the close error should be returned. This always happens when receiving a CONNECTION_CLOSE from the server.
Furthermore, this adds a missing break statement in the case when receiving an invalid HTTP request.

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -97,24 +97,23 @@ func (c *client) handleHeaderStream() {
 	decoder := hpack.NewDecoder(4096, func(hf hpack.HeaderField) {})
 	h2framer := http2.NewFramer(nil, c.headerStream)
 
-	var lastStream protocol.StreamID
-
+	var err error
 	for {
-		frame, err := h2framer.ReadFrame()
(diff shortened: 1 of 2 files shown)


h2quic/client.go
```diff
@@ -97,44 +97,45 @@ func (c *client) handleHeaderStream() {
...
+			err = fmt.Errorf("response channel for stream %d not found", lastStream)
 			break
 		}
 
-		rsp, err := responseFromHeaders(mhframe)
+		var rsp *http.Response
+		rsp, err = responseFromHeaders(mhframe)
 		if err != nil {
-			c.headerErr = qerr.Error(qerr.InternalError, err.Error())
+			break
```
lucas-clemente: A lot of breaks are often a sign that the loop body should be refactored into a separate function. WDYT?

marten-seemann: SGTM.

It seems like there's a missing break statement for the last if of the loop. I'll send a new PR.

PR #1092, opened 2018-01-01
Few small customizations

costinm:
- allow a user of the library to use a custom dialer
- allow bi-directional streaming, don't wait for the request to be fully sent before returning.
- in case of transient errors (channel closed), retry the request.

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -26,6 +26,11 @@ type roundTripperOpts struct {
 
 var dialAddr = quic.DialAddr
 
+// SetQuicDialer overrides the dialer used for connecting to quic servers
+func SetQuicDialer(qd func(addr string, tlsConf *tls.Config, config *quic.Config) (quic.Session, error)) {
+	dialAddr = qd
+}
+
 // client is a HTTP2 client doing QUIC requests
 type client struct {
 	mutex sync.RWMutex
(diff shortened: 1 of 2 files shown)


marten-seemann:
Hello @costinm, thanks for this PR. The first PR of the new year :)

A few comments: 

- First of all, I'd prefer separate PRs for separate issues. It makes code review a lot easier.
- What is the use case for setting a custom dialer?
- I don't understand the bidirectional streaming. It seems to be contained in one of the other commits. Can you please send a new PR just containing this change?
- It's application-layer decision how to handle connection timeouts, so request retries shouldn't be in h2quic. h2quic currently doesn't handle closed clients correctly (there's an issue for that: #765).

costinm:
Sure, will break it apart.

On the other comments:
- use case is allowing the client to control the local port, or use a custom mechanism to resolve the name. I'm working on a 'device mesh' project using quic, and I need a small hack to use a multicast IP6 instead of the normal IP6 to work around some bugs in Android P2P.
- for bi-directional: the problem was that 'do' was waiting for the request body to be sent before returning.
After the change, the method return after headers are sent - so caller can read the response body while generating more request body data.
- connection timeouts should be application-layer decision, but this is a transient error happening at the low-level in QUIC.

Issue #747, opened 2017-07-19
implement DialContext in h2quic.RoundTripper


yinqiwen:
If set, 
1. Users can use other dns lib instead of the default one which may have  polluted dns result.
2. Log & monitor the handshake latency.

lucas-clemente:
LGTM, though I'd argue for implementing only `DialContext`, as `DialAddr` is deprecated for on [http.Transport](https://golang.org/pkg/net/http/#Transport).

marten-seemann:
What would the signature of our `DialContext` be?
```DialContext func(ctx context.Context, network, addr string) (quic.Session, error)```, right?

lucas-clemente:
No, it should return a `net.PacketConn`.

marten-seemann:
Makes sense. And this `net.PacketConn` would then be used for calling `quic.Dial`.

I'm marking this issue as want-help, since I don't think I will have the time to write a fix for this any time soon.

marten-seemann:
I just wrote a fix for this, and came to the conclusion that returning a `quic.Session` is indeed the right thing to do. Please see #1120 for details.

PR #1093, opened 2018-01-02
Allows using a custom dialer

costinm:
Use cases:
- listen to specific interface (for example IP6 link local)
- listen to specific port - for example if certain ports are configured with higher NAT timeouts
- customize how the name is resolved

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -26,6 +26,11 @@ type roundTripperOpts struct {
 
 var dialAddr = quic.DialAddr
 
+// SetQuicDialer overrides the dialer used for connecting to quic servers
+func SetQuicDialer(qd func(addr string, tlsConf *tls.Config, config *quic.Config) (quic.Session, error)) {
+	dialAddr = qd
+}
+
 // client is a HTTP2 client doing QUIC requests
 type client struct {
 	mutex sync.RWMutex


marten-seemann:
I think you can achieve all of this by directly initializing a [h2quic.Server](https://godoc.org/github.com/lucas-clemente/quic-go/h2quic#Server), can't you?

marten-seemann:
I'm closing this PR for now. Feel free to reopen if you you have a use case that can't be achieved by initializing the h2quic.Server directly.

costinm:
I don't see how you can achieve this using h2quic.Server - this sets the Dialer interface, used in client mode. There is nothing related to server used. 

Client uses a UDP socket - and that needs to be customized, just like the server-side socket. In particular, if quic is used for link local it needs a different way to setup, also if some ports are whitelisted
in firewalls or work better they need to be set instead of any ('0'). There are other things as well.

marten-seemann:
I was wrong. Of course, the server doesn't help you with dialing new connections... sorry for that, and thanks for the explanation!

I think the right way to fix this is by implementing something analogous to the `http.Transport.DialContext`. In fact, we already have an issue for that: #747. Please ping that issue if you plan to contribute a fix.

PR #1094, opened 2018-01-02
Stop waiting for request body sent, allowing 2-directional streaming.

costinm:
See net/http/h2_bundle.go scheduleBodyWrite() in the standard h2 API - the body send is scheduled in a separate thread, but the client doesn't wait for entire body to be sent, only for response headers to 
be received.

This allows quic to be used for example in gRPC, which requires the bidirectional streaming from h2.

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -44,6 +44,7 @@ type client struct {
 	headerErrored chan struct{} // this channel is closed if an error occurs on the header stream
 	requestWriter *requestWriter
 
+	// Responses pending header receipt.
 	responses map[protocol.StreamID]chan *http.Response
 }
 
@@ -180,34 +181,27 @@ func (c *client) RoundTrip(req *http.Request) (*http.Response, error) {
 		return nil, err
 	}
 
-	resc := make(chan error, 1)
+	// This will write the request body in a separate goroutine.
 	if hasBody {
 		go func() {
-			resc <- c.writeRequestBody(dataStream, req.Body)
+			c.writeRequestBody(dataStream, req.Body)
 		}()
 	}
 
 	var res *http.Response
 
 	var receivedResponse bool
-	var bodySent bool
 
-	if !hasBody {
-		bodySent = true
-	}
-
-	for !(bodySent && receivedResponse) {
+	// Wait until the response headers are fully received or an error happened.
+	// The request and response bodies may still be open.
+	for !(receivedResponse) {
 		select {
 		case res = <-responseChan:
 			receivedResponse = true
 			c.mutex.Lock()
 			delete(c.responses, dataStream.StreamID())
 			c.mutex.Unlock()
-		case err := <-resc:
-			bodySent = true
-			if err != nil {
-				return nil, err
-			}
+			// Response headers received - request may still be sending.
 		case <-c.headerErrored:
 			// an error occured on the header stream
 			_ = c.CloseWithError(c.headerErr)


marten-seemann:
That sounds like a really useful change. Thanks for the PR!

I always like have a running example, so I can see that the code is actually doing the right thing. I wanted to see how bidirectional streaming works in Go HTTP/2, so I tried to set up a minimal example: https://gist.github.com/marten-seemann/b1f04723bebc82f7108611c563f053e2. However, it doesn't work as intended: The client only receives the response once the server finished writing the body. Do you know what's going wrong here?

costinm:
On a quick look - are you missing the flush() ( using http.Flusher ) ? This is the most frequent problem I've seen ...

marten-seemann:
You're right! My example code works now :)

This is pretty cool. I'll have a look at your code later.

marten-seemann:
Hi @costinm,

I tried your code, and it works exactly as the H2/TCP example, so I'd like to merge this change.

I noticed that you ignore the error in `writeRequestBody`. I think this is generally the right thing to do, but we might have to be a bit more careful.

An error in `writeRequestBody` can occur:

1. when the connection errors. The next call to stream.Read will return the close error.
2. in gQUIC: when the server resets the stream (by sending a RST_STREAM). This would reset the steam in both directions, and be noticed the next time the application reads from the stream.
3. in IETF QUIC: when the server resets the stream (by sending a STOP_SENDING). Since STOP_SENDING only resets the (client's) send direction of the stream, this error would go unnoticed.
4. when `http.Request.Body.Read` errors.

Here's what I think about this:
1. Is correct.
2. Is inherent to gQUIC, and there's nothing we can do about this.
3. Is correct as well: A STOP_SENDING is sent when a server is not interested in the client's request body, for example if the client starts a POST to a site that would 404. Nobody is going to read that body, so the server sends the STOP_SENDING. The right behavior is to ignore the `Stream.Write` error, and return the response the server sent (otherwise we could not read the 404).
4. is the problematic case. We probably should probably reset the data stream in that case.

I'd like to hear what @lucas-clemente thinks about this.

costinm:
The code in h2 is similar. In general, there is no guarantee in the protocol that errors sending the request body will be visible. In most cases the server will detect the error and not return 200 - but even with HTTP1.1 it is possible for a server to return 200 before getting the full body (for uploads or streaming). 
Even when a write succeeds, and the packet is sent on the wire - it may still fail to be processed.

So I don't think it's a problem that quic needs to handle in a special way - in most cases ( normal POST, etc) the server will wait for the request body and return error if it can't get it. For bi-directional streaming like gRPC - the protocol takes care of this in other ways ( the error-in-trailers, etc).

BTW - I think you should reconsider the other patch ( handling of internal 'connection' reuse errors ), 
probably retrying in RoundTriper is not perfect but something is needed, it's bad behavior to have an 
error when nothing wrong happened. Application logic will get confused, and apps using http interfaces should not be aware of quic internal errors.

marten-seemann:
I made another example to check how H2 behaves when an error is encountered when reading the `http.Request.Body`: https://gist.github.com/marten-seemann/7f496f2bb98c096f799ae13cb150181c.
 In H2, nothing happens (it doesn't even close the stream), so I think it's okay if we just ignore the error.

@costinm: Would it be ok for you if I take your commit, make ignoring the error more explicit and also fix the test cases?

costinm:
Sure, sorry I've been busy - feel free to change it any way you want...

jared2501:
Hey guys, this would be pretty useful for me. Anything I can do to help push this along?

PR #1095, opened 2018-01-02
Retry the roundtrip on low-level public reset/idle errors.

costinm:
Should address Issue #765 

Public reset / idle are low level protocol behavior allowing client or server to reclaim resources, similar with 
h2 or http connection close. Most http clients handle this transparently, by opening new TCP connections,
since it is not an actual error. The change attempts to do the same in quic - it may be better to handle it
at a lower level, but this seems pretty clean and still allows lower levels to expose this state.

### DIFF
--- a/h2quic/roundtrip.go
+++ b/h2quic/roundtrip.go
@@ -11,6 +11,7 @@ import (
 
 	quic "github.com/lucas-clemente/quic-go"
 
+	"github.com/lucas-clemente/quic-go/qerr"
 	"golang.org/x/net/lex/httplex"
 )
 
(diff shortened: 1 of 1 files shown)


costinm:
A 'connection timeout' is caused by the connect failing - usually because the remote end can't be reached (DNS down, lack of network, etc). Agreed, like in all http clients it needs to be propagated.

The exceptions I'm handling are equivalent with a TCP 'keep alive' socket getting closed, or a H2 socket 
getting max streams or 'go away'. It is not handled as a 'connection timeout' in H2 case - just as internal
socket management, not exposed to the app.

marten-seemann:
The idle timeout actually is a connection timeout (and you can configure the timeout duration in the `quic.Config`), and in QUIC a Public Reset is sent when the server doesn't have the cryptographic state to decrypt a packet (e.g. after a reboot). Both of these events need to be handled at the application layer.

costinm:
Of course those events need to be handled at application - if they happen after the roundtrip, 
when the app is reading/writing to the body.  

Both public reset and quick session can happen when there is no stream (http2 request) active, and
a brand new connection is attempted. That's part of the rountrip method, and what this PR is fixing. 

Once the request body is sent and initial response received - RoundTrip is done, and the user code will
handle the 2 events. 

That's consistent with other http clients - if the keep-alive connection can't be used to send a new request, because of timeout or server reset/close, the client automatically opens a new TCP.

heri16:
@marten-seemann I would agree that it would be good to handle at application layer. However, the way the RoundTripper API is designed does not allow this to be easy to implement. Even if the application layer can catch the error, it seems there is no way for the application to redial the connection via the RoundTripper API. Last I looked at this was 2 months ago. Not sure whether the situation has changed.

marten-seemann:
@heri16 Again, this is #765.

PR #958, opened 2017-11-25
WIP implement send pacing

marten-seemann:
EDIT: I found some flaws in this, and I'm working on a fix. No need to review this PR at the current moment.

Fixes #355.

This PR adds send pacing according to the IETF Loss Recovery draft.
The most important change is in `session.sendPacket`. This function now only sends at most one packet at a time. Furthermore, it now has two return values, one bool indicating if a packet was sent, and the error. **Only if** a packet was sent, we now set the timer to the next time a packet should be sent according to the packet packer. This way, if we have a lot of data to send, we'll send it at the desired pacing rate. If there's no data to send, we avoid going through the run loop every single time.

## Benchmark results

The benchmarks were performed by having Chrome upload a 50 MB file to 1. a TCP/HTTPS server, a quic-go server and by having a quic-go client upload to a quic-go server.

without packet pacing:
```
+-----------------+----------------------------+--------------------------+--------------------------+
|                 | Chrome => TCP/HTTPS server | Chrome => quic-go server | quic-go client => server |
+-----------------+----------------------------+--------------------------+--------------------------+
| direct transfer | 156.35 ± 37.83             | 63.25 ± 8.66             | 109.16 ± 2.89            |
| 10ms RTT        | 69.50 ± 4.62               | 34.71 ± 3.82             | 12.14 ± 0.91             |
| 25ms RTT        | 30.17 ± 1.86               | 21.19 ± 7.02             | 4.06 ± 0.44              |
| 50ms RTT        | 15.44 ± 0.47               | 13.79 ± 1.18             | 2.17 ± 0.11              |
| 100ms RTT       | 7.84 ± 0.34                | 7.30 ± 0.11              | 1.62 ± 0.35              |
+-----------------+----------------------------+--------------------------+--------------------------+
Based on 4 samples of 50 MB. All values in MB/s.
```

with packet pacing:

```
+-----------------+----------------------------+--------------------------+--------------------------+
|                 | Chrome => TCP/HTTPS server | Chrome => quic-go server | quic-go client => server |
+-----------------+----------------------------+--------------------------+--------------------------+
| direct transfer | 120.58 ± 41.98             | 68.50 ± 7.22             | 23.03 ± 1.48             |
| 10ms RTT        | 70.38 ± 4.39               | 32.91 ± 5.87             | 22.39 ± 0.52             |
| 25ms RTT        | 27.98 ± 1.21               | 19.62 ± 3.54             | 19.77 ± 1.54             |
[...]

PR #1097, opened 2018-01-02
implement send pacing

### LINKED ISSUE #355: Add send pacing
We should implement pacing of sent packets, i.e. try to reduce us sending bursts of packets.

I think the best way to implement this is by returning a delay from the sentPacketHandler and using the existing timer in the session to delay sending.

See also https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-recovery.md (search for pacing).

marten-seemann:
Fixes #355. This is an improved version of #958.

This PR adds send pacing as it was described in the IETF Loss Recovery draft (Jana recently removed the details, since this draft is not the right place to specify a specific pacing algorithm).

The congestion calculates the delay between two packets, based on the RTT estimate and the congestion window.
In the session, there are two optimisations to make this perform better:
1. Since the pacing delay can be short O(10 microseconds), we don't want to wake up the run loop that often. We only set the pacing timer if we sent out a packet the last time the run loop ran (assuming that if one packet was sent, we have more data to send).
2. If the delay is to short, pacing is done in larger blocks. The minimum delay 100 mus, so if the pacing delay is supposed to be 20 mus, we'd send 5 packets at once.

### Benchmark results:

Both results rebased on #1081.

before:
```
+-----------------+----------------------------+--------------------------+--------------------------+
|                 | Chrome => TCP/HTTPS server | Chrome => quic-go server | quic-go client => server |
+-----------------+----------------------------+--------------------------+--------------------------+
| direct transfer | 124.33 ± 29.20             | 68.52 ± 10.16            | 114.52 ± 3.05            |
| 5ms RTT         | 111.12 ± 20.70             | 39.36 ± 6.03             | 26.31 ± 2.42             |
| 10ms RTT        | 69.08 ± 12.75              | 31.23 ± 3.62             | 12.53 ± 3.26             |
| 25ms RTT        | 31.66 ± 1.76               | 25.50 ± 1.29             | 3.81 ± 0.21              |
| 50ms RTT        | 15.83 ± 0.60               | 12.29 ± 1.98             | 2.70 ± 0.56              |
| 100ms RTT       | 8.67 ± 0.42                | 7.38 ± 0.08              | 1.35 ± 0.09              |
+-----------------+----------------------------+--------------------------+--------------------------+
Based on 5 samples of 50 MB. All values in MB/s.
```

after:
```
+-----------------+----------------------------+--------------------------+--------------------------+
|                 | Chrome => TCP/HTTPS server | Chrome => quic-go server | quic-go client => server |
+-----------------+----------------------------+--------------------------+--------------------------+
| direct transfer | 115.45 ± 40.58             | 83.29 ± 16.04            | 23.69 ± 1.38             |
| 5ms RTT         | 117.17 ± 6.87              | 37.93 ± 6.10             | 39.97 ± 2.55             |
| 10ms RTT        | 63.88 ± 6.37               | 31.66 ± 5.27             | 35.86 ± 4.37             |
| 25ms RTT        | 26.60 ± 0.79               | 23.23 ± 2.72             | 24.59 ± 1.06             |
| 50ms RTT        | 13.63 ± 0.45               | 11.90 ± 2.77             | 13.42 ± 0.67             |
| 100ms RTT       | 7.85 ± 0.93                | 7.39 ± 0.09              | 7.27 ± 0.12              |
+-----------------+----------------------------+--------------------------+--------------------------+
Based on 5 samples of 50 MB. All values in MB/s.
```


marten-seemann:
@lucas-clemente: I rebased this PR on top of all the performance improving PRs that you reviewed yesterday. This should be ready for review and merging now.

session.go
```diff
@@ -335,6 +335,9 @@ func (s *session) run() error {
 
 	var closeErr closeError
 	handshakeEvent := s.handshakeEvent
+	var pacingDeadline time.Time
```
lucas-clemente: Can you make this a member var as all the other deadlines? Then you can drop the param to resetTimer() and simplify some of the code below iiuc.

congestion/cubic_sender.go
```diff
@@ -76,15 +79,22 @@ func NewCubicSender(clock Clock, rttStats *RTTStats, reno bool, initialCongestio
 	}
 }
 
+// TimeUntilSend returns when the next packet should be sent.
 func (c *cubicSender) TimeUntilSend(now time.Time, bytesInFlight protocol.ByteCount) time.Duration {
```
lucas-clemente: I think it would be cleaner if this function returned an instant, not a duration. Then you wouldn't need the now param and could drop the double diff calculation here and in the session.

session.go
```diff
@@ -384,21 +385,66 @@ runLoop:
...
+			// Set the timer and restart the run loop.
+			s.maybeResetTimer(pacingDeadline)
+			continue
+		}
+
+		var startPacingTimer bool
+		numPackets, pacingDelay := s.sentPacketHandler.TimeUntilSend(now)
+		packetsSent := 1
+		if !s.sentPacketHandler.SendingAllowed() { // if congestion limited, at least try sending an ACK frame
+			if err := s.maybeSendAckOnlyPacket(); err != nil {
```
lucas-clemente: I only realized this now, but I think we have a bug here. SendingAllowed() also checks for the max tracked packet number. If that is limiting us (and not the congestion), we aren't allowed to send another packet.

session.go
```diff
@@ -384,21 +385,66 @@ runLoop:
...
 			s.keepAlivePingSent = true
+		} else if !pacingDeadline.IsZero() && now.Before(pacingDeadline) {
+			// If we get to this point before the pacing deadline, we should wait until that deadline.
+			// This can happen when scheduleSending is called, or a packet is received.
+			// Set the timer and restart the run loop.
+			s.maybeResetTimer(pacingDeadline)
+			continue
+		}
+
+		var startPacingTimer bool
```
lucas-clemente: This whole logic is too complicated to be inlined into the run loop. Can you extract into a separate function?

lucas-clemente: I don't think you need this bool if you just set the deadline directly.

session.go
```diff
@@ -384,21 +385,66 @@ runLoop:
...
+		numPackets, pacingDelay := s.sentPacketHandler.TimeUntilSend(now)
+		packetsSent := 1
```
lucas-clemente: It's weird to see this set to 1 before a packet is sent. Better: set it to 0 here and increment it in the loop.

session.go
lucas-clemente: I don't understand why sending an ACK-only packet doesn't have an effect on pacing. It can change state in the SPH, so you need similar logic to the normal case, right?

marten-seemann: What do you suggest?

marten-seemann: If we're congestion limited, we don't care about pacing (since pacing is not the limiting factor anyway).

marten-seemann:
#1112 merges the first three commits of this PR, #1108 addresses @lucas-clemente's review comments.

PR #1098, opened 2018-01-03
refactor the loop reading the header stream in the h2quic client

marten-seemann:
As @lucas-clemente suggested in #1090.

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -98,45 +98,43 @@ func (c *client) handleHeaderStream() {
 	h2framer := http2.NewFramer(nil, c.headerStream)
 
 	var err error
-	for {
-		var frame http2.Frame
-		frame, err = h2framer.ReadFrame()
-		if err != nil {
-			break
-		}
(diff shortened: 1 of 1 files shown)


h2quic/client.go
```diff
@@ -97,46 +97,45 @@ func (c *client) handleHeaderStream() {
...
-		hframe, ok := frame.(*http2.HeadersFrame)
-		if !ok {
-			err = errors.New("not a headers frame")
-			break
-		}
-		mhframe := &http2.MetaHeadersFrame{HeadersFrame: hframe}
-		mhframe.Fields, err = decoder.DecodeFull(hframe.HeaderBlockFragment())
-		if err != nil {
-			err = fmt.Errorf("cannot read header fields: %s", err.Error())
+		if err := c.readResponse(h2framer, decoder); err != nil {
```
lucas-clemente:
```
var err error
for err == nil { err = c.readResponse(...) }
// ...
```

Issue #952, opened 2017-11-13
implement Stateless Retries for gQUIC


marten-seemann:

marten-seemann:
Closing since we're not supporting gQUIC any more (in the master, that is).

PR #1101, opened 2018-01-03
remove the Range method in the streamsMap

marten-seemann:
`streamsMap.Range` was only used to pass the stream flow control window to all open streams after receiving the peer's transport parameters.

We still need to pass the stream flow control limits to the streams map, since 0-RTT, the client sends an unencrypted packet containing the full CHLO (including all transport parameters), and immediately sends an encrypted packet potentially containing stream data. Since the crypto stream is handled by a separate go routine, the transport parameters are not guaranteed to be processed before the second packet. Note that we can get rid of this when we support stateless connection establishment for gQUIC (#952).

### DIFF
--- a/session.go
+++ b/session.go
@@ -688,18 +688,13 @@ func (s *session) handleCloseError(closeErr closeError) error {
 
 func (s *session) processTransportParameters(params *handshake.TransportParameters) {
 	s.peerParams = params
-	s.streamsMap.UpdateMaxStreamLimit(params.MaxStreams)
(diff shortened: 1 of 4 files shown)


marten-seemann:
I think there's a race condition here: For 0-RTT, the client sends an unencrypted packet containing the full CHLO (including all transport parameters), and immediately sends an encrypted packet potentially containing stream data. Since the crypto stream is handled by a separate go routine, the transport parameters are not guaranteed to be processed before the second packet.

PR #1108, opened 2018-01-09
implement send pacing

### LINKED ISSUE #355: Add send pacing

marten-seemann:
Fixes #355.

This PR addresses the points raised in the review of #1097.

before:
```
+-----------------+----------------------------+--------------------------+--------------------------+
|                 | Chrome => TCP/HTTPS server | Chrome => quic-go server | quic-go client => server |
+-----------------+----------------------------+--------------------------+--------------------------+
| direct transfer | 154.48 ± 12.94             | 64.30 ± 9.24             | 95.82 ± 5.24             |
| 5ms RTT         | 105.47 ± 14.86             | 37.36 ± 6.19             | 27.05 ± 3.29             |
| 10ms RTT        | 65.81 ± 9.68               | 30.88 ± 3.44             | 11.77 ± 0.92             |
| 25ms RTT        | 28.92 ± 1.96               | 25.47 ± 1.77             | 4.13 ± 0.50              |
| 50ms RTT        | 14.16 ± 0.33               | 14.15 ± 0.44             | 2.34 ± 0.19              |
| 100ms RTT       | 7.37 ± 0.26                | 7.44 ± 0.05              | 1.52 ± 0.19              |
+-----------------+----------------------------+--------------------------+--------------------------+
Based on 5 samples of 50 MB. All values in MB/s.
```

after:
```
+-----------------+----------------------------+--------------------------+--------------------------+
|                 | Chrome => TCP/HTTPS server | Chrome => quic-go server | quic-go client => server |
+-----------------+----------------------------+--------------------------+--------------------------+
| direct transfer | 134.52 ± 23.89             | 50.99 ± 8.09             | 82.63 ± 6.30             |
| 5ms RTT         | 97.37 ± 20.27              | 40.55 ± 3.46             | 35.79 ± 8.09             |
| 10ms RTT        | 62.27 ± 7.43               | 30.81 ± 4.44             | 39.04 ± 7.12             |
| 25ms RTT        | 28.81 ± 1.06               | 24.82 ± 2.81             | 25.35 ± 1.59             |
| 50ms RTT        | 14.56 ± 0.23               | 14.61 ± 0.13             | 13.89 ± 0.61             |
| 100ms RTT       | 7.29 ± 0.14                | 6.87 ± 1.01              | 7.21 ± 0.03              |
+-----------------+----------------------------+--------------------------+--------------------------+
Based on 5 samples of 50 MB. All values in MB/s.
```

### DIFF
--- a/ackhandler/interfaces.go
+++ b/ackhandler/interfaces.go
@@ -15,6 +15,7 @@ type SentPacketHandler interface {
 	SetHandshakeComplete()
 
 	SendingAllowed() bool
+	TimeUntilSend() time.Time
 	GetStopWaitingFrame(force bool) *wire.StopWaitingFrame
 	GetLowestPacketNotConfirmedAcked() protocol.PacketNumber
 	ShouldSendRetransmittablePacket() bool
--- a/ackhandler/interfaces.go
+++ b/ackhandler/interfaces.go
@@ -14,8 +14,21 @@ type SentPacketHandler interface {
 	ReceivedAck(ackFrame *wire.AckFrame, withPacketNumber protocol.PacketNumber, encLevel protocol.EncryptionLevel, recvTime time.Time) error
 	SetHandshakeComplete()
 
+	// SendingAllowed says if a packet can be sent.
+	// Sending packets might not be possible because:
+	// * we're congestion limited
+	// * we're tracking the maximum number of sent packets
 	SendingAllowed() bool
+	// TimeUntilSend is the time when the next packet should be sent.
+	// It is used for pacing packets.
 	TimeUntilSend() time.Time
+	// ShouldSendNumPackets returns the number of packets that should be sent immediately.
+	// It always returns a number greater or equal than 1.
+	// A number greater than 1 is returned when the pacing delay is smaller than the minimum pacing delay.
+	// Note that the number of packets is only calculated based on the pacing algorithm.
+	// Before sending any packet, SendingAllowed() must be called to learn if we can actually send it.
+	ShouldSendNumPackets() int
+
 	GetStopWaitingFrame(force bool) *wire.StopWaitingFrame
 	GetLowestPacketNotConfirmedAcked() protocol.PacketNumber
 	ShouldSendRetransmittablePacket() bool
--- a/ackhandler/sent_packet_handler.go
+++ b/ackhandler/sent_packet_handler.go
@@ -3,6 +3,7 @@ package ackhandler
 import (
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/lucas-clemente/quic-go/congestion"
@@ -34,7 +35,7 @@ var ErrDuplicateOrOutOfOrderAck = errors.New("SentPacketHandler: Duplicate or ou
 
 type sentPacketHandler struct {
 	lastSentPacketNumber protocol.PacketNumber
-	lastPacketSentTime   time.Time
+	nextPacketSendTime   time.Time
 	skippedPackets       []protocol.PacketNumber
 
 	numNonRetransmittablePackets int // number of non-retransmittable packets since the last retransmittable packet
@@ -125,7 +126,6 @@ func (h *sentPacketHandler) SentPacket(packet *Packet) error {
 
 	now := time.Now()
 	h.lastSentPacketNumber = packet.PacketNumber
-	h.lastPacketSentTime = now
 
 	var largestAcked protocol.PacketNumber
 	if len(packet.Frames) > 0 {
@@ -154,6 +154,8 @@ func (h *sentPacketHandler) SentPacket(packet *Packet) error {
 		packet.Length,
 		isRetransmittable,
 	)
+
+	h.nextPacketSendTime = utils.MaxTime(h.nextPacketSendTime, now).Add(h.congestion.TimeUntilSend(h.bytesInFlight))
 
 	h.updateLossDetectionAlarm(now)
 	return nil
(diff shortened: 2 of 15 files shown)


marten-seemann:
@lucas-clemente: I merged the changes you approved in #1097, so this PR only contains the new commits.

ackhandler/sent_packet_handler.go
```diff
@@ -386,6 +390,18 @@ func (h *sentPacketHandler) SendingAllowed() bool {
...
+func (h *sentPacketHandler) TimeUntilSend() time.Time {
+	return h.nextPacketSendTime
+}
+
+func (h *sentPacketHandler) ShouldSendNumPackets() int {
+	delay := h.congestion.TimeUntilSend(h.bytesInFlight)
+	if delay == 0 || delay > protocol.MinPacingDelay {
+		return 1
+	}
+	return int(math.Ceil(float64(protocol.MinPacingDelay) / float64(delay)))
```
lucas-clemente: Nit: Why Ceil() here? If the float is 1.000000001, don't we want it to come out as 1?

session.go
```diff
@@ -355,15 +356,15 @@ func (s *session) run() error {
 
 runLoop:
 	for {
+		s.maybeResetTimer()
```
lucas-clemente: Please keep this below the error abort condition. We're potentially in an invalid state if we have an error, and we should do as little work as possible in that case.

session.go
```diff
@@ -741,6 +740,28 @@ func (s *session) processTransportParameters(params *handshake.TransportParamete
...
+			return err
+		}
+		// If no packet was sent, or we're congestion limit, we're done here.
+		if !sentPacket || !s.sentPacketHandler.SendingAllowed() {
+			return nil
+		}
+	}
+	// Only start the pacing timer if actually a packet was sent.
+	// If one packet was sent, there will probably be more to send when calling sendPacket again.
+	s.pacingDeadline = s.sentPacketHandler.TimeUntilSend()
```
lucas-clemente: This won't be triggered if we sent less than the max number of packets. Is that intentional? Maybe update the comment to be more clear in this case.

session.go
```diff
@@ -407,36 +408,31 @@ runLoop:
...
-		if s.config.KeepAlive && s.handshakeComplete && time.Since(s.lastNetworkActivityTime) >= s.peerParams.IdleTimeout/2 {
+		var pacingDeadline time.Time
+		if s.pacingDeadline.IsZero() { // the timer didn't have a pacing deadline set
+			pacingDeadline = s.sentPacketHandler.TimeUntilSend()
+		}
+		if s.config.KeepAlive && !s.keepAlivePingSent && s.handshakeComplete && time.Since(s.lastNetworkActivityTime) >= s.peerParams.IdleTimeout/2 {
 			// send the PING frame since there is no activity in the session
 			s.packer.QueueControlFrame(&wire.PingFrame{})
 			s.keepAlivePingSent = true
+		} else if !pacingDeadline.IsZero() && now.Before(pacingDeadline) {
```
lucas-clemente: You can get rid of the var in the run loop scope by merging it into the if condition and checking for the member there too. I think that will improve readability.

session.go
```diff
@@ -114,6 +114,7 @@ type session struct {
 
 	sessionCreationTime     time.Time
 	lastNetworkActivityTime time.Time
+	pacingDeadline          time.Time
```
lucas-clemente: Can you add a comment here?

internal/utils/timer.go
```diff
@@ -21,7 +21,7 @@ func (t *Timer) Chan() <-chan time.Time {
 
 // Reset the timer, no matter whether the value was read or not
 func (t *Timer) Reset(deadline time.Time) {
-	if deadline.Equal(t.deadline) {
+	if deadline.Equal(t.deadline) && !t.read {
```
lucas-clemente: This is a quite fundamental change that should definitely be in its own commit, with description and tests.

ackhandler/interfaces.go
```diff
@@ -14,7 +14,19 @@ type SentPacketHandler interface {
...
+	// * we're congestion limited
+	// * we're tracking the maximum number of sent packets
 	SendingAllowed() bool
+	// TimeUntilSend is the time when the next packet should be sent.
+	// It is used for pacing packets.
+	TimeUntilSend() time.Time
+	// ShouldSendNumPackets returns the number of packets that should be sent immediately.
+	// It always returns a number greater or equal than 1.
+	// A number greater than 1 is returned when the pacing delay is smaller than the minimum pacing delay.
+	ShouldSendNumPackets() int
```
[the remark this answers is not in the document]
marten-seemann: We could, but I don't think that makes a nicer API.

internal/utils/timer.go
marten-seemann: Done.

session.go
marten-seemann: Fixed.

session.go
marten-seemann: I don't think this works.

session.go
marten-seemann: Yes, this is intentional. If we don't send all packets that we were allowed to send, there's not more data available for sending, and no need to wake up for the next pacing deadline.
I updated the comment.

session.go
marten-seemann: Done.

ackhandler/sent_packet_handler.go
marten-seemann: Because otherwise we'd set deadlines shorter than `protocol.MinPacingDeadline`. To be precise, if we round that value, we might set deadlines as short as `protocol.MinPacingDeadline / 1.5`.

marten-seemann:
@lucas-clemente: PHAL.

PR #1114, opened 2018-01-10
expose the ConnectionState in the Session

### LINKED ISSUE #746: Expose ConnectionState() to quic Session and h2quic Request handler
I intend to implement a SNI handler on top of h2quic.Server. 
Originally(means tls/http2), I could use  `req.TLS.ServerName` to identify the client's SNI.

For now, req.TLS in h2quic always is `nil`.

To get it work, seems that we need expose a method `ConnectionState()` in quic.Session and implement it in `cryptoSetup.HandleCryptoStream()` ? 

Is it a reasonable thought?

marten-seemann:
Fixes #746.

This will be needed for the libp2p integration (to access the certificate chain). I created a PR to mint to expose the certificate chain: https://github.com/bifurcation/mint/pull/162.

The ConnectionState contains basic details about the QUIC connection. I noticed that [mint](https://godoc.org/github.com/bifurcation/mint#ConnectionState) as well as [tls-tris](https://godoc.org/github.com/cloudflare/tls-tris#ConnectionState) define their own `ConnectionState` struct, so maybe that's what we can go with in quic-go as well, at least as long as we're not using the TLS 1.3 implementation that will be added to the Go standard library at some point (maybe).
The `ConnectionState` now has a minimal feature set, and we can certainly add more fields there. I marked the API as unstable, so we don't need to assume any responsibility for breaking other people's builds ;).

### DIFF
--- a/internal/handshake/crypto_setup_server.go
+++ b/internal/handshake/crypto_setup_server.go
@@ -23,6 +23,8 @@ type KeyExchangeFunction func() crypto.KeyExchange
 
 // The CryptoSetupServer handles all things crypto for the Session
 type cryptoSetupServer struct {
+	mutex sync.RWMutex
+
 	connID               protocol.ConnectionID
 	remoteAddr           net.Addr
 	scfg                 *ServerConfig
@@ -51,7 +53,7 @@ type cryptoSetupServer struct {
 
 	params *TransportParameters
 
-	mutex sync.RWMutex
+	sni string // need to fill out the ConnectionState
 }
 
 var _ CryptoSetup = &cryptoSetupServer{}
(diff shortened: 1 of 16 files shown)


internal/handshake/crypto_setup_server.go
```diff
@@ -23,6 +23,8 @@ type KeyExchangeFunction func() crypto.KeyExchange
 
 // The CryptoSetupServer handles all things crypto for the Session
 type cryptoSetupServer struct {
+	mutex sync.RWMutex
```
lucas-clemente: This mutex is never used. Should it be, if users can now make calls into the crypto setup?

marten-seemann: I just moved that line up. The mutex is used a bunch of times, for example for opening and sealing packets.
We need to lock it for `ConnectionState` calls, in all crypto setups. I'll update the PR.

lucas-clemente: Sorry, didn't catch that. LGTM.

marten-seemann: Done.

PR #1120, opened 2018-01-15
implement a Dial function for the h2quic.RoundTripper

### LINKED ISSUE #747: implement DialContext in h2quic.RoundTripper

marten-seemann:
Fixes #747.

If `Dial` is set, it will be used for dialing new QUIC connections. If it is nil, `quic.DialAddr` will be used.

Contrary to what we discussed in #747, I think that returning a `quic.Session` is the right thing to do. Only then can the application use the full flexibilty that `quic.Dial` provides (using a custom `net.PacketConn` **and** taking care of the resolution of the remote `net.Addr`).
In fact, the standard library (kind of) does the same in the `http2.Transport` [here](https://godoc.org/golang.org/x/net/http2#Transport), where the returned `net.Conn` for `DialTLS` has to implement some of the methods a `tls.Conn`.

@lucas-clemente: I still need to fix the tests, but before doing that, I'd like to come to an agreement on the right signature.
I'd also like to hear what @costinm thinks about this, since this is the more idiomatic way of #1093.

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -37,6 +37,7 @@ type client struct {
 	hostname     string
 	handshakeErr error
 	dialOnce     sync.Once
+	dialer       func(network, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.Session, error)
 
 	session       quic.Session
 	headerStream  quic.Stream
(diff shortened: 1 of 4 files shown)


marten-seemann:
I added some tests.
@lucas-clemente: This should be ready for review now.

costinm:
LGTM, thanks for fixing this !

marten-seemann:
@lucas-clemente: Please review this PR.

PR #1123, opened 2018-01-22
ignore the delay in an ACK if it results in an RTT less than minRTT

### LINKED ISSUE #965: implement a sanity check for the RTT calculation using the ACK delay
There seems to be a vulnerability, where a peer could send us ACK delays designed to keep our RTT estimate close to 0. There's a fix for that on the way in https://github.com/quicwg/base-drafts/pull/975, which we should implement asap.

marten-seemann:
Fixes #965.

This is basically just the translation of https://chromium-review.googlesource.com/c/chromium/src/+/873930/1/net/quic/core/congestion_control/rtt_stats_test.cc into Go (while at the same time removing the flag protection).

### DIFF
--- a/congestion/rtt_stats.go
+++ b/congestion/rtt_stats.go
@@ -98,10 +98,10 @@ func (r *RTTStats) UpdateRTT(sendDelta, ackDelay time.Duration, now time.Time) {
 	r.updateRecentMinRTT(sendDelta, now)
 
 	// Correct for ackDelay if information received from the peer results in a
-	// positive RTT sample. Otherwise, we use the sendDelta as a reasonable
-	// measure for smoothedRTT.
+	// an RTT sample at least as large as minRTT. Otherwise, only use the
+	// sendDelta.
 	sample := sendDelta
-	if sample > ackDelay {
+	if sample-r.minRTT >= ackDelay {
 		sample -= ackDelay
 	}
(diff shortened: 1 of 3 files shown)


congestion/rtt_stats.go
```diff
@@ -98,10 +98,10 @@ func (r *RTTStats) UpdateRTT(sendDelta, ackDelay time.Duration, now time.Time) {
 	r.updateRecentMinRTT(sendDelta, now)
 
 	// Correct for ackDelay if information received from the peer results in a
-	// positive RTT sample. Otherwise, we use the sendDelta as a reasonable
-	// measure for smoothedRTT.
+	// an RTT sample at least as large as minRTT. Otherwise, only use the
+	// sendDelta.
 	sample := sendDelta
-	if sample > ackDelay {
+	if sample-r.minRTT >= ackDelay {
```
lucas-clemente: Is there a reason why we are ignoring the ackDelay if it would result in a value smaller than the min? Why not `max(sample - ackDelay, minRTT)`?

marten-seemann: That's a good question. It's specified like this in the draft: https://quicwg.github.io/base-drafts/draft-ietf-quic-recovery.html#rfc.section.3.4.5

PR #1128, opened 2018-01-30
update CIs to Go 1.9.3 and Go 1.10rc1

marten-seemann:


marten-seemann:
I ran the tests a couple of hundred times (in a separate branch), and I couldn't reproduce the failure.
I'm merging this now, and *if* there's any problem with Go 1.10, we can still fix it later.

PR #1137, opened 2018-02-01
make sure that at least every 20th ACK is retransmittable

### LINKED ISSUE #1135: make sure every nth packet containing an ACK is retransmittable
Found while debugging a failure of the benchmark test on AppVeyor, where the test fails with a *TooManyOutstandingReceivedPackets* error.

We currently have some logic that makes sure that at least one out of 20 packets is retransmittable. This logic fails to raise the lower boundary in the received packet history, if for example we send 19 packets only containing an ACK, and then one packet containing a MAX_DATA frame.

marten-seemann:
Fixes #1135. We should merge this before releasing 0.7 (#1139).

This is important because we need to make sure that we can remove old ACK ranges from the received packet history. The logic we implemented before was not correct, since we only made sure that every 20th packet would be retransmittable, but we didn't have any guarantee that this packet would also contain an ACK frame.


packet_packer.go
```diff
@@ -169,14 +169,18 @@ func (p *packetPacker) PackPacket() (*packedPacket, error) {
...
 		return nil, nil
 	}
-	// check if this packet only contains an ACK and / or STOP_WAITING
-	if !ackhandler.HasRetransmittableFrames(payloadFrames) {
-		if p.makeNextPacketRetransmittable {
-			payloadFrames = append(payloadFrames, &wire.PingFrame{})
-			p.makeNextPacketRetransmittable = false
+	if p.ackFrame != nil {
+		// check if this packet only contains an ACK (and maybe a STOP_WAITING)
+		if len(payloadFrames) == 1 || (p.stopWaiting != nil && len(payloadFrames) == 2) {
```
lucas-clemente: Why can't we use HasRetransmittableFrames here?

marten-seemann: We could, but it's ugly. There's no need to scan through the whole slice of `payloadFrames`. The second part of this if is going away very soon, when STOP_WAITING is (finally) removed by Chrome.

Issue #1168, opened 2018-02-06
don't retransmit the Initial packet when receiving a reply from the server


marten-seemann:
Annoyingly, the Initial packet is never acked in IETF QUIC. However, we're relying on the normal retransmission logic of the `SentPacketHandler` to retransmit it in case it's lost.

PR #1169, opened 2018-02-06
remove handshake packets from the history when the handshake completes

marten-seemann:
We already remove handshake packets from the retransmission queue, but we should also remove them from the packet history.

This mitigates the bad effects of #1168, since we're removing the Initial packet now once the handshake completes (before it would be retransmitted indefinitely).


marten-seemann:
This PR is wrong, since we would never retransmit the TLS Finished message.

PR #1170, opened 2018-02-07
don't retransmit Initial packets after receiving a packet from the server

### LINKED ISSUE #1168: don't retransmit the Initial packet when receiving a reply from the server

marten-seemann:
Fixes #1168.

This is not a trivial problem to solve. When we retransmit an Initial packet (and we do that quickly, to reduce handshake latency), multiple things can happen:

1.  the first one can actually turn out to be lost
2. the second one can get lost
3. both can arrive and get a reply:

     1. both replies are a Handshake
     2. one reply is a Retry, one is a Handshake

Note that the case where we receive one Retry packet is not a problem, since we're recreating the session (thereby canceling all retransmissions of the first Initial packet).

We need to handle all 3 cases correctly:

1. Is not a problem, normal retransmission logic takes of this.
2. Is more tricky. The server creates a new session when receiving the Initial packet. When the retransmission of second Initial packet arrives, it will just drop it, and we'll never get an ACK for that. That's why I had to introduce the `SentPacketHandler.ReceivedFirstPacket` that cancels *all* Initial retransmissions immediately.
3. i and ii are race conditions in the protocol, see https://github.com/quicwg/base-drafts/issues/1085.

Note that this problem also exists in gQUIC, but doesn't occur, since we're not doing stateless processing.

### DIFF
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -55,7 +55,8 @@ type sentPacketHandler struct {
 	congestion congestion.SendAlgorithm
 	rttStats   *congestion.RTTStats
 
-	handshakeComplete bool
+	receivedFirstPacket bool
+	handshakeComplete   bool
 	// The number of times the handshake packets have been retransmitted without receiving an ack.
 	handshakeCount uint32
 
@@ -92,6 +93,10 @@ func (h *sentPacketHandler) lowestUnacked() protocol.PacketNumber {
 		return f.Value.PacketNumber
 	}
 	return h.largestAcked + 1
+}
+
+func (h *sentPacketHandler) ReceivedFirstPacket() {
+	h.receivedFirstPacket = true
 }
 
 func (h *sentPacketHandler) SetHandshakeComplete() {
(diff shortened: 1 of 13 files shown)


internal/ackhandler/sent_packet_handler.go
```diff
@@ -348,15 +353,19 @@ func (h *sentPacketHandler) onPacketAcked(packetElement *PacketElement) {
 }
 
 func (h *sentPacketHandler) DequeuePacketForRetransmission() *Packet {
-	if len(h.retransmissionQueue) == 0 {
-		return nil
+	for len(h.retransmissionQueue) > 0 {
```
lucas-clemente: By the time the packet ends up in the retransmission queue, we've already applied penalties due to the loss. Could that be a performance issue?

marten-seemann: Yes, and I don't really know what to do about this. Probably this will have to wait for the rewrite of the loss recovery.

marten-seemann: So the right way to solve this is to keep track of which packet is a retransmission of which one (which will be part of the rewrite). As soon as *any one* of the Initials (the original one or the retransmission) gets acked, we would remove *all* of them from the history.
I'm not yet entirely sure what to do with the retransmission queue. It seems to be a general case, that we need to cancel a retransmission, even if it's queued.

Issue #410, opened 2017-02-07
Read timeout values from http.Server


lucas-clemente:
https://tip.golang.org/pkg/net/http/#Server

ohoushyar:
Hi there, just wanted to reach out and see if the ticket as the status said still needs help?

ohoushyar:
Hey there, I updated server side to read timeout from http.Server. I was also thinking to create a new ticket(s) to address server and client separately, but then I realized that client side implementation had been completed after #2091 (maybe I'm wrong though). If my assumption is right, then I can proceed and send a PR without new ticket. Any objection/comment?

marten-seemann:
@ohoushyar Sure, always happy about PRs :)

PR #1172, opened 2018-02-07
Make h2quic RoundTripper.RoundTrip(...) return if client timeout expires

lorenzosaino:
Currently, the implementation of `h2quic.RoundTripper.RoundTrip(req *http.Request)` ignores the context of `req`. As a result, if the `RoundTripper` is used as transport of an `http.Client` with a timeout value set, that is ignored.

For example, in the following snippet, client.Do(req) does not promptly return if the task takes more than client.Timeout to complete.

```
    client := http.Client{
        Timeout: 50 * time.Millisecond,
        Transport = &h2quic.RoundTripper{}
    }
    req, err := http.NewRequest("GET", "https://www.example.com", nil)
    response, err := client.Do(req)
```

This commit updates `h2quic.client.RoundTrip(req *http.Request)` to return an error if the request is cancelled.

This partially addresses #410

### DIFF
--- a/h2quic/client.go
+++ b/h2quic/client.go
@@ -202,6 +202,7 @@ func (c *client) RoundTrip(req *http.Request) (*http.Response, error) {
 		bodySent = true
 	}
 
+	ctx := req.Context()
 	for !(bodySent && receivedResponse) {
 		select {
 		case res = <-responseChan:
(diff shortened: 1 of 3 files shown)


lorenzosaino:
I signed it!

lorenzosaino:
Hello @marten-seemann, thanks a lot for reviewing my PR. I rebased my branch onto master and implemented the changes you requested. Please let me know if there is further changes you'd like me to make. Thanks!

PR #1187, opened 2018-02-21
use pointer to byte slices in the buffer pool

marten-seemann:
https://staticcheck.io/docs/staticcheck#SA6002 suggests to use pointers to objects in the `sync.Pool`. It seems like we can avoid copying the two ints of the slice (for length and capacity) when using a pointer, at the cost of an additional pointer dereference.

I created a [benchmark](https://gist.github.com/marten-seemann/07dd737cd7ed3fb190c8e6833a184d22) for this, but the results are inconclusive. I'm not seeing any reduction in allocations, and the performance seems pretty much the same.

### DIFF
--- a/buffer_pool.go
+++ b/buffer_pool.go
@@ -8,19 +8,20 @@ import (
 
 var bufferPool sync.Pool
 
-func getPacketBuffer() []byte {
-	return bufferPool.Get().([]byte)
+func getPacketBuffer() *[]byte {
+	return bufferPool.Get().(*[]byte)
 }
 
-func putPacketBuffer(buf []byte) {
-	if cap(buf) != int(protocol.MaxReceivePacketSize) {
+func putPacketBuffer(buf *[]byte) {
+	if cap(*buf) != int(protocol.MaxReceivePacketSize) {
 		panic("putPacketBuffer called with packet of wrong size!")
 	}
-	bufferPool.Put(buf[:0])
+	bufferPool.Put(buf)
 }
 
 func init() {
 	bufferPool.New = func() interface{} {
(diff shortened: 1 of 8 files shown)


buffer_pool.go
```diff
@@ -9,18 +9,20 @@ import (
 var bufferPool sync.Pool
 
 func getPacketBuffer() []byte {
```
lucas-clemente: Can you run a benchmark with this changed to return *[]byte? That way you wouldn't need to allocate another copy of the empty slice before reslicing it.

marten-seemann:
@lucas-clemente I updated the PR, such that the pool returns `*[]byte` instead of `[]byte`. I also updated the benchmark.
```
BenchmarkBufferPoolPointer-8   	    5000	    586164 ns/op	  110814 B/op	     211 allocs/op
BenchmarkBufferPoolDirect-8    	    3000	    922887 ns/op	  434221 B/op	   10218 allocs/op
```
Now this is a big difference. 

We're now not reslicing when putting the slice back, so we have to take extra care when getting the slice, i.e. every time we get a slice, we need to reslice it. There are two ways in which we use the buffer in quic-go:
1. As an input to a `net.Conn.Read`. We're already reslicing there.
2. As the underlying buffer for a `bytes.Buffer`. We now need to make sure to reslice to the length of 9 there.

Or is it all not worth it? Are we over optimizing here based on a micro benchmark?

lucas-clemente:
Can you run a "real" benchmark with this change? To see whether it affects transfer rates at all?

marten-seemann:
I don't know how. The benchmark test has a really high variance (I can't figure out why, every time I add logging it disappears).

PR #1191, opened 2018-02-22
add some debug logging for the proxy

marten-seemann:


integrationtests/tools/proxy/proxy.go
```diff
@@ -221,18 +225,21 @@ func (p *QuicProxy) runConnection(conn *connection) error {
 		packetCount := atomic.AddUint64(&conn.outgoingPacketCounter, 1)
 
 		if p.dropPacket(DirectionOutgoing, packetCount) {
+			utils.Debugf("dropping outgoing packet %d (%d bytes)", packetCount, n)
 			continue
 		}
+		utils.Debugf("forwarding outgoing packet %d (%d bytes) to %s", packetCount, n, conn.ClientAddr)
```
lucas-clemente: This might affect performance and therefore the tests, since you're allocating a lot of slices here. Maybe move the common cases behind an if?

PR #1193, opened 2018-02-23
implement the recent key schedule changes

### LINKED ISSUE #1129: implement key schedule changes
Some labels were changed in https://github.com/quicwg/base-drafts/pull/1004/.
This PR might be useful: https://github.com/quicwg/base-drafts/pull/1077.

marten-seemann:
Fixes #1129.


internal/crypto/key_derivation.go
```diff
@@ -16,6 +19,14 @@ type TLSExporter interface {
 	ComputeExporter(label string, context []byte, keyLength int) ([]byte, error)
 }
 
+func qhkdfExpand(secret []byte, label string, length int) []byte {
+	qlabel := make([]byte, 2+1+5+len(label)+1)
```
lucas-clemente: Where does the final +1 come from?

marten-seemann: The last 0x00 here: https://quicwg.github.io/base-drafts/draft-ietf-quic-tls.html#one-rtt-secrets. Since Go initializes the slice to 0, we don't need to set it later.

lucas-clemente: Can you leave a comment that this is intentional?

PR #1194, opened 2018-02-23
log the TLS Transport Parameters

marten-seemann:


internal/handshake/transport_parameters.go
```diff
@@ -174,3 +174,9 @@ func (p *TransportParameters) getTransportParameters() []transportParameter {
 	}
 	return params
 }
+
+// String returns a string representation, intended for logging.
+// It should only used for IETF QUIC.
+func (p *TransportParameters) String() string {
+	return fmt.Sprintf("&handshake.TransportParameters{StreamFlowControlWindow: %#x, ConnectionFlowControlWindow: %#x, MaxBidiStreamID: %d, MaxUniStreamID: %d, OmitConnectionID: %t, IdleTimeout: %s}", p.StreamFlowControlWindow, p.ConnectionFlowControlWindow, p.MaxBidiStreamID, p.MaxUniStreamID, p.OmitConnectionID, p.IdleTimeout)
```
lucas-clemente: Why not %#v?

marten-seemann: Because it displays durations as floats.

lucas-clemente: Is that worth the risk of not remembering to add new fields to the log statement?

marten-seemann: I'd say so. I don't expect (m)any additions to the transport parameters anyway.

marten-seemann:
@lucas-clemente: LGTY?

PR #1209, opened 2018-02-26
reject unknown versions in the quic.Config

### LINKED ISSUE #1017: Error when an unsupported version is seen in the config

marten-seemann:
Fixes #1017.

### DIFF
--- a/integrationtests/self/handshake_test.go
+++ b/integrationtests/self/handshake_test.go
@@ -51,10 +51,18 @@ var _ = Describe("Handshake tests", func() {
 	}
 
 	Context("Version Negotiation", func() {
+		var supportedVersions []protocol.VersionNumber
+
+		BeforeEach(func() {
+			supportedVersions = protocol.SupportedVersions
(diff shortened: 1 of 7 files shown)


integrationtests/self/handshake_test.go
```diff
@@ -51,10 +51,18 @@ var _ = Describe("Handshake tests", func() {
...
+
+		BeforeEach(func() {
+			supportedVersions = protocol.SupportedVersions
+			protocol.SupportedVersions = append(protocol.SupportedVersions, []protocol.VersionNumber{7, 8, 9, 10}...)
+		})
+
+		AfterEach(func() {
+			protocol.SupportedVersions = supportedVersions
+		})
+
```
lucas-clemente: Can you run this with the race detector to make sure there aren't any leftovers after the tests that might still access those?

marten-seemann: I ran the self-tests more than 10 times with race detector in randomized order, and everything seems ok.

PR #1212, opened 2018-02-26
refactor retransmissions

marten-seemann:
We should bundle retransmissions with new data, but retransmit a packet as a whole. This is the first step of fixing our loss recovery. In the future, we will keep track of which packet is a retransmission of which packet. If we then receive an acknowledgement for a packet that we already retransmitted, we know that there's no need to retransmit the retransmission, even if it gets lost.

Note that there's a special case here: If the header grows larger for the retransmission (e.g. because we need to use a longer packet number encoding), a retransmission might not fit into a single packet. In gQUIC, this will even happen more often, since we need to add a STOP_WAITING to every retransmitted packet. In these case, we need to split up the packet into two.

### DIFF
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -157,10 +157,22 @@ func (p *packetPacker) PackRetransmission(packet *ackhandler.Packet) ([]*packedP
 			frames = append(frames, frame)
 			controlFrames = controlFrames[1:]
 		}
+
+		// temporarily increase the maxFrameSize by the (minimum) length of the DataLen field
+		// this leads to a properly sized packet in all cases, since we do all the packet length calculations with StreamFrames that have the DataLen set
+		// however, for the last STREAM frame in the packet, we can omit the DataLen, thus yielding a packet of exactly the correct size
+		// for gQUIC STREAM frames, DataLen is always 2 bytes
+		// for IETF draft style STREAM frames, the length is encoded to either 1 or 2 bytes
+		if p.version.UsesIETFFrameFormat() {
+			maxSize++
+		} else {
+			maxSize += 2
+		}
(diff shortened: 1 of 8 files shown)


packet_packer.go
```diff
@@ -99,11 +99,112 @@ func (p *packetPacker) PackAckPacket() (*packedPacket, error) {
...
+
+		header := p.getHeader(encLevel)
+		headerLength, err := header.GetLength(p.perspective, p.version)
+		if err != nil {
+			return nil, err
+		}
+		maxSize := protocol.MaxPacketSize - protocol.ByteCount(sealer.Overhead()) - headerLength
+
+		// for gQUIC: add a STOP_WAITING for *every* retransmission
+		if !p.version.UsesIETFFrameFormat() {
```
lucas-clemente: Can you introduce a UsesStopWaitingFrames here? That makes it easier to support a future gQUIC version where they are dropped.

session.go
```diff
@@ -838,39 +838,30 @@ func (s *session) sendPacket() (bool, error) {
...
-		utils.Debugf("\tDequeueing retransmission for packet 0x%x", retransmitPacket.PacketNumber)
-		// resend the frames that were in the packet
-		for _, frame := range retransmitPacket.GetFramesForRetransmission() {
-			// TODO: only retransmit WINDOW_UPDATEs if they actually enlarge the window
-			switch f := frame.(type) {
-			case *wire.StreamFrame:
-				s.streamFramer.AddFrameForRetransmission(f)
-			default:
-				s.packer.QueueControlFrame(frame)
+		if !s.version.UsesIETFFrameFormat() {
```
lucas-clemente: Same here

PR #1213, opened 2018-03-02
implement ACK decimation

### LINKED ISSUE #389: implement ACK decimation
ACK decimation is a way to reduce the number of ACKs when there is minor reordering on the wire.
The details can be found in Chromium's `QuicConnection::MaybeQueueAck`.

marten-seemann:
Fixes #389.

This should lead to a significant increase in throughput.

### DIFF
--- a/internal/ackhandler/received_packet_handler.go
+++ b/internal/ackhandler/received_packet_handler.go
@@ -25,11 +25,18 @@ type receivedPacketHandler struct {
 	version protocol.VersionNumber
 }
 
+const (
+	// ackSendDelay is the maximum delay that can be applied to an ACK for a retransmittable packet
+	ackSendDelay = 25 * time.Millisecond
+	// retransmittablePacketsBeforeAck is the number of retransmittable that an ACK is sent for
+	retransmittablePacketsBeforeAck = 10
+)
+
 // NewReceivedPacketHandler creates a new receivedPacketHandler
 func NewReceivedPacketHandler(version protocol.VersionNumber) ReceivedPacketHandler {
 	return &receivedPacketHandler{
 		packetHistory: newReceivedPacketHistory(),
-		ackSendDelay:  protocol.AckSendDelay,
+		ackSendDelay:  ackSendDelay,
 		version:       version,
 	}
 }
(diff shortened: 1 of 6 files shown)


internal/ackhandler/received_packet_handler.go
```diff
@@ -65,35 +86,63 @@ func (h *receivedPacketHandler) IgnoreBelow(p protocol.PacketNumber) {
...
+	highestRange := h.packetHistory.GetHighestAckRange()
+	return highestRange.First >= h.lastAck.LargestAcked && highestRange.Len() <= maxPacketsAfterNewMissing
+}
 
-	// if the packet number is smaller than the largest acked packet, it must have been reported missing with the last ACK
-	// note that it cannot be a duplicate because they're already filtered out by ReceivedPacket()
-	if h.lastAck != nil && packetNumber < h.lastAck.LargestAcked {
+// maybeQueueAck queues an ACK, if necessary.
+// It is implemented analogously to Chrome's QuicConnection::MaybeQueueAck()
+// in ACK_DECIMATION_WITH_REORDERING mode.
```
lucas-clemente: I think it would still be valuable to have a description of the algorithm somewhere in our comments. Chrome's implementation might diverge or change completely, and then it'll be very hard to figure our what you meant.

internal/ackhandler/received_packet_handler.go
```diff
@@ -65,35 +86,63 @@ func (h *receivedPacketHandler) IgnoreBelow(p protocol.PacketNumber) {
...
+func (h *receivedPacketHandler) maybeQueueAck(packetNumber protocol.PacketNumber, rcvTime time.Time, shouldInstigateAck, wasMissing bool) {
```
lucas-clemente: This function doesn't actually "queue" anything. Maybe rename it and h.ackQueued? setAckRequired() and h.ackRequired?

internal/ackhandler/received_packet_handler.go
```diff
@@ -65,35 +86,63 @@ func (h *receivedPacketHandler) IgnoreBelow(p protocol.PacketNumber) {
 	h.packetHistory.DeleteBelow(p)
 }
 
-func (h *receivedPacketHandler) maybeQueueAck(packetNumber protocol.PacketNumber, rcvTime time.Time, shouldInstigateAck bool) {
-	h.packetsReceivedSinceLastAck++
-
-	if shouldInstigateAck {
-		h.retransmittablePacketsReceivedSinceLastAck++
+func (h *receivedPacketHandler) isMissing(p protocol.PacketNumber) bool {
```
lucas-clemente: Add a comment that missing = gap.

internal/ackhandler/received_packet_handler.go
marten-seemann: setAckRequired() doesn't sound correct, that would have to be maybeSetAckRequired(). I'd prefer to keep it as it is though, since we're following Chrome so closely here.

internal/ackhandler/received_packet_handler.go
marten-seemann: I added some more comments.

PR #1218, opened 2018-03-05
refactor sending of retransmission in the session

marten-seemann:
This won't be the last refactoring, but it's going in the right direction.

### DIFF
--- a/session.go
+++ b/session.go
@@ -768,12 +768,30 @@ func (s *session) processTransportParameters(params *handshake.TransportParamete
 }
 
 func (s *session) sendPackets() error {
+	s.packer.SetLeastUnacked(s.sentPacketHandler.GetLeastUnacked())
 	s.pacingDeadline = time.Time{}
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -768,12 +768,30 @@ func (s *session) processTransportParameters(params *handshake.TransportParamete
...
+	// * the maximum number of packets was reached
+	for ; numPacketsSent < numPackets; numPacketsSent++ {
+		sentPacket, err := s.maybeSendRetransmission()
+		if err != nil {
+			return err
+		}
+		if !sentPacket { // no more retransmission to send. Proceed to send new data.
+			break
+		}
+		if !s.sentPacketHandler.SendingAllowed() {
```
lucas-clemente: Is this correct? I vaguely remember that we wanted to allow retransmissions even when congestion limited?

marten-seemann: That's handled by https://github.com/lucas-clemente/quic-go/blob/master/internal/ackhandler/sent_packet_handler.go#L384-L388.
I'm currently refactoring the logic of `SendingAllowed`, to make it less hacky, and I'll send a PR later this week.

PR #1220, opened 2018-03-05
implement the IETF draft 10 header changes

### LINKED ISSUE #1217: implement recent changes to short header packet types
https://github.com/quicwg/base-drafts/pull/1120

marten-seemann:
Fixes #1217.

### DIFF
--- a/internal/wire/ietf_header_test.go
+++ b/internal/wire/ietf_header_test.go
@@ -115,7 +115,7 @@ var _ = Describe("IETF draft Header", func() {
 		Context("short headers", func() {
 			It("reads a short header with a connection ID", func() {
 				data := []byte{
-					0x40 ^ 0x1,                                     //
+					0x10,                                           // 1 byte packet number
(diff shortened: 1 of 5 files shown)


internal/wire/ietf_header_test.go
```diff
@@ -133,7 +133,7 @@ var _ = Describe("IETF draft Header", func() {
 
 			It("reads the Key Phase Bit", func() {
 				data := []byte{
-					0x20 ^ 0x1,
+					0x10 ^ 0x40 ^ 0x20,
```
lucas-clemente: It's a bit surprising to see an XOR here. Why not a normal OR?

marten-seemann: We're using XOR everywhere in these tests, and even in the header writing code. I can change that to OR everywhere if you prefer.

lucas-clemente: It obviously doesn't make a difference, but XOR is unusual for this use case, which is what tripped me up.

PR #1229, opened 2018-03-07
don't loop over all sent packets to update the RTT

marten-seemann:


internal/ackhandler/sent_packet_handler.go
```diff
@@ -186,6 +180,10 @@ func (h *sentPacketHandler) ReceivedAck(ackFrame *wire.AckFrame, withPacketNumbe
 			if p.Value.largestAcked != 0 {
 				h.lowestPacketNotConfirmedAcked = utils.MaxPacketNumber(h.lowestPacketNotConfirmedAcked, p.Value.largestAcked+1)
 			}
+			if p.Value.PacketNumber == ackFrame.LargestAcked {
+				h.rttStats.UpdateRTT(rcvTime.Sub(p.Value.sendTime), ackFrame.DelayTime, rcvTime)
+				h.congestion.MaybeExitSlowStart()
```
lucas-clemente: This causes a behavior change in the congestion. Before, if an ACK would cause us to exit slow start, all newly ACKed packets already contribute in a non-slowstart way to the congestion. With this change they won't.

I don't understand the details enough to judge whether that's relevant. What's the reasoning behind this commit? Does it fix a hotspot in the profiler?

marten-seemann: No, this is just an attempt to simplify the code. Apparently, I must have missed that point. :(

Issue #497, opened 2017-03-27
Implement tail loss probing (TLP) in loss recovery


lucas-clemente:
See the wg draft.

PR #1232, opened 2018-03-09
implement packet send modes to determine what kind of packets are sent

marten-seemann:
The `SentPacketHandler` uses different "modes" to communicate to the session what kind of packets can be sent:

* `None`: No packets can be sent. This only happens if we reach the limit of tracked packets.
* `Ack`: ACK-only packets can be sent. This happens if we're congestion limited, or if we're keeping track of too many outstanding packets (note that this is a lower limit than the number of tracked packets, since as soon as we keep track of retransmissions, those will not count towards the outstanding, but the tracked packets).
* `Retransmission`: A retransmission should be sent. This is returned every time the retransmission queue is not empty (and we're not limited by the number of tracked packets).
* `Any`: Any kind of packet can be sent.

This design will make it easy to implement TLPs (#497) by introding a TLP mode (and a method to dequeue the first outstanding packet for retransmission). According to the loss recovery draft, a sender should send new data as a probe packet, and only send a retransmission if no new data is available.

### DIFF
--- a/internal/ackhandler/interfaces.go
+++ b/internal/ackhandler/interfaces.go
@@ -14,11 +14,8 @@ type SentPacketHandler interface {
 	ReceivedAck(ackFrame *wire.AckFrame, withPacketNumber protocol.PacketNumber, encLevel protocol.EncryptionLevel, recvTime time.Time) error
 	SetHandshakeComplete()
 
-	// SendingAllowed says if a packet can be sent.
-	// Sending packets might not be possible because:
-	// * we're congestion limited
-	// * we're tracking the maximum number of sent packets
-	SendingAllowed() bool
+	// The SendMode determines if and what kind of packets can be sent.
+	SendMode() SendMode
 	// TimeUntilSend is the time when the next packet should be sent.
 	// It is used for pacing packets.
 	TimeUntilSend() time.Time
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -357,18 +357,31 @@ func (h *sentPacketHandler) GetStopWaitingFrame(force bool) *wire.StopWaitingFra
 	return h.stopWaitingManager.GetStopWaitingFrame(force)
 }
 
-func (h *sentPacketHandler) SendingAllowed() bool {
-	cwnd := h.congestion.GetCongestionWindow()
-	congestionLimited := h.bytesInFlight > cwnd
-	maxTrackedLimited := protocol.PacketNumber(len(h.retransmissionQueue)+h.packetHistory.Len()) >= protocol.MaxTrackedSentPackets
-	if congestionLimited {
-		utils.Debugf("Congestion limited: bytes in flight %d, window %d", h.bytesInFlight, cwnd)
+func (h *sentPacketHandler) SendMode() SendMode {
+	numTrackedPackets := len(h.retransmissionQueue) + h.packetHistory.Len()
+
+	// Don't send any packets if we're keeping track of the maximum number of packets.
+	// Note that since MaxOutstandingSentPackets is smaller than MaxTrackedSentPackets,
+	// we will stop sending out new data when reaching MaxOutstandingSentPackets,
+	// but still allow sending of retransmissions and ACKs.
+	if numTrackedPackets >= protocol.MaxTrackedSentPackets {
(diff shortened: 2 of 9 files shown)


marten-seemann:
@lucas-clemente wdyt?

internal/ackhandler/sent_packet_handler.go
```diff
@@ -357,18 +357,31 @@ func (h *sentPacketHandler) GetStopWaitingFrame(force bool) *wire.StopWaitingFra
...
+	// but still allow sending of retransmissions and ACKs.
+	if numTrackedPackets >= protocol.MaxTrackedSentPackets {
+		utils.Debugf("Limited by the number of tracked packets: tracking %d packets, maximum %d", numTrackedPackets, protocol.MaxTrackedSentPackets)
+		return SendNone
+	}
+	// Send retransmissions first, if there are any.
+	if len(h.retransmissionQueue) > 0 {
+		return SendRetransmission
+	}
+	// Only send ACKs, when  we're congestion limited or
```
lucas-clemente: nit: grammar

session.go
```diff
@@ -766,40 +766,54 @@ func (s *session) processTransportParameters(params *handshake.TransportParamete
...
+			sentPacket, err := s.sendPacket()
+			if err != nil {
+				return err
+			}
+			if !sentPacket {
+				break sendLoop
+			}
+			numPacketsSent++
+		default:
+			return errors.New("BUG: invalid send mode")
```
lucas-clemente: Please include the value in the error message.

session.go
```diff
@@ -766,40 +766,54 @@ func (s *session) processTransportParameters(params *handshake.TransportParamete
...
+			return s.maybeSendAckOnlyPacket()
+		case ackhandler.SendRetransmission:
+			sentPacket, err := s.maybeSendRetransmission()
+			if err != nil {
+				return err
+			}
+			if !sentPacket {
+				// This can happen if a retransmission queued, but it wasn't necessary to send it.
+				// e.g. when an Initial is queued, but we already received a packet from the server.
+				break // breaks the switch statement, not the for sendLoop
```
lucas-clemente: This would be cleaner as `if sentPacket { numPacketsSent++ }`. Then you can drop the break.

session.go
```diff
@@ -766,40 +766,54 @@ func (s *session) processTransportParameters(params *handshake.TransportParamete
...
-		sentPacket, err := s.maybeSendRetransmission()
-		if err != nil {
-			return err
+sendLoop:
+	for {
+		switch sendMode {
+		case ackhandler.SendNone:
+			break sendLoop
+		case ackhandler.SendAck:
+			return s.maybeSendAckOnlyPacket()
```
lucas-clemente: Why do you return here and not break? Can you add a comment?

internal/ackhandler/interfaces.go
```diff
@@ -14,11 +14,8 @@ type SentPacketHandler interface {
 	ReceivedAck(ackFrame *wire.AckFrame, withPacketNumber protocol.PacketNumber, encLevel protocol.EncryptionLevel, recvTime time.Time) error
 	SetHandshakeComplete()
 
-	// SendingAllowed says if a packet can be sent.
-	// Sending packets might not be possible because:
-	// * we're congestion limited
-	// * we're tracking the maximum number of sent packets
-	SendingAllowed() bool
+	// The SendMode determines if and what kind of packets can be sent.
+	SendMode() SendMode
```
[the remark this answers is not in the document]
marten-seemann: Since we're calling this function every time before sending a packet, I don't see a need for overlapping modes.

marten-seemann:
@lucas-clemente: Addressed all your points. PHAL.

session.go
```diff
@@ -766,40 +766,55 @@ func (s *session) processTransportParameters(params *handshake.TransportParamete
...
-			return err
+sendLoop:
+	for {
+		switch sendMode {
+		case ackhandler.SendNone:
+			break sendLoop
+		case ackhandler.SendAck:
+			// We can at most send a single ACK only packet.
+			// There will only be a new ACK after receiving new packets.
+			return s.maybeSendAckOnlyPacket()
```
lucas-clemente: So just to be clear, return here means that ACK-only packets don't contribute to pacing.

PR #1237, opened 2018-03-10
use the max_packet_size transport parameter

### LINKED ISSUE #1233: use the max_packet_size transport parameter
We're currently sending the max_packet_size transport parameter, but we're not reading or acting on it.

marten-seemann:
Fixes #1233.

### DIFF
--- a/internal/handshake/transport_parameters.go
+++ b/internal/handshake/transport_parameters.go
@@ -20,6 +20,8 @@ type TransportParameters struct {
 	StreamFlowControlWindow     protocol.ByteCount
 	ConnectionFlowControlWindow protocol.ByteCount
 
+	MaxPacketSize protocol.ByteCount
+
 	MaxBidiStreamID protocol.StreamID // only used for IETF QUIC
 	MaxUniStreamID  protocol.StreamID // only used for IETF QUIC
 	MaxStreams      uint32            // only used for gQUIC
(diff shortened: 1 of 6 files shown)


internal/handshake/transport_parameters.go
```diff
@@ -137,6 +139,15 @@ func readTransportParamters(paramsList []transportParameter) (*TransportParamete
 				return nil, fmt.Errorf("wrong length for omit_connection_id: %d (expected empty)", len(p.Value))
 			}
 			params.OmitConnectionID = true
+		case maxPacketSizeParameterID:
+			if len(p.Value) != 2 {
+				return nil, fmt.Errorf("wrong length for max_packet_size: %d (expected 2)", len(p.Value))
+			}
+			maxPacketSize := protocol.ByteCount(binary.BigEndian.Uint16(p.Value))
+			if maxPacketSize < 1200 {
+				return nil, fmt.Errorf("invalid value for max_packet_size: %d (minimum 1200)", maxPacketSize)
```
lucas-clemente: Should this be a qerr?

marten-seemann: It doesn't matter. This will end up as a TLS Internal Error alert anyway.

marten-seemann:
We allocate buffers with the `MaxReceivePacketSize`.
The maximum size we use for packing packets is `MaxPacketSizeIPv4`, which is smaller than the `MaxReceivePacketSize`. The parameter sent in the TLS transport parameters can only lead to a reduction in the packet size for packets that we send, see https://github.com/lucas-clemente/quic-go/pull/1237/files#diff-4edc6b25b26b03f5ea2847eb0d3487e8R524.

lucas-clemente:
ic, thx for clarifying.

PR #1246, opened 2018-03-14
keep track of retransmissions

marten-seemann:
Replaces #1230.

We now keep track which packet was sent as a retransmission of which other packet. When an ACK arrives for the original packet, we can delete the retransmission(s) from the packet history.

I ran some benchmarks with reordered packets, and I don't get a statistically significant improvement in transfer speed here. At this point, I'm not worried about this, since we still have a bunch of loss recovery optimizations outstanding, especially spurious RTO detection (#687) and TLPs (#497).

### DIFF
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -150,6 +150,12 @@ func (h *sentPacketHandler) SentPacket(packet *Packet) {
 	h.updateLossDetectionAlarm(now)
 }
 
+func (h *sentPacketHandler) SentPacketsAsRetransmission(packets []*Packet, retransmissionOf protocol.PacketNumber) {
+	for _, packet := range packets {
+		h.SentPacket(packet)
+	}
+}
+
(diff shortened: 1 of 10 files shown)


internal/ackhandler/sent_packet_handler.go
```diff
@@ -294,40 +309,98 @@ func (h *sentPacketHandler) detectLostPackets(now time.Time) {
...
-		h.detectLostPackets(now)
+		err = h.detectLostPackets(now)
 	} else {
 		// RTO
-		h.retransmitOldestTwoPackets()
+		err = h.queueRTOs()
 		h.rtoCount++
 	}
 
 	h.updateLossDetectionAlarm(now)
```
lucas-clemente: It's weird that we continue to do work after an error was returned – but I guess it doesn't really matter in this case.

marten-seemann: I agree, and it's easy to fix. Just one additional `if`.

PR #1255, opened 2018-03-17
calculate loss alarm based on send time of last retransmittable packet

marten-seemann:


internal/ackhandler/sent_packet_handler.go
```diff
@@ -273,13 +276,13 @@ func (h *sentPacketHandler) updateLossDetectionAlarm(now time.Time) {
 
 	// TODO(#497): TLP
 	if !h.handshakeComplete {
-		h.alarm = now.Add(h.computeHandshakeTimeout())
+		h.alarm = h.lastSentRetransmittablePacketTime.Add(h.computeHandshakeTimeout())
```
lucas-clemente: Is lastSentRetransmittablePacketTime guaranteed to be non-zero by the time we get here? What if we get an ACK in the first packet from a client?

marten-seemann: That's not guaranteed, and the same problem exists in the loss recovery spec.
Maybe the best solution is to initialize `lastSentRetransmittablePacket` to `time.Now()`?

lucas-clemente: Yes, that should work.

marten-seemann: Actually, the client can't send an ACK in the first packet, since the ACK format doesn't allow to send ACKs that don't ack any packets.

PR #1260, opened 2018-03-20
don't reduce the bytes in flight for RTO probe packets

marten-seemann:
We should **only** reduce the `bytes_in_flight` when declaring a packet lost (or of course when receiving an ACK).

Packets that we retransmit as RTO probe packets still count towards the `bytes_in_flight` until:
* we receive an ACK for the original packet (i.e. we detect a spurious RTO, see #687)
* we receive an ACK for the probe packet (i.e. we detect a non-spurious RTO), and we declare all packets below the probe packet lost

### DIFF
--- a/internal/ackhandler/packet.go
+++ b/internal/ackhandler/packet.go
@@ -20,6 +20,7 @@ type Packet struct {
 	sendTime     time.Time
 
 	queuedForRetransmission bool
+	includedInBytesInFlight bool
 	retransmittedAs         []protocol.PacketNumber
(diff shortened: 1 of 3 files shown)


marten-seemann:
That doesn't really work. When dequeuing a packet for retransmission, in some cases we need to reduce the bytes in flight, and in other case we don't.

lucas-clemente:
Yeah, you wouldn't remove it from the history until the bif is actually reduced. Until then, it's just marked as retransmitted.

Anyway, we can think about this some other time.

PR #1262, opened 2018-03-24
Add reverse proxy example

fholzer:
Providing a demo reverse proxy makes it much easier for non-devs to start playing with this project.


fholzer:
I signed it

fholzer:
Well, again, it's not meant to show how to use your library. It's a very simple demo for non-devs that allows playing with quic and a webserver of choice, without the need for a fully fledged webserver like caddy.

marten-seemann:
I talked about this with @lucas-clemente today. We agreed that this doesn't really belong into the examples, but a we could add an example Caddy config (which would only be a few lines) either to our wiki or to the README.

PR #1264, opened 2018-03-27
don't do trial decryption for IETF QUIC

marten-seemann:
One of the cool features of IETF QUIC is that we don't need to do trial decryption, because the encryption level can be learned from the packet header.
This PR refactors a bunch of stuff, such that the TLS crypto setup doesn't need to do trial decryption any more.


mint_utils.go
```diff
@@ -108,16 +108,23 @@ func tlsToMintConfig(tlsConf *tls.Config, pers protocol.Perspective) (*mint.Conf
 // unpackInitialOrRetryPacket unpacks packets Initial and Retry packets
 // These packets must contain a STREAM_FRAME for the crypto stream, starting at offset 0.
 func unpackInitialPacket(aead crypto.AEAD, hdr *wire.Header, data []byte, version protocol.VersionNumber) (*wire.StreamFrame, error) {
-	unpacker := &packetUnpacker{aead: &nullAEAD{aead}, version: version}
-	packet, err := unpacker.Unpack(hdr.Raw, hdr, data)
+	buf := *getPacketBuffer()
+	buf = buf[:0]
+	defer putPacketBuffer(&buf)
+
+	decrypted, err := aead.Open(buf, data, hdr.PacketNumber, hdr.Raw)
```
lucas-clemente: Is there a reason why you aren't using the (more efficient) in-place decryption here?

packet_unpacker.go
```diff
@@ -50,6 +44,84 @@ func (u *packetUnpacker) Unpack(headerBinary []byte, hdr *wire.Header, data []by
...
+		packetUnpackerBase: packetUnpackerBase{version: version},
+		aead:               aead,
+	}
+}
+
+func (u *packetUnpackerGQUIC) Unpack(headerBinary []byte, hdr *wire.Header, data []byte) (*unpackedPacket, error) {
+	buf := *getPacketBuffer()
+	buf = buf[:0]
+	defer putPacketBuffer(&buf)
+	decrypted, encryptionLevel, err := u.aead.Open(buf, data, hdr.PacketNumber, headerBinary)
```
lucas-clemente: Same here?

mint_utils.go
marten-seemann: I don't think there's a reason, but this is not something that this PR introduced. I opened a separate issue for that: #1277. I'd suggest to do the fix separately.

lucas-clemente: SGTM, thx

PR #1266, opened 2018-03-28
use the FNV 128a from the standard library

marten-seemann:
It doesn't really make our code easier, but at least we can get rid of a dependency.

### DIFF
--- a/internal/crypto/null_aead_fnv128a.go
+++ b/internal/crypto/null_aead_fnv128a.go
@@ -1,10 +1,11 @@
 package crypto
 
 import (
-	"encoding/binary"
+	"bytes"
 	"errors"
+	"fmt"
+	"hash/fnv"
 
-	"github.com/lucas-clemente/fnv128a"
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 )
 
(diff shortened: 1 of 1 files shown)


internal/crypto/null_aead_fnv128a.go
```diff
@@ -21,21 +22,21 @@ func (n *nullAEADFNV128a) Open(dst, src []byte, packetNumber protocol.PacketNumb
...
 	if n.perspective == protocol.PerspectiveServer {
 		hash.Write([]byte("Client"))
 	} else {
 		hash.Write([]byte("Server"))
 	}
-	testHigh, testLow := hash.Sum128()
+	sum := make([]byte, 0, 16)
+	hash.Sum(sum)
+	sum = sum[:16]
+	reverse(sum)
```
lucas-clemente: Can you explain / document why this reverse is necessary?

internal/crypto/null_aead_fnv128a.go
```diff
@@ -21,21 +22,21 @@ func (n *nullAEADFNV128a) Open(dst, src []byte, packetNumber protocol.PacketNumb
...
 	hash.Write(src[12:])
 	if n.perspective == protocol.PerspectiveServer {
 		hash.Write([]byte("Client"))
 	} else {
 		hash.Write([]byte("Server"))
 	}
-	testHigh, testLow := hash.Sum128()
+	sum := make([]byte, 0, 16)
+	hash.Sum(sum)
+	sum = sum[:16]
```
lucas-clemente: This should be unnecessary.

internal/crypto/null_aead_fnv128a.go
marten-seemann: The hash is written in little endian, so we have to reverse the slice.

internal/crypto/null_aead_fnv128a.go
marten-seemann: It's not if we use the return value. We should do that.

PR #1278, opened 2018-04-03
create a logger interface and use it everywhere

marten-seemann:


marten-seemann:
@lucas-clemente: Done. PTAL.

PR #1285, opened 2018-04-05
only allow sending of retransmissions if these are RTO probe packets

### LINKED ISSUE #1283: only send probe packets when congestion limited
Currently, we're allowing any retransmissions to be sent when congestion limited.

marten-seemann:
Fixes #1283.

### DIFF
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -61,6 +61,8 @@ type sentPacketHandler struct {
 
 	// The number of times an RTO has been sent without receiving an ack.
 	rtoCount uint32
+	// The number of RTO probe packets that should be sent.
+	numRTOs int
 
 	// The time at which the next packet will be considered lost based on early transmit or exceeding the reordering window in time.
 	lossTime time.Time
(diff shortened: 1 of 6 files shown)


internal/ackhandler/sent_packet_handler.go
```diff
@@ -455,15 +458,19 @@ func (h *sentPacketHandler) SendMode() SendMode {
 		h.logger.Debugf("Limited by the number of tracked packets: tracking %d packets, maximum %d", numTrackedPackets, protocol.MaxTrackedSentPackets)
 		return SendNone
 	}
-	// Send retransmissions first, if there are any.
-	if len(h.retransmissionQueue) > 0 {
-		return SendRetransmission
+	if h.numRTOs > 0 {
+		h.numRTOs--
```
lucas-clemente: This makes this function non-pure, which isn't clear from the name. Either change the name or do the state change where you actually send the RTO.

marten-seemann: @lucas-clemente `NextPacketSendMode`?

lucas-clemente: That doesn't solve the issue. I think the best fix is decrementing numRTOs when actually queueing them.

Sad that Go doesn't support const functions 🙄

PR #1286, opened 2018-04-05
expose the gQUIC 39 version number in the quic package

### LINKED ISSUE #1271: Expose version information
At the moment all the protocol version related structs are concealed in `internal/protocol` so applications using this library have no way of
1. detecting supported version (might be useful for printing a server's version information via e.g. ` caddy -version`)
2. specifying which versions the server should support via `h2quic.Server{ QuicConfig: &quic.Config{Versions: versions} }`

marten-seemann:
Fixes #1271.

Not sure about the best variable name. Feel free to make better suggestions.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -16,6 +16,9 @@ type StreamID = protocol.StreamID
 // A VersionNumber is a QUIC version number.
 type VersionNumber = protocol.VersionNumber
 
+// VersionGQUIC39 is gQUIC version 39.
+const VersionGQUIC39 = protocol.Version39
+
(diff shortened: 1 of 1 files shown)


interface.go
```diff
@@ -16,6 +16,9 @@ type StreamID = protocol.StreamID
 // A VersionNumber is a QUIC version number.
 type VersionNumber = protocol.VersionNumber
 
+// VersionGQUIC39 is gQUIC version 39.
+var VersionGQUIC39 = protocol.Version39
```
lucas-clemente: const?

Issue #1289, opened 2018-04-11
VersionTLS does not close open streams on non-graceful disconnect


jared2501:
Hey @marten-seemann --

I'm deploying quic with VertionsTLS (since I need client auth), and I think I've found a bug with the streams_map. 

In the streams_map_legacy, the CloseWithError method will iterate over all streams and close invoke closeForShutdown():

![image](https://user-images.githubusercontent.com/2866738/38597758-5758fb74-3d0e-11e8-88ab-92b5cc0a1b51.png)

But in the streams_map, this doesn't happen:

![image](https://user-images.githubusercontent.com/2866738/38597779-7adce2d6-3d0e-11e8-98e5-44f5eaf21c6a.png)

Since the streams_map is code-generated, my Go-fu isn't good enough to figure out a good way to fix this. Any suggestions?

jared2501:
Note, by copying the code from the legacy streams map I can fix the bug, so I'm pretty sure this is the issue!

PR #1290, opened 2018-04-11
Close streams in streamsMap when connection closed

jared2501:
An attempt at fixing #1289.

### DIFF
--- a/streams_map.go
+++ b/streams_map.go
@@ -217,8 +217,16 @@ func (m *streamsMap) UpdateLimits(p *handshake.TransportParameters) {
 }
 
 func (m *streamsMap) CloseWithError(err error) {
-	m.outgoingBidiStreams.CloseWithError(err)
-	m.outgoingUniStreams.CloseWithError(err)
-	m.incomingBidiStreams.CloseWithError(err)
-	m.incomingUniStreams.CloseWithError(err)
+	for _, s := range m.outgoingBidiStreams.CloseWithError(err) {
+		s.closeForShutdown(err)
+	}
+	for _, s := range m.outgoingUniStreams.CloseWithError(err) {
+		s.closeForShutdown(err)
+	}
+	for _, s := range m.incomingBidiStreams.CloseWithError(err) {
+		s.closeForShutdown(err)
+	}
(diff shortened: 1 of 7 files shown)


jared2501:
@marten-seemann - feel free to fork this change

marten-seemann:
Thanks for the PR!
Why do you return a map on CloseWithError? I’d prefer to call closeForShutdown() in the respective streams map.

Regarding the generated code: Just edit the _generic.go and run go generate. This will generate all required code.

jared2501:
I returned the items since the item is a `generic.Type` so we can't invoke any methods on it. Not sure if there's a better way to do this?

And thanks for the tip!

marten-seemann:
Code generation sucks...
I’m not sur if it works, but did you try adding the close method to the generic.Type (e.g. by creating an interface that has this method and inherits generic.Type)?
If that works, this would give us a lot cleaner code.

jared2501:
Hmm, unfortunately you can't do that ;(

This code in the generic map:

```
type closeableStream interface {
	item
	closeForShutdown(error)
}
```

Results in this generated code with an error:

image

jared2501:
To be honest, having the close method return the list of closed streams doesn't seem that bad to me!

marten-seemann:
I think it's not a good solution because 1. is violates the separation of concern (streams are created and should be closed in the same struct) and 2. it involves a lot of unnecessary allocations.

jared2501:
Ah yeah I see, makes sense. Will play around with this a bit tonight. I
think perhaps I can make the streams inherit the interface and it will work
as you suggest.

marten-seemann:
Hi @jared2501,

I gave it a try myself, and submitted a fix in #1295. The key change is redefining the `item`:
```go
type item interface {
    generic.Type
    closeForShutdown(error)
}
```
and to make sure, that this definition doesn't end up in any auto-generated files (by putting the definition in a separate file). 
As I said before, code generation sucks. This is so much effort to get it right, and it feels so unnatural. I wish there was another way to do that, but I don't see one.

jared2501:
Thanks for that Marten! That solution looks much better. Maybe one day the
language will make this kind of thing easier :)

PR #1291, opened 2018-04-11
Copy over RootCAs to mint config

jared2501:


jared2501:
@marten-seemann - done!

PR #1295, opened 2018-04-16
close all streams when closing the IETF QUIC streams map

### LINKED ISSUE #1289: VersionTLS does not close open streams on non-graceful disconnect

marten-seemann:
Fixes #1289. Closes #1290.

### DIFF
--- a/streams_map_generic_helper.go
+++ b/streams_map_generic_helper.go
@@ -1,0 +1,11 @@
+package quic
+
+import "github.com/cheekybits/genny/generic"
+
+// In the auto-generated streams maps, we need to be able to close the streams.
+// Therefore, extend the generic.Type with the stream close method.
+// This definition must be in a file that Genny doesn't process.
+type item interface {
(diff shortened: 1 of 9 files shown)


streams_map_generic_helper.go
```diff
@@ -0,0 +1,8 @@
+package quic
+
+import "github.com/cheekybits/genny/generic"
+
+type item interface {
```
lucas-clemente: This definitely needs a comment. Link to the discussion on the other PR?

Issue #1302, opened 2018-04-16
limit the size of ACK frames


marten-seemann:
IETF QUIC ACK frames can be arbitrarily large. We need to limit the size such that they fit into one packet.

PR #1303, opened 2018-04-17
refactor the ACK frame

marten-seemann:

This PR removes the `LargestAcked` and `LowestAcked` fields from the `AckFrame` struct.
Every `AckFrame`, even if it doesn't report any packets missing, now contains at least one `AckRange`.

This way, we can get rid with some awkward consistency conditions (e.g. for ACKs that reported packets missing, the `LargestAcked` value had to match the largest packet number of the highest range, and the `LowestAcked` value had to match the lowest packet number of the lowest range).

This will make it easier to implement the size limit for the ACK frame (#1302).


internal/wire/ack_frame.go
```diff
@@ -215,23 +190,29 @@ func (f *AckFrame) validateAckRanges() bool {
...
 		return false
 	}
 
-	if f.HasMissingRanges() {
-		// TODO: this could be implemented as a binary search
-		for _, ackRange := range f.AckRanges {
-			if p >= ackRange.Smallest && p <= ackRange.Largest {
-				return true
-			}
+	// TODO: this could be implemented as a binary search
```
lucas-clemente: There's a binary search implementation in the stdlib btw: https://golang.org/pkg/sort/#Search

marten-seemann: Will implement that in a separate PR.

PR #1304, opened 2018-04-17
implement a buffer optimized for writing to preallocated slices

marten-seemann:
We will need to move away from the `bytes.Buffer` for writing packets when implementing #1302, since it doesn't implement the `io.WriterAt`. We will need to write a specific offset, since in the ACK frame, we **first** write the number of ACK ranges, and **later** we write the ACK ranges. However, we don't know in advance how many ranges we will write, so we will need go back and update the counter.

This can be easily achieved by using our own interface, which combines `io.Writer`, `io.ByteWriter` and `io.WriterAt`. For now, this PR only implements `io.Writer` and `io.ByteWriter`, the `io.WriterAt` will be added in the actual fix for #1302.

The `utils.Buffer` is considerably faster than the `bytes.Buffer`, since it doesn't need to perform any checks if the underlying byte slice needs to be grown (for 1<<15 repetitions).

```
BenchmarkBytesBufferSingleByte-8   	   30000	    106479 ns/op
BenchmarkUtilsBufferSingleByte-8   	  100000	     49962 ns/op
BenchmarkBytesBuffer2Bytes-8       	    5000	    461520 ns/op
BenchmarkUtilsBuffer2Bytes-8       	   10000	    298070 ns/op
BenchmarkBytesBuffer10Bytes-8      	    5000	    521583 ns/op
BenchmarkUtilsBuffer10Bytes-8      	   10000	    343500 ns/op
BenchmarkBytesBuffer50Bytes-8      	    2000	   1545412 ns/op
BenchmarkUtilsBuffer50Bytes-8      	    2000	   1417985 ns/op
BenchmarkBytesBuffer200Bytes-8     	    1000	   2741885 ns/op
BenchmarkUtilsBuffer200Bytes-8     	    1000	   2691521 ns/op
```

### DIFF
--- a/internal/utils/buffer.go
+++ b/internal/utils/buffer.go
@@ -1,0 +1,65 @@
+package utils
+
+import (
+	"io"
+)
+
+// A ByteWriter combines the io.Writer and io.ByteWriter interface
+type ByteWriter interface {
+	io.ByteWriter
+	io.Writer
+}
+
+// The Buffer implements the ByteWriter interface.
+// It is supposed to be used with pre-allocated slices.
+// It does not grow the underlying slice.
+// Every write that exceeds the capacity will panic.
+type Buffer struct {
+	buf []byte
+}
+
+var _ ByteWriter = &Buffer{}
+
+// NewBuffer creates a new Buffer
+func NewBuffer(buf []byte) *Buffer {
+	return &Buffer{buf}
+}
+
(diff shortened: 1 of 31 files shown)


marten-seemann:
We'll find a better solution for the ACK frame, which doesn't require jumping around in the buffer.

PR #1313, opened 2018-04-22
implement the payload length field in the Long Header

### LINKED ISSUE #1305: implement payload length for long headers
See https://github.com/quicwg/base-drafts/pull/1262 and https://github.com/quicwg/base-drafts/pull/1294.

marten-seemann:
Fixes #1305.


internal/wire/header.go
```diff
@@ -33,6 +33,7 @@ type Header struct {
 	Type         protocol.PacketType
 	IsLongHeader bool
 	KeyPhase     int
+	PayloadLen   protocol.ByteCount
```
lucas-clemente: Maybe it's worth splitting the long header and short header? Or have a substruct with only the long header / short header specific fields?

PR #1321, opened 2018-04-24
fix deadlock when receiving two packets with diversification nonces

### LINKED ISSUE #1308: deadlock when receiving multiple diversification nonces
I use gomobile to convert quic-go into a lib and running on android. Occasionally it will hang on the follwing DialAddr method.

config := &quic.Config{KeepAlive: true}
session, err := quic.DialAddr(ipAddressUrl, &tls.Config{InsecureSkipVerify: true}, config)
stream, err := session.OpenStreamSync()

After further investigation， it seems session.SetDiversificationNonce and cryptoSetupClient.HandleCryptoStream called from seesion.run() has a deadlock on channel divNonceChan and paramsChan. 

func (h *cryptoSetupClient) SetDiversificationNonce(data []byte) {
	h.divNonceChan <- data
}

func (h *cryptoSetupClient) HandleCryptoStream() error {
     ......
      case TagSHLO:
			params, err := h.handleSHLOMessage(message.Data)
			if err != nil {
				return err
			}
			// blocks until the session has received the parameters
			h.paramsChan <- *params
			h.handshakeEvent <- struct{}{}
			close(h.handshakeEvent)
   ......
}

Please suggest, thanks

marten-seemann:
Fixes #1308.

### DIFF
--- a/internal/handshake/crypto_setup_client.go
+++ b/internal/handshake/crypto_setup_client.go
@@ -38,7 +38,7 @@ type cryptoSetupClient struct {
 	lastSentCHLO     []byte
 	certManager      crypto.CertManager
 
-	divNonceChan         <-chan []byte
+	divNonceChan         chan struct{}
 	diversificationNonce []byte
 
 	clientHelloCounter int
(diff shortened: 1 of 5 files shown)


marten-seemann:
Test.

Issue #776, opened 2017-07-31
Pluggable congestion control


isaachess:
Right now this package uses an implementation of TCP-cubic for congestion control. It would be nice if the code easily allowed different congestion control algos to be used.

For example, if I wanted to use a LEDBAT congestion control algorithm, if I were to write my own implementation of the `SendAlgorithm` interface I still don't see a way to plug that in when I create a new session.

So it would be nice, even if we do not yet have officially supported congestion control other than cubic, to have an easy way to write one and use it instead of cubic.

isaachess:
The reason I'm asking this is specifically for LEDBAT. We need the ability to use cubic cc for some connections, but LEDBAT for others. So if there is already a simple way to do that with this QUIC implementation I'd be interested in that.

lucas-clemente:
I think the options are either having an enum for the congestion controllers we support (which wouldn't allow it to be different on a connection basis), or adding a `func (connectionID int64) SendAlgorithm` to the quic.Config.

We also have plans to support BBR (#341), and for that we'll need to change the SendAlgorithm interface – so there will be breaking changes at some point.

If you plan to do benchmarks please be aware that some performance-relevant parts of our implementation are still lacking (e.g. loss recovery).

isaachess:
I think having different congestion-control per-connection is not really what I need (sorry for being imprecise), but rather being able to change it per-process.

And yes, I agree having an enum for officially supported congestion controllers is a fantastic idea. I just think even now, before you get around to officially supporting a bunch of different congestion controllers, the code could be rewritten a little to make it simpler to write your own congestion controller and plug it into the session.

lucas-clemente:
SGTM :) If forking is an acceptable solution for now, the relevant line you need to change is https://github.com/lucas-clemente/quic-go/blob/master/ackhandler/sent_packet_handler.go#L69.

Feel free to send PRs to simplify things wherever you think it makes sense – what specifically did you have in mind?

qiulaidongfeng:
Hi All:

For those who are concerned about this issue
https://github.com/qiulaidongfeng/quic-go
The experiment to solve this issue by exporting SendAlgorithmWithDebugInfos and adding the Congestion field in Config is now available. Those interested are welcome to experiment with different congestion control algorithms, and feedback is highly appreciated.

[...]

Issue #341, opened 2016-11-01
implement BBR congestion control


marten-seemann:

glen-dai:
When will be quic-go support BBR congestion control?

marten-seemann:
It's still on our list, but we have a lot of higher-priority issues at the moment.

LibinLiu0189:
is BBR implemented in any quic-go branch now? Thanks a lot.

marten-seemann:
No, and we're currently not planning to implement it anytime soon.

LibinLiu0189:
> No, and we're currently not planning to implement it anytime soon.

Got it. Thanks.

jinleileiking:
I do not know why this thing is not on the top list. Bbr is so critical in quic cc

netroby:
any news on add it to quic-go？

marten-seemann:
There’s no need to regularly ping us on this issue. It will be updated once we start working on it.

moranno:
It's been almost 3 years...

bt90:
https://github.com/lucas-clemente/quic-go/issues/3417#issuecomment-1127961210 still stands

marten-seemann:
Update on BBR: We're not planning to implement BBR in quic-go. Over the years, BBR has been shown to not be the silver bullet to solve all of our performance problems. BBR remains a highly complex algorithm with an enormous set of magic configuration values. It has also not seen the widespread adoption that one might have expected after BBR's initial introduction.

In the mean time, different congestion controller algorithms have been explored by the congestion control community, with promising results. quic-go might experiment with those at some point in the future. Until then, there are lots of improvements outside of congestion control that will have a significant performance impact, for example #3526, #2482, #3563.

marten-seemann:
BBRv3 support is now tracked in #4565.

PR #1325, opened 2018-04-27
changed: congestion control can be selected via the config

TobiasVi:
Issue #776


TobiasVi:
I signed it!

Nikolasel:
I signed it!

Nikolasel:
Please rerun.

TobiasVi:
I signed it!

marten-seemann:
Hi @TobiasVi, thank you for this PR!
We're excited about trying new congestion control algorithms in QUIC, and implementing BBR (#341) has long been on our ToDo list.
However, the only reason we currently have Reno support is because we translated the congestion control code from Chromium to Go, and it contained this Reno flag. It's completely untested, and definitely not intended for production use.
Therefore, I'd rather not offer a config option as this point, but instead wait until we have BBR support.

PR #1332, opened 2018-05-01
improve logging of ACK frames with ranges

marten-seemann:


internal/wire/log.go
```diff
@@ -22,9 +27,13 @@ func LogFrame(logger utils.Logger, frame Frame, sent bool) {
 		}
 	case *AckFrame:
 		if len(f.AckRanges) > 1 {
-			logger.Debugf("\t%s &wire.AckFrame{LargestAcked: 0x%x, LowestAcked: 0x%x, AckRanges: %#v, DelayTime: %s}", dir, f.LargestAcked(), f.LowestAcked(), f.AckRanges, f.DelayTime.String())
+			ackRanges := make([]string, len(f.AckRanges))
+			for i, r := range f.AckRanges {
+				ackRanges[i] = fmt.Sprintf("{Largest: %#x, Smallest: %#x}", r.Largest, r.Smallest)
+			}
+			logger.Debugf("\t%s &wire.AckFrame{LargestAcked: %#x, LowestAcked: %#x, AckRanges: {%s}, DelayTime: %s}", dir, f.LargestAcked(), f.LowestAcked(), strings.Join(ackRanges, ", "), f.DelayTime.String())
```
lucas-clemente: We should probably build the string and call Debugf once, so that multiple lines don't interleave.

marten-seemann: I don’t understand. Debugf is only called once per branch.

lucas-clemente: Right. Jetlag, sorry.

PR #1340, opened 2018-05-06
schedule sending when queueing a connection-level window update

marten-seemann:
It is not sufficient to check for connection-level window updates every time a packet is sent. When a connection-level window update needs to be sent, we need to make sure that it gets sent immediately (i.e. call `scheduleSending()` in the session).
Otherwise we might end up in a situation where the connection flow controller knows that a window update needs to be sent, but the session does not. The window update would be sent with the next packet, but the next packet to send might be the CONNECTION_CLOSE after hitting the idle timeout.


marten-seemann:
@lucas-clemente: This is the most important PR to review.

PR #1343, opened 2018-05-06
log all connection flow control window increases

marten-seemann:


internal/flowcontrol/connection_flow_controller.go
```diff
@@ -78,6 +78,7 @@ func (c *connectionFlowController) GetWindowUpdate() protocol.ByteCount {
 func (c *connectionFlowController) EnsureMinimumWindowSize(inc protocol.ByteCount) {
 	c.mutex.Lock()
 	if inc > c.receiveWindowSize {
+		c.logger.Debugf("Increasing receive flow control window for the connection to %d kB, in response to stream flow control window increase", c.receiveWindowSize/(1<<10))
```
lucas-clemente: How often do we expect this to trigger? Should we maybe put it behind an if logger.Debug()?

PR #1345, opened 2018-05-07
add logging to the ackhandler

marten-seemann:


internal/ackhandler/received_packet_handler.go
```diff
@@ -82,8 +89,12 @@ func (h *receivedPacketHandler) ReceivedPacket(packetNumber protocol.PacketNumbe
 // IgnoreBelow sets a lower limit for acking packets.
 // Packets with packet numbers smaller than p will not be acked.
 func (h *receivedPacketHandler) IgnoreBelow(p protocol.PacketNumber) {
+	if p <= h.ignoreBelow {
+		return
+	}
 	h.ignoreBelow = p
 	h.packetHistory.DeleteBelow(p)
+	h.logger.Debugf("\tIgnoring all packets below %#x.", p)
```
lucas-clemente: Most of the log statements (all that get called on O(number packets)) here should go behind if Debug() checks to avoid unnecessary garbage.

PR #1358, opened 2018-05-15
make the peer completing the handshake last send a forward-secure packet

marten-seemann:
In gQUIC, the server completes the handshake first (after sending the SHLO). In TLS 1.3, the client completes the handshake first (after sending the CFIN).
They need to have some way to learn that the peer successfully completed the handshake, so they can stop retransmitting the handshake packets containing the SHLO or the CFIN, respectively. To achieve that, the peer finishing the handshake last will send a PING frame in a forward-secure packet.

### DIFF
--- a/session.go
+++ b/session.go
@@ -579,12 +579,18 @@ func (s *session) handleHandshakeEvent(completed bool) {
 	}
 	s.handshakeComplete = true
 	s.handshakeEvent = nil // prevent this case from ever being selected again
-	if !s.version.UsesTLS() && s.perspective == protocol.PerspectiveClient {
-		// In gQUIC, there's no equivalent to the Finished message in TLS
-		// The server knows that the handshake is complete when it receives the first forward-secure packet sent by the client.
-		// We need to make sure that the client actually sends such a packet.
-		s.packer.QueueControlFrame(&wire.PingFrame{})
-		s.scheduleSending()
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -579,12 +579,18 @@ func (s *session) handleHandshakeEvent(completed bool) {
...
+	// In gQUIC, the server completes the handshake first (after sending the SHLO).
+	// In TLS 1.3, the client completes the handshake first (after sending the CFIN).
+	// We need to make sure they learn about the peer completing the handshake,
+	// in order to stop retransmitting handshake packets.
+	// They will stop retransmitting handshake packets when receiving the first forward-secure packet.
+	// We need to make sure that a retransmittable forward-secure packet is sent,
+	// independent from the application protocol.
+	if (!s.version.UsesTLS() && s.perspective == protocol.PerspectiveClient) ||
+		(s.version.UsesTLS() && s.perspective == protocol.PerspectiveServer) {
+		s.queueControlFrame(&wire.PingFrame{})
```
lucas-clemente: Is there a reason why you dropped the scheduleSending here?

marten-seemann: Because `s.queueControlFrame`, as opposed to `s.packer.QueueControlFrame`, already calls `s.scheduleSending`.

PR #1360, opened 2018-05-15
Use http.Request.Host only for authority header and not for dialing

nogoegst:
As from `net/http` documentation:
```text
    // For client requests Host optionally overrides the Host
    // header to send. If empty, the Request.Write method uses
    // the value of URL.Host.
```

Though `h2quic` uses `http.Request.Host` as the address to connect to. This is wrong as it has to pick the address from the URL. `hostnameFromRequest()` is being used only to get the hostname to dial and there is similar (but correct) logic in `requestWriter.encodeHeaders()`.

### DIFF
--- a/h2quic/request.go
+++ b/h2quic/request.go
@@ -70,9 +70,6 @@ func requestFromHeaders(headers []hpack.HeaderField) (*http.Request, error) {
 }
 
 func hostnameFromRequest(req *http.Request) string {
-	if len(req.Host) > 0 {
-		return req.Host
-	}
 	if req.URL != nil {
 		return req.URL.Host
 	}
--- a/h2quic/request_test.go
+++ b/h2quic/request_test.go
@@ -101,16 +101,16 @@ var _ = Describe("Request", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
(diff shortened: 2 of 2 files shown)


marten-seemann:
Thanks for this PR.
I'm not sure I understand the issue. Aren't we doing exactly what the spec says?
Doesn't the standard library do the same in https://github.com/golang/net/blob/2491c5de3490fced2f6cff376127c667efeed857/http2/transport.go#L1180-L1183?

nogoegst:
@marten-seemann, right, it is all correct in https://github.com/lucas-clemente/quic-go/blob/master/h2quic/request_writer.go#L64-L67 as it was just copied from `x/net/http2`. There is no problem here. The problem is that `Request.Host` should be used only for setting the `Host`/`:authority` headers and not for dialing.

ltucker:
I think h2quic does diverge in the treatment of the request.Host field for client requests.   "regular" h2 will only modify the Host header, not change the host that is contacted.

Here is a quick demonstration of the difference: [strangedial.go](https://gist.github.com/ltucker/62926ee60cbb41e62b1ec3dfae4b4da3)

Issue #1364, opened 2018-05-17
Deadlock between onStreamCompleted and Session.Close


ltucker:
There appears to be a deadlock between calling `Session.Close(...)` and various `Stream` calls (eg `Read`) which may trigger a call to `onStreamCompleted` while holding a Stream lock.  They each hold a lock the other needs since they acquire them in different orders (eg legacy_stream_map lock and receive_stream lock)

Here is a (finnecky, contrived) reproduction of a particular instance of this problem: [deadlock_test.go](https://gist.github.com/ltucker/0480861abc84592f32da3a8dfe4957fb)

When it hits the issue, the goroutines are deadlocked like this:

[Full output](https://gist.github.com/ltucker/7779a7d31634fb77312929d37a788e13)
```
goroutine 107673 [semacquire, 9 minutes]:
sync.runtime_SemacquireMutex(0xc420740384, 0x0)
    /usr/local/Cellar/go/1.10/libexec/src/runtime/sema.go:71 +0x3d
sync.(*Mutex).Lock(0xc420740380)
    /usr/local/Cellar/go/1.10/libexec/src/sync/mutex.go:134 +0x108
sync.(*RWMutex).Lock(0xc420740380)
    /usr/local/Cellar/go/1.10/libexec/src/sync/rwmutex.go:93 +0x2d
github.com/lucas-clemente/quic-go.(*streamsMapLegacy).DeleteStream(0xc420740380, 0x2d, 0x0, 0x0)
    /Users/ltucker/go/src/github.com/lucas-clemente/quic-go/streams_map_legacy.go:224 +0x39
github.com/lucas-clemente/quic-go.(*session).onStreamCompleted(0xc420711400, 0x2d)
    /Users/ltucker/go/src/github.com/lucas-clemente/quic-go/session.go:1219 +0x3e
github.com/lucas-clemente/quic-go.(*stream).checkIfCompleted(0xc420654000)
    /Users/ltucker/go/src/github.com/lucas-clemente/quic-go/stream.go:173 +0x6f
github.com/lucas-clemente/quic-go.newStream.func2()
    /Users/ltucker/go/src/github.com/lucas-clemente/quic-go/stream.go:119 +0x52
github.com/lucas-clemente/quic-go.(*uniStreamSender).onStreamCompleted(0xc420590000, 0x2d)
    /Users/ltucker/go/src/github.com/lucas-clemente/quic-go/stream.go:41 +0x29
[...]

PR #1365, opened 2018-05-17
call onStreamCompleted without holding stream locks

ltucker:
An example fix for #1364 ... in this case I tried to preserve the property that the onStreamCompleted callback has been called / completed before returning from the triggering call by calling it deferred once the stream lock has been released.


ltucker:
I signed it!

ltucker:
@marten-seemann great thanks!

PR #1372, opened 2018-05-22
add support for QUIC 42

### LINKED ISSUE #845: add support for QUIC 42

marten-seemann:
Fixes #845.

QUIC 42 accepts overlapping stream data. We've been doing this for some time already, so we don't need any additional changes.

### DIFF
--- a/internal/protocol/version.go
+++ b/internal/protocol/version.go
@@ -18,7 +18,8 @@ const (
 
 // The version numbers, making grepping easier
 const (
-	Version39       VersionNumber = gquicVersion0 + 3*0x100 + 0x9 + iota
+	Version39       VersionNumber = gquicVersion0 + 3*0x100 + 0x9
+	Version42       VersionNumber = gquicVersion0 + 4*0x100 + 0x2
 	VersionTLS      VersionNumber = 101
 	VersionWhatever VersionNumber = 0 // for when the version doesn't matter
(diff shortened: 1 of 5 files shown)


marten-seemann:
The problem with the Travis build is that Travis is still using Ubuntu 14.04 (!!!), which apparently comes with Chrome 62. Of course, this ancient version doesn't support QUIC 42, which is why the build is failing.
As far as I can see, there's no way to get an up-to-date build environment if we want to use Travis' container infrastructure (which speeds up the builds by 20 seconds or so). I'm not really sure what to do about this.

@lucas-clemente: We should solve this. Chrome will ship QUIC 43 very soon, and I'd prefer to have CI integration tests with that. Any ideas?

internal/protocol/version.go
```diff
@@ -18,7 +18,8 @@ const (
 
 // The version numbers, making grepping easier
 const (
-	Version39       VersionNumber = gquicVersion0 + 3*0x100 + 0x9 + iota
+	Version39       VersionNumber = gquicVersion0 + 3*0x100 + 0x9
```
lucas-clemente: How is this more readable than 0x0309? ;)

marten-seemann:
@lucas-clemente: any ideas regarding the Travis issue?

lucas-clemente:
Containers sound like out best option.

PR #1376, opened 2018-05-28
Remove dead clients from roundtripper

hpidcock:
Fixing an issue where clients that previously error'd at the quic level were not removed from future requests through the RoundTripper.


hpidcock:
I signed it!

marten-seemann:
As described in #765, this is not the correct fix. Closing, also due to age.

PR #1403, opened 2018-06-13
only accept a single Retry packet for a connection

### LINKED ISSUE #1401: IETF QUIC handshake fails for RTTs > 200 ms
The error is: `closing session in order to recreate after a retry`.

marten-seemann:
Fixes #1401.

### DIFF
--- a/client.go
+++ b/client.go
@@ -24,6 +24,8 @@ type client struct {
 	conn     connection
 	hostname string
 
+	receivedRetry bool
+
 	versionNegotiated                bool // has the server accepted our version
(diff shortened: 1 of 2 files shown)


marten-seemann:
All other packets are idempotent, that’s why it doesn’t hurt to process duplicates both (although we have some duplicate detection for performance reasons). This doesn’t hold for Retry packets, as #1401 shows. What’s your suggestion for dealing with this issue?

marten-seemann:
@lucas-clemente: I changed the logic such that only authenticated Retry packets are accepted. PHAL.

Issue #561, opened 2017-04-23
make it possible to run Dialer and Listener on the same UDP socket


marten-seemann:
Not sure yet how we can make this work, but it'd be a nice feature.

lucas-clemente:
I don't understand what you want to do here and why ^^

AudriusButkevicius:
This would be useful for protocols that rely on stun for NAT punchthrough.

It would be awesome if this existed, yet it can be done outside of the library if quic framing is deterministic and sessions are easily identifiable with each Write() call.

We already do something simillar in syncthing and kcp where we have a single UDP connection and then filter incoming stun and KCP traffic based on the signatures into virtual connections using something like https://github.com/AudriusButkevicius/pfilter.

net.PacketConn ->

marten-seemann:
This will not be supported by QUIC v1, see https://github.com/quicwg/base-drafts/issues/714#issuecomment-359995274.

marten-seemann:
Good news. It looks like with the current new QUIC header format, this will be possible.

AudriusButkevicius:
Superb! Are there plans to suppprt this feature in this implementation?

marten-seemann:
Sure, that's what this issue is there for ;)
I don't have a timeline yet though.

marten-seemann:
It's probably the most complicated issues we currently have.
I've been working on #1324, which is a subset of this issue, and I already spent two complete days refactoring stuff, and it's still a long way to go...

AudriusButkevicius:
Given the multiplexer is now merged, would it be possible to reuse the multiplexer for mixing server and client sockets?

My idea is have a type exposed:
```
type Multiplexer interface {
     Accept() (Session, error)
     Dial(...) (Session, error)
     ReadFrom([]byte) (int, net.Addr, error)
}
```

It would work as follows:

1. Dial normally registers the connection with the mux, and forward connections to the right session.
2. Any valid quic packets sent for sessions the mux is not aware of, would go onto an Accept queue, which can be consumer by calling Accept(), at which point they would get registered with the mux. Not sure that everything that is needed for `newsession` is available at that point.
3. Any non-quic packets would go on to the backlog channel, which can then be consumer by calling ReadFrom(), this way we can piggyback stun on the same net.PacketConn.

marten-seemann:
Header parsing is currently a mess, which is blocking this issue. There are just too many different header formats at the moment...
[...]

PR #1407, opened 2018-06-16
multiplex multiple clients on one packet conn

### LINKED ISSUE #1324: allow multiple clients to run on the same port
When given a `net.PacketConn`. In this case, we'd have to use the connection ID to demultiplex the connections, similar to what the server is already doing.

marten-seemann:
Fixes #1324.


Changelog.md
```diff
@@ -6,6 +6,7 @@
 - Add a `quic.Config` option for the maximum number of incoming streams.
 - Add support for QUIC 42 and 43.
 - Add dial functions that use a context.
+- Multiplex clients on a net.PacketConn, when using Dial(Context).
```
lucas-clemente: Dial(conn)?

client_multiplexer.go
```diff
@@ -0,0 +1,94 @@
...
+		m.conns[c] = sessions
+	}
+	sessions.Add(connID, handler)
+	if ok {
+		return
+	}
+
+	// If we didn't know this packet conn before, listen for incoming packets
+	// and dispatch them to the right sessions.
+	go func() {
```
lucas-clemente: This should go into a separate function.

client_multiplexer.go
```diff
@@ -0,0 +1,94 @@
...
+			rcvTime := time.Now()
+
+			r := bytes.NewReader(data)
+			hdr, err := wire.ParseHeaderSentByServer(r)
+			// drop the packet if we can't parse the header
+			if err != nil {
+				m.logger.Debugf("error parsing packet from %s: %s", addr, err)
+				continue
+			}
+			hdr.Raw = data[:len(data)-r.Len()]
```
lucas-clemente: This code looks familiar – can we maybe share it with the existing implementation?

marten-seemann: The problem is that the header parser still needs the perspective. We'll need to get creative if we want to fix this (which will be needed for #561).

client_multiplexer_test.go
```diff
@@ -0,0 +1,106 @@
...
+		getClientMultiplexer().Add(conn, connID2, packetHandler2)
+		Eventually(handledPacket1).Should(BeClosed())
+		Eventually(handledPacket2).Should(BeClosed())
+		// makes the listen go routine return
+		packetHandler1.EXPECT().Close(gomock.Any()).AnyTimes()
+		packetHandler2.EXPECT().Close(gomock.Any()).AnyTimes()
+		close(conn.dataToRead)
+	})
+
+	It("drops unparseable packets", func() {
```
AudriusButkevicius: It would be nice if there was a way to read these, especially if you are performing stun over the same net.PacketConn, so you could essentially filter out stun traffic out of quic traffic.

marten-seemann: Stun multiplexing is not the task of a QUIC library. You’ll have to filter those out before passing them on to quic-go.

AudriusButkevicius:
Sure, it's just a lot of effort, where as the library already knows which packets it can't parse and could just call a callback with the bytes.

PR #1415, opened 2018-06-18
only use the diversification nonce after authenticating a packet

marten-seemann:


marten-seemann:
This PR is wrong. We need the diversification nonce to unpack the packet containing the SHLO, so we can't try decrypting it without setting it before.

PR #1464, opened 2018-08-01
multiplex server and client on the same packet conn

### LINKED ISSUE #561: make it possible to run Dialer and Listener on the same UDP socket

marten-seemann:
Fixes #561.


internal/protocol/perspective.go
```diff
@@ -9,6 +9,11 @@ const (
 	PerspectiveClient Perspective = 2
 )
 
+// Opposite returns the perspective of the peer
+func (p Perspective) Opposite() Perspective {
+	return 3 - p
```
lucas-clemente: wat.

server.go
```diff
@@ -67,6 +67,9 @@ type server struct {
 	config  *Config
 
 	conn net.PacketConn
+	// If the server is started with ListenAddr, we create a packet conn.
+	// If it is started with Listen, we take a packet conn as a parameter.
+	createdPacketConn bool
```
lucas-clemente: Nit: Maybe shouldCloseConn?

marten-seemann: We have the same in the client, so I'd prefer to leave it like this.

packet_handler_map.go
```diff
@@ -67,13 +105,94 @@ func (h *packetHandlerMap) Close() error {
...
+}
+
+func (h *packetHandlerMap) handlePacket(addr net.Addr, data []byte) error {
+	rcvTime := time.Now()
+
+	r := bytes.NewReader(data)
+	iHdr, err := wire.ParseInvariantHeader(r, h.connIDLen)
+	// drop the packet if we can't parse the header
+	if err != nil {
+		return fmt.Errorf("error parsing invariant header: %s", err)
```
AudriusButkevicius: Could we add a "sink" for these packets? This would make piggybacking a DHT on the same connection possible.

PR #1477, opened 2018-08-12
implement the new Retry mechanism

### LINKED ISSUE #1431: implement the new Retry mechanism
This should simplify things a bit. After https://github.com/quicwg/base-drafts/pull/1498 lands.

marten-seemann:
Fixes #1431. Depends on #1475 and #1476.

This is a big change, that implements the first part of the draft-13 changes.
The Stateless Retry is now independent of the TLS layer. This means we don't need to mess with TLS Cookies anymore, which cleans up a lot interdependence between the `quic` and the `handshake` package.


internal/handshake/crypto_stream_conn.go
```diff
@@ -1,23 +1,43 @@
...
 import (
+	"bytes"
 	"io"
 	"net"
 	"time"
 )
 
 type cryptoStreamConn struct {
-	io.ReadWriter
+	buffer *bytes.Buffer
```
lucas-clemente: Are you aware of https://golang.org/pkg/bufio/?

PR #1522, opened 2018-09-20
remove dependency on mint for HKDF

marten-seemann:
mint exposes a HKDF extraction and expansion methods, which we used in our `crypto` package. Tris doesn't expose those methods, and it's just around 40 LoC (which can easily be tested using the cases provided in the [HKDF RFC](https://tools.ietf.org/html/rfc5869), so it makes sense to copy those to quic-go.


lucas-clemente:
Is there a reason why we can't use https://godoc.org/golang.org/x/crypto/hkdf ?

marten-seemann:
The standard library `hkdf` package doesn't expose an extract and expand function.

PR #1539, opened 2018-10-05
Replace (streamI)(nil) with streamI(nil)

ludweeg:
Found using https://go-critic.github.io/overview#typeUnparen-ref


marten-seemann:
Thanks for this PR.
This is the standard way of testing if an interface implements another interface, see https://stackoverflow.com/questions/27803654/explanation-of-checking-if-value-implements-interface and https://splice.com/blog/golang-verify-type-implements-interface-compile-time/, so I don't see any reason to change that.

PR #1542, opened 2018-10-15
use a modified version of tls-tris instead of mint

marten-seemann:
This is part of the changes of the stream 0 design team.
It replaces stream 0 by the three dedicated crypto streams that send data in CRYPTO frames.
The crypto setup now uses my fork of tls-tris, named qtls, for the handshake. Unfortunately, tris pulls in a bunch of dependencies, and vendoring those increases our LoC by more than 10k.

There are still a few TODOs here, the biggest one is the implementation of separate packet number spaces (#1534).


marten-seemann:
@lucas-clemente: This PR is ready for review now.

marten-seemann:
@lucas-clemente: Can you please review this PR?

internal/handshake/aead.go
```diff
@@ -0,0 +1,56 @@
...
+	"encoding/binary"
+
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+)
+
+type sealer struct {
+	iv   []byte
+	aead cipher.AEAD
+
+	nonceBuf []byte
```
lucas-clemente: Why is this a member and not a local var?

internal/handshake/crypto_setup_tls.go
```diff
@@ -1,168 +1,474 @@
...
+	var buf *bytes.Buffer
+	switch encLevel {
+	case protocol.EncryptionInitial:
+		buf = &h.initialReadBuf
+	case protocol.EncryptionHandshake:
+		buf = &h.handshakeReadBuf
+	default:
+		return fmt.Errorf("received handshake data with unexpected encryption level: %s", encLevel)
+	}
+	buf.Write(data)
```
lucas-clemente: WDYT about factoring wire stuff out into a separate helper / file.

vendor/vendor.json
```diff
@@ -44,6 +38,18 @@
 			"revision": "d2f86524cced5186554df90d92529757d22c1cb6",
 			"revisionTime": "2016-08-23T09:51:56Z"
 		},
+		{
+			"checksumSHA1": "tMUbiAxGypX/0kCJpIFg31Q6vOU=",
+			"path": "github.com/marten-seemann/qtls",
```
lucas-clemente: This needs a license. Since it's forked from tris, does tris have one?

marten-seemann: It doesn't. I opened an issue there: https://github.com/cloudflare/tls-tris/issues/147.

internal/handshake/aead.go
marten-seemann: This avoids allocating the `nonceBuf` again and again, for every packet that we send.

internal/handshake/crypto_setup_tls.go
marten-seemann: #1552 moves this to the quic package.

internal/handshake/aead.go
marten-seemann: I added a comment.

PR #1544, opened 2018-10-16
optimize stream deadline timers

### LINKED ISSUE #1543: use a single timer in the stream
Currently, we're using `time.After` in the `sendStream` and the `receiveStream`. This creates a lot of allocations which aren't garbage collected until the timer expires (which can be far in the future). It's better to use a single timer and reset it.
See https://gist.github.com/marten-seemann/33a80e3b84c0a38e5416b51f18426326 for a benchmark:
```
BenchmarkTimeAfter-12     	 5000000	       304 ns/op	     253 B/op	       3 allocs/op
BenchmarkTimerReset-12    	20000000	        89.3 ns/op	       0 B/op	       0 allocs/op
```

marten-seemann:
Fixes #1543. Thanks to @kubuxu for pointing this out.

We're currently calling `time.After` for every change of the deadline, as long as a `Read` or a `Write` call is blocking. This creates a new timer which is only GCed after the deadline expires, and consumes about 250 bytes of data each.
By using a single timer, we can avoid any allocations for setting a new deadline.

### DIFF
--- a/receive_stream.go
+++ b/receive_stream.go
@@ -43,8 +43,9 @@ type receiveStream struct {
 	canceledRead      bool // set when CancelRead() is called
 	resetRemotely     bool // set when HandleRstStreamFrame() is called
 
-	readChan     chan struct{}
-	readDeadline time.Time
+	readChan      chan struct{}
+	deadline      time.Time
+	deadlineTimer *time.Timer // initialized by SetReadDeadline()
 
 	flowController flowcontrol.StreamFlowController
 	version        protocol.VersionNumber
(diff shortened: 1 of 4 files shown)


receive_stream.go
```diff
@@ -272,13 +272,22 @@ func (s *receiveStream) onClose(offset protocol.ByteCount) {
...
+	s.deadline = t
 	s.mutex.Unlock()
-	// if the new deadline is before the currently set deadline, wake up Read()
-	if t.Before(oldDeadline) {
+	if s.deadline.IsZero() { // skip if there's no deadline to set
 		s.signalRead()
+		return nil
+	}
+	// Lazily initialize the deadline timer.
+	if s.deadlineTimer == nil {
```
lucas-clemente: Does this not need to be protected by the mutex? (same in the other commit)

marten-seemann: Good catch. I think we can just use a `defer` here, since the deadline calls won't be used that often.

Kubuxu: If you want it to be thread safe, then `sync.Once` might be the best option.

marten-seemann: It would, if we didn't have the `deadline`, which also needs to be thread-safe.

PR #1560, opened 2018-10-29
drop support for gQUIC

marten-seemann:
Should be merged after #1558, #1559.

@lucas-clemente: Please also have a look at the new README.


example/client/main.go
```diff
@@ -30,9 +29,6 @@ func main() {
 	logger.SetLogTimeFormat("")
 
 	versions := protocol.SupportedVersions
```
lucas-clemente: I'm guessing we could delete this line now.

example/main.go
```diff
@@ -136,9 +135,6 @@ func main() {
 	logger.SetLogTimeFormat("")
 
 	versions := protocol.SupportedVersions
```
lucas-clemente: Same here.

internal/handshake/crypto_setup_tls.go
```diff
@@ -46,7 +46,7 @@ func (m messageType) String() string {
 	}
 }
 
-type cryptoSetupTLS struct {
```
lucas-clemente: Rename the file too.

PR #1576, opened 2018-11-02
fix go routine leak in crypto setup test

marten-seemann:


marten-seemann:
Merging without review, since this is blocking the other PRs.

Issue #1563, opened 2018-10-31
use counts for stream limits


marten-seemann:
https://github.com/quicwg/base-drafts/pull/1906/files

PR #1587, opened 2018-11-08
fix stream limit increase in the case of many open streams

marten-seemann:


marten-seemann:
The fix for #1563 will make this obsolete.

PR #1593, opened 2018-11-10
use more reasonable flow control parameters

marten-seemann:
1. We used to have different values for the maximum flow control values for server and client. This is kind of arbitrary, and very dependent on the use case. This PR removes the distinction between server and client. Applications that need to set different values, can do so via the `quic.Config`.
2. This PR also increases the initial stream flow control window from 32 kB to 512 kB (and equivalently for the connection flow control window). This will allow us to better use the BDP of a connection, without having to wait for several RTT to increase the flow control window.

### DIFF
--- a/internal/protocol/server_parameters.go
+++ b/internal/protocol/server_parameters.go
@@ -24,25 +24,21 @@ const InitialCongestionWindow ByteCount = 32 * DefaultTCPMSS
 // session queues for later until it sends a public reset.
 const MaxUndecryptablePackets = 10
 
-// InitialMaxStreamData is the stream-level flow control window for receiving data
-// This is the value that Google servers are using
-const InitialMaxStreamData = (1 << 10) * 32 // 32 kB
-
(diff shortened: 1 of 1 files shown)


internal/protocol/server_parameters.go
```diff
@@ -24,33 +24,25 @@ const InitialCongestionWindow ByteCount = 32 * DefaultTCPMSS
 // session queues for later until it sends a public reset.
 const MaxUndecryptablePackets = 10
 
+// ConnectionFlowControlMultiplier determines how much larger the connection flow control windows needs to be relative to any stream's flow control window
+// This is the value that Chromium is using
+const ConnectionFlowControlMultiplier = 1.5
+
 // InitialMaxStreamData is the stream-level flow control window for receiving data
 // This is the value that Google servers are using
```
lucas-clemente: Maybe remove this comment line now? ;) Same below.

PR #1620, opened 2018-11-22
Add Go module support

### LINKED ISSUE #1547: Add Go modules support
Add go modules support as a part of Go 1.11. Easy setup without GOPATH.

xyproto:
* Add `go.mod` and `go.sum`
* Use `go mod vendor` to vendor dependencies
* Should resolve #1547
* All tests pass


marten-seemann:
Why does this add almost half a million LOC to the code base?

xyproto:
@marten-seemann Because the dependencies are vendored.

marten-seemann:
We're already vendoring our dependencies right now, so I'd like to understand where these *additional* 500k LOC come from.

marten-seemann:
Neither. We currently have our complete set of dependencies vendored, so I don't expect adding Go mod support to add anything.

xyproto:
The dependencies were also updated with `go get -u`, if that helps explain things?

marten-seemann:
No way this could have added 500k.

xyproto:
Well, then I'm afraid your expectations do not match what actually happened.

xyproto:
```
% go version
go version go1.11.2 linux/amd64
```

lucas-clemente:
I think the difference here is that you're also vendoring the test-only deps (I see ginkgo in there), which we currently aren't doing.

Anyway, this might be a misunderstanding of go modules, but why is vendoring even required at all when using modules?

xyproto:
@lucas-clemente

Thank you for a measured response.

I have not specifically vendored the test dependencies, this was done automatically as part of `go mod init` / `go get -u` / `go mod vendor`.

Vendoring in general is not required, for any project. It's a matter of preference.

Some projects like to be able to build and test everything, even if dependencies are unavailable or projects are abandoned and removed from the web in the future, while other projects prefer not to.

If you prefer not to vendor dependencies, I can update this pull request to only add `go.mod` and `go.sum`.

xyproto:
Got no reply in 14 days. Closing pull request.

Issue #1335, opened 2018-05-02
implement header encryption


marten-seemann:
The QUIC WG decided to move forward with packet number encryption: https://github.com/quicwg/base-drafts/pull/1079

PR #1626, opened 2018-11-26
move header parsing to the session

marten-seemann:
Merge after #1624, #1625 and #1627.

This is a prerequisite for #1335. Only the session has the keys for unprotecting the header.


client.go
```diff
@@ -369,7 +369,7 @@ func (c *client) handleVersionNegotiationPacket(hdr *wire.Header) error {
 
 func (c *client) handleRetryPacket(hdr *wire.Header) {
 	c.logger.Debugf("<- Received Retry")
-	hdr.Log(c.logger)
+	(&wire.ExtendedHeader{Header: *hdr}).Log(c.logger)
```
marten-seemann: I don't really like this, but I don't see a better solution.

PR #1676, opened 2018-12-18
only copy stream data to write when popping a STREAM frame

### LINKED ISSUE #1675: don't copy the whole slice on stream.Write
Currently, we copy the whole slice on `stream.Write`. It should be sufficient to just copy on `stream.getDataForWriting`.

marten-seemann:
Fixes #1675.

`stream.Write` can be called with arbitrarily large slices of data. We should avoid copying all that data up front since this can take a long time. Instead, we can copy the data that is dequeued when a STREAM frame is popped.

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -95,8 +95,7 @@ func (s *sendStream) Write(p []byte) (int, error) {
 		return 0, nil
 	}
 
-	s.dataForWriting = make([]byte, len(p))
-	copy(s.dataForWriting, p)
+	s.dataForWriting = p
 	s.sender.onHasStreamData(s.streamID)
 
 	var bytesWritten int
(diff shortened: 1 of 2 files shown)


send_stream.go
```diff
@@ -202,10 +201,12 @@ func (s *sendStream) getDataForWriting(maxBytes protocol.ByteCount) ([]byte, boo
...
 	var ret []byte
 	if protocol.ByteCount(len(s.dataForWriting)) > maxBytes {
-		ret = s.dataForWriting[:maxBytes]
+		ret = make([]byte, int(maxBytes))
+		copy(ret, s.dataForWriting[:maxBytes])
 		s.dataForWriting = s.dataForWriting[maxBytes:]
 	} else {
-		ret = s.dataForWriting
+		ret = make([]byte, len(s.dataForWriting))
+		copy(ret, s.dataForWriting)
```
lucas-clemente: Move this above the if with a min in the length?

marten-seemann: I thought about that, but I don't think it makes the code more legible. Probably the opposite.

send_stream.go
```diff
@@ -202,10 +201,12 @@ func (s *sendStream) getDataForWriting(maxBytes protocol.ByteCount) ([]byte, boo
 
 	var ret []byte
 	if protocol.ByteCount(len(s.dataForWriting)) > maxBytes {
-		ret = s.dataForWriting[:maxBytes]
+		ret = make([]byte, int(maxBytes))
```
VictoriaRaymond: It would be nice if the `make([]byte)` call can be replaced with `getPacketBuffer()`. I guess this can be done later.

marten-seemann: Funny, I just discussed this with @lucas-clemente half an hour ago. The complicated part about this when to return the buffer. I opened #1677 to track this.

PR #1686, opened 2018-12-25
update qtls

marten-seemann:
I rebased the *qlts* branch in https://github.com/marten-seemann/qtls onto the current release of tris. This pulls in the SIDH changes, which unfortunately means a lot of additional vendored code.


marten-seemann:
Note that the additional code is only temporary. Once Go 1.12 is released, I can rebase my QUIC related TLS changes onto the standard library TLS code, and we can get rid of tris.

Issue #1691, opened 2018-12-28
implement persistent congestion detection


marten-seemann:
There are still a few issues in the draft, e.g. https://github.com/quicwg/base-drafts/issues/2243 and https://github.com/quicwg/base-drafts/issues/2255.

PR #1692, opened 2018-12-28
unify TLP and RTO

### LINKED ISSUE #1673: unify TLP and RTO
see https://github.com/quicwg/base-drafts/pull/2114

marten-seemann:
Fixes #1673.

This doesn't yet implement the persistent congestion detection (#1691), since the specification of that still is in flux.


internal/ackhandler/sent_packet_handler.go
```diff
@@ -56,17 +47,13 @@ type sentPacketHandler struct {
...
 	cryptoCount uint32
-
-	// The number of times a TLP has been sent without receiving an ack.
-	tlpCount uint32
-	allowTLP bool
-
-	// The number of times an RTO has been sent without receiving an ack.
-	rtoCount uint32
+	// The number of times a PTO has been sent without receiving an ack.
+	ptoCount uint32
```
lucas-clemente: Can you rename this and the following var to make their semantics more clear?

ptoCount and numPTOs next to each other are hard to understand.

PR #1706, opened 2019-01-02
add an integration test for tail loss recovery

### LINKED ISSUE #1434: add integration tests that trigger PTOs
Currently, none of our integration tests triggers an RTO (a verified RTO, not a spurious one).

marten-seemann:
Fixes #1434.

### DIFF
--- a/integrationtests/self/deadline_test.go
+++ b/integrationtests/self/deadline_test.go
@@ -1,4 +1,4 @@
-package self
+package self_test
 
 import (
 	"crypto/tls"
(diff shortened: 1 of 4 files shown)


integrationtests/self/client_test.go
```diff
@@ -1,4 +1,4 @@
-package self_test
```
lucas-clemente: Why are you dropping the _test suffix? I think go tools (mod, vendor) treat test packages differently.

Now that I think about it, maybe missing _test suffixes were the problem with the vendoring of test-only files?

marten-seemann: Then we probably should rename everything to `self_test` instead of `self`.

PR #1724, opened 2019-01-07
implement the ack_delay_exponent

### LINKED ISSUE #975: implement the ack_delay_exponent transport parameter
I'm not exactly sure when this was added, but there's a new transport parameter for encoding the ACK delay exponent: https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#rfc.section.7.4.1. We will need this when we support the new varint frame types (#967).

marten-seemann:
Fixes #975.

### DIFF
--- a/internal/handshake/transport_parameters.go
+++ b/internal/handshake/transport_parameters.go
@@ -25,6 +25,7 @@ const (
 	initialMaxStreamDataUniParameterID        transportParameterID = 0x7
 	initialMaxStreamsBidiParameterID          transportParameterID = 0x8
 	initialMaxStreamsUniParameterID           transportParameterID = 0x9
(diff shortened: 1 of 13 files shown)


internal/handshake/transport_parameters.go
```diff
@@ -147,6 +151,11 @@ func (p *TransportParameters) readNumericTransportParameter(
 			return fmt.Errorf("invalid value for max_packet_size: %d (minimum 1200)", val)
 		}
 		p.MaxPacketSize = protocol.ByteCount(val)
+	case ackDelayExponentParameterID:
+		if val > 20 {
```
lucas-clemente: Make the magic number a constant?

Issue #1709, opened 2019-01-04
Multiplexer never lets go of old connections --> memory leak


Eichhoernchen:
When establishing many connections as a client or when running for a long time, the connection multiplexer consumes more and more memory.

New packet conns are added to the multiplexer but are never let go again even if the underlying socket is closed.
Probably one would like to cleanup the multiplexer when the underlying packetHandlerMap has no connections for a period of time.

marten-seemann:
That's right, we need to clean up the map. I think the right time to do that is on `packetHandlerMap.close`. That way, the behavior would be completely deterministic.

Eichhoernchen:
I tried out your proposed solution, this fixes my memory leak.
I added a ref counter to the `connManager` which I think is required to not remove a multiplexed connection and further added a `RemConn` method similar to the `AddConn` method (both accounting for the ref count) that is called from the `packetHandlerMap.close`.

If you deem this an appropriate solution I'm happy to file a PR.

marten-seemann:
I'm not sure I understand everything from your description. Please send a PR, then we can do a proper review.

marten-seemann:
Fixed by #1738.

PR #1728, opened 2019-01-08
delete unused connections from the multiplexer

Eichhoernchen:
This fixes issue #1709 by counting how many packet handlers use a packet conn and once this count reaches zero, the connection is removed from the multiplexer

### DIFF
--- a/multiplexer.go
+++ b/multiplexer.go
@@ -15,6 +15,7 @@ var (
 
 type multiplexer interface {
 	AddConn(net.PacketConn, int) (packetHandlerManager, error)
+	RemoveConn(net.PacketConn) error
 }
 
 type connManager struct {
@@ -60,4 +61,18 @@ func (m *connMultiplexer) AddConn(c net.PacketConn, connIDLen int) (packetHandle
 		return nil, fmt.Errorf("cannot use %d byte connection IDs on a connection that is already using %d byte connction IDs", connIDLen, p.connIDLen)
 	}
 	return p.manager, nil
+}
+
+func (m *connMultiplexer) RemoveConn(c net.PacketConn) error {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	_, ok := m.conns[c]
+	if !ok {
+		return fmt.Errorf("cannote remove connection, connection is unknown")
+	}
+
+	delete(m.conns, c)
+
+	return nil
 }
--- a/packet_handler_map.go
+++ b/packet_handler_map.go
@@ -139,7 +139,7 @@ func (h *packetHandlerMap) close(e error) error {
 	}
 	h.mutex.Unlock()
 	wg.Wait()
-	return nil
+	return getMultiplexer().RemoveConn(h.conn)
 }
 
 func (h *packetHandlerMap) listen() {


marten-seemann:
Hi @Eichhoernchen,
thanks for the PR. Why do you need the `useCount`? A `net.PacketConn` is only used from a single `packetHandlerMap`, so closing of the `packetHandlerMap` should be sufficient to remove it from the `multiplexer`, shouldn't it?

Eichhoernchen:
I was unsure about this, when I looked at the `AddConn` method, I figured if I called it twice with the same conn it should work, at least it won't give an error so I thought all Adds should have equal removes...

If this is not a problem, I'm happy to remove the `useCount`, but why not raise an error if the conn is already added?

marten-seemann:
`AddConn` can indeed be called multiple times with the same packet conn, and it is every time you dial on the same connection: https://github.com/lucas-clemente/quic-go/blob/6fbe8d563eca005c6faae2f39faace5c5d035d48/client.go#L117-L127
However, it's only added once: https://github.com/lucas-clemente/quic-go/blob/6fbe8d563eca005c6faae2f39faace5c5d035d48/multiplexer.go#L49-L62
One reason is that the `packetHandlerMap` listens on the packet conn, and it only makes sense to read from a connection from one go routine.

Eichhoernchen:
alright, then this seems unnecessary, i'll remove the counting part

marten-seemann:
I think this might be related to a race condition I'm seeing in the handshake integration tests from time to time. We should fix this issue asap.

Eichhoernchen:
I removed the counting part and reverted back to storing the object not a pointer.

multiplexer.go
```diff
@@ -15,6 +15,7 @@ var (
 
 type multiplexer interface {
 	AddConn(net.PacketConn, int) (packetHandlerManager, error)
+	RemConn(net.PacketConn) (error)
```
marten-seemann: better: `RemoveConn`

multiplexer.go
```diff
@@ -61,3 +62,17 @@ func (m *connMultiplexer) AddConn(c net.PacketConn, connIDLen int) (packetHandle
...
+func (m *connMultiplexer) RemConn(c net.PacketConn) (error) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	_, ok := m.conns[c]
+	if !ok {
+        	return fmt.Errorf("cannote remove connection, connection is unknown")
+	}
+
+       	delete(m.conns, c)
```
marten-seemann: There are too many spaces here. Can you please run `gofmt`?

multiplexer.go
```diff
@@ -61,3 +62,17 @@ func (m *connMultiplexer) AddConn(c net.PacketConn, connIDLen int) (packetHandle
 	}
 	return p.manager, nil
 }
+
+func (m *connMultiplexer) RemConn(c net.PacketConn) (error) {
```
marten-seemann: No brackets around `(error)`.

Eichhoernchen:
This should address your comments, thanks for brining `gofmt` to my attention🥇

marten-seemann:
Can also please run `go generate .` in the root directory? That's the reason the CIs are currently failing... This will regenerate the mock implementation of the multiplexer. It would also be great if you could squash all commits into a single commit.

Eichhoernchen:
Hm, is there a trick to `go generate`? I fetched the missing mockgen and goimports but I still get errors, I think triggered through qtls. 
```
/src/github.com/marten-seemann/qtls/13.go:24:2: cannot find package "github_com/cloudflare/sidh/sidh" in any of:
	/usr/local/Cellar/go/1.11.4/libexec/src/github_com/cloudflare/sidh/sidh (from $GOROOT)
	/var/folders/gc/vw0t42j16_76lpvsdlq6m_9r0000gn/T/tmp.853M4luX/src/github_com/cloudflare/sidh/sidh (from $GOPATH)
	/Users/jan/gopath/src/github_com/cloudflare/sidh/sidh
```

I tried to build qtls but this seems to fail on OS X...

marten-seemann:
qtls should be vendored, and the vendored version shouldn't trigger any errors.

If it's easier for you, I can manually rebase your PR and fix the mocks.

marten-seemann:
Closed in favor of #1738.

PR #1729, opened 2019-01-09
correctly set the 1-RTT header protection mask

protocol7:
Ensure that the rigth mask is picked for header protection.

### DIFF
--- a/internal/handshake/crypto_setup.go
+++ b/internal/handshake/crypto_setup.go
@@ -417,20 +417,15 @@ func (h *cryptoSetup) SetReadKey(suite *qtls.CipherSuite, trafficSecret []byte) 
 	if err != nil {
 		panic(fmt.Sprintf("error creating new AES cipher: %s", err))
 	}
-	opener := newOpener(
-		suite.AEAD(key, iv),
-		hpDecrypter,
-		h.readEncLevel == protocol.Encryption1RTT,
-	)
 
 	switch h.readEncLevel {
 	case protocol.EncryptionInitial:
 		h.readEncLevel = protocol.EncryptionHandshake
-		h.handshakeOpener = opener
+		h.handshakeOpener = newOpener(suite.AEAD(key, iv), hpDecrypter, false)
 		h.logger.Debugf("Installed Handshake Read keys")
 	case protocol.EncryptionHandshake:
 		h.readEncLevel = protocol.Encryption1RTT
-		h.opener = opener
+		h.opener = newOpener(suite.AEAD(key, iv), hpDecrypter, true)
 		h.logger.Debugf("Installed 1-RTT Read keys")
 	default:
 		panic("unexpected read encryption level")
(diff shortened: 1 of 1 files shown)


protocol7:
I signed it!

protocol7:
This fixes a bug where the wrong mask is picked for header protection. I'm not sure that this would be the way to you want to fix it, so certainly wouldn't mind closing this PR for a better patch :)

marten-seemann:
Thanks for the PR. Can you please explain what you're trying to do?
The argument to `newSealer` and `newOpener` is `is1RTT`, so `encLevel == protocol.EncryptionLevel1RTT` should be correct, right? Or am I missing something?

protocol7:
When we're creating the opener/sealer for 1RTT, the state of h has not yet progressed to be encLevel == protocol.EncryptionLevel1RTT. It does that a few lines down. So that means that is1RTT will be false for the 1RTT opener/sealer before this fix. I'm not sure that this is the best fix though, as the fix creates some pretty non-obvious code.

marten-seemann:
Good catch! Of course you're right. That's highly non-obvious indeed.

Thinking about it, I'd suggest the following as the probably most readable solution:
```go
func (h *cryptoSetup) SetWriteKey(suite *qtls.CipherSuite, trafficSecret []byte) {
	key := qtls.HkdfExpandLabel(suite.Hash(), trafficSecret, []byte{}, "quic key", suite.KeyLen())
	iv := qtls.HkdfExpandLabel(suite.Hash(), trafficSecret, []byte{}, "quic iv", suite.IVLen())
	hpKey := qtls.HkdfExpandLabel(suite.Hash(), trafficSecret, []byte{}, "quic hp", suite.KeyLen())
	hpEncrypter, err := aes.NewCipher(hpKey)
	if err != nil {
		panic(fmt.Sprintf("error creating new AES cipher: %s", err))
	}

	switch h.writeEncLevel {
	case protocol.EncryptionInitial:
		h.writeEncLevel = protocol.EncryptionHandshake
		h.handshakeSealer = newSealer(suite.AEAD(key, iv), hpEncrypter, false)
		h.logger.Debugf("Installed Handshake Write keys")
	case protocol.EncryptionHandshake:
		h.writeEncLevel = protocol.Encryption1RTT
		h.sealer = newSealer(suite.AEAD(key, iv), hpEncrypter, true)
		h.logger.Debugf("Installed 1-RTT Write keys")
	default:
		panic("unexpected write encryption level")
	}
	h.receivedWriteKey <- struct{}{}
}
```

Do you want to update your PR (please squash with the last commits before submitting)?

protocol7:
Pushed an updated version based on your proposal

marten-seemann:
I was about to merge this right now, but it looks like the linter is complaining.
It's because of the tabs / spaces at the beginning of the line. Could you fix this please?

protocol7:
Sorry for having to walk me though this, but should pass now :)

marten-seemann:
No worries. Everything looks good now. Thanks for the fix!

PR #1741, opened 2019-01-16
Allow CONNECT method

klzgrad:
:path is empty in CONNECT method.


klzgrad:
It looks like I need to fix the test cases too. I can't really sign the CLA. So closing.

PR #1752, opened 2019-01-23
cancel reading from the response stream when the response body is closed

### LINKED ISSUE #1618: Stream exhaustion in h2quic client
I'm building an application with the ```h2quic.Server``` and ```h2quic.RoundTripper```.  My application has a benchmarking mode where it sends continuous requests to the server. Since ```h2quic.RoundTripper``` caches connections to the same host, it opens a new stream in the same session for every request sent. This makes the client run into a ```NetworkIdleTimeout: No recent network activity``` error after the ```DefaultMaxIncomingStreams``` of size 100 is reached. Any subsequent requests time out and return this error. It appears to me that the opened streams don't get closed after the response was received.

marten-seemann:
Fixes #1618.

In order to release the flow control credit for HTTP responses that were already sent and received, but will never be read because the `http.Response.Body` is closed, we need to call `CancelRead` on the response stream.


h2quic/response_body.go
```diff
@@ -0,0 +1,31 @@
...
+	n, err := rb.dataStream.Read(b)
+	if err == io.EOF {
+		rb.eofRead.Set(true)
+	}
+	return n, err
+}
+
+func (rb *responseBody) Close() error {
+	if !rb.eofRead.Get() {
+		rb.dataStream.CancelRead(0)
```
lucas-clemente: Should CancelRead be a noop after an EOF?

lucas-clemente: Also, we should probably document this pitfall.

marten-seemann: I'll make `CancelRead` a noop in a separate PR.

Issue #1756, opened 2019-01-23
HTTP integration test that downloads files and doesn't read them is failing


marten-seemann:

marten-seemann:
We might want to postpone this until we implement HTTP/3 (#1241). Fixing the issue is easy, but fixing the tests is a mess.

marten-seemann:
This is obsolete now that we have HTTP/3.

PR #1757, opened 2019-01-23
implement a workaround in the h2quic server that accepts data streams

marten-seemann:
The problem with #1756 was introduced in #1754. There, we changed the streams map such that a stream is not deleted from the map until it has been accepted. This fix is correct, and the problem lies (again...) in the layer violation between gQUIC's H2 mapping and the QUIC transport.
The workaround here is to accept (and immediately forget) the response stream (and all lower streams in case of reordering) just to make sure that the streams map can delete a stream. Note that in the case of reordering, lower streams will stay in the streams map until the client closes it (again, because of the layer violation this implicit via `CloseRemote`), i.e. until they're used for a request. Thus, this shouldn't open us up to a DoS vulnerability.

### DIFF
--- a/h2quic/server.go
+++ b/h2quic/server.go
@@ -54,6 +54,8 @@ type Server struct {
 
 	supportedVersionsAsString string
 
+	highestAcceptedStream protocol.StreamID
+
 	logger utils.Logger // will be set by Server.serveImpl()
 }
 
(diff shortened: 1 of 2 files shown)


marten-seemann:
We support HTTP/3, so this fix is now obsolete.

PR #1758, opened 2019-01-24
make Stream.Cancel{Read,Write} a no-op if called after close, remove return value

marten-seemann:


h2quic/response_body.go
```diff
@@ -4,28 +4,15 @@ import (
 	"io"
 
 	quic "github.com/lucas-clemente/quic-go"
-	"github.com/lucas-clemente/quic-go/internal/utils"
 )
 
 type responseBody struct {
```
lucas-clemente: You can probably drop the struct here.

marten-seemann: That doesn't work because `quic.Stream` is an interface.

PR #1760, opened 2019-01-24
internal/protocol: fix doc comment formatting

Quasilyte:
Put a space between "//" and doc comment text.

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


marten-seemann:
I don't think this needs to be fixed.

PR #1773, opened 2019-01-28
Add TLS and PeerCertificates. 

costinm:
ValidatedCerts and other TLS fields requires additional work.

qtls/common seems to have most of the information that is needed.


marten-seemann:
Outdated after #1836.

PR #1776, opened 2019-02-02
handle Retries in the session

### LINKED ISSUE #1775: use the same ClientHello after a Retry
Decided in Tokyo: https://github.com/quicwg/base-drafts/issues/2180#issuecomment-459192852

marten-seemann:
Fixes #1775.

Unfortunately, this adds a bunch of code to the session. However, it fixes two issues:
* we need to send the right error code when processing the transport parameters fails. Since they're now handled in the session, this is easy. Before, the error would be passed to the TLS stack, and sent as a TLS error.
* we now just effectively retransmit the ClientHello when receiving a Retry (#1775)

Question: Should we remove the `error` return value from `ReceivedExtensions` (which is passed to qtls)? @lucas-clemente, wdyt?


marten-seemann:
Will do so in a separate PR.

PR #1789, opened 2019-02-15
Add go mod support

anacrolix:
Adds functioning go mod support.


marten-seemann:
Thanks for the PR. Since we’re about to switch from tls-tris to a standard library based TLS implementation, this PR will be moot as soon as I merge #1780.
There’s little point in using the development version of quic-go for anything other than testing anyway, since the version number doesn’t guarantee interoperability.
As I’ve said on other issues, I’ll take care of go mod support once quic-go switches to Go 1.12. Although I won’t merge this PR now, I’m sure it will be very helpful for me then, because I can see if I’m using go mod correctly.

raulk:
@anacrolix For now we can refer to a specific working commit of quic-go from the go.mod file of https://github.com/libp2p/go-libp2p-quic-transport.

marten-seemann:
#1799 is the way to go.

PR #1794, opened 2019-02-23
move coalesced packet handling to the session

### LINKED ISSUE #1770: don't parse coalesced packet if version on the first packet is unknown

marten-seemann:
Fixes #1770.
Also fixes the problem that if a sender coalesced a lot of Initial packets with unsupported versions, we would've sent a lot of VN packets. Now we're just sending one.


server_test.go
```diff
@@ -187,55 +191,53 @@ var _ = Describe("Server", func() {
...
-				hdr: &wire.Header{
-					Type:    protocol.PacketTypeInitial,
-					Token:   []byte("foobar"),
-					Version: serv.config.Versions[0],
-				},
-				data: bytes.Repeat([]byte{0}, protocol.MinInitialPacketSize),
-			}))
-			Eventually(done).Should(BeClosed())
-		})
+		// It("passes an empty cookie to the callback, if decoding fails", func() {
```
lucas-clemente: Commented test?

marten-seemann: Fixed.

PR #1799, opened 2019-03-01
add Go mod support

### LINKED ISSUE #1547: Add Go modules support

marten-seemann:
Fixes #1547.


go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/lucas-clemente/quic-go
```
lucas-clemente: Do we have a way to detect updates in these packages? Is there a GH badge or something similar?

marten-seemann: I’m not sure yet how this is supposed to work.

Sean-Der:
woohoo, nice work @marten-seemann this is awesome thanks for moving!

PR #1817, opened 2019-03-07
rework crypto errors

### LINKED ISSUE #1804: return the TLS error locally, send the correct error code on the wire
As of #1800, we now return the TLS alert locally.
We should send the TLS alert, and return the TLS error locally.

We should probably also have an integration test for that.

marten-seemann:
Fixes #1804.

This PR:
* uses the IETF QUIC error codes (0x100 + TLS alert code) for crypto errors
* introduces a string representation for crypto errors
* makes sure that `Dial` returns the local TLS error (which is more detailed), but only sends the QUIC error code on the wire (without any message)


internal/qerr/error_codes.go
```diff
@@ -52,9 +51,10 @@ func (e ErrorCode) String() string {
 		return "PROTOCOL_VIOLATION"
 	case InvalidMigration:
 		return "INVALID_MIGRATION"
-	case CryptoError:
-		return "CRYPTO_ERROR"
 	default:
+		if e >= 0x100 && e < 0x200 {
+			return fmt.Sprintf("CRYPTO_ERROR %d", e-0x100)
```
lucas-clemente: Does it make sense to print these in decimal if the spec is hexa?

marten-seemann: TLS error codes are decimal, but it makes sense for the "unknown error code" message.

PR #1836, opened 2019-03-25
expose the tls.ConnectionState

marten-seemann:
I'm leaving the API instability warning for this in our documentation, since I'm not sure if we don't want at some time introduce a `quic.ConnectionState`, which might contain additional information, as well as a `tls.ConnectionState`.

Alternatively, we could also just change `quic.ConnectionState` to be an alias of `tls.ConnectionState`. WDYT, @lucas-clemente?

### DIFF
--- a/internal/handshake/crypto_setup.go
+++ b/internal/handshake/crypto_setup.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"unsafe"
 
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 	"github.com/lucas-clemente/quic-go/internal/qerr"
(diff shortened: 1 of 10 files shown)


internal/handshake/crypto_setup.go
```diff
@@ -541,13 +542,9 @@ func (h *cryptoSetup) GetOpener(level protocol.EncryptionLevel) (Opener, error)
...
-func (h *cryptoSetup) ConnectionState() ConnectionState {
-	connState := h.conn.ConnectionState()
-	return ConnectionState{
-		HandshakeComplete: connState.HandshakeComplete,
-		ServerName:        connState.ServerName,
-		PeerCertificates:  connState.PeerCertificates,
-	}
+func (h *cryptoSetup) ConnectionState() tls.ConnectionState {
+	cs := h.conn.ConnectionState()
+	return *(*tls.ConnectionState)(unsafe.Pointer(&cs))
```
lucas-clemente: Add a comment here explaining the rationale, and why this is the best option.

Many people audit uses of "unsafe", and it's very helpful to have local docs on that.

lucas-clemente: Add a comment here explaining the rationale, and why this is the best option.

Many people audit uses of "unsafe" thoughout their deps.

PR #1842, opened 2019-03-31
implement TLS session resumption

### LINKED ISSUE #981: implement session resumption
This will be part of the 3rd implementation draft: https://github.com/quicwg/base-drafts/wiki/Third-Implementation-Draft.

It will require a few changes to mint, see https://github.com/bifurcation/mint/issues/148.

marten-seemann:
Fixes #981.


internal/handshake/qtls.go
```diff
@@ -2,10 +2,37 @@ package handshake
 
 import (
 	"crypto/tls"
+	"unsafe"
```
lucas-clemente: Maybe refer to the unsafe.go file for documentation somewhere here.

PR #1847, opened 2019-04-01
Fix memory leak in packetHandlerMap closure

matzf:
The closure used to cleanup sessions after the ClosedSessionDeleteTimeout would keep a reference to the
packetHandlerMap, preventing it from being garbage collected until the timer actually runs. However, once the packetHandlerMap is closed, this becomes pointless.
When rapidly opening and closing many sessions, this delay in garbage collecting the packetHandlerMap led to large memory leaks.

---

It seems fairly hard to add a testcase that covers this.

In order to demonstrate that this patch is effective, I ran the following program that quickly connects to the local "example" server. Without the proposed changes, the memory usage of the program quickly shoots up. When I run it here, it seems to stabilize at around 150M (resident set size). With this PR applied, the observed memory usage stays constant (at around 10M).

Note: In our use case, we call `quic.Dial` with a somewhat fatter type for `net.PacketConn` which made the leak much more apparent.

```go
package main

import (
        "crypto/tls"
        "time"

        "github.com/lucas-clemente/quic-go/internal/testdata"
)

func main() {
        cfg := quic.Config{
                HandshakeTimeout: 5 * time.Millisecond,
        }
        tlsCfg := tls.Config{RootCAs: testdata.GetRootCA()}
        for {
                session, _ := quic.DialAddr("localhost:6121", &tlsCfg, &cfg)
                if session != nil {
                        _ = session.Close()
                }
        }
}
```

### DIFF
--- a/packet_handler_map.go
+++ b/packet_handler_map.go
@@ -33,7 +33,7 @@ type packetHandlerMap struct {
 	closed    bool
 
 	deleteRetiredSessionsAfter time.Duration
-	self                       **packetHandlerMap
+	deletionTimers             map[string]*time.Timer
 
 	statelessResetEnabled bool
 	statelessResetHasher  hash.Hash
@@ -60,11 +60,11 @@ func newPacketHandlerMap(
 		handlers:                   make(map[string]packetHandler),
 		resetTokens:                make(map[[16]byte]packetHandler),
 		deleteRetiredSessionsAfter: protocol.RetiredConnectionIDDeleteTimeout,
+		deletionTimers:             make(map[string]*time.Timer),
 		statelessResetEnabled:      len(statelessResetKey) > 0,
 		statelessResetHasher:       hmac.New(sha256.New, statelessResetKey),
 		logger:                     logger,
 	}
-	m.self = &m
 	go m.listen()
 	return m
 }
@@ -90,13 +90,12 @@ func (h *packetHandlerMap) Retire(id protocol.ConnectionID) {
 }
 
 func (h *packetHandlerMap) retireByConnectionIDAsString(id string) {
-	weakPtr := h.self
-	time.AfterFunc(h.deleteRetiredSessionsAfter, func() {
-		m := *weakPtr
-		if m != nil {
-			m.removeByConnectionIDAsString(id)
-		}
+	timer := time.AfterFunc(h.deleteRetiredSessionsAfter, func() {
+		h.removeByConnectionIDAsString(id)
 	})
+	h.mutex.Lock()
+	h.deletionTimers[id] = timer
+	h.mutex.Unlock()
 }
 
 func (h *packetHandlerMap) AddResetToken(token [16]byte, handler packetHandler) {
@@ -152,7 +151,10 @@ func (h *packetHandlerMap) close(e error) error {
 		return nil
 	}
 	h.closed = true
-	*h.self = nil
+
+	for _, timer := range h.deletionTimers {
+		timer.Stop()
+	}
 
 	var wg sync.WaitGroup
 	for _, handler := range h.handlers {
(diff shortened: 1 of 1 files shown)


marten-seemann:
I’m not convinced this needs to be fixed at all, since it’s not an actual memory leak, and the timer just runs for 5 seconds.

If we actually wanted to do something about this, I think that canceling the the AfterFuncs would be the cleaner solution.

matzf:
Changed it to stop the timers instead.

In v0.10.1 the timeout is 1 minute, so the issue was much more apparent there. 
While not a "traditional" leak, it's still possible to open and close many connections even in only 5 seconds.

matzf:
As it's somewhat related, I'll ask here instead of creating a separate issue: would you consider creating a v0.10.2 version with some already merged patches for other memory leaks and this PR? Please see [here](https://github.com/matzf/quic-go/compare/release-0.10...matzf:fixes-0.10.1) for the commits that we applied to use v0.10.1 without significant leaks.

matzf:
Of course! :man_facepalming: Sorry, fixed.

packet_handler_map.go
```diff
@@ -86,10 +87,10 @@ func (h *packetHandlerMap) Retire(id protocol.ConnectionID) {
 }
 
 func (h *packetHandlerMap) retireByConnectionIDAsString(id string) {
+	h.mutex.Lock()
```
marten-seemann: I don't think it's necessary to move the locking of the mutex here.

matzf: The idea was that this would avoid a potential race condition; if `h.deleteRetiredSessionsAfter` is 0, the remove-function could run immediately and execute `delete(h.deletionTimers, id)` *before* the insertion happens, and therefore leak memory. Also, I can't think of any downsides to this; creating the timer should be fairly quick, so the lock is not held significantly longer.
If you're not convinced, I'm happy to move this back though.

marten-seemann: You’re right, thanks for the explanation. Since this is not obvious, could you maybe add a comment explaining why you lock the mutex there already?

marten-seemann:
@matzf In order to get this PR merged, can you also please take care of the CLA (see https://github.com/lucas-clemente/quic-go/pull/1847#issuecomment-478486558)?

marten-seemann:
@matzf I'm planning to cut a new quic-go release **very soon**, if you want this PR included there, can  you please make sure this is mergeable asap?

packet_handler_map.go
matzf: Done.

matzf:
I've added a comment to explain the locking.
Unfortunately, I'm still blocked trying to figure out the internal procedure here to accept the CLA. Sorry.

marten-seemann:
Then this PR will have to wait for the next release.

marten-seemann:
@matzf Any progress on the CLA?

matzf:
@marten-seemann Sorry for the delay. Unfortunately the lawyers here decided that I cannot accept the CLA. They proposed that I could upload the patch to our own public platform (the [ETH Data Archive](https://www.library.ethz.ch/en/ms/Research-Data-Management-and-Digital-Curation/ETH-Data-Archive)), with the copyright by ETH under MIT License. You could then download and apply the patch (*). Would this be a workable (one time) solution for you? Otherwise I'll close this PR and just report this as an issue. *sigh*

(*) I don't actually understand the point of this exercise.

matzf:
@marten-seemann ^ Could this work for you?

lucas-clemente:
Hey Matthias,

the reason for the CLA mess is that we're hoping to eventually contribute quic-go into the Go standard library. For that to be an option, Google lawyercats tell me that we need to have everybody sign the CLA. My understanding (IANAL) is that that's required to be able to relicense it, if that becomes required at any point.

To be frank: I don't feel comfortable judging whether the exercise you describe above meets those requirements, so I'm making the judgement call that we'd rather not risk it.

I'm really sorry about that, and the time you put into this. I know how frustrating this probably sounds, but I don't see an easy way out here :(

PR #1852, opened 2019-04-06
implement HTTP/3

### LINKED ISSUE #1241: implement HTTP/3
It appears that the HTTP/2 implementation implements an earlier version of quic's HTTP mapping. Was curious what the timeline/blockers are for moving to the IETF mapping?

marten-seemann:
Fixes #1241.

There are still a few TODOs, especially around the error handling (which is still being worked on in the spec) and DoS protection.

@lucas-clemente Please add the HTTP/3 tag to our repo.


http3/body.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	var n int
+	var err error
+	if r.bytesRemainingInFrame < uint64(len(b)) {
+		n, err = r.str.Read(b[:r.bytesRemainingInFrame])
+	} else {
+		n, err = r.str.Read(b)
+	}
+	r.bytesRemainingInFrame -= uint64(n)
```
lucas-clemente: Is there a risk in doing this before we check the error value?

Right now we don't have any temporary errors I believe, but if we added those in the future there could be an underflow here I imagine?

http3/request_writer.go
```diff
@@ -0,0 +1,312 @@
...
+	hf.Write(buf)
+	if _, err := io.Copy(buf, w.headerBuf); err != nil {
+		return nil, err
+	}
+	w.headerBuf.Reset()
+	return buf.Bytes(), nil
+}
+
+func (w *requestWriter) sendRequestBody(req io.ReadCloser, str quic.Stream) error {
+	b := make([]byte, 8*1024)
```
lucas-clemente: Are the 8k taken from the spec?

marten-seemann: The spec doesn't give any guidance about this. Ideally, we'd have the request body as a `[]byte`, and just send a single DATA frame. However, `http.Request.Body` is a `io.ReadCloser`, so we need to copy the data and frame it. 8k is just a number I made up.

http3/body.go
marten-seemann: `io.Reader` says:
>  It returns the number of bytes read (0 <= n <= len(p)) and any error encountered.

So I believe this should be safe with any `io.Reader`.

PR #1859, opened 2019-04-10
Bugfixed:Modify commment of h2quic.client

fengjinchen:


fengjinchen:
Some checks were not successful,Try again?

PR #1860, opened 2019-04-10
Bugfixed:Modify commment of h2quic.client

fengjinchen:


marten-seemann:
I don’t think this is a bug, neither does it need fixing. Furthermore, this package is going to be replaced by HTTP/3 very soon.

PR #1879, opened 2019-04-17
replace aes12 with aes standard lib,fix #1877

### LINKED ISSUE #1877: gquic:replace aes12 to standard aes lib
the standard lib crypto/cipher  applies an export method cipher.NewGCMWithTagSize, we can replace aes12 with standard lib now

kordenlu:
replace aes12 with aes standard lib


kordenlu:
I signed it!

PR #1886, opened 2019-04-26
fix support for gQUIC 44

marten-seemann:
No idea why the integration test so far wasn't failing.
Maybe Google changed something in the wire format after rolling out the experiment.


marten-seemann:
Blocked by #1897.

PR #1891, opened 2019-05-03
Add function to use *tls.Config with HTTP3 ListenAndServe

alexandrestein:
Adds **http3.ListenAndServeTLSConfig** function which is a clone of **http3.ListenAndServe** where ***tls.Config** struct is used instead of **certFile, keyFile _string_** to start a http3.Server.

My goal is to have full access to the *tls.Config capabilities.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -293,6 +293,13 @@ func ListenAndServe(addr, certFile, keyFile string, handler http.Handler) error 
 		Certificates: certs,
 	}
 
+	return ListenAndServeTLSConfig(addr, config, handler)
+}
+
+// ListenAndServeTLSConfig does the same as ListenAndServe (listens on the
+// given network address for both, TLS and QUIC connetions in parallel)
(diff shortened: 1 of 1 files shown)


marten-seemann:
Thanks for the PR. In the HTTP3 package, we’re following the Go standard library as closely as possible. The Go standard library way to specify the tls.Config is to setup the http.Server directly.

alexandrestein:
I see your point. 

But in the case I want to use TLS and quic in the same time with a complex *tls.Config (client auth for specific hosts, using autocert and other configurations...)

I may copy what you made in this function.
Open the UDP and TCP ports, init the corresponding goroutines and setup everything which is here.

I completely see your point; you are trying to keep as close as possible to standard http package.
But the two stack protocol make sense to me to slightly drift from the standard package in this particular case.

I hop to see this evolve in the future maybe 😊

Thank you for all the work done.

PR #1904, opened 2019-05-14
http3/request: Fix URL parsing of leading double slashes after authority

ironsteel:
This is the same fix merged  into the gquic branch, see https://github.com/lucas-clemente/quic-go/pull/1898


marten-seemann:
Thank you!

ironsteel:
No problem! Can I ask a question, I know that this is not the right place but here goes: 
Will there be any video recordings of the next QUIC WG interop next week?

marten-seemann:
There’s no video recording of the interops, but you can remote participate
in the interim meeting itself.

PR #1908, opened 2019-05-20
add quic-trace support

marten-seemann:


internal/ackhandler/interfaces.go
```diff
@@ -39,6 +39,9 @@ type SentPacketHandler interface {
 
 	GetAlarmTimeout() time.Time
 	OnAlarm() error
+
+	// report some congestion statistics. For tracing only.
+	GetCurrentState() *State
```
lucas-clemente: Maybe call this stats instead, to make it more obvious that this is informational only?

quictrace/interface.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/lucas-clemente/quic-go/internal/ackhandler"
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/wire"
+)
+
+// A Tracer traces a QUIC connection
+type Tracer interface {
+	Trace(protocol.ConnectionID, Event)
+	EmitAll() map[string][]byte
```
lucas-clemente: What does this method do?

internal/ackhandler/interfaces.go
marten-seemann: Done.

quictrace/interface.go
marten-seemann: It returns all traces. I'll rename it to `GetAllTraces`, and document it better.

PR #1930, opened 2019-05-29
introduce invalid stream ID to simplify the outgoing streams map

marten-seemann:


internal/protocol/stream_id.go
```diff
@@ -1,7 +1,11 @@
 package protocol
 
 // A StreamID in QUIC
-type StreamID uint64
+type StreamID int64
```
lucas-clemente: What if the peer sends us a number that overflows int64? I think with this we need to validate that it fits on every cast, e.g. [here](https://github.com/lucas-clemente/quic-go/blob/master/internal/wire/stream_frame.go#L38).

marten-seemann: Stream IDs are always sent as varints, so the maximum value is MAX_UINT62.

lucas-clemente: Ah, okay. I was confused by the readvarint fn returning an uint64.

PR #1933, opened 2019-05-30
return OpenStreamSync calls in the order they were called

### LINKED ISSUE #1927: make OpenStreamSync() calls return streams in the order they were called
We're currently using a `sync.Cond`, and doing a `Broadcast()` when the stream limit is increased. This will randomly select any of the go routines that are blocking on the cond.
We should return `OpenStreamSync()` calls FIFO.

marten-seemann:
Fixes #1927. Depends on #1930.

### DIFF
--- a/internal/protocol/stream_id.go
+++ b/internal/protocol/stream_id.go
@@ -5,7 +5,7 @@ type StreamID int64
 
 // InvalidPacketNumber is a stream ID that is invalid.
 // The first valid stream ID in QUIC is 0.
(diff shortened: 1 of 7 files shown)


marten-seemann:
This PR is inherently racy. When calling `OpenStream` in between receiving a MAX_STREAM_ID frame and before all `OpenStreamSync` calls have returned, `OpenStream` might snatch a stream, and one of the `OpenStreamSync` calls might error.

PR #1952, opened 2019-06-10
add contexts to all blocking functions

### LINKED ISSUE #1915: No way to set a timeout/deadline for Listener.Accept()
Other than closing a connection, which makes it tricky to handle shutdowns.
Ideally Accept/Listen should take a context.

marten-seemann:
Fixes #1915.

@AudriusButkevicius What do you think?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,7 +8,7 @@
 - Enforce application protocol negotiation (via `tls.Config.NextProtos`).
 - Use a varint for error codes.
 - Add support for [quic-trace](https://github.com/google/quic-trace).
-- Add a context to `Listener.Accept`.
+- Add a context to `Listener.Accept` and `Session.Accept{Uni}Stream`.
 
 ## v0.11.0 (2019-04-05)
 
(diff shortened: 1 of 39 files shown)


AudriusButkevicius:
Looks good, yet I wonder for the benefits of this for the async stuff.

marten-seemann:
I'm not sure I understand what you mean. Can you elaborate?

AudriusButkevicius:
You've added context support for non-blocking openstream et al, I wonder the benefit of those.

marten-seemann:
I added a context for `Open{Uni}StreamSync`, which is blocking if the endpoint reached the peers stream limit. `Open{Uni}Stream` is non-blocking (and returns an error when no new stream can be opened at that moment).

AudriusButkevicius:
But Open[Uni]Stream would never block right? What would be the purpose of of the context in that case? I guess cancellation?

marten-seemann:
`Open{Uni}Stream` never blocks, and it doesn't take a context.

AudriusButkevicius:
Sorry, I was looking at accept which does block and now does take a context. Lgtm.

marten-seemann:
Ok, thanks for the feedback!

PR #1983, opened 2019-07-02
don't pace when congestion limited

marten-seemann:


marten-seemann:
#2615 is the better solution for this.

PR #1993, opened 2019-07-06
allow host without port passed as 'host' argument in Dial function.

spacewander:
Previously, if the given host doesn't contain port, dial with it will
result in error "missing port in address".

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 	"sync"
 
 	"github.com/lucas-clemente/quic-go/internal/handshake"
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -151,11 +152,16 @@ func newClient(
 		tlsConf = &tls.Config{}
 	}
 	if tlsConf.ServerName == "" {
-		var err error
-		tlsConf.ServerName, _, err = net.SplitHostPort(host)
-		if err != nil {
-			return nil, err
+		sni := host
+		if strings.IndexByte(sni, ':') != -1 {
```
marten-seemann: Why not `strings.Contains`?

spacewander: @marten-seemann 
`IndexByte` could be faster, so I prefer it to `Contains` when the target is only a single character.

Just wrote a benchmark:
```go
package main
import (
	"testing"
	"strings"
)

func BenchmarkMethod(b *testing.B) {
	var ok bool
	s := strings.Repeat("0", 10)
	for i := 0; i < b.N; i++ {
		//ok = strings.Contains(s, "a")
		ok = strings.IndexByte(s, 'a') != -1
	}
	println(ok)
}
```

`IndexByte` is 10% faster in my machine.

spacewander:
CLA signed.

client.go
marten-seemann: Fair enough. I don't think it matters, since this is only executed a single time per connection. But if you insist, I'm ok with keeping `IndexByte`.

PR #2003, opened 2019-07-16
update the initial salt for draft-22

marten-seemann:
Unfortunately, the draft didn't contain any new test vectors, so I had to disable a few unit tests.


marten-seemann:
Thank you @Lekensteyn.

PR #2023, opened 2019-07-27
implement sending of 0.5-RTT data

marten-seemann:


interface.go
```diff
@@ -171,6 +171,16 @@ type Session interface {
 	ConnectionState() tls.ConnectionState
 }
 
+// An EarlySession is a session that is handshaking.
```
lucas-clemente: Document or at least link to the security guarantees here.

server.go
```diff
@@ -86,19 +88,41 @@ type server struct {
...
 
 	logger utils.Logger
 }
 
-var _ Listener = &server{}
-var _ unknownPacketHandler = &server{}
+var _ Listener = &baseServer{}
+var _ unknownPacketHandler = &baseServer{}
+
+type earlyServer struct{ *baseServer }
```
lucas-clemente: Maybe my Go is rusty, but can't this just be an alias?

server.go
```diff
@@ -126,7 +150,17 @@ func Listen(conn net.PacketConn, tlsConf *tls.Config, config *Config) (Listener,
...
 }
 
-func listen(conn net.PacketConn, tlsConf *tls.Config, config *Config) (*server, error) {
+// ListenEarly works like Listen, but it returns sessions before the handshake completes.
+func ListenEarly(conn net.PacketConn, tlsConf *tls.Config, config *Config) (EarlyListener, error) {
+	s, err := listen(conn, tlsConf, config)
+	if err != nil {
+		return nil, err
+	}
+	s.acceptEarlySessions = true
```
lucas-clemente: Is there a race condition here if a client connects before this is set?

marten-seemann: Right, this should better be a parameter to `listen`.

server.go
marten-seemann: I think I need a struct here to define the `Accept()` method, don't I?

PR #2024, opened 2019-07-28
remove complex checks performed on the tls.Config

marten-seemann:
There are just too many option. For example, for the server, it should be valid to set `Certificates`, `GetCertificate` or `GetConfigForClient` (which then has to set `Certificates` or `GetCertificate` for every single call).


lucas-clemente:
This one feels easier to miss than the certs, but I'll leave it up to you.

lucas-clemente:
This one feels easier to miss than the certs, but I'll leave it up to you.

PR #2029, opened 2019-07-30
 never keep track of more than 500 ACK ranges 

### LINKED ISSUE #2010: What is likely cause of "Too many outstanding received ACK ranges" error?
I got this error on v0.11.2 after leaving up a quic connection for about 70hours.  Not much was going on over the connection.  Both hosts sitting in AWS EC2.   Could this error indicate a transient network connectivity issue, or is it more likely to be misconfiguration?  Is the appropriate remediation to just open a new stream?

 I'm attempting to use quic for  long running connections in a VPN-style setup. I'm bringing up a single stream on a session at the start and just treating it like a socket.  I know this is sub-optimal, and I will switch to using multiple shorter lived streams.  But this approach allowed me to simply drop quic in where TCP used to be, and I figured that this setup would be ok for initial testing.  Am I already running into trouble by using this approach I wonder?

Thanks for any tips!

marten-seemann:
Fixes #2010. Builds on top of #2028.

### DIFF
--- a/internal/ackhandler/received_packet_history.go
+++ b/internal/ackhandler/received_packet_history.go
@@ -22,28 +22,25 @@ func newReceivedPacketHistory() *receivedPacketHistory {
 }
 
 // ReceivedPacket registers a packet with PacketNumber p and updates the ranges
-func (h *receivedPacketHistory) ReceivedPacket(p protocol.PacketNumber) error {
+func (h *receivedPacketHistory) ReceivedPacket(p protocol.PacketNumber) {
 	// ignore delayed packets, if we already deleted the range
 	if p < h.deletedBelow {
-		return nil
+		return
 	}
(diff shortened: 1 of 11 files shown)


internal/ackhandler/received_packet_history.go
```diff
@@ -75,10 +75,15 @@ func (h *receivedPacketHistory) ReceivedPacket(p protocol.PacketNumber) error {
 
 	// create a new range at the beginning
 	h.ranges.InsertBefore(utils.PacketInterval{Start: p, End: p}, h.ranges.Front())
-
 	return nil
 }
 
+func (h *receivedPacketHistory) maybeDeleteOldRanges() {
```
lucas-clemente: Leave a comment here that this is a DOS defense?

marten-seemann: Done.

PR #2033, opened 2019-08-01
Successful early-handshake injection attacks

tatianab:
I added a number of tests that show how an adversary who can inject packets before real packets arrive can cause a connection to fail during the handshake. The ultimate goal is of course to find suitable mitigations to these attacks, but the first step is to show that they exist.

I'm very open to feedback and suggestions, as I am new to this codebase and to open source contributing.

Note that the github diff is not working well for the file mitm_test.go because I indented existing code. To make this easier to read, you can look at the diff between either of the commits "added successful pre-handshake injection attacks" and the previous commit.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -663,4 +663,53 @@ var _ = Describe("Client", func() {
 		Expect(cl.version).ToNot(BeZero())
 		Expect(cl.GetVersion()).To(Equal(cl.version))
 	})
+
+	Context("handling potentially injected packets", func() {
+		// NOTE: We hope these tests as written will fail once mitigations for injection adversaries are put in place.
+
+		// Illustrates that adversary who injects any packet quickly can
+		// cause a real version negotiation packet to be ignored.
+		It("version negotiation packets ignored if any other packet is received", func() {
+			// Copy of existing test "recognizes that a non Version Negotiation packet means that the server accepted the suggested version"
+			sess := NewMockQuicSession(mockCtrl)
+			sess.EXPECT().handlePacket(gomock.Any())
+			cl.session = sess
+			cl.config = &Config{}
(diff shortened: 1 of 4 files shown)


session_test.go
```diff
@@ -1681,4 +1682,100 @@ var _ = Describe("Client Session", func() {
...
+				encryptionLevel: protocol.EncryptionInitial,
+				hdr:             hdr1,
+				data:            []byte{0}, // one PADDING frame
+			}, nil)
+			Expect(sess.handlePacketImpl(getPacket(hdr1, nil))).To(BeTrue())
+			// The next packet has to be ignored, since the source connection ID doesn't match.
+			Expect(sess.handlePacketImpl(getPacket(hdr2, nil))).To(BeFalse())
+		})
+
+		// Illustrates that an injected Initial with an ACK frame for an unsent causes
```
marten-seemann: unsent packet

client_test.go
```diff
@@ -668,4 +668,53 @@ var _ = Describe("Client", func() {
...
+			// Version negotiation is now ignored
+			cl.config = &Config{}
+			ver := cl.version
+			cl.handlePacket(composeVersionNegotiationPacket(connID, []protocol.VersionNumber{1234}))
+			Expect(cl.version).To(Equal(ver))
+		})
+
+		// Illustrates that adversary that injects a version negotiation packet
+		// with no supported versions can break a connection.
+		It("connection fails if no supported versions are found in version negotation packet", func() {
```
marten-seemann: Is this just a copy of the existing test? Wouldn't it be better to just add the comment to the original test then?

tatianab:
Thanks! I don't have write access, could you merge this in?

PR #2042, opened 2019-08-08
implement a store for address validation tokes

### LINKED ISSUE #2040: save and use tokens

marten-seemann:
Fixes #2040.


interface.go
```diff
@@ -27,6 +27,23 @@ type Token struct {
...
 
+// A ClientToken is a token received by the client.
+// It can be used to skip address validation on future connection attempts.
+type ClientToken struct {
+	data []byte
+}
+
+type TokenCache interface {
+	// Get searches for a ClientToken associated with the given key.
+	// Since tokens should not be reused, it should remove the token from the cache.
```
lucas-clemente: should not or must not?

interface.go
```diff
@@ -27,6 +27,23 @@ type Token struct {
...
+	// On return, ok is true if one was found.
+	Get(key string) (token *ClientToken, ok bool)
```
lucas-clemente: Maybe call it Pop, to make it clear that this modifies the cache?

interface.go
```diff
@@ -27,6 +27,23 @@ type Token struct {
...
+// A ClientToken is a token received by the client.
+// It can be used to skip address validation on future connection attempts.
+type ClientToken struct {
+	data []byte
+}
+
+type TokenCache interface {
+	// Get searches for a ClientToken associated with the given key.
+	// Since tokens should not be reused, it should remove the token from the cache.
+	// On return, ok is true if one was found.
```
lucas-clemente: Why not just return nil instead?

interface.go
```diff
@@ -27,6 +27,23 @@ type Token struct {
 	SentTime     time.Time
 }
 
+// A ClientToken is a token received by the client.
+// It can be used to skip address validation on future connection attempts.
+type ClientToken struct {
+	data []byte
+}
+
+type TokenCache interface {
```
lucas-clemente: Be explicit that this is conceptually a multimap.

I'm also not happy about the name cache – this isn't caching, it's storing (and deleting).

interface.go
marten-seemann: https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#validate-future

> A client SHOULD NOT reuse a token in different connections. Reusing a token allows connections to be linked by entities on the network path (see Section 9.5). A client MUST NOT reuse a token if it believes that its point of network attachment has changed since the token was last used; that is, if there is a change in its local IP address or network interface

marten-seemann:
I was trying to mimic the `tls.ClientSessionCache` here, but it seems like the subtle differences (session tickets can be reused, unless you're doing 0-RTT) would justify minting a different interface.

interface.go
marten-seemann: I'll change the text so that it doesn't sound like normative language: "Since tokens are not supposed to be reused, it must remove..."

interface.go
marten-seemann: I'm not sure I understand what you mean here.

PR #2062, opened 2019-08-14
mitigation for immediate-failure injection attacks

tatianab:
Implements mitigation for the simplest class of injection attacks: attacks which cause an immediate failure, and that can be performed by injecting a single packet without keeping state.

Mitigation strategy is to simply ignore malformed packets, e.g, those that have no payload, bad frames or ACKs for unsent packets. For packets that should cause failure if received from a legitimate peer, such as version negotiation with no supported versions or CONNECTION_CLOSE frame, the connection closes if no packets are received after a timeout.

Mitigation is currently turned off by default, but can be turned on by assigning a positive value to the AttackTimeout field of a Config.

@bemasc

### DIFF
--- a/client.go
+++ b/client.go
@@ -348,14 +348,11 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
 	c.logger.Infof("Received a Version Negotiation packet. Supported Versions: %s", hdr.SupportedVersions)
 	newVersion, ok := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
 	if !ok {
-		shutdown := func(err error) {
-			c.session.destroy(err)
-		}
 		err := fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
 		if c.config.AttackTimeout <= 0 {
-			shutdown(err)
+			c.session.destroy(err)
 		} else {
-			c.session.maybeRecover(shutdown, err)
+			c.session.registerInsecureClose(&closeError{err: err})
 		}
 		return
 	}
--- a/client.go
+++ b/client.go
@@ -348,6 +348,7 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
 	c.logger.Infof("Received a Version Negotiation packet. Supported Versions: %s", hdr.SupportedVersions)
 	newVersion, ok := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
 	if !ok {
+		//nolint:stylecheck
 		err := fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
 		if c.config.AttackTimeout <= 0 {
 			c.session.destroy(err)
--- a/client.go
+++ b/client.go
@@ -215,6 +215,12 @@ func populateClientConfig(config *Config, createdPacketConn bool) *Config {
 	if config.IdleTimeout != 0 {
 		idleTimeout = config.IdleTimeout
 	}
+	attackTimeout := protocol.DefaultAttackTimeout
+	if config.AttackTimeout > 0 {
+		attackTimeout = config.AttackTimeout
+	} else if config.AttackTimeout < 0 {
+		attackTimeout = 0
+	}
 
 	maxReceiveStreamFlowControlWindow := config.MaxReceiveStreamFlowControlWindow
 	if maxReceiveStreamFlowControlWindow == 0 {
@@ -245,6 +251,7 @@ func populateClientConfig(config *Config, createdPacketConn bool) *Config {
 		Versions:                              versions,
 		HandshakeTimeout:                      handshakeTimeout,
 		IdleTimeout:                           idleTimeout,
+		AttackTimeout:                         attackTimeout,
 		ConnectionIDLength:                    connIDLen,
 		MaxReceiveStreamFlowControlWindow:     maxReceiveStreamFlowControlWindow,
 		MaxReceiveConnectionFlowControlWindow: maxReceiveConnectionFlowControlWindow,
@@ -341,9 +348,17 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
 	c.logger.Infof("Received a Version Negotiation packet. Supported Versions: %s", hdr.SupportedVersions)
 	newVersion, ok := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
 	if !ok {
-		//nolint:stylecheck
-		c.session.destroy(fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions))
-		c.logger.Debugf("No compatible QUIC version found.")
+		fail := func(err error) { c.session.destroy(err) }
+		errStr := fmt.Sprintf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
+		errImmediate := fmt.Errorf(errStr)
+		errTimeout := fmt.Errorf("Timed out: %v", errStr)
+		sess, ok := c.session.(*session)
+		if !ok { // this catches MockQuicSession
+			fail(errImmediate)
+			return
+		}
+		sess.maybeRecover(fail, errImmediate, errTimeout)
+		// c.session.maybeRecover(fail, errImmediate, errTimeout)
 		return
 	}
 	c.receivedVersionNegotiationPacket = true
--- a/client.go
+++ b/client.go
@@ -348,14 +348,14 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
 	c.logger.Infof("Received a Version Negotiation packet. Supported Versions: %s", hdr.SupportedVersions)
 	newVersion, ok := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
 	if !ok {
-		fail := func(err error) {
+		shutdown := func(err error) {
 			c.session.destroy(err)
 		}
 		err := fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
 		if c.config.AttackTimeout <= 0 {
-			fail(err)
+			shutdown(err)
 		} else {
-			c.session.maybeRecover(fail, err)
+			c.session.maybeRecover(shutdown, err)
 		}
 		return
 	}
--- a/client.go
+++ b/client.go
@@ -348,17 +348,15 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
 	c.logger.Infof("Received a Version Negotiation packet. Supported Versions: %s", hdr.SupportedVersions)
 	newVersion, ok := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
 	if !ok {
-		fail := func(err error) { c.session.destroy(err) }
-		errStr := fmt.Sprintf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
-		errImmediate := fmt.Errorf(errStr)
-		errTimeout := fmt.Errorf("Timed out: %v", errStr)
-		sess, ok := c.session.(*session)
-		if !ok { // this catches MockQuicSession
-			fail(errImmediate)
-			return
+		fail := func(err error) {
+			c.session.destroy(err)
 		}
-		sess.maybeRecover(fail, errImmediate, errTimeout)
-		// c.session.maybeRecover(fail, errImmediate, errTimeout)
+		err := fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
+		if c.config.AttackTimeout <= 0 {
+			fail(err)
+		} else {
+			c.session.maybeRecover(fail, err)
+		}
 		return
 	}
 	c.receivedVersionNegotiationPacket = true
--- a/interface.go
+++ b/interface.go
@@ -227,7 +227,7 @@ type Config struct {
 	// received without any incoming network activity, used for experimental mitigation
 	// against injection attacks. If the timeout is exceeded, the connection is closed.
 	// This timeout should be longer than any reasonable round trip time, but excessively
-	// long timeouts with delay reporting of rejected connections.
+	// long timeouts will delay reporting of rejected connections.
 	// If this value is zero or negative, mitigation will be turned off.
 	AttackTimeout time.Duration
 	// AcceptToken determines if a Token is accepted.
--- a/interface.go
+++ b/interface.go
@@ -223,6 +223,13 @@ type Config struct {
 	// If the timeout is exceeded, the connection is closed.
 	// If this value is zero, the timeout is set to 30 seconds.
 	IdleTimeout time.Duration
+	// AttackTimeout is the maximum duration that may pass after a suspicious packet is
+	// received without any incoming network activity, used for experimental mitigation
+	// against injection attacks. If the timeout is exceeded, the connection is closed.
+	// This timeout should be longer than any reasonable round trip time, but excessively
+	// long timeouts with delay reporting of rejected connections.
+	// If this value is zero or negative, mitigation will be turned off.
+	AttackTimeout time.Duration
 	// AcceptToken determines if a Token is accepted.
 	// It is called with token = nil if the client didn't send a token.
 	// If not set, a default verification function is used:
--- a/interface.go
+++ b/interface.go
@@ -228,7 +228,8 @@ type Config struct {
 	// against injection attacks. If the timeout is exceeded, the connection is closed.
 	// This timeout should be longer than any reasonable round trip time, but excessively
 	// long timeouts will delay reporting of rejected connections.
-	// If this value is zero or negative, mitigation will be turned off.
+	// If this value is zero, the default mitigation behavior will be used (currently off).
+	// If this value is negative, mitigation will be disabled.
 	AttackTimeout time.Duration
 	// AcceptToken determines if a Token is accepted.
 	// It is called with token = nil if the client didn't send a token.
--- a/internal/qerr/quic_error.go
+++ b/internal/qerr/quic_error.go
@@ -11,8 +11,7 @@ type QuicError struct {
 	ErrorMessage       string
 	isTimeout          bool
 	isApplicationError bool
-	recoverable        bool
-	ignorable          bool
+	delayed            bool
 }
 
 var _ net.Error = &QuicError{}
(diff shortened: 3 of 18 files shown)


client.go
```diff
@@ -340,8 +347,18 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
...
-		c.session.destroy(fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions))
-		c.logger.Debugf("No compatible QUIC version found.")
+		fail := func(e error) {
+			c.session.destroy(e)
+			c.logger.Debugf("No compatible QUIC version found.")
+		}
+		errImmediate := fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
+		errTimeout := fmt.Errorf("Timed out: No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
+		sess, ok := c.session.(*session)
+		if !ok { // this catches MockQuicSession
```
bemasc: How hard would it be to make `MockQuicSession` implement `maybeRecover`?  Would that allow us to remove this conditional?

interface.go
```diff
@@ -206,6 +206,11 @@ type Config struct {
 	// If the timeout is exceeded, the connection is closed.
 	// If this value is zero, the timeout is set to 30 seconds.
 	IdleTimeout time.Duration
+	// AttackTimeout is the maximum duration that may pass after a suspicious packet is
+	// received without any incoming network activity, used for experimental mitigation
+	// against injection attacks. If the timeout is exceeded, the connection is closed.
+	// If this value is zero or negative, mitigation will be turned off.
+	AttackTimeout time.Duration
```
bemasc: I would add
> This timeout should be longer than any reasonable round trip time, but excessively long timeouts will delay reporting of rejected connections.

session.go
```diff
@@ -699,9 +720,22 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
...
+	var containsIgnoredFrames bool
+	var ignorePacket bool
+
+parseLoop:
 	for {
 		frame, err := s.frameParser.ParseNext(r, packet.encryptionLevel)
 		if err != nil {
+			if s.config.AttackTimeout > 0 && packet.encryptionLevel != protocol.Encryption1RTT {
+				errStr := err.Error()
+				if strings.Contains(errStr, "FRAME_ENCODING_ERROR") {
```
bemasc: Conditioning on the error string is not ideal.  Your type assertion for `handleUnpackedPacket` is a better model, even if that needs changes to `FrameParser`.

session.go
```diff
@@ -731,6 +777,18 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
 		})
 	}
 
+	if containsIgnoredFrames && !containsValidFrames {
```
bemasc: I think it's probably simpler and more correct to reject the whole packet if any frames were rejected.  How hard would it be to ensure that within a packet either all frames are accepted or none are?

session.go
```diff
@@ -677,6 +689,10 @@ func (s *session) handleRetryPacket(p *receivedPacket, hdr *wire.Header) bool /*
 
 func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time) error {
 	if len(packet.data) == 0 {
+		// Ignore empty packets if mitigation is turned on.
+		if s.config.AttackTimeout > 0 && packet.encryptionLevel != protocol.Encryption1RTT {
+			return qerr.Error(qerr.NoError, "empty packet")
```
bemasc: Something feels strange about using `NoError` to describe these recoverable error situations, since they are in fact protocol errors.  I think you might be better off defining a new error type like `qerr.Recoverable`*, or perhaps adding an attribute like `QuicError.Ignorable`.

*Perhaps using the new [`Unwrap` pattern](https://blog.golang.org/experiment).

session.go
```diff
@@ -808,14 +866,34 @@ func (s *session) handlePacketAfterClosed(p *receivedPacket) {
...
+// or fail(errTimeout) if no packets are received before AttackTimeout elapses
+func (s *session) maybeRecover(fail func(error), errImmediate error, errTimeout error) {
+	if s.config.AttackTimeout <= 0 {
+		fail(errImmediate)
+	} else {
+		timerStart := time.Now()
+		time.AfterFunc(s.config.AttackTimeout, func() {
+			s.mutex.Lock()
+			defer s.mutex.Unlock()
+			if (s.lastPacketReceivedTime).Before(timerStart) {
```
bemasc: Nit: Unnecessary parentheses

session.go
```diff
@@ -808,14 +866,34 @@ func (s *session) handlePacketAfterClosed(p *receivedPacket) {
...
+				fail(errTimeout)
```
bemasc: `fail` runs while `s.mutex` is locked.  Is that what you want?

bemasc (changes requested):
I think this is a good start!

client.go
marten-seemann: `MockQuicSession` is [generated from](https://github.com/lucas-clemente/quic-go/blob/7df59b855f69adebbfaa95803a5976ac35cc49e0/mockgen.go#L17) `quicSession`, so you'd have to add `maybeRecover` to that interface. Generating the mocks can be done by running
```bash
GO111MODULE=off go generate .
```
I know it's messy to turn module support off for that, but mockgen is unfortunately not as smooth as I wish it would be, and it requires us to jump through a bunch of hoops to generate mocks for private methods (and even more for internal packages).

client.go
```diff
@@ -340,8 +347,18 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
...
 	newVersion, ok := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
 	if !ok {
-		c.session.destroy(fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions))
-		c.logger.Debugf("No compatible QUIC version found.")
+		fail := func(e error) {
+			c.session.destroy(e)
+			c.logger.Debugf("No compatible QUIC version found.")
+		}
+		errImmediate := fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
+		errTimeout := fmt.Errorf("Timed out: No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions)
```
lucas-clemente: Nit: Maybe avoid duplication by just printing the above error here.

session.go
```diff
@@ -151,6 +152,8 @@ type session struct {
 	receivedFirstPacket bool
 
 	sessionCreationTime time.Time
+
+	mutex sync.Mutex // protects lastPacketReceivedTime
```
lucas-clemente: The mutex will also need to be locked wherever lastPacketReceivedTime is read. It might be easier to just turn it into a function?

session.go
```diff
@@ -808,14 +866,34 @@ func (s *session) handlePacketAfterClosed(p *receivedPacket) {
 	}
 }
 
+// MaybeRecover runs fail(errImmediate) if injection attack mitigation is turned off,
+// or fail(errTimeout) if no packets are received before AttackTimeout elapses
+func (s *session) maybeRecover(fail func(error), errImmediate error, errTimeout error) {
```
lucas-clemente: The session struct has a "run loop" model and a central timer that is used to handle all wake-up things. That's more efficient, and avoids many many mutexes. Maybe you can fit this timeout into that structure too?

tatianab:
Hi Ben, Lucas, and Marten,

Thank you for your comments! I believe I have addressed all of them.

Most importantly, I overhauled the timeout logic, so that it better integrates with the runLoop structure, though feedback on it would be great as I'm not 100% confident that it fully integrates properly.

Though the default attack timeout is set to 0, the full suite of tests passes when the default is non-zero.

I also ran fuzz testing (I haven't committed the infrastructure) that sets up the MITM attack like the tests in integrationtests/self and injects correctly encrypted Initial packets with fuzzed payloads, and no crashes were found.

Please take a look when you get the chance - I appreciate your feedback!

-Tatiana

session.go
```diff
@@ -718,6 +793,51 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
...
+			parsedFirstFrame = true
+			if err := s.validateFrame(frame, packet.packetNumber, packet.encryptionLevel); err != nil {
+				quicErr, ok := err.(*qerr.QuicError)
+				if ok {
+					if quicErr.IsIgnorable() {
+						ignorableError = quicErr
+						break
+					}
+					if quicErr.IsRecoverable() {
+						recoverable = true
```
bemasc: Does this logic actually work?  It looks like a packet that generates a recoverable error is still fully processed through `s.handleFrame`, which could commit irrevocable changes to the session.

client.go
```diff
@@ -215,6 +215,12 @@ func populateClientConfig(config *Config, createdPacketConn bool) *Config {
 	if config.IdleTimeout != 0 {
 		idleTimeout = config.IdleTimeout
 	}
+	attackTimeout := protocol.DefaultAttackTimeout
+	if config.AttackTimeout > 0 {
+		attackTimeout = config.AttackTimeout
+	} else if config.AttackTimeout < 0 {
+		attackTimeout = 0
```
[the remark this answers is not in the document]
tatianab: You're right - it's a little odd. I did this so that a 0 value for AttackTimeout would indicate the default, which could be non-zero, and a negative value would indicate 0. But maybe this is fragile and could be improved by a mitigationOn flag in the config.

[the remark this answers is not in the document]
bemasc: That makes sense; it just needs to be documented.  I think a comment explaining the intended semantics would be enough.

session.go
```diff
@@ -722,6 +794,56 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
...
+	var recoverableFrame wire.Frame
+
+	if s.mitigationOn(packet.encryptionLevel) {
+		var ignorableError *qerr.QuicError
+		var parsedFirstFrame bool
+		validatingReader := bytes.NewReader(packet.data)
+		for {
+			frame, err := s.frameParser.ParseNext(validatingReader, packet.encryptionLevel)
+			if err != nil {
+				quicErr, ok := err.(*qerr.QuicError)
```
bemasc: Nit: can be combined with the `if` statement.

session.go
```diff
@@ -722,6 +794,56 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
...
+			}
+			if frame == nil {
+				if !parsedFirstFrame && !updateSrcConnID {
+					return qerr.ToIgnorableError(fmt.Errorf("Packet contains only padding frames."))
+				}
+				break
+			}
+			parsedFirstFrame = true
+			if err := s.validateFrame(frame, packet.packetNumber, packet.encryptionLevel); err != nil {
+				quicErr, ok := err.(*qerr.QuicError)
```
bemasc: Nit: Combine with `if` statement.

session.go
```diff
@@ -722,6 +794,56 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
 
 	r := bytes.NewReader(packet.data)
 	var isAckEliciting bool
+	var recoverableFrame wire.Frame
```
bemasc: I think this can be moved inside `if s.mitigationOn(...) {`

session.go
tatianab: done

session.go
tatianab: done

session.go
tatianab: done

session.go
```diff
@@ -691,22 +764,49 @@ func (s *session) handleRetryPacket(p *receivedPacket, hdr *wire.Header) bool /*
...
-	// The server can change the source connection ID with the first Handshake packet.
-	if s.perspective == protocol.PerspectiveClient && !s.receivedFirstPacket && packet.hdr.IsLongHeader && !packet.hdr.SrcConnectionID.Equal(s.destConnID) {
+	updateDestConnID := s.perspective == protocol.PerspectiveClient && !s.receivedFirstPacket && packet.hdr.IsLongHeader && !packet.hdr.SrcConnectionID.Equal(s.destConnID)
+	if updateDestConnID {
 		s.logger.Debugf("Received first packet. Switching destination connection ID to: %s", packet.hdr.SrcConnectionID)
 		s.destConnID = packet.hdr.SrcConnectionID
 		s.packer.ChangeDestConnectionID(s.destConnID)
 	}
 
+	// The server can change the source connection ID with the first Handshake packet.
```
bemasc: I think this comment was better in the original location.

session.go
```diff
@@ -718,6 +818,43 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
...
+
+		validatingReader := bytes.NewReader(packet.data)
+		for {
+			frame, err := s.frameParser.ParseNext(validatingReader, packet.encryptionLevel)
+			if err != nil {
+				return err
+			}
+			if frame == nil {
+				if !parsedFirstFrame && !updateDestConnID {
+					return fmt.Errorf("packet contains only padding frames")
```
bemasc: Why is this only true if `!updateDestConnID`?  Are Retry packets permitted to contain padding?

session.go
```diff
@@ -133,7 +139,10 @@ type session struct {
 
 	closeOnce sync.Once
 	// closeChan is used to notify the run loop that it should terminate
-	closeChan chan closeError
+	closeChan   chan closeError
+	delayedChan chan delayedError
+
+	prevState *recoveryState
```
bemasc: It seems like you don't need this here.  Instead, you can just have `saveState` return a `recoveryState`, and have `restoreState` take an argument.

session.go
```diff
@@ -647,6 +707,19 @@ func (s *session) handleSinglePacket(p *receivedPacket, hdr *wire.Header) bool /
 		packet.hdr.Log(s.logger)
 	}
 
+	if s.mitigationOn(packet.encryptionLevel) {
+		s.saveState()
+		if err := s.handleUnpackedPacket(packet, p.rcvTime); err != nil {
+			s.logger.Debugf("Ignoring or trying to recover from suspicious/invalid packet, rolling back relevant state.")
+			s.restoreState()
```
bemasc: What kind of packet would both (1) cause a connection ID change and (2) cause an error here?  An Initial containing a CONNECTION_CLOSE with a different connection ID?

Could we just ignore the server connection ID change, if the packet closes the connection anyway?

internal/qerr/quic_error.go
```diff
@@ -11,6 +11,7 @@ type QuicError struct {
 	ErrorMessage       string
 	isTimeout          bool
 	isApplicationError bool
+	delayed            bool
```
bemasc: It looks like this is only ever true for (1) remote close before handshake completes and (2) version negotiation failure.  I propose we replace "delay" with "isInsecureClose" or similar throughout.  I think that would be clearer.

session.go
```diff
@@ -807,14 +958,32 @@ func (s *session) handlePacket(p *receivedPacket) {
 	}
 }
 
+func (s *session) mitigationOn(encLevel protocol.EncryptionLevel) bool {
+	return s.config.AttackTimeout > 0 && !s.handshakeComplete && encLevel != protocol.Encryption1RTT
+}
+
+// MaybeRecover runs fail(err) immediately if injection attack mitigation is turned off,
+// or after AttackTimeout elapses if no packets are received
+func (s *session) maybeRecover(fail func(error), err error) {
```
bemasc: I think `fail` here and throughout should probably be called `shutdown` or similar, since it's not necessarily a failure at all, but normally a deliberate remote close.

session.go
```diff
@@ -133,7 +139,10 @@ type session struct {
 
 	closeOnce sync.Once
 	// closeChan is used to notify the run loop that it should terminate
-	closeChan chan closeError
+	closeChan   chan closeError
+	delayedChan chan delayedError
```
bemasc: I think you could avoid the need for this channel by sending insecure close errors over closeChan, appropriately marked, and checking those errors when they're received in the mainloop instead of `break`ing unconditionally.

client.go
```diff
@@ -341,8 +348,15 @@ func (c *client) handleVersionNegotiationPacket(p *receivedPacket) {
 	c.logger.Infof("Received a Version Negotiation packet. Supported Versions: %s", hdr.SupportedVersions)
 	newVersion, ok := protocol.ChooseSupportedVersion(c.config.Versions, hdr.SupportedVersions)
 	if !ok {
-		c.session.destroy(fmt.Errorf("No compatible QUIC version found. We support %s, server offered %s", c.config.Versions, hdr.SupportedVersions))
-		c.logger.Debugf("No compatible QUIC version found.")
+		fail := func(err error) {
+			c.session.destroy(err)
```
bemasc: Instead of calling `session.destroy` here, could we send `err` to `closeChan` instead?  That would allow some simplification.

session.go
tatianab: We could - this would mean that "firstPacketReceived" would be set to true, so an attacker who could inject this kind of packet could cause the connection to fail if the server wanted to send a Retry packet. But an attacker could also send a legit Initial packet and get the same result. It's a question of whether its significantly easier for an attacker to construct an all-padding Initial vs a real Initial, and in practice it's probably not that different

tatianab: It might be nice to have a timer that starts when we think we have received the first packet and doesn't block others until the timer is up, but that's probably beyond the scope of this PR

client.go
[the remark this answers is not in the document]
lucas-clemente: @marten-seemann what's your take on this? Should it be a bool vs a timeout with negative values?

interface.go
```diff
@@ -223,6 +223,14 @@ type Config struct {
...
 	// If this value is zero, the timeout is set to 30 seconds.
 	IdleTimeout time.Duration
+	// AttackTimeout is the maximum duration that may pass after a suspicious packet is
+	// received without any incoming network activity, used for experimental mitigation
+	// against injection attacks. If the timeout is exceeded, the connection is closed.
+	// This timeout should be longer than any reasonable round trip time, but excessively
+	// long timeouts will delay reporting of rejected connections.
+	// If this value is zero, the default mitigation behavior will be used (currently off).
+	// If this value is negative, mitigation will be disabled.
+	AttackTimeout time.Duration
```
lucas-clemente: Thinking: Maybe it would be more useful to define this in terms of RTTs. That would also make it easier to switch this to a bool, since we could then pick a reasonable default value.

Unless of course an attacker could influence our RTT estimate. Is that the case?

lucas-clemente (approved):

lucas-clemente:
LGTM, but I'd like @marten-seemann to take another look before merging.

interface.go
tatianab: It's safest to assume that the attacker has control over the RTT estimate until we get the 1-RTT keys. That said, there may be a more intuitive way to represent this AttackTimeout.

PR #2064, opened 2019-08-15
add basic support for 0-RTT

### LINKED ISSUE #438: implement 0-RTT handshakes
The server already handles 0-RTT connections correctly.
Most of the logic for 0-RTT handshakes is client-side. This is a proposal for implementing 0-RTT connections.

## quic package

We need a way to store `ServerConfig`s such that they can be reused between requests. A good way to that would be to offer a `ServerConfigPool` similar to the [x509.CertPool](https://golang.org/pkg/crypto/x509/#CertPool). Before establishing the connection, `Dial` would search the `ServerConfigPool` for matching `ServerConfig`s and attempt a 0-RTT if it finds one.

## h2quic package

There needs to be a way to mark requests as "0-RTT safe", while the default is that they are not (i.e. only get executed forward-secure). 
I see two options here:

1. Add a `Is0RTTSafe(*http.Request) bool` callback to the `QuicRoundTripper`, which defaults to returning `false`. In this callback, an application could for example inspect if the `http.Request.Method` and only execute GET-requests as 0-RTT requests.
2. Add an explicit function `DoAs0Rtt(*http.Request)` which would mark a single request as 0-RTT safe.

marten-seemann:
Fixes #438. Depends on https://github.com/marten-seemann/qtls/pull/6 (please also review that PR).

This PR only adds very basic support for 0-RTT. Although the 0-RTT handshake works in general, it should not be considered production-ready yet. I opened a few follow-up issues: #2065, #2066, #2067, #2068.

### DIFF
--- a/internal/handshake/transport_parameters.go
+++ b/internal/handshake/transport_parameters.go
@@ -369,6 +369,51 @@ func (p *TransportParameters) marshalVarintParam(b *bytes.Buffer, id transportPa
 	utils.WriteVarInt(b, val)
 }
 
+// MarshalForSessionTicket marshals the transport parameters we save in the session ticket.
+// When sending a 0-RTT enabled TLS session tickets, we need to save the transport parameters.
+// The client will remember the transport parameters used in the last session,
+// and apply those to the 0-RTT data it sends.
+// Saving the transport parameters in the ticket gives the server the option to reject 0-RTT
+// if the transport parameters changed.
+// Since the session ticket is encrypted, the serialization format is defined by the server.
+// For convenience, we use the same format that we also use for sending the transport parameters.
+func (p *TransportParameters) MarshalForSessionTicket() []byte {
+	b := &bytes.Buffer{}
+	b.Write([]byte{0, 0}) // length. Will be replaced later
+
+	// initial_max_stream_data_bidi_local
+	p.marshalVarintParam(b, initialMaxStreamDataBidiLocalParameterID, uint64(p.InitialMaxStreamDataBidiLocal))
+	// initial_max_stream_data_bidi_remote
+	p.marshalVarintParam(b, initialMaxStreamDataBidiRemoteParameterID, uint64(p.InitialMaxStreamDataBidiRemote))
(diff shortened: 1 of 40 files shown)


marten-seemann:
@lucas-clemente Can you please review this PR?

internal/handshake/transport_parameters.go
```diff
@@ -258,6 +258,49 @@ func (p *TransportParameters) marshalVarintParam(b *bytes.Buffer, id transportPa
...
 
+// When sending a 0-RTT enabled TLS session tickets, we need to save the transport parameters.
+// The client will remember the transport parameters used in the last session,
+// and apply those to the 0-RTT data it sends.
+// Saving the transport parameters in the ticket gives the server the option to reject 0-RTT
+// if the transport parameters changed.
+// Since the session ticket is encrypted, the serialization format is defined by the server.
+// For convenience, we use the same format that we also use for sending the transport parameters.
+func (p *TransportParameters) MarshalForSessionTicket() []byte {
+	b := &bytes.Buffer{}
```
lucas-clemente: Should we have a version number or something similar here?

internal/handshake/unsafe.go
```diff
@@ -19,6 +19,9 @@ func init() {
 	if !structsEqual(&tls.ClientSessionState{}, &qtls.ClientSessionState{}) {
 		panic("qtls.ClientSessionState not compatible with tls.ClientSessionState")
 	}
+	if !structsEqual(&tls.ClientSessionState{}, &clientSessionState{}) {
+		panic("clientSessionState not compatible with tls.ClientSessionState")
```
lucas-clemente: Should we add a note here to file a bug if this ever happens?

marten-seemann: I think we'll always be aware of this, since our CI will catch it.

internal/handshake/transport_parameters.go
marten-seemann: Yes, that definitely makes sense. I'll add one.

PR #2069, opened 2019-08-16
grease transport parameters

marten-seemann:
https://github.com/quicwg/base-drafts/pull/2873/ was finally merged.

### DIFF
--- a/internal/handshake/transport_parameters.go
+++ b/internal/handshake/transport_parameters.go
@@ -6,12 +6,17 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/handshake/transport_parameters.go
```diff
@@ -204,6 +205,15 @@ func (p *TransportParameters) Marshal() []byte {
 	b := &bytes.Buffer{}
 	b.Write([]byte{0, 0}) // length. Will be replaced later
 
+	// add a greased value
+	utils.BigEndian.WriteUint16(b, uint16(27+31*rand.Intn(100)))
+	rand.Seed(time.Now().UTC().UnixNano())
```
lucas-clemente: Seed once, e.g. in init(). Seeding this way may be expensive depending on the implementation.

PR #2074, opened 2019-08-17
gracefully handle invalid ACK delay values when parsing ACK frames

marten-seemann:


internal/wire/ack_frame.go
```diff
@@ -37,7 +38,13 @@ func parseAckFrame(r *bytes.Reader, ackDelayExponent uint8, version protocol.Ver
 	if err != nil {
 		return nil, err
 	}
-	frame.DelayTime = time.Duration(delay*1<<ackDelayExponent) * time.Microsecond
+
+	delayTime := time.Duration(delay*1<<ackDelayExponent) * time.Microsecond
+	if delayTime < 0 {
+		// If the delay time overflows, set it to the maximum encodable value.
+		delayTime = time.Duration(math.MaxInt64)
```
lucas-clemente: Wouldn't zero be a better default value?

marten-seemann: It doesn’t really matter, because no compliant peer would send such a large value anyway (unless it really experiences an ACK delay larger than 290 years...). My reasons for choosing the maximum value are:
* The only reason why this would underflow is that the peer actually sent a value > 290 years, so its the best approximation.
* This will result in a negative RTT sample for this ACK, and we‘re ignoring all RTT samples smaller than minimum RTT recorded for this connection.

Issue #1778, opened 2019-02-06
Use public variable for "server closed" error when Listener is closed


swdee:
Given the following code the Listener is closed and Accept() returns errors.New("server closed") at https://github.com/lucas-clemente/quic-go/blob/707e9f7b79d12a3aeee7db60ce7793ebc99be736/server.go#L298

```
ln, err := quic.ListenAddr(addr, tslCfg, nil)

go func() {
    time.Sleep(5*time.Second)
    ln.Close()
}

for {
    sess, err := ln.Accept()

    if err != nil {
        if err.Error() == "server closed" {
            break
        }
    }

    go handleSession(sess)
}
```

Rather than matching on the error string value "server closed", which is susceptible to changing in future versions, instead the code should define a public error variable for matching on, as is done in the standard lib net/http package, eg: https://github.com/golang/go/blob/ccd9d9d4cef16618d84bae9cba9f1b264278c8ab/src/net/http/server.go#L2826

Enabling something like;
```
sess, err := ln.Accept()

if err != nil {
    if err == quic.ErrServerClosed {
        break
    }
}
```

marten-seemann:
I'm not opposed to this idea, but I'm not sure if HTTP is the right comparison here, since it's an application protocol.
Looking at TCP, Go doesn't return any specific error code, and you have to do a partial string match on "use of closed network connection", if I remember the error correctly. Any idea why this is?

swdee:
Not sure on the TCP specifics on returned error code, however my thoughts on opening the issue were more idiomatic with packages defining public vars for error handling which can be picked up in an IDE so typing "quic.Err" quickly reveals the available error codes, versus having to grep through the code base to find out what the returned errors are.

marten-seemann:
@swdee I was wondering about your code
```go
for {
    sess, err := ln.Accept()

    if err != nil {
        if err.Error() == "server closed" {
            break
        }
    }

    go handleSession(sess)
}
```

Why would you only break the loop in case the server is closed. `Accept` never returns an error, unless something has gone wrong - you definitely shouldn't handle the session in that case (it might be `nil` anyway), and for sure further class to `Accept` will also fail.
The following code should be safe:
```go
for {
    sess, err := ln.Accept()
    if err != nil {
       break
    }

    go handleSession(sess)
}
```

mholt:
Is this something that Go 1.13's error wrapping can address?

marten-seemann:
[...]

PR #2109, opened 2019-09-03
add server closed error

keizo042:
hi, I'd like you to give me a chance to handle server error.

I read #1778 and checked your code snippets.  I'd like to handle error as below.
I think server closed error can be treaded as normal termination but I think it's a little hard for someone who is not familiar with quic-go internal. 

```

func (l *MyListener) Accept() error {
	for {
		sess, err := ln.Accept()
		if err == quic.ErrServerClosed {
			return nil
		}
		if err != nil {
			return err
		}
		go handleSession(sess)
	}
}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -19,6 +19,8 @@ import (
 	"github.com/lucas-clemente/quic-go/internal/wire"
 )
 
+var ErrServerClosed = errors.New("server closed")
+
 // packetHandler handles packets
 type packetHandler interface {
 	handlePacket(*receivedPacket)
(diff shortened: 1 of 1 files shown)


marten-seemann:
As I explained in #1778, I don't see any need to expand our API here.

Do you have any new information to add to that discussion?

marten-seemann:
Closing this PR for now. As mentioned in https://github.com/lucas-clemente/quic-go/issues/1778#issuecomment-529322513, I'm happy to revisit the introduction of a new error code (or an interface, or something similar) when a compelling use case is presented.
My reluctance to introduce something now is that we're trying hard to keep quic-go's API as small and as simple as possible.

Issue #1743, opened 2019-01-18
gquic: Received a packet with an unexpected connection ID


mholt:
I'm trying to get QUIC working in Caddy after a USR1 graceful reload; this means that Caddy gets a copy of the file descriptors underlying the sockets and then uses them in the new, reloaded instance to create listeners/PacketConns to set up "new" sockets. For TCP, this enables graceful restarts; for UDP, this should just move packets over to the new process/socket.

You can see the code for this dance here: https://sourcegraph.com/github.com/mholt/caddy@771dcf3d403fe862f5afa040309fa1db7b067554/-/blob/caddy.go#L705

The problem I'm having is that after doing this, QUIC no longer functions. The browser hangs for a few seconds while it renegotiates a regular TCP/TLS connection. It works fine before the reload.

The very first errors are:

```
muxer error handling packet from <ip>:1026: received a packet with an unexpected connection ID 0x538e276ef9b47552
```

And then they repeat every time I try refreshing in the browser.

I have verified that this connection ID are always new, even when starting a very new session in the browser (i.e. going incognito or opening Guest Mode). In other words, the connection IDs before and after the reload are different.

It seems like the QUIC library is refusing to make new servers for these new connection IDs, but whether the socket file is new or reused shouldn't really matter, should it?

The error appears to be emitted from here: https://github.com/lucas-clemente/quic-go/blob/714f38d5d0aff85894fd890718b991e361f03e7d/packet_handler_map.go#L165

So the server is nil, so there is no handler for the given connection ID. But shouldn't QUIC make one the first time it sees a new connection ID? (I'm not totally clear how gQUIC protocol works under the hood.)

Any tips would be helpful, if you can. Thank you!

marten-seemann:
I added some more logging to the `packetHandlerMap`, and this is what I'm getting:
```
2019/01/23 09:29:15 [INFO] Successfully loaded TLS assets from /Users/marten/src/go/src/github.com/lucas-clemente/quic-go/example/fullchain.pem and /Users/marten/src/go/src/github.com/lucas-clemente/quic-go/example/privkey.pem
Activating privacy features... done.
https://quic.clemente.io
2019/01/23 09:29:15 https://quic.clemente.io
WARNING: File descriptor limit 7168 is too low for production servers. At least 8192 is recommended. Fix with `ulimit -n 8192`.
[...]

PR #2111, opened 2019-09-03
identify connections by their local addr when adding to the multiplexer

### LINKED ISSUE #2108: multiplexer doesn't recognize the same connection
This seems to be what's causing the issue in https://github.com/caddyserver/caddy/pull/2727.

Test case here: https://github.com/lucas-clemente/quic-go/blob/d9cb20538b083093e6ea91546e374b36917cac06/multiplexer_test.go#L22-L30

marten-seemann:
Fixes #2108.
Fixes the issue described in https://github.com/caddyserver/caddy/pull/2727#issuecomment-526856566.

@mholt Could you run your test with this branch?

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -25,6 +25,7 @@ type mockPacketConn struct {
 
 func newMockPacketConn() *mockPacketConn {
 	return &mockPacketConn{
+		addr:        &net.UDPAddr{IP: net.IPv6zero, Port: 0x42},
 		dataToRead:  make(chan []byte, 1000),
 		dataWritten: make(chan mockPacketConnWrite, 1000),
 	}
--- a/multiplexer.go
+++ b/multiplexer.go
@@ -30,7 +30,7 @@ type connManager struct {
 type connMultiplexer struct {
 	mutex sync.Mutex
 
-	conns                   map[net.PacketConn]connManager
+	conns                   map[string] /* LocalAddr().String() */ connManager
 	newPacketHandlerManager func(net.PacketConn, int, []byte, utils.Logger) packetHandlerManager // so it can be replaced in the tests
 
 	logger utils.Logger
@@ -41,7 +41,7 @@ var _ multiplexer = &connMultiplexer{}
 func getMultiplexer() multiplexer {
 	connMuxerOnce.Do(func() {
 		connMuxer = &connMultiplexer{
-			conns:                   make(map[net.PacketConn]connManager),
+			conns:                   make(map[string]connManager),
 			logger:                  utils.DefaultLogger.WithPrefix("muxer"),
 			newPacketHandlerManager: newPacketHandlerMap,
 		}
@@ -57,7 +57,8 @@ func (m *connMultiplexer) AddConn(
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
-	p, ok := m.conns[c]
+	laddr := c.LocalAddr().String()
+	p, ok := m.conns[laddr]
 	if !ok {
 		manager := m.newPacketHandlerManager(c, connIDLen, statelessResetKey, m.logger)
 		p = connManager{
@@ -65,7 +66,7 @@ func (m *connMultiplexer) AddConn(
 			statelessResetKey: statelessResetKey,
 			manager:           manager,
 		}
-		m.conns[c] = p
+		m.conns[laddr] = p
 	}
 	if p.connIDLen != connIDLen {
 		return nil, fmt.Errorf("cannot use %d byte connection IDs on a connection that is already using %d byte connction IDs", connIDLen, p.connIDLen)
@@ -80,10 +81,11 @@ func (m *connMultiplexer) RemoveConn(c net.PacketConn) error {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
-	if _, ok := m.conns[c]; !ok {
+	laddr := c.LocalAddr().String()
+	if _, ok := m.conns[laddr]; !ok {
 		return fmt.Errorf("cannote remove connection, connection is unknown")
 	}
 
-	delete(m.conns, c)
+	delete(m.conns, laddr)
 	return nil
 }
--- a/multiplexer_test.go
+++ b/multiplexer_test.go
@@ -1,15 +1,34 @@
 package quic
 
 import (
+	"net"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
+
+type testConn struct {
+	counter int
+	net.PacketConn
+}
 
 var _ = Describe("Client Multiplexer", func() {
 	It("adds a new packet conn ", func() {
 		conn := newMockPacketConn()
 		_, err := getMultiplexer().AddConn(conn, 8, nil)
 		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("recognizes when the same connection is added twice", func() {
+		pconn := newMockPacketConn()
+		pconn.addr = &net.UDPAddr{IP: net.IPv4(1, 2, 3, 4), Port: 4321}
+		conn := testConn{PacketConn: pconn}
+		_, err := getMultiplexer().AddConn(conn, 8, nil)
+		Expect(err).ToNot(HaveOccurred())
+		conn.counter++
+		_, err = getMultiplexer().AddConn(conn, 8, nil)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(getMultiplexer().(*connMultiplexer).conns).To(HaveLen(1))
 	})
 
 	It("errors when adding an existing conn with a different connection ID length", func() {
--- a/quic_suite_test.go
+++ b/quic_suite_test.go
@@ -1,6 +1,8 @@
 package quic
 
 import (
+	"sync"
+
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -17,6 +19,9 @@ var mockCtrl *gomock.Controller
 
 var _ = BeforeEach(func() {
 	mockCtrl = gomock.NewController(GinkgoT())
+
+	// reset the sync.Once
+	connMuxerOnce = *new(sync.Once)
 })
 
 var _ = AfterEach(func() {


mholt:
Thanks for the quick work!

Unfortunately, the problem still manifests. It seems to just not like being handed a PacketConn that it has seen before. (I'm creating all new http3.Servers, by the way, just giving each one the same underlying PacketConn. I'm not reusing the http3.Server instances.)

mholt:
Aha, I was finally able to get it to happen with `QUIC_GO_LOG_LEVEL=debug` enabled on the server! It (the server) reports this error over and over before the client panics:

```
muxer received a packet with an unexpected connection ID 0xd78dd894048b0c33d84d9c738a31b348
```

over the course of a few seconds.

marten-seemann:
Great, that's a different error now!
Maybe #1743 is back? I'll look into it right now.

mholt:
Ah, yes, now that you mention it, it does seem similar to what we saw in Caddy 1.

Fortunately I can hold out on merging it into Caddy 2 until we get it figured out. :) Let me know how I can help! Thank you for looking into it more! It'll be exciting to get this into people's deployments.

marten-seemann:
Can you help me reproduce it?
I'm running Caddy with the config file that you gave me in https://github.com/caddyserver/caddy/pull/2727#issuecomment-526856566 with this branch of quic-go, and I've reloaded the config 5 times or so, both with and without debug mode. In both cases the downloads work without any problems, I'm not getting any error messages.

mholt:
@marten-seemann Sure, I'll see what I can do. I just pushed the latest, sorry, I forgot to do that >_< -- try pulling from the http3 branch of the caddy repo. The latest commit should be this: https://github.com/caddyserver/caddy/commit/dde3cd102080e7d80bdf6d45606dbe67bf77df4b 
 (it's where I keep track of the listeners myself and close them, since closing the http3 server doesn't close the listener).

marten-seemann:
@mholt I pulled dde3cd1, but everything still works here. I applied a few new configurations, and all transfers are succeeding.

mholt:
@marten-seemann 

Hmm... 

here's my full output (from the client) - in particular, this is the output from the run where I was able to reproduce it with the debug mode enabled at the server:

```
GET https://example.local:2443
client Starting new connection to example.local ([::]:52577 -> 127.0.0.1:2443), source connection ID (empty), destination connection ID 0xd78dd894048b0c33d84d9c738a31b348, version TLS dev version (WIP)
client Destroying session 0xd78dd894048b0c33d84d9c738a31b348: NO_ERROR: Handshake did not complete in time
client Connection (empty) closed.
panic: Get https://example.local:2443: NO_ERROR: Handshake did not complete in time

goroutine 3 [running]:
main.main.func1(0xc0001162a0, 0x1560a80, 0xc000020c90, 0xc000024641, 0xc000118370, 0x7ffeefbffa13, 0x23)
	/Users/matt/Dev/src/github.com/lucas-clemente/quic-go/example/client/main.go:50 +0x434
created by main.main
	/Users/matt/Dev/src/github.com/lucas-clemente/quic-go/example/client/main.go:47 +0x462
exit status 2
```

(notice that the connection ID is the same as the one the server reported was unexpected)

Okay, after trying it a few more times, I'm pretty convinced it is easier to reproduce with debug mode OFF at the server. In a few separate but identical trials, I was able to have it succeed after reloads like you are reporting (enabling debug mode helped it seems). But once debug mode was disabled again, the problem reappeared.

What are your commands and their outputs? Mine, for the server (you can see the reload happen in the middle of this output):

```
$ go run main.go run --config ../../_gitignore/v2/caddy2.json 
2019/09/04 00:10:42 Caddy 2 admin endpoint listening on localhost:2019
2019/09/04 00:10:42 [INFO][cache:0xc0000f21e0] Started certificate maintenance routine
2019/09/04 00:10:42 [INFO][example.local] Skipping automatic certificate management because a certificate with that SAN is already loaded
2019/09/04 00:10:42 [INFO] Enabling automatic HTTPS certificates for []
2019/09/04 00:10:42 [INFO] Enabling automatic HTTP->HTTPS redirects for [example.local]
2019/09/04 00:10:42 [DEBUG] Started HTTP/3 listener on :2443
2019/09/04 00:10:42 Caddy 2 serving initial configuration
2019/09/04 00:10:45 [INFO][cache:0xc00016c730] Started certificate maintenance routine
2019/09/04 00:10:45 [INFO][example.local] Skipping automatic certificate management because a certificate with that SAN is already loaded
2019/09/04 00:10:45 [INFO] Enabling automatic HTTPS certificates for []
2019/09/04 00:10:45 [INFO] Enabling automatic HTTP->HTTPS redirects for [example.local]
2019/09/04 00:10:45 [DEBUG] udp/:2443: Usage counter should not go above 2 or maybe 3, is now: 2
2019/09/04 00:10:45 [DEBUG] Started HTTP/3 listener on :2443
2019/09/04 00:10:45 [INFO][cache:0xc0000f21e0] Stopped certificate maintenance routine
2019/09/04 00:10:45 [DEBUG] Fake-closing underlying packet conn
```

For the client (before config reload):

```
$ GO111MODULE=on go run main.go -insecure https://example.local:2443
GET https://example.local:2443
client Starting new connection to example.local ([::]:56208 -> 127.0.0.1:2443), source connection ID (empty), destination connection ID 0x86a63fa879c159856af90eb596, version TLS dev version (WIP)
Got response for https://example.local:2443: &http.Response{Status:"200 OK", StatusCode:200, Proto:"HTTP/3", ProtoMajor:3, ProtoMinor:0, Header:http.Header{"Server":[]string{"Caddy"}}, Body:(*http3.body)(0xc0000b01e0), ContentLength:0, TransferEncoding:[]string(nil), Close:false, Uncompressed:false, Trailer:http.Header(nil), Request:(*http.Request)(nil), TLS:(*tls.ConnectionState)(nil)}
Request Body:
Booyah: FOO URI: '/test/abc?foo=bar' PATH: '/test/abc' QUERY: 'foo=bar'
client Closing session.
client Connection (empty) closed.
```

The config reload:

```
$ curl -X POST -d @caddy.json -H "Content-Type: application/json" "http://localhost:2019/load"
```

Then, the client again after the reload:

```
$ GO111MODULE=on go run main.go -insecure https://example.local:2443
GET https://example.local:2443
client Starting new connection to example.local ([::]:56209 -> 127.0.0.1:2443), source connection ID (empty), destination connection ID 0x0c59e95b17c0be8760ee5b80bbc9cda419, version TLS dev version (WIP)
client Destroying session 0x0c59e95b17c0be8760ee5b80bbc9cda419: NO_ERROR: Handshake did not complete in time
client Connection (empty) closed.
panic: Get https://example.local:2443: NO_ERROR: Handshake did not complete in time

goroutine 34 [running]:
main.main.func1(0xc0000bf3e0, 0x1560a80, 0xc0000beb10, 0xc0000bc599, 0xc0000bc910, 0x7ffeefbffa13, 0x23)
	/Users/matt/Dev/src/github.com/lucas-clemente/quic-go/example/client/main.go:50 +0x434
created by main.main
	/Users/matt/Dev/src/github.com/lucas-clemente/quic-go/example/client/main.go:47 +0x462
exit status 2
```

marten-seemann:
I tried it again a couple of times and now the error occurs. Seems to be pretty shaky. I'll keep digging, this definitely should not happen at all.

marten-seemann:
@mholt I'm really having problems reproducing this. I got it once, and it never occurred on my ~100 attempts afterwards.

The error you're getting
```
muxer received a packet with an unexpected connection ID 0xd78dd894048b0c33d84d9c738a31b348
```
should only occur if no server is set: https://github.com/lucas-clemente/quic-go/blob/d1489e5045f4270ebd7be356fee9951a12c0b450/packet_handler_map.go#L214-L217
so I'm wondering if there might be a deadlock in `SetServer` https://github.com/lucas-clemente/quic-go/blob/d1489e5045f4270ebd7be356fee9951a12c0b450/packet_handler_map.go#L104-L108

I simplified the shutdown logic a bit in #2115. I doubt that it fixes the issue, but could you give it a try and run it with this branch?

mholt:
I'm glad you were able to reproduce it! ... once, anyway. I feel a little better knowing that. 😅  I checked out that branch and did some trials with it but unfortunately the behavior was the same and the problem still manifests. (With debug mode enabled at the server I was able to see the "unexpected connection ID" error again.)

If it helps any, so you can see what we're doing in Caddy, the only place we interact with the http3 servers is here: https://sourcegraph.com/github.com/caddyserver/caddy@dde3cd102080e7d80bdf6d45606dbe67bf77df4b/-/blob/modules/caddyhttp/caddyhttp.go#L188-198

Old servers are garbage-collected on the next config reload because the App goes out of scope.

mholt:
If this helps:

You can see references to h.server here: https://sourcegraph.com/github.com/lucas-clemente/quic-go/-/blob/packet_handler_map.go#L214:7&tab=references

The only other place that sets server to nil is here, which leads into the function that you pointed out in that linked PR: https://sourcegraph.com/github.com/lucas-clemente/quic-go@d1489e5045f4270ebd7be356fee9951a12c0b450/-/blob/packet_handler_map.go#L112:4

But I'll see if I have a chance to look into potential deadlock there.

mholt:
It looks like SetServer is in fact called here: https://sourcegraph.com/github.com/lucas-clemente/quic-go@d1489e5045f4270ebd7be356fee9951a12c0b450/-/blob/server.go#L193:17 and the debug output shows that it is listening. I'm a little lost as to where sessionHandler goes (it doesn't look like it is ever used) -- but I see lots of interfaces. Still, is it possible that the sessionHandler which has a server added to it is a different copy from the one that is used to handle requests? (Just throwing ideas out there, feel free to ignore me haha, I don't have much of an idea what is going on this deep.)

mholt:
Somewhat significant finding, I think: If I comment out these lines in Caddy: https://sourcegraph.com/github.com/caddyserver/caddy@dde3cd102080e7d80bdf6d45606dbe67bf77df4b/-/blob/modules/caddyhttp/caddyhttp.go#L230-237

Then the problem _appears_ to go away. In other words: NOT closing the http3.Server and _only_ "closing" the listener (remember, in a reload scenario in Caddy, "Close()" on the PacketConn doesn't actually close the underlying listener) seems to work. In other words, not closing anything seems to work. I also made changes to the config on each reload (changed the response body) and verified that every time, the latest changes were printed by the client.

So, am I leaking resources then? But why is that working?

I hope that helps!

marten-seemann:
@mholt You're right, it's a bit confusing. The complexity is necessary to be able to handle a server and outgoing connections on the same `net.PacketConn`...

> Then the problem appears to go away. In other words: NOT closing the http3.Server and only "closing" the listener (remember, in a reload scenario in Caddy, "Close()" on the PacketConn doesn't actually close the underlying listener) seems to work

That's an interesting observation. I think we might have had a logical race in the `http3.Server`, but I'm not entirely sure. I just fixed another issue with the same code (as of #2122, the server can now serve on multiple packet conns at the same time), which might have solved that issue as well. Running Caddy and reloading the config at least didn't cause any further hangs any more.

I'm now waiting for @lucas-clemente to review all these PRs. It's getting confusing now with so many PRs stacked on top of each other.

marten-seemann:
@mholt Now that all the relevant PRs are merged, can you please try again?
And since this issue here is now closed, let's move the discussion back to https://github.com/caddyserver/caddy/pull/2727.

PR #2118, opened 2019-09-05
simplify mutex usage in the http3.Server

marten-seemann:


marten-seemann:
Closing in favor of #2122.

PR #2125, opened 2019-09-07
implement ChaCha20 header protection

### LINKED ISSUE #2123: Wrong header protection key derived for AEAD_CHACHA20_POLY1305
Whenever the ChaCha20-Poly1305 TLS 1.3 cipher suite is negotiated, quic-go uses AES anyway for header protection which breaks decryption:
https://github.com/lucas-clemente/quic-go/blob/9d92550e4e4b26e1a67b23ae7e703d15ffe38d0c/internal/handshake/aead.go#L103-L110

ChaCha20-based header protection must be applied instead as described at
https://tools.ietf.org/html/draft-ietf-quic-tls-22#section-5.4.4

This issue was originally caught by Wireshark (where decryption with an AES cipher works, but ChaCha20-Poly1305 fails) and later reproduced with ngtcp2.

## Reproducer 1
crypto/tls (on which qtls is based) has some "funny" behavior. If the CPU supports AES acceleration instructions, it will default to AES ciphersuites. Otherwise, it will use ChaCha20-Poly1305.

It is unfortunately not easy to force a specific TLS 1.3 cipher with OpenSSL or crypto/tls, the developers wanted to reduce the possibility of misconfigured software at the cost of flexibility for testing.
To force the use of ChaCha20-Poly1305 with standard crypto/tls, one can set `GODEBUG=cpu.all=off`. Unfortunately, x/sys/cpu does not respect this (see https://github.com/golang/go/issues/33963), so in order to make this override work I used this patch in qtls v0.3.2:
```diff
diff --git a/common.go b/common.go
index d3c66fd..2350061 100644
--- a/common.go
+++ b/common.go
@@ -1082,6 +1082,12 @@ func initDefaultCipherSuites() {
 		hasGCMAsm = hasGCMAsmAMD64 || hasGCMAsmARM64 || hasGCMAsmS390X
 	)
 
+	// x/sys/cpu does not respect GODEBUG=cpu.all=off. As a workaround,
+	// check it here. See https://github.com/golang/go/issues/33963
+	if strings.Contains(os.Getenv("GODEBUG"), "cpu.all=off") {
+		hasGCMAsm = false
+	}
+
 	if hasGCMAsm {
 		// If AES-GCM hardware is provided then prioritise AES-GCM
 		// cipher suites.
```

For generating a TLS keylog file for use with Wireshark I used:
```diff
diff --git a/example/echo/echo.go b/example/echo/echo.go
index 7c1ae280..d25e40ca 100644
--- a/example/echo/echo.go
+++ b/example/echo/echo.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"log"
 	"math/big"
+	"os"
 
 	quic "github.com/lucas-clemente/quic-go"
 )
@@ -106,8 +107,16 @@ func generateTLSConfig() *tls.Config {
 	if err != nil {
 		panic(err)
 	}
+	var keyLogWriter io.Writer
+	if keylogfile := os.Getenv("SSLKEYLOGFILE"); keylogfile != "" {
+		keyLogWriter, err = os.OpenFile(keylogfile, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0600)
+		if err != nil {
+			panic(err)
+		}
+	}
 	return &tls.Config{
 		Certificates: []tls.Certificate{tlsCert},
 		NextProtos:   []string{"quic-echo-example"},
+		KeyLogWriter: keyLogWriter,
 	}
 }
diff --git a/go.mod b/go.mod
index 15a9b90d..762e743f 100644
--- a/go.mod
+++ b/go.mod
@@ -12,4 +12,5 @@ require (
 	github.com/onsi/gomega v1.4.3
 	golang.org/x/crypto v0.0.0-20190228161510-8dd112bcdc25
 	golang.org/x/net v0.0.0-20190228165749-92fc7df08ae7
+	golang.org/x/sys v0.0.0-20190228124157-a34e9553db1e
 )
```

I then built and ran the echo client with:
```
go mod download && go mod vendor
go build -mod=readonly ./example/echo
sudo tcpdump -i any -p -w quic-go-v0.12.0.pcap -f udp &
# normal AES cipher
SSLKEYLOGFILE=quic-go-v0.12.0.keys ./echo
# Force ChaCha20-Poly1305
GODEBUG=cpu.all=off SSLKEYLOGFILE=quic-go-v0.12.0.keys ./echo
```

Wireshark (with some patches) will be able to decrypt the first, but not the second session.

## Reproducer 2
Interop between quic-go v0.12.0 server and ngtcp2 client with the following versions:
https://github.com/ngtcp2/ngtcp2 draft-22 9c1805430d4a
https://github.com/ngtcp2/nghttp3 master 6964d8e771ec
https://github.com/tatsuhiro-t/openssl openssl-quic-draft-22 90ccde5d7d53

Reproducer:
```
# run in quic-go directory
go build -mod=vendor -o example-main ./example
./example-main -v

# run in ngtcp2 build dir
cd examples
dumpcap -i lo -f 'udp port 6121' -w quic-interop.pcapng -p &
SSLKEYLOGFILE=quic-interop.keys ./client 127.0.0.1 6121 http://example.com --ciphers=TLS_CHACHA20_POLY1305_SHA256
```

Without the `--ciphers` override, Wireshark is able to decrypt the traffic and interop succeeds because AES is in use.

With the `--ciphers` override, ChaCha20-Poly1305 will be used, and interop fails. Wireshark shows decryption failure and quic-go reports this as well:
```
server Dropping packet that could not be unpacked. Error: decryption failed
```

marten-seemann:
Fixes #2123.

Interop tests with picoquic.


go.mod
```diff
@@ -3,13 +3,15 @@ module github.com/lucas-clemente/quic-go
 go 1.13
 
 require (
+	github.com/alangpierce/go-forceexport v0.0.0-20160317203124-8f1d6941cd75
```
marten-seemann: This is only needed for tests, since qtls doesn't need to expose the constructor for the ChaCha AEAD.

Lekensteyn:
Thanks, I can confirm that decryption works again in Wireshark with https://code.wireshark.org/review/34469 and the capture from https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=13881#c242.

marten-seemann:
Thanks for testing this!

PR #2127, opened 2019-09-09
Fix nil pointer crash in http3.client.Close()

fbzhong:
client.session would be nil when handshake err. So it's better to check `session != nil` in client.Close() function.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -120,6 +120,9 @@ func (c *client) setupSession() error {
 }
 
 func (c *client) Close() error {
+	if c.session == nil {
+		return nil
+	}
 	return c.session.Close()
 }
 


marten-seemann:
Where does this crash occur?

fbzhong:
@marten-seemann when calling http3.RoundTripper.Close().
I call RoundTripper.Close() to cleanup closed clients. However, those clients are closed, session is nil, then crash.

marten-seemann:
Can you share a code snippet? I still don’t understand how this can happen.

fbzhong:
I tested on the mater branch, no crash anymore. My crash was occurred in tag 0.12.0.
Sorry for this.

fbzhong:
Hi @marten-seemann ,

I'd like re-open this PR. I found a scenario to re-produce this crash on master branch:

```go
package main

import (
	"crypto/tls"
	"net/http"

	"log"

	"github.com/lucas-clemente/quic-go/internal/testdata"
)

func main() {
	var cfg *quic.Config
	url := "https://127.0.0.2:6060"

	roundTripper := &http3.RoundTripper{
		QuicConfig: cfg,
		TLSClientConfig: &tls.Config{
			RootCAs:            testdata.GetRootCA(),
			InsecureSkipVerify: true,
		},
	}

	hclient := &http.Client{
		Transport: roundTripper,
	}

	var err error
	// remote url is unavailable, error occurred.
	_, err = hclient.Get(url)
	log.Printf("[1/4] Get %s return err: %v", url, err)
	_, err = hclient.Get(url)
	log.Printf("[2/4] Get %s return err: %v", url, err)

	// Cleanup cached clients in http3.RoundTripper and wait for remote server is back online.
	roundTripper.Close() // crashed here.
	_, err = hclient.Get(url)
	log.Printf("[3/4] Get %s return err: %v", url, err)

	// Try again, because we don't know when the server will be back online.
	roundTripper.Close() // crashed here.
	_, err = hclient.Get(url)
	log.Printf("[4/4] Get %s return err: %v", url, err)
}
```

The root cause is because of those snippet:
1. If there is no cached client, new a client (`session == nil` by default ) and put into `clients`
https://github.com/lucas-clemente/quic-go/blob/fc962d18373aa730d5f84786cebec841f5e7ef22/http3/roundtrip.go#L133-L143

2. Calling `quic.DialAddr` to open a session to remote server. If remote server is unavailable, `nil, err` will be returned, `c.session == nil` again.
https://github.com/lucas-clemente/quic-go/blob/fc962d18373aa730d5f84786cebec841f5e7ef22/http3/client.go#L83-L92

Then when we call `RoundTripper.Close()` to clean all cached clients again, crashed.

marten-seemann:
This is yet another incarnation of #765. I'd suggest to fix this issue, instead of adding piecemeal fixes to work around it.

PR #2128, opened 2019-09-09
add DialAddrFromPacketConn() func.

fbzhong:
In my case, I need to dial to addr from a conn created by `net.ListenConfig.ListenPacket()`. Because I need to set some socket options for the conn.

This `DialAddrFromPacketConn` func is wrapper of `DialAddrContext` with setting `createdPacketConn = true`, which will close the conn after session closed.

### DIFF
--- a/client.go
+++ b/client.go
@@ -85,6 +85,22 @@ func DialAddrContext(
 	return dialContext(ctx, udpConn, udpAddr, addr, tlsConf, config, true)
 }
 
+// DialAddrFromPacketConn establishes a new QUIC connection to a server using the provided net.PacketConn and context.
+// See DialAddr for details.
+func DialAddrFromPacketConn(
+	ctx context.Context,
+	conn net.PacketConn,
+	addr string,
+	tlsConf *tls.Config,
+	config *Config,
+) (Session, error) {
+	udpAddr, err := net.ResolveUDPAddr("udp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return dialContext(ctx, conn, udpAddr, addr, tlsConf, config, true)
+}
+
 // Dial establishes a new QUIC connection to a server using a net.PacketConn.
 // The same PacketConn can be used for multiple calls to Dial and Listen,
 // QUIC connection IDs are used for demultiplexing the different connections.


marten-seemann:
I don’t think we yet another dial function. You can easily achieve this by just resolving the address yourself, it’s literally a single line.

fbzhong:
@marten-seemann all public function of Dial with PacketConn has set `createdPacketConn = false`. How to pass PacketConn and set `createdPacketConn = true`? I don't find that function.

marten-seemann:
DialAddr.

fbzhong:
@marten-seemann The `DialAddr` function create a new PacketConn inside it.
But in my case, I need to set socks options to that PacketConn. The `DialAddr` don't return `pconn`, either `quic.session`.
So I have to write new function.

marten-seemann:
Then you created the packet conn yourself. quic-go is behaving correctly in that case.

fbzhong:
@marten-seemann If call `DialContext` function, how can I get to know: quic session is timeout or remove address is unavailable -> 1. close packet conn 2. release packet handler in quic runtime?

marten-seemann:
The error return value of `DialContext` implements the `net.Error` interface, which should allow you to distinguish between these cases.

fbzhong:
I am encountering the case when DialContext is finished. http3. RoundTripper already create `client` instance associated with my `packet conn`.

At first, `http3.client` works well. Then after a while, the remote server is unavailable, or `IdleTimeout` is trigger, the underlying `quic.session` is closed, but the `packet conn` is still available, it leads `socket leakage` in my server application.

I do tried to trace every `request, response` from `client`, if no active request/resp and timeout is trigger, close `packet conn` and `http3.RoundTripper`. It works, but ugly...

Btw, the `crash when close client` PR is found from that solution...

marten-seemann:
That sounds like another occurrence of #765.

fbzhong:
I think it have some differences.

In this PR, I just want to close `packet conn` at the right time. I think the best way is to let quic manage it, like created one in `DialAddr` function.

marten-seemann:
I see that your PR is doing this, and I think the logic is wrong, which is why I closed it.
If an application hands a packet conn to quic-go, it’s not quic-go’s responsibility to close the connection.

fbzhong:
Yes, I understand it.

And, can you please tell me how to close `packet conn` properly?

marten-seemann:
Hi @fbzhong, I feel that my comments yesterday were a bit too short, so allow me to offer a more comprehensive explanation here.

We're dealing with two layers here:
1. the transport layer, which corresponds to the quic-go package
2. the application layer, with HTTP/3 as one example

Since QUIC allows you to demultiplex connections by their connection ID, you can run a QUIC server and multiple outgoing connections **on the same packet conn**. That's why quic-go doesn't close the underlying connection if it was passed into quic-go by the caller (i.e. `createdPacketConn = false`). Only in the case where quic-go created the packet conn itself (i.e. `createdPacketConn = true`), which for example happens when you run `DialAddr` will quic-go close the packet conn after the connection is done.

So how is an application supposed to close the packet conn? Well, that depends on the application. In principle, an application can reuse the same packet conn to dial to multiple servers (in sequence or even in parallel, again, QUIC demultiplexes connections by their connection IDs). In that case, all you'd have to do is create a single packet conn in the beginning, then use it for as many QUIC connections as you like, and only close it during shutdown of your program.

fbzhong:
@marten-seemann Thanks for your explanation, very appreciate.

**Run a QUIC server and multiple outgoing connections on the same packet conn** that's really cool~ So I don't need to close the packet conn anymore, one packet conn connects to them all~

PR #2130, opened 2019-09-11
fix request cancelation gorouting leaking when client.RoundTrip() returns error

fbzhong:
The goroutine for Request Cancelation is started too early. When client.RounTrip() returns error, there is no way to quit that goroutine.
It's best to create `reqDone := make(chan struct{})` and start that goroutine before call `newResponseBody`. The caller could stop it by calling `res.Body.Close()`.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -179,6 +179,12 @@ func (c *client) doRequest(req *http.Request, str quic.Stream) (*http.Response, 
 		case <-reqDone:
 		}
 	}()
+	noCloseReqDone := false
+	defer func() {
+		if !noCloseReqDone {
+			close(reqDone)
+		}
+	}()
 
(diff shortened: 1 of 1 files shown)


marten-seemann:
LGTM. Can you please fix the tests?

fbzhong:
Sure~

marten-seemann:
@fbzhong Are you still trying to fix this?

fbzhong:
@marten-seemann Sorry for late reply.
Sure I will fix it.

marten-seemann:
The fix you had the first time was correct. Please revert to this state and just fix the test case.

fbzhong:
Ok. Let me take a look.

marten-seemann:
Closing due to age. Will fix it myself.

PR #2148, opened 2019-09-25
fix leaking request cancelation go routine in case an error occurred

marten-seemann:
Replacement for #2130.


marten-seemann:
I don't think that is correct.

PR #2162, opened 2019-10-06
implement the datagram draft

marten-seemann:
Implements draft-01 of https://datatracker.ietf.org/doc/draft-ietf-quic-datagram/.

Open questions:
* how to expose the maximum DATAGRAM frame size (which is announced by peer in its transport parameters)
* how to handle messages larger than the maximum size (at the moment we just return an error)

Note that the maximum datagram size can change at any moment (in theory, not as it's implemented right now): The path MTU could change, or we could use a connection ID with a different length, leaving more / less space in the packet for the payload.

Idea: Return a `quic.ErrMessageTooLarge` error. Callers of `SendMessage` are required to check for this error and should not propagate this error, but repackage the message into smaller pieces.

### DIFF
--- a/datagram_queue.go
+++ b/datagram_queue.go
@@ -1,47 +1,73 @@
 package quic
 
 import (
-	"sync"
-
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/lucas-clemente/quic-go/internal/wire"
 )
 
 type datagramQueue struct {
-	mutex sync.Mutex
-	queue chan *wire.DatagramFrame
+	sendQueue chan *wire.DatagramFrame
+	rcvQueue  chan []byte
 
 	closeErr error
 	closed   chan struct{}
 
 	hasData func()
+
+	logger utils.Logger
 }
 
-func newDatagramQueue(hasData func()) *datagramQueue {
+func newDatagramQueue(hasData func(), logger utils.Logger) *datagramQueue {
 	return &datagramQueue{
-		queue:   make(chan *wire.DatagramFrame),
-		hasData: hasData,
-		closed:  make(chan struct{}),
+		hasData:   hasData,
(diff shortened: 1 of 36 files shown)


marten-seemann:
Updated to the new draft version: https://tools.ietf.org/html/draft-pauly-quic-datagram-04.

apmattil:
I hope this is merged to master.

internal/protocol/protocol.go
```diff
@@ -44,11 +44,14 @@ const (
 )
 
 // A ByteCount in QUIC
-type ByteCount uint64
+type ByteCount int64
```
lucas-clemente, line 47: This seems scary to me :S Are we sure this doesn't break any assumptions around non-negative numbers sprinkled throughout the codebase?

datagram_queue.go
```diff
@@ -1,50 +1,76 @@
...
+	case f := <-h.sendQueue:
 		return f
 	default:
 		return nil
 	}
 }
 
+// HandleDatagramFrame handles a received DATAGRAM frame.
+func (h *datagramQueue) HandleDatagramFrame(f *wire.DatagramFrame) {
+	data := make([]byte, len(f.Data))
```
lucas-clemente, line 55: use a buf here?

lucas-clemente (approved):

datagram_queue.go
marten-seemann: STREAM frames use frame-specific buffer pools, so maybe we need a DATAGRAM-pool here? Anyway, let's keep this for a future PR.

internal/protocol/protocol.go
marten-seemann: It did uncover a uint underflow, fortunately it was just in test code (in the flow controller). I'm not sure why this is scary, as QUIC's varint encoding makes sure that all numbers that can practically be used are limited to `MaxUint64 / 4`.

PR #2166, opened 2019-10-12
Fix alt-svc header

juliens:
### What does this PR do?
This PR changes the `alt-svc` header format.

### Motivation
as specified https://quicwg.org/base-drafts/draft-ietf-quic-http.html#rfc.section.3.1, we need to send `h3-#draft-version#=:#port#` instead of `quic=:#port#`

and

as specified https://quicwg.org/base-drafts/draft-ietf-quic-http.html#rfc.appendix.B.22, we need to send supported version in `quic=#supported#` instead of `v=#supported#`

### More
- [x] Added/updated tests

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -334,7 +334,7 @@ func (s *Server) SetQuicHeaders(hdr http.Header) error {
 		s.supportedVersionsAsString = strings.Join(versions, ",")
 	}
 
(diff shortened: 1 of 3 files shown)


marten-seemann:
Thank you for this PR!

The changes you're suggestion look correct, but it looks like we also need to fix the encoding of the version number. If I understand the spec correctly, it should be encoded as a hex value, right? Can you change `protocol.VersionNumber.ToAltSvc()`?

juliens:
You're right, good catch!
And it's done :wink:

PR #2192, opened 2019-11-01
remove unneccessary destroy of the session when retiring a connection ID

marten-seemann:
The only time `packetHandlerMap.Retire` is called is from `session.handleCloseError` https://github.com/lucas-clemente/quic-go/blob/afc7c1091896219294a0b6754c3de302c1244d43/session.go#L976.
At this point, the session should already be closed, so there's no need to call `session.destroy`.


marten-seemann:
This PR was wrong. We needed the call to `destroy` to close the `closedSession`.
Closing a session is becoming way too complicated. I'm beginning to think we need to refactor stuff here, but I'm not sure yet how to simplify things.

PR #2226, opened 2019-11-22
implement the retry integrity tag

marten-seemann:
See https://github.com/quicwg/base-drafts/pull/3120.

### DIFF
--- a/internal/handshake/retry.go
+++ b/internal/handshake/retry.go
@@ -1,0 +1,42 @@
+package handshake
+
+import (
+	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
+	"fmt"
+
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+)
+
+var retryAEAD cipher.AEAD
+
+func init() {
+	var key = [16]byte{0xf5, 0xed, 0x46, 0x42, 0xe0, 0xe4, 0xc8, 0xd8, 0x78, 0xbb, 0xbc, 0x8a, 0x82, 0x88, 0x21, 0xc9}
+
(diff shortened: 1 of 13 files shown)


internal/handshake/retry.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
+	"fmt"
+
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+)
+
+var retryAEAD cipher.AEAD
```
lucas-clemente: cipher.AEAD doesn't seem to make any thread safety guarantees. I think you might need a mutex here, or make this a server local var?

marten-seemann: I'm not sure I understand what you mean. `retryAEAD` is only set in `init()`, after we just call `Seal`.
`Seal` doesn't modify any internal state (at least not that I'm aware of), so it should be safe to use it concurrently, right?

lucas-clemente: As long as that’s not guaranteed by the interface or the constructor, you can’t make that assumption.

marten-seemann:
@lucas-clemente I added a mutex around the retry AEAD. This allows us to use pre-allocate some global state, which leads to considerably fewer allocations as well as a significant speedup (in the non-contended case):
```
pkg: github.com/lucas-clemente/quic-go/internal/handshake
BenchmarkRetry-16       	15345738	        71.8 ns/op	      16 B/op	       1 allocs/op
BenchmarkRetryOld-16    	 9577706	       112 ns/op	      56 B/op	       4 allocs/op

PR #2248, opened 2019-11-25
route retransmissions of the client's Initial to the right session

### LINKED ISSUE #2243: also use the client-chosen DCID to route packets to sessions
If the server's Initial is lost, the client will retransmit its Initial. If the server only uses the server-chosen connection ID for routing, each retransmission will cause a new session to be established.

marten-seemann:
Fixes #2243.

### DIFF
--- a/packet_handler_map.go
+++ b/packet_handler_map.go
@@ -99,7 +99,19 @@ func (h *packetHandlerMap) Add(id protocol.ConnectionID, handler packetHandler) 
 	h.mutex.Lock()
 	h.handlers[string(id)] = handler
 	h.mutex.Unlock()
-	return h.getStatelessResetToken(id)
+	return h.GetStatelessResetToken(id)
+}
+
+func (h *packetHandlerMap) AddIfNotTaken(id protocol.ConnectionID, handler packetHandler) bool /* was added */ {
+	sid := string(id)
+	h.mutex.Lock()
+	defer h.mutex.Unlock()
+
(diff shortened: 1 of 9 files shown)


marten-seemann:
Looks like I accidentally deleted the branch, which closed the PR.

marten-seemann:
There's a problem with this PR though: It's racy. If two Initials arrive at the same time (which might happen frequently, since a PTO would send two copies of the same Initial), it can happen that `newSession` is called before the dest conn ID is inserted into the packet handler map.

marten-seemann:
@lucas-clemente I fixed the race condition: When the server creates a new session, it calls `packetHandlerMap.AddIfTaken()`. This function checks if we already have an entry for that session. In that case, the server will discard the newly created session.

packet_handler_map.go
```diff
@@ -99,7 +99,19 @@ func (h *packetHandlerMap) Add(id protocol.ConnectionID, handler packetHandler)
 	h.mutex.Lock()
 	h.handlers[string(id)] = handler
 	h.mutex.Unlock()
-	return h.getStatelessResetToken(id)
+	return h.GetStatelessResetToken(id)
+}
+
+func (h *packetHandlerMap) AddIfNotTaken(id protocol.ConnectionID, handler packetHandler) bool /* was added */ {
+	var added bool
```
lucas-clemente: Nit: defer would allow you to avoid the bool var.

PR #2279, opened 2020-01-03
add a test that waits for handshake completion after sending 0-RTT data

marten-seemann:


marten-seemann:
Which other PR? This is just testing that `HandshakeComplete()` works, and we don't send out data intended for 1-RTT as 0-RTT.

lucas-clemente:
Ah, okay, I thought thiss was related to the broken 0RTT support. Case in point, better explanation what the test checks would be helpful ;)

PR #2280, opened 2020-01-03
queue 0-RTT for 100ms to allow the corresponding Initial to arrive

### LINKED ISSUE #2065: buffer 0-RTT packets for a short while
This helps with
* reordering on the wire
* the time we need to create the session after receiving after the Initial

The buffering time should be rather short, maybe 20ms.

marten-seemann:
Fixes #2065.

### DIFF
--- a/server.go
+++ b/server.go
@@ -68,6 +68,7 @@ type baseServer struct {
 
 	tokenGenerator *handshake.TokenGenerator
 
+	zeroRTTQueue   *zeroRTTQueue
 	sessionHandler packetHandlerManager
 
 	receivedPackets chan *receivedPacket
@@ -175,6 +176,7 @@ func listen(conn net.PacketConn, tlsConf *tls.Config, config *Config, acceptEarl
 		config:              config,
 		tokenGenerator:      tokenGenerator,
 		sessionHandler:      sessionHandler,
+		zeroRTTQueue:        newZeroRTTQueue(),
 		sessionQueue:        make(chan quicSession),
 		errorChan:           make(chan struct{}),
 		receivedPackets:     make(chan *receivedPacket, 1000),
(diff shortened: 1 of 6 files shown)


server.go
```diff
@@ -352,17 +350,26 @@ func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* was the packet
...
 	}
-	if hdr.IsLongHeader && hdr.Type != protocol.PacketTypeInitial {
-		// Drop long header packets.
-		// There's litte point in sending a Stateless Reset, since the client
-		// might not have received the token yet.
-		s.logger.Debugf("Dropping long header packet of type %s (%d bytes)", hdr.Type, len(p.data))
-		return false
+	if hdr.IsLongHeader {
+		if hdr.Type == protocol.PacketType0RTT {
+			s.zeroRTTQueue.Enqueue(hdr.DestConnectionID, p)
```
lucas-clemente: This means you always queue initial packets, even if we already have a session for them?

server.go
```diff
@@ -336,11 +338,7 @@ func (s *baseServer) handlePacket(p *receivedPacket) {
 	}()
 }
 
-func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* was the packet passed on to a session */ {
-	if len(p.data) < protocol.MinInitialPacketSize {
-		s.logger.Debugf("Dropping a packet that is too small to be a valid Initial (%d bytes)", len(p.data))
-		return false
-	}
+func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* was the packet handled */ {
```
lucas-clemente: OT question: This function is called in a separate go routine above, so it may run in parallel for initial packets intended for the same session. Wouldn't we create two sessions in that case?

server.go
marten-seemann: No, if we already have a session, the `packetHandlerMap` will already direct the packets there.

server.go
marten-seemann: https://github.com/lucas-clemente/quic-go/blob/54d9daba4468d719b086bfd37773b8e14f215d9c/server.go#L456-L462 is preventing this.

server.go
```diff
@@ -428,6 +435,13 @@ func (s *baseServer) handleInitialImpl(p *receivedPacket, hdr *wire.Header) (qui
 	if sess != nil {
 		sess.handlePacket(p)
 	}
+	for {
+		p := s.zeroRTTQueue.Dequeue(hdr.DestConnectionID)
+		if p == nil {
+			break
+		}
+		sess.handlePacket(p)
```
lucas-clemente: `sess` can be nil here.

server.go
lucas-clemente: Not if you have two go routines at this point in the code in parallel. One creates the session, the other one then ends up queuing a packet which will never be read.

I find the overall structure here somewhat hard to reason about, but I think you need a mutex around checking whether we have a session and queueing, or an additional (possibly noop) dequeue after each queue.

server.go
marten-seemann: This was resolved in #2299.

server.go
marten-seemann: Fixed.

marten-seemann:
Most of the problems should have been solved by #2299. Can you please re-review, @lucas-clemente?

PR #2294, opened 2020-01-13
don't generate new packets when the send queue is full

### LINKED ISSUE #2258: Network outage causes Session to freeze
This seems to be caused by the internal send buffer of the UDP socket becoming full causing Linux kernel to return EAGAIN for writes. The golang netpoller handles this by waiting until the socket is writable ([source](https://github.com/golang/go/blob/release-branch.go1.13/src/internal/poll/fd_unix.go#L331)). This eventually causes Session to freeze on SendQueue.Send().

marten-seemann:
Fixes #2258, which was deleted by Github. Depends on #2293.

### DIFF
--- a/session.go
+++ b/session.go
@@ -117,8 +117,11 @@ type session struct {
 	version        protocol.VersionNumber
 	config         *Config
 
-	conn      connection
-	sendQueue *sendQueue
+	conn connection
+
+	sendQueue          *sendQueue
+	sendQueueBlocked   bool
+	sendQueueAvailable <-chan struct{}
 
 	streamsMap      streamManager
 	connIDManager   *connIDManager
@@ -409,7 +412,7 @@ var newClientSession = func(
 }
 
 func (s *session) preSetup() {
-	s.sendQueue = newSendQueue(s.conn)
+	s.sendQueue, s.sendQueueAvailable = newSendQueue(s.conn)
 	s.retransmissionQueue = newRetransmissionQueue(s.version)
 	s.frameParser = wire.NewFrameParser(s.version)
 	s.rttStats = &congestion.RTTStats{}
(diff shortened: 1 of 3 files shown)


session.go
```diff
@@ -490,24 +493,36 @@ runLoop:
...
+		var sendQueueChan <-chan struct{}
+		if s.sendQueueBlocked {
+			sendQueueChan = s.sendQueueAvailable
+		}
+
 		select {
 		case closeErr = <-s.closeChan:
 			break runLoop
+		case <-sendQueueChan:
+			// fmt.Println("runLoop: send queue available")
```
lucas-clemente: commented code

session.go
```diff
@@ -490,24 +493,36 @@ runLoop:
 
 		s.maybeResetTimer()
 
+		var sendQueueChan <-chan struct{}
```
lucas-clemente: Maybe add a comment on why we don't always wait on this chan, even if we're not blocked? I think it's because of spurious unblocks, but I'm not sure.

session.go
```diff
@@ -117,8 +117,11 @@ type session struct {
 	version        protocol.VersionNumber
 	config         *Config
 
-	conn      connection
-	sendQueue *sendQueue
+	conn connection
+
+	sendQueue          *sendQueue
+	sendQueueBlocked   bool
```
lucas-clemente: Can we avoid this bool by exposing the sendQueue's blocked bool?

marten-seemann: I don't think that works, since that bool could change before the run loop is started the next time.

session.go
marten-seemann: Done.

marten-seemann:
Replaced by #2971.

PR #2295, opened 2020-01-13
avoid segfault when closing the client after dialing fails

spacewander:
Make it safe to call `defer client.Close()` just after creating the
client.


marten-seemann:
Duplicate of https://github.com/lucas-clemente/quic-go/pull/2127?

PR #2305, opened 2020-01-22
update Retry key and nonce to the values used in draft-25

marten-seemann:
Because changing constants is so much fun: https://github.com/quicwg/base-drafts/pull/3365.


marten-seemann:
Interop successfully tested with quicly (thanks to @kazuho).

marten-seemann:
Added test vector from the draft (see https://github.com/quicwg/base-drafts/pull/3394).

Issue #2311, opened 2020-01-25
allow 0-RTT http3 requests


marten-seemann:

marten-seemann:
Ideally, we'd have a `SafeFor0RTT` field on the `http.Request`, and the `http.RoundTripper` could then use this to determine if it can issue this request in 0-RTT or if it has to block it until completion of the handshake.

Unfortunately, we don't have this field.

A workaround would be to add a `RoundTripEarly()` to our implementation of the `http.RoundTripper`. This would work, but limit users to actually use this interface explicitly, i.e. they wouldn't be able to use `http.Get`, and not even the `http.Client`.

marten-seemann:
Another option would be to (ab)use the `Method` field on the `http.Request`, by defining a "GET_0RTT". This would make it clear than *only* GET requests can be used for 0-RTT requests (other requests are not idempotent).

This hack would allow users the continue using the `http.Client` and to define which requests are safe for 0-RTT.

PR #2312, opened 2020-01-25
use 0.5-RTT data to open the control stream in the http3 server

marten-seemann:
The CircleCI test is failing due to #2313.


http3/server.go
```diff
@@ -173,6 +173,10 @@ func (s *Server) handleConn(sess quic.Session) {
 	(&settingsFrame{}).Write(buf)
 	str.Write(buf.Bytes())
 
+	// Wait for completion of the handshake.
+	// TODO(#2311): allow 0-RTT requests.
+	<-sess.HandshakeComplete().Done()
```
marten-seemann: There's no real reason to wait for completion of the handshake here, other than keeping the (behavioral) diff small in this PR. I think it will be safe to remove these lines in #2311.

PR #2320, opened 2020-01-26
add 0-RTT support for HTTP/3

### LINKED ISSUE #2311: allow 0-RTT http3 requests

marten-seemann:
Fixes #2311. Depends on #2315.


http3/client.go
```diff
@@ -151,11 +155,14 @@ func (c *client) RoundTrip(req *http.Request) (*http.Response, error) {
 		return nil, c.handshakeErr
 	}
 
-	// wait for the handshake to complete
-	select {
-	case <-c.session.HandshakeComplete().Done():
-	case <-req.Context().Done():
-		return nil, req.Context().Err()
+	// Immediately send out this request, if this is a 0-RTT request.
+	if req.Method != MethodGet0RTT {
```
lucas-clemente: Do we want to rewrite the method name here, so that it doesn't get sent to the server like this?

marten-seemann: Good catch. Forgot that.

PR #2323, opened 2020-01-27
add basic support for qlog

### LINKED ISSUE #2221: add support for qlog
The spec is here: https://github.com/quiclog/internet-drafts
Awesome visualization tools are here: https://github.com/quiclog/qvis

marten-seemann:
Fixes #2221.


marten-seemann:
@rmarx noticed that our traces are not exactly ordered, which is required by the qlog specification. The reason is that we're exporting both send and receive events. We record the receive time of a packet in the go routine that's doing the `Read`, whereas the go routine handling the session record sent packets.

That means that the slice of events that we're keeping in the `qlog.Tracer` is *almost* ordered, i.e. most events are in order, but some events might be swapped over a short time frame.

I did some benchmarking of Go's `sort.Sort` implementation (see https://gist.github.com/marten-seemann/30251742b378318097e5400ea170c00f):

```
❯ go test -bench . -benchmem -run Bench -benchtime=3s
goos: darwin
goarch: amd64
pkg: github.com/lucas-clemente/quic-go/qlog
BenchmarkRandomSort-16                       253          13652995 ns/op              32 B/op          1 allocs/op
BenchmarkRandomSortStable-16                  81          40883706 ns/op              32 B/op          1 allocs/op
BenchmarkOrderedSort-16                      645           5385259 ns/op              32 B/op          1 allocs/op
BenchmarkOrderedSortStable-16               3322           1115647 ns/op              32 B/op          1 allocs/op
BenchmarkAlmostOrderedSort-16                604           5626393 ns/op              32 B/op          1 allocs/op
BenchmarkAlmostOrderedSortStable-16         2776           1233624 ns/op              32 B/op          1 allocs/op
BenchmarkReversedSort-16                     567           5614487 ns/op              32 B/op          1 allocs/op
BenchmarkReversedSortStable-16               406           8972660 ns/op              32 B/op          1 allocs/op
PASS
ok      github.com/lucas-clemente/quic-go/qlog  65.466s
```

It looks like the stable sort is significantly faster if the slice is already ordered or contains a small amount of reordering.

qlog/frame.go
```diff
@@ -0,0 +1,360 @@
...
+	case protocol.StreamTypeUni:
+		return "unidirectional"
+	case protocol.StreamTypeBidi:
+		return "bidirectional"
+	default:
+		panic("unknown stream type")
+	}
+}
+
+func escapeStr(str string) []byte { return []byte("\"" + str + "\"") }
```
lucas-clemente: This will break if the string has quotation marks?

qlog/frame.go
```diff
@@ -0,0 +1,360 @@
...
+
+type streamFrame struct {
+	StreamID protocol.StreamID
+	Offset   protocol.ByteCount
+	Length   protocol.ByteCount
+	FinBit   bool
+}
+
+func transformFrame(wf wire.Frame) *frame {
+	// We don't want to store CRYPTO and STREAM frames, for multiple reasons:
```
lucas-clemente: Why not make a copy of the struct and zero out the slices? That would allow you to reuse the structs?

fuzzing/handshake/fuzz.go
```diff
@@ -0,0 +1,46 @@
...
+	if params.MaxAckDelay > 365*24*time.Hour {
+		return 0
+	}
+	marshaled := params.Marshal()
+	params2 := &handshake.TransportParameters{}
+	if err := params2.Unmarshal(marshaled, perspective); err != nil {
+		fmt.Printf("%#v\n", params2)
+		panic(err)
+	}
+	// if !reflect.DeepEqual(params, params2) {
```
lucas-clemente: Commented code.

(I'm also not sure why this change is in this commit at all?)

marten-seemann: Committed it unintentionally. Deleted.

qlog/frame.go
marten-seemann: Because we still need the length of the frame. `wire.StreamFrame` does implement a `DataLen` method, but it just wraps `len(frame.Data)`.

qlog/frame.go
marten-seemann: Yes, and I removed this function in the commit that switched from standard library JSON to GoJay.

Issue #2326, opened 2020-01-28
Panic related to ChaCha20-based header protection


rod-hynes:
We're running v0.14.0 (fork with minor changes) in production and we encounter the following panic from time to time.

```
panic: chacha20: counter overflow

goroutine 53078166 [running]:
.../vendor/github.com/marten-seemann/chacha20.(*Cipher).XORKeyStream(0xc0031697d8, 0xc00d4d2dc0, 0x5, 0x5, 0xc00d4d2dc0, 0x5, 0x5)
	/go/src/.../vendor/github.com/marten-seemann/chacha20/chacha_generic.go:167 +0xc3b
.../vendor/github.com/marten-seemann/chacha20.XORKeyStream(0xc00d4d2dc0, 0x5, 0x5, 0xc00d4d2dc0, 0x5, 0x5, 0xc00d4d2de5, 0xc00d4d2dc5)
	/go/src/.../vendor/github.com/marten-seemann/chacha20/chacha_generic.go:237 +0x12d
.../vendor/github.com/Psiphon-Labs/quic-go/internal/handshake.(*chachaHeaderProtector).apply(0xc00d4d2dc0, 0xc00e208009, 0x10, 0x5a3, 0xc00e208000, 0xc00e208005, 0x2, 0x5a7)
	/go/src/.../vendor/github.com/Psiphon-Labs/quic-go/internal/handshake/header_protector.go:109 +0xa9
.../vendor/github.com/Psiphon-Labs/quic-go/internal/handshake.(*chachaHeaderProtector).EncryptHeader(0xc00d4d2dc0, 0xc00e208009, 0x10, 0x5a3, 0xc00e208000, 0xc00e208005, 0x2, 0x5a7)
	/go/src/.../vendor/github.com/Psiphon-Labs/quic-go/internal/handshake/header_protector.go:98 +0x77
.../vendor/github.com/Psiphon-Labs/quic-go/internal/handshake.(*updatableAEAD).EncryptHeader(0xc0054fab40, 0xc00e208009, 0x10, 0x5a3, 0xc00e208000, 0xc00e208005, 0x2, 0x5a7)
	/go/src/.../vendor/github.com/Psiphon-Labs/quic-go/internal/handshake/updatable_aead.go:246 +0x8a
[...]

PR #2331, opened 2020-01-31
add a workaround for the ChaCha20 bug

marten-seemann:
Don't try to apply ChaCha20 header protection if the nonce value will end up to be 0xffffffff, which would cause a panic in the ChaCha20 library.  See #2326.

@rod-hynes, can you have a look at this? Would this prevent the crash?

### DIFF
--- a/internal/handshake/header_protector.go
+++ b/internal/handshake/header_protector.go
@@ -3,6 +3,7 @@ package handshake
 import (
 	"crypto/aes"
 	"crypto/cipher"
+	"crypto/rand"
 	"fmt"
 
 	"github.com/marten-seemann/chacha20"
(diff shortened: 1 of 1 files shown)


rod-hynes:
> Would this prevent the crash?  

Looking at the code, I believe so, yes.

What happens in the workaround case where the packet ends up undecryptable? Will the QUIC protocol retry with sample data that's likely to be different? Or will the QUIC connection terminate?

marten-seemann:
The packet will end up undecryptable and be dropped. Loss detection will kick in and the frames will be retransmitted.
Since frames (and not packets) are retransmitted, and retransmission have a different packet number, the ciphertext (and therefore the nonce) will be different.

PR #2334, opened 2020-02-01
save the RTT along the session ticket and use it on resumed connections

### LINKED ISSUE #2068: save RTT stats with the session ticket
Having an RTT estimate is particularly important when establishing a 0-RTT connection, since the RTT determines the pacing rate. Currently, our estimate is 0, so we're not pacing at all, but dumping a full congestion window of bytes on the wire immediately.

marten-seemann:
Fixes #2068.

### DIFF
--- a/internal/handshake/client_session_cache.go
+++ b/internal/handshake/client_session_cache.go
@@ -3,7 +3,7 @@ package handshake
 import (
 	"bytes"
 	"crypto/tls"
-	"encoding/asn1"
+	"io"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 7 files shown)


internal/handshake/client_session_cache.go
```diff
@@ -43,12 +58,22 @@ func (c *clientSessionCache) Get(sessionKey string) (*qtls.ClientSessionState, b
 	var session clientSessionState
 	sessBytes := (*[unsafe.Sizeof(session)]byte)(unsafe.Pointer(&session))[:]
 	copy(sessBytes, tlsSessBytes)
+	if len(session.nonce) < 4 {
+		fmt.Println(1)
```
lucas-clemente: ?

(same below)

Issue #153, opened 2016-05-30
Implement graceful shutdown of servers and connections


lucas-clemente:
For normal shutdowns, we should send `GOAWAY` frames, stop accepting new connections, and let all requests finish before shutting down the server.

fastest963:
Now that Golang 1.8 brought graceful shutdown [1], hopefully this is simpler?

[1] https://golang.org/doc/go1.8#http_shutdown

lucas-clemente:
Yes, we could implement a similar API for both quic and h2quic. I didn't start working on this yet, if anyone wants to help please ping the issue and I will provide more context.

ocastx:
I would like to give a hand here

Kh4n:
@lucas-clemente Is this issue still available? I see that the file has moved to `http3/server.go`, but the stub is still there.

marten-seemann:
@Kh4n Yes, we haven't worked on this yet.

Kh4n:
Looking at the most recent draft version for QUIC ([v25](https://www.ietf.org/id/draft-ietf-quic-transport-25.html)), GOAWAY frames are no longer part of the spec ([section C.21](https://www.ietf.org/id/draft-ietf-quic-transport-25.html#name-since-draft-ietf-quic-transport-04)). Should we now send a [STOP_SENDING](https://www.ietf.org/id/draft-ietf-quic-transport-25.html#frame-stop-sending) frame with an appropriate application error code?

marten-seemann:
GOAWAY is a HTTP feature, not a QUIC feature, see https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-goaway.

Kh4n:
That makes sense. I have added a GOAWAY frame into frames.go, following how the headersFrame is constructed. Looking at the spec, I need to send this frame on a control stream, but I do not have access to any of the sessions or their respective streams. There are a variety of ways I think we can track the sessions and streams; is there a preferred method of doing this that you have in mind?

marten-seemann:
The control stream is opened here: https://github.com/lucas-clemente/quic-go/blob/62d3a4166a86de0f4a6a31f66411a6c3446f28fb/http3/server.go#L166-L171.

It would probably make sense to save it in a member of the `server`. Furthermore, you'll need to accept the control stream (by calling `Session.AcceptUniStream()`) when a new connection is established, and then process frames arriving on this stream.

Does that make sense? Let me know if you run into any problems, I'm happy to help!

[...]

PR #2335, opened 2020-02-04
Graceful shutdown (#153)

Kh4n:
Implements graceful shutdown with GOAWAY frames, as well has some handling for control streams


Kh4n:
@googlebot I signed it!

go.mod
```diff
@@ -6,13 +6,16 @@ require (
...
 	github.com/onsi/gomega v1.8.1
-	golang.org/x/crypto v0.0.0-20190829043050-9756ffdc2472
-	golang.org/x/net v0.0.0-20190404232315-eb5bcb51f2a3
-	golang.org/x/sync v0.0.0-20190423024810-112230192c58
+	golang.org/x/crypto v0.0.0-20200128174031-69ecbb4d6d5d
+	golang.org/x/net v0.0.0-20200202094626-16171245cfb2
+	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
+	golang.org/x/sys v0.0.0-20200202164722-d101bd2416d5 // indirect
+	golang.org/x/text v0.3.2 // indirect
+	google.golang.org/genproto v0.0.0-20180831171423-11092d34479b // indirect
```
marten-seemann: I don't think you need to make any changes to go.mod.

http3/client.go
```diff
@@ -98,32 +110,92 @@ func (c *client) dial() error {
...
+		return newConnError(errorInternalError, err)
+	}
+
+	br, ok := controlStreamIn.(byteReader)
+	if !ok {
+		br = &byteReaderImpl{controlStreamIn}
+	}
+	t, err := utils.ReadVarInt(br)
+	if t != 0x0 {
+		c.logger.Debugf("First stream must be a control stream")
```
marten-seemann: The first stream doesn't necessarily need to be the control stream.

http3/client.go
```diff
@@ -98,32 +110,92 @@ func (c *client) dial() error {
...
+		return newConnError(errorMissingSettings, nil)
+	}
+	// TODO: do something with the settings frame
+	c.logger.Debugf("Got settings frame: %+v", sf)
+	go func() {
+		for {
+			frame, err := parseNextFrame(controlStreamIn)
+			if err != nil {
+				// Hack
+				if err.Error() == "Application error 0x100" {
```
marten-seemann: What is this?

http3/client.go
```diff
@@ -98,32 +110,92 @@ func (c *client) dial() error {
...
+			}
+			switch f := frame.(type) {
+			case *goawayFrame:
+				c.logger.Debugf("Received goaway frame, halting requests with streamID >= %d", f.StreamID)
+				c.clientCancel()
+				c.mutex.Lock()
+				c.logger.Debugf("streams %+v", c.clientStreams)
+				c.clientStreams.Range(func(str, _ interface{}) bool {
+					if str.(quic.Stream).StreamID() >= f.StreamID {
+						str.(quic.Stream).SetDeadline(time.Now().Add(-42 * time.Second))
```
marten-seemann: Where do 42 seconds come from?

http3/client.go
```diff
@@ -83,6 +92,9 @@ func newClient(
 		opts:          opts,
 		dialer:        dialer,
 		logger:        logger,
+		clientContext: ctx,
+		clientCancel:  cancel,
+		clientStreams: &sync.Map{},
```
marten-seemann: You don't need any map here.

http3/client.go
```diff
@@ -98,32 +110,92 @@ func (c *client) dial() error {
 	}
 
 	// run the sesssion setup using 0-RTT data
-	go func() {
-		if err := c.setupSession(); err != nil {
-			c.logger.Debugf("Setting up session failed: %s", err)
-			c.session.CloseWithError(quic.ErrorCode(errorInternalError), "")
-		}
-	}()
+	// go func() {
```
marten-seemann: Why?

http3/frames.go
```diff
@@ -131,3 +131,32 @@ func (f *settingsFrame) Write(b *bytes.Buffer) {
...
+
+func parseGoawayFrame(r io.Reader, l uint64) (*goawayFrame, error) {
+	buf := make([]byte, l)
+	if _, err := io.ReadFull(r, buf); err != nil {
+		if err == io.ErrUnexpectedEOF {
+			return nil, io.EOF
+		}
+		return nil, err
+	}
+	b := bytes.NewReader(buf)
```
marten-seemann: Reading the frame in a buffer and then reading from the buffer seems unnecessary.

http3/server.go
```diff
@@ -34,11 +37,38 @@ type requestError struct {
...
+	if r.streamErr != 0 {
+		ret = append(ret, r.streamErr.String())
+	}
+	if r.connErr != 0 {
+		ret = append(ret, r.connErr.String())
+	}
+	return strings.Join(ret, ", ")
+}
+
+type ctrlStream struct {
```
marten-seemann: Why do you define a struct that combines the two streams?

http3/server.go
```diff
@@ -205,6 +259,85 @@ func (s *Server) handleConn(sess quic.EarlySession) {
 	}
 }
 
+func (s *Server) addCtrlStream(c *ctrlStream) {
+	s.mutex.Lock()
+	if s.ctrlStreams == nil {
+		s.ctrlStreams = make(map[*ctrlStream]struct{})
+	}
+	s.ctrlStreams[c] = struct{}{}
```
marten-seemann: I don't think you need any map here.

marten-seemann (changes requested):
Thanks for the PR. I left some high-level comments on the code.

Could you also please rebase everything on top of master? Not having any merge commits in a PR makes reviewing a lot easier.

http3/client.go
Kh4n: I can change it, but the only reason I made it like that was because of this line in the spec: 
"Each side MUST initiate a single control stream at the beginning of the connection".

http3/client.go
Kh4n: When the connection closes normally, it does so with an error that says that. I couldnt find the specific error anywhere (otherwise I would've used `errors.Is`) so I just compared the message directly. Not doing this will debug print the error, which could be misleading

http3/client.go
Kh4n: Just a time in the past as was suggested. Is there a preferred way to do this?

http3/client.go
Kh4n: It was causing issues in testing because of the control streams racing to connect before the request completed

http3/frames.go
Kh4n: This is exactly how it was done in `parseSettingsFrame`, but I can look into alternatives

http3/client.go
Kh4n: Actually I found a better way to handle this entirely

http3/server.go
Kh4n: To make it easier to keep track of the max stream ID for each connection. I can remove it if necessary, but it make the relevant code cleaner

http3/server.go
Kh4n: Yes I have removed it. Now using a channel to signal to control streams

http3/client.go
Kh4n: I removed it as well, but i realized a (potential) issue. Can the client call RoundTrip for multiple different hosts and maintain those connections simultaneously?

Kh4n:
I think i did the rebase correctly. Not sure though. Let me know if something needs to be changed

marten-seemann:
Looks like the rebase didn't work properly.

Kh4n:
What was the issue with it? Is it because we are both authors on the commits in-between?

marten-seemann:
> What was the issue with it? Is it because we are both authors on the commits in-between?

I didn't author any commits. That's why I asked you to rebase.

Kh4n:
Haha, I see my mistake. Not good with Git. Is that better?

go.sum
```diff
@@ -38,8 +38,12 @@ github.com/golang/mock v1.2.0/go.mod h1:oTYuIxOrZwtPieC+H1uAHpcLFnEyAGVDL/k47Jfb
 github.com/golang/mock v1.4.0 h1:Rd1kQnQu0Hq3qvJppYSG0HtP+f5LPPUiDswTLiEegLg=
 github.com/golang/mock v1.4.0/go.mod h1:UOMv5ysSaYNkG+OFQykRIcU/QvvxJf3p21QfJ2Bt3cw=
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/golang/protobuf v1.3.0 h1:kbxbvI4Un1LUWKxufD+BiE6AEExYYgkQLQmLFqA1LFk=
+github.com/golang/protobuf v1.3.0/go.mod h1:Qd/q+1AKNOZr9uGQzbzCmRO6sUih6GTPZv6a1/R87v0=
```
marten-seemann: I don't think you need to change this file.

http3/client.go
marten-seemann: This definitely needs to be done concurrently, otherwise you'll have to wait for at least one round trip before you can send a request.

http3/client.go
marten-seemann: That's exactly what the type byte is there for. Just because you have to do it at the beginning doesn't mean that this necessarily is the first stream.

http3/client.go
marten-seemann: I'd prefer to print the error.

http3/client.go
```diff
@@ -201,6 +262,34 @@ func (c *client) RoundTrip(req *http.Request) (*http.Response, error) {
 	return rsp, rerr.err
 }
 
+func (c *client) requestCancelled(str quic.Stream, req *http.Request, reqDone <-chan struct{}) {
+	select {
+	case id := <-c.goawayChan:
```
marten-seemann: This won't work if there's more than a single request.

http3/server.go
```diff
@@ -205,6 +261,85 @@ func (s *Server) handleConn(sess quic.EarlySession) {
...
+		s.logger.Debugf("First incoming frame parsed was not a settings frame")
+		return nil, newConnError(errorMissingSettings, nil)
+	}
+	// TODO: do something with the settings frame
+	s.logger.Debugf("Got settings frame: %+v", sf)
+
+	s.clients.Add(1)
+	connDone := make(chan struct{})
+	go func() {
+		defer ginkgo.GinkgoRecover()
```
marten-seemann: No.

http3/server.go
```diff
@@ -205,6 +261,85 @@ func (s *Server) handleConn(sess quic.EarlySession) {
 	}
 }
 
+func (s *Server) handleControlStreams(sess quic.Session) (*ctrlStream, requestError) {
+	// send a SETTINGS frame
+	controlStreamOut, err := sess.OpenUniStream()
+	if err != nil {
+		return nil, newConnError(errorMissingSettings, err)
```
marten-seemann: That's not correct. The peer didn't do anything wrong here.

http3/server.go
```diff
@@ -205,6 +261,85 @@ func (s *Server) handleConn(sess quic.EarlySession) {
...
+	}
+
+	pair := &ctrlStream{in: controlStreamIn, out: controlStreamOut}
+
+	br, ok := controlStreamIn.(byteReader)
+	if !ok {
+		br = &byteReaderImpl{controlStreamIn}
+	}
+	t, err := utils.ReadVarInt(br)
+	if t != 0x0 {
```
marten-seemann: This is not correct, see above.

http3/server.go
```diff
@@ -205,6 +261,85 @@ func (s *Server) handleConn(sess quic.EarlySession) {
...
+	(&settingsFrame{}).Write(buf)
+	controlStreamOut.Write(buf.Bytes())
+
+	controlStreamIn, err := sess.AcceptUniStream(context.Background())
+	if err != nil {
+		s.logger.Debugf("Accepting the incoming control stream failed.")
+		return nil, newConnError(errorMissingSettings, err)
+	}
+
+	pair := &ctrlStream{in: controlStreamIn, out: controlStreamOut}
```
marten-seemann: I don't understand what pairing the streams achieves.

marten-seemann (commented):
Thanks for updating the PR! I left some comments on the code.

Please also note the CI failure, which detected a race condition in the server.

http3/server.go
Kh4n: To which part? the `ginkgo.GinkgoRecover`, the go routine, clients waitgroup?

marten-seemann: The `GinkgoRecover`. Ginkgo is a testing framework and doesn't belong into non-test code.

Kh4n: ok. there are other instance of ginkgorecover in that file, should I leave them there then? (they were there before)

marten-seemann: I'll take care of those. Must have been commit erroneously.

marten-seemann: See https://github.com/lucas-clemente/quic-go/pull/2372. Now with a custom linter to make sure this doesn't slip through again.

Kh4n:
Question about Gomock: To satisfy the requirement that the first incoming unistream may or may not be a control stream, I have it setup a go routine that waits on incoming uni streams. What i need is for it to accept the first stream, and then for the rest of calls just hang. I have achieved this by the following code:
```
sess.EXPECT().AcceptUniStream(gomock.Any()).DoAndReturn(func(ctx context.Context) (quic.ReceiveStream, int) {
	select {}
}).AnyTimes()
```
but that feels weird. Is that ok to write? The tests pass just fine with it.

http3/client.go
Kh4n: Do you mean multiple GOAWAY frames or multiple client requests on the same connection? I have found a way to handle multiple GOAWAY frames, but this code should work for multiple requests (further down in that code it puts the id back on to the channel so that other requests can read it as well)

marten-seemann: Please don't do that, that's highly non-obvious. It would be a lot cleaner to have a `chan struct{}`, close that channel, and then read from a member variable of the client.

marten-seemann:
> Question about Gomock: To satisfy the requirement that the first incoming unistream may or may not be a control stream, I have it setup a go routine that waits on incoming uni streams. What i need is for it to accept the first stream, and then for the rest of calls just hang. I have achieved this by the following code:
> 
> ```
> sess.EXPECT().AcceptUniStream(gomock.Any()).DoAndReturn(func(ctx context.Context) (quic.ReceiveStream, int) {
> 	select {}
> }).AnyTimes()
> ```
> 
> but that feels weird. Is that ok to write? The tests pass just fine with it.

That should work, but you should make sure that all Go routines return when the test finishes. So probably it would be better to have a `<-done` instead of the `select{}`, and then `close(done)` at the end of the test.

http3/client.go
Kh4n: Yes, that is true. I have done as you said, with a slight difference: since the client can receive multiple go away frames, I am using a slice of `chan struct{}` (not one for each client! just one for each goaway frame sent). I cannot see any alternative that does not involve managing client channels with a map of sorts. Is this ok?

marten-seemann: I don't think that's necessary. I'd assume something along these lines makes sense. You add member variables to the client:
```go
type client {
    // other stuff
    goawayChan chan struct{}
    goawayStreamNum protocol.StreamID
}
```

When receiving a GOAWAY frame, you

- `if c.goawayStreamNum == 0`, this is the first GOAWAY frame, so you `close(c.goawayChan)`
- `else`, you check that the new GOAWAY frame carries a smaller stream number than the one previously received (and throw an error otherwise), and update the value of `c.goawayStreamNum`

Does that make sense?

Kh4n: Yes, and that is a solution that I considered. However, while it works for when
- Client is about to reach `requestCancelled`
- Client `requestCancelled` goroutine is already sleeping when the first goaway frame is received
it will not work when client the `requestCancelled` goroutine is sleeping and the first goaway frame has already been received, and the streamID was < than that goaway frame. In that case, it will still be sleeping, and since we closed the channel there is no way to communicate with them if another goaway frame is received. If you believe handling this case is not necessary, I can go with the simpler one channel solution (closing client connections on goaway is optional after all)
I have a demo of what I am doing in the code here: https://play.golang.org/p/boPrANTKCnM

marten-seemann: The `[]chan struct` looks kind of complicated, I'd prefer to avoid that.

What about using a `sync.Cond` and then using a `Broadcast()` every time a GOAWAY is received?

Kh4n: That was also something I tried, however `sync.Cond` uses `Wait()`, which I couldn't see a way of using alongside the existing channel mechanism we use (`<-reqDone`). sync.Cond is kind of inflexible in this aspect. I couldn't see a way of feeding it a context somehow; if that were possible I think I could make it work. Looking through docs right now. Basically we need a mutex that takes a context: if we have that then `sync.Cond` is workable.

Kh4n:
Got it. Solved it with:
```
sess.EXPECT().AcceptUniStream(gomock.Any()).Return(controlStrIn, nil).MaxTimes(1)
sess.EXPECT().AcceptUniStream(gomock.Any()).DoAndReturn(func(ctx context.Context) (quic.ReceiveStream, error) {
	select {
	case <-uniDone:
	}
	return nil, errors.New("done")
}).AnyTimes()
sess.EXPECT().CloseWithError(gomock.Any(), "done").MaxTimes(1)
sess.EXPECT().CloseWithError(gomock.Any(), "").MaxTimes(1)
```
Something similar is also in `server_test.go`

marten-seemann:
You don't even need the `select`:
```go
sess.EXPECT().AcceptUniStream(gomock.Any()).DoAndReturn(func(ctx context.Context) (quic.ReceiveStream, error) {
	<-uniDone
	return nil, errors.New("done")
}).AnyTimes()
```

Kh4n:
Ah, great catch

http3/client.go
Kh4n: Ok, I have only one possible solution through `sync.Cond`, though it is arguably more obtuse than my own. Essentially we use two goroutines, the original `requestCancelled` goroutine along with another one that uses `sync.Cond.Wait()`, and when the `requestCancelled` goroutine finishes, we `Broadcast()` to it a special termination signal (say -2) that tells that goroutine to terminate. I am not sure of the safety of `Broadcast()` concurrently, but it should be workable. 

Spinning two goroutines is mandatory for `sync.Cond`, as it uses `Wait()`  semantics instead of channels, so you need it on its own goroutine.

The only other option besides this and my channel slice is a map of client goroutines (at least after my analysis). I just can't see a way of telling the `sync.Cond` to stop other than to Broadcast to it every time

Kh4n: @lucas-clemente Which of these options do you prefer?

marten-seemann: I think this can be solved a lot simpler. So if I see this correctly, the problem we're trying to solve is how to deal with requests that are blocked on `OpenStreamSync` when a GOAWAY is received. Requests that are already sent to the server don't need to be canceled, and requests for which the server is already sending a response can also just continue as if nothing happened.

I think it would be a valid behavior to return an error from all those requests, even if they might have gotten a stream ID lower than the limit in the GOAWAY frame. GOAWAY is used for graceful shutdown, after all.

Kh4n: The issue is when multiple GOAWAYs are sent. When just one is sent, it is no issue. Multiple GOAWAYs require something more complex to handle them. We can always just only parse the first one, and have the server reject any further client conns itself while also sending the GOAWAY. This would not be against the spec because it says that 

> Some requests might already be in transit. If the client has already sent requests on streams with a Stream ID greater than or equal to that indicated in the GOAWAY frame, those requests will not be processed and MAY be retried by the client on a different connection. The client **MAY** cancel these requests. *It is RECOMMENDED that the server explicitly reject such requests (see Section 4.1.2) in order to clean up transport state for the affected streams.*

Emphases mine

amery:
any plans regarding graceful shutdown? No updates in 4y is pretty depressing

PR #2339, opened 2020-02-09
add a Clone() function to the Config

marten-seemann:


config.go
```diff
@@ -0,0 +1,89 @@
+package quic
+
+import "github.com/lucas-clemente/quic-go/internal/protocol"
+
+// Clone clones a Config
+func (c *Config) Clone() *Config {
```
lucas-clemente: If you're copying every value, why not just `copy := *c; return &copy`?

marten-seemann: Yes, that's a lot nicer. I was just copying what the standard library does for the `tls.Config`.

PR #2344, opened 2020-02-11
Add trailers

### LINKED ISSUE #2266: HTTP trailer support
any plan  H3 to support  trailer? It's important for file transfer streaming over H3 to present final status

juliens:
This PR adds Trailers support.

Fix #2266

This implementation is based on http2 trailers, so to send a trailer, you must:
- announce Trailer in the Trailer header (before `WriteHeader`) and then (after` WriteHeader`) add your Trailer value in ResponseWriter.Header ()
- or you can use TrailerPrefix, by simply adding headers (after `WriteHeader`) with the TrailerPrefix (` Trailer: `)

I could not test it with another server / client implementation.

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -36,9 +36,9 @@ var _ io.ReadCloser = &body{}
 
 func newRequestBody(str quic.Stream, onFrameError func()) *body {
 	return &body{
-		str:            str,
-		onFrameError:   onFrameError,
-		isRequest:      true,
+		str:          str,
+		onFrameError: onFrameError,
+		isRequest:    true,
 	}
 }
 
@@ -93,7 +93,8 @@ func (r *body) readImpl(b []byte) (int, error) {
 					r.resp.Trailer.Add(trailer.Name, trailer.Value)
 				}
 
-				continue
+				// Trailer Frame is the last frame.
+				return 0, nil
 			case *dataFrame:
 				r.bytesRemainingInFrame = f.Length
 				break parseLoop
--- a/http3/body.go
+++ b/http3/body.go
@@ -3,8 +3,10 @@ package http3
 import (
 	"fmt"
 	"io"
+	"net/http"
 
 	"github.com/lucas-clemente/quic-go"
+	"github.com/marten-seemann/qpack"
 )
 
 // The body of a http.Request or http.Response.
@@ -22,6 +24,8 @@ type body struct {
 	onFrameError func()
 
 	bytesRemainingInFrame uint64
+
+	resp *http.Response
 }
 
 var _ io.ReadCloser = &body{}
@@ -34,11 +38,12 @@ func newRequestBody(str quic.Stream, onFrameError func()) *body {
 	}
 }
 
-func newResponseBody(str quic.Stream, done chan<- struct{}, onFrameError func()) *body {
+func newResponseBody(str quic.Stream, done chan<- struct{}, onFrameError func(), resp *http.Response) *body {
 	return &body{
 		str:          str,
 		onFrameError: onFrameError,
 		reqDone:      done,
+		resp:         resp,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


http3/response_writer.go
```diff
@@ -81,6 +88,75 @@ func (w *responseWriter) Write(p []byte) (int, error) {
...
+		// Forbidden by RFC 7230, section 4.1.2.
+		w.logger.Debugf("ignoring invalid trailer %q", k)
+		return
+	}
+	if !strSliceContains(w.trailers, k) {
+		w.trailers = append(w.trailers, k)
+	}
+}
+
+func (w *responseWriter) hasNonemptyTrailers() bool {
```
marten-seemann: Nit: `hasNonEmptyTrailers`

http3/response_writer.go
```diff
@@ -81,6 +88,75 @@ func (w *responseWriter) Write(p []byte) (int, error) {
...
+
+func (w *responseWriter) writeTrailers() {
+	w.promoteTrailer()
+
+	if !w.hasNonemptyTrailers() {
+		return
+	}
+
+	var headers bytes.Buffer
+	enc := qpack.NewEncoder(&headers)
```
marten-seemann: Same as above. Should this be the same encoder?

http3/body.go
```diff
@@ -60,7 +65,21 @@ func (r *body) readImpl(b []byte) (int, error) {
 			}
 			switch f := frame.(type) {
 			case *headersFrame:
```
marten-seemann: This means that trailers can appear in the middle of a body? Is that valid?

http3/response_writer.go
```diff
@@ -48,6 +52,9 @@ func (w *responseWriter) WriteHeader(status int) {
 
 	for k, v := range w.header {
 		for index := range v {
+			if k == "Trailer" {
```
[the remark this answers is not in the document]
juliens: As `TrailerPrefix` is for "not" announced trailers, there are only handle after `WriteHeader`

http3/response_writer.go
```diff
@@ -48,6 +52,9 @@ func (w *responseWriter) WriteHeader(status int) {
...
+				w.declareTrailer(v[index])
```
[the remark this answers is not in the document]
juliens: We need to keep the header named `Trailer` because the client will need it to know the trailers the server had announced.

http3/response_writer.go
juliens: I'm not sure we can't because the `NewEncoder` needs a specific `io.Writer`.

http3/body.go
```diff
@@ -60,8 +71,30 @@ func (r *body) readImpl(b []byte) (int, error) {
...
+
+				for _, trailer := range trailers {
+					if r.resp.Trailer == nil {
+						r.resp.Trailer = http.Header{}
+					}
+					r.resp.Trailer.Add(trailer.Name, trailer.Value)
+				}
+
+				// Trailer Frame is the last frame.
+				return 0, nil
```
marten-seemann: I'm not sure if returning here is the right thing to do. Shouldn't we continue reading from the stream until we get an EOF, and throw an error if there's any other frame after the trailers?
Furthermore, would it be valid to send two HEADER frames that both contain trailers or is this not possible?

juliens: https://tools.ietf.org/html/rfc7540#section-8.1
```
optionally, one HEADERS frame, followed by zero or more
       CONTINUATION frames containing the trailer-part, if present
```
So for me, if I understand well, it's only one HEADERS frame.

For the return, when I return it will make a `requestDone` and I thought it was enough, but I need to analyse more to be sure.

I took a look at the http2 implementation in the standard library, and, they control that there is an `EndStream` in the `HeaderFrame`, if there is not they raise an error, and then they `endStream` the current stream. I can do the same and it will be enough. WDYT?

joliveirinha:
@marten-seemann are there any plans to include trailers into h3 package?

Having this support is a must have to proxy gRPC, and it would be nice to have it. 
This PR is already a too old, so, not sure if it should start from scratch.

If there are not plans, I can try and open a PR for this in the near future. Let me know if that would be ok.

marten-seemann:
> @marten-seemann are there any plans to include trailers into h3 package?

It would definitely be a nice-to-have. I'm not actively working on this, but if you want to submit a PR, that would be great!

PR #2350, opened 2020-02-14
Read timeout value from http.Server

ohoushyar:
Server considers timeout values from http.Server in handleRequest.

Most of the changes is quite similar to the way net/http/server.go deals with timeouts.

Addresses server side of #410 .

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -137,6 +137,12 @@ func (s *Server) serveImpl(tlsConf *tls.Config, conn net.PacketConn) error {
 		}
 	}
 
+	// Update QUIC MaxIdleTimeout to server IdleTimeout
+	if s.IdleTimeout > 0 {
+		s.QuicConfig.MaxIdleTimeout = s.IdleTimeout
+		s.logger.Debugf("Updated QUIC MaxIdleTimeout to IdleTimeout: %v", s.IdleTimeout)
+	}
+
 	var ln quic.EarlyListener
 	var err error
 	if conn == nil {
@@ -230,6 +236,37 @@ func (s *Server) maxHeaderBytes() uint64 {
 }
 
 func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpack.Decoder, onFrameError func()) requestError {
+
+	var (
+		wholeReadDeadline time.Time
+		hdrDeadline       time.Time
+
+		setReadDeadline bool
+	)
+	t0 := time.Now()
+	// set read timeout
+	if d := s.ReadTimeout; d != 0 {
+		wholeReadDeadline = t0.Add(d)
+		err := str.SetReadDeadline(wholeReadDeadline)
+		if err != nil {
+			s.logger.Errorf("failed to SetReadDeadline on stream %v: %s", str.StreamID(), err)
+			return newStreamError(errorInternalError, err)
+		}
+		s.logger.Debugf("Set read deadline to the ReadTimeout %v on StreamID %v", d, str.StreamID())
+		setReadDeadline = true
+	}
+	// set header timeout
+	if d := s.ReadHeaderTimeout; d != 0 {
+		hdrDeadline = t0.Add(d)
+		err := str.SetReadDeadline(hdrDeadline)
+		if err != nil {
+			s.logger.Errorf("failed to SetReadDeadline on stream %v: %s", str.StreamID(), err)
+			return newStreamError(errorInternalError, err)
+		}
+		s.logger.Debugf("Set read deadline to the ReadHeaderTimeout %v on stream %v", d, str.StreamID())
+		setReadDeadline = true
+	}
+
 	frame, err := parseNextFrame(str)
 	if err != nil {
 		return newStreamError(errorRequestIncomplete, err)
@@ -265,6 +302,16 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
 		s.logger.Infof("%s %s%s", req.Method, req.Host, req.RequestURI)
 	}
 
+	// Adjust the read deadline if necessary
+	if setReadDeadline && !hdrDeadline.Equal(wholeReadDeadline) {
+		err := str.SetReadDeadline(wholeReadDeadline)
+		if err != nil {
+			s.logger.Errorf("failed to SetReadDeadline on stream %v: %s", str.StreamID(), err)
+			return newStreamError(errorInternalError, err)
+		}
+		s.logger.Debugf("Adjusted read deadline to the ReadTimeout %v on StreamID %v", s.ReadTimeout, str.StreamID())
+	}
+
 	ctx := str.Context()
 	ctx = context.WithValue(ctx, ServerContextKey, s)
 	ctx = context.WithValue(ctx, http.LocalAddrContextKey, sess.LocalAddr())
@@ -274,6 +321,16 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
 	handler := s.Handler
 	if handler == nil {
 		handler = http.DefaultServeMux
+	}
+
+	// Done with header, now set the write timeout
+	if d := s.WriteTimeout; d != 0 {
+		err = str.SetWriteDeadline(time.Now().Add(d))
+		if err != nil {
+			s.logger.Errorf("failed to SetWriteDeadline on stream %v: %s", str.StreamID(), err)
+			return newStreamError(errorInternalError, err)
+		}
+		s.logger.Debugf("Set write deadline to %v on stream %v", d, str.StreamID())
 	}
 
 	var panicked, readEOF bool
@@ -304,6 +361,7 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
 	if !readEOF {
 		str.CancelRead(quic.ErrorCode(errorEarlyResponse))
 	}
+
 	return requestError{}
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ main
 mockgen_tmp.go
 *.qtr
 *.qlog
+.DS_Store
+*~
 
 fuzzing/*/*.zip
 fuzzing/*/coverprofile
@@ -14,3 +16,4 @@ fuzzing/*/corpus/
 !fuzzing/frames/single-frame*
 !fuzzing/frames/multiple-frame*
 !fuzzing/header/header*
+
--- a/example/main.go
+++ b/example/main.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	_ "net/http/pprof"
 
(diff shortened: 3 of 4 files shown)


ohoushyar:
@googlebot I signed it!

http3/server.go
```diff
@@ -213,6 +213,23 @@ func (s *Server) maxHeaderBytes() uint64 {
...
 func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpack.Decoder, onFrameError func()) requestError {
+
+	var (
+		hdrDeadline      time.Time
+		wholeReqDeadline time.Time
+		writeDeadline    time.Time
+	)
+
+	t0 := time.Now()
+	if d := s.ReadHeaderTimeout; d != 0 {
```
marten-seemann: The documentation of `http.Server` says:
```
// ReadHeaderTimeout is the amount of time allowed to read
// request headers. The connection's read deadline is reset
// after reading the headers and the Handler can decide what
// is considered too slow for the body. If ReadHeaderTimeout
// is zero, the value of ReadTimeout is used. If both are
// zero, there is no timeout.
```

So don't you need to read the `ReadTimeout` if `ReadHeaderTimeout` is 0?

http3/server.go
```diff
@@ -240,6 +257,15 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
 	}
 
 	req.RemoteAddr = sess.RemoteAddr().String()
+
```
marten-seemann: I think you need to reset the deadline after reading the header (if a deadline was set in the first place), right?

http3/server.go
```diff
@@ -248,6 +274,17 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
 		s.logger.Infof("%s %s%s", req.Method, req.Host, req.RequestURI)
 	}
 
+	// Now set the write timeout
+	t0 = time.Now()
+	if d := s.WriteTimeout; d != 0 {
```
marten-seemann:
```
// WriteTimeout is the maximum duration before timing out
// writes of the response. It is reset whenever a new
// request's header is read. Like ReadTimeout, it does not
// let Handlers make decisions on a per-request basis.
```

I'm not sure I fully understand this, but don't we need to reset the deadline when a new header is read?

http3/server.go
```diff
@@ -213,6 +213,23 @@ func (s *Server) maxHeaderBytes() uint64 {
 }
 
 func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpack.Decoder, onFrameError func()) requestError {
+
+	var (
+		hdrDeadline      time.Time
+		wholeReqDeadline time.Time
+		writeDeadline    time.Time
+	)
```
marten-seemann: It looks like these variables could be scoped to their respective if blocks.

ohoushyar:
Thanks a lot for review. Sure, I'll push the new changes asap.

http3/server.go
ohoushyar: I reset the timeout at the end of handleRequest. I have the feeling that it may not be necessary cause the stream will be closed after handleRequest. But it seems safer in this way.

marten-seemann: I was referring to
```
It is reset whenever a new request's header is read.
```

Does that mean we need to reset the deadline when the server receives a new request?

ohoushyar: We are not reusing stream, are we? The server in handleConn, accepts a Stream, handles it and then closes it. I believe this phrase make more sense for http2 which handles multiple requests in single connection. Please correct me if I'm wrong.

marten-seemann: We're not reusing streams, and HTTP/2 is neither. As HTTP/2, HTTP/3 also handles multiple requests on a single connection.

ohoushyar: Correct. The deadline that we set here is stream level though.

Let me research more Marten.

marten-seemann: Thank you @ohoushayr! I'm not sure how it's supposed to work, just reading the Godoc here. I guess the best thing would be to see how this is implemented in Go's HTTP/2 implementation.

marten-seemann:
@ohoushyar Any updates?

ohoushyar:
Working on it, will explain the process in more detail, stay tuned ...

http3/server.go
ohoushyar: In HTTP/2 at the beginning of ServeConn, session level write timeout will be disarmed. Then the stream level write timeout will be set after reading header frame, at newStream. It will never reset. In stream closing however, the write timeout timer will be reset.

The Godoc is only talks about HTTP/1.x which wouldn't be completely applicable to our implementation.

ohoushyar:
Sorry for the late update, a lot is going on these days.

I've done my research and below is the sequence of handling timeout in HTTP/2:
http1 -> http2
- serve a new connection
- if tls 
    - set read and write deadline
    - Do tls handshake
    - if nextproto -> http2 : upgrade to http2 and handle
- http2:
    In http2 there is no read deadline per stream. There is only
    writedeadline.
    - disarm connection write deadline which was set in h1 as write deadline
    is set per stream in serverConn.newStream
    - set connection/session level h2 idle timeout to h1 idle timeout or h1 read timeout
    - disarm the connection level read deadline at the end of each processHeaderFrame
    - create new stream onProcessHeaderFrame 
       - on new stream set stream level write deadline to write timeout
    - stop write deadline timer on close stream

So I believe the code looks good now. I also removed reset timeouts at the end of handleRequest as the stream will be closed after handleRequest.

My only question is about idleTimeout which is connection/session level. Is it taken care of by QUIC?

Thanks.

ohoushyar:
Hi @marten-seemann, were you able to look at the changes? Does it make sense? Do you think I'm missing something?

http3/server.go
```diff
@@ -230,6 +230,35 @@ func (s *Server) maxHeaderBytes() uint64 {
...
+
+		setReadDeadline bool
+	)
+	t0 := time.Now()
+	// set read timeout
+	if d := s.ReadTimeout; d != 0 {
+		wholeReadDeadline = t0.Add(d)
+		err := str.SetReadDeadline(wholeReadDeadline)
+		if err != nil {
+			return newStreamError(errorSettingsError, err)
```
marten-seemann: This should be an internal error. Or maybe we need a timer here instead of a deadline (see comment)?

http3/server.go
```diff
@@ -265,6 +294,23 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
 		s.logger.Infof("%s %s%s", req.Method, req.Host, req.RequestURI)
 	}
 
+	// Adjust the read deadline if necessary
+	if setReadDeadline && !hdrDeadline.Equal(wholeReadDeadline) {
```
marten-seemann: I believe we can remove this, since H2 doesn't use read deadlines. What do you think?

http3/server.go
```diff
@@ -265,6 +294,23 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
...
+		if err != nil {
+			return newStreamError(errorSettingsError, err)
+		}
+		s.logger.Debugf("Adjusted read deadline to the ReadTimeout on StreamID %v", str.StreamID())
+	}
+	// Done with header, now set the write timeout
+	if d := s.WriteTimeout; d != 0 {
+		err = str.SetWriteDeadline(time.Now().Add(d))
+		if err != nil {
+			return newStreamError(errorSettingsError, err)
```
marten-seemann: This should be an internal error.

ohoushyar:
Thanks @marten-seemann . I'll update it soon.

http3/server.go
ohoushyar: Yes we can remove it and let quic handle ReadTimeout by setting MaxIdleTimeout. Make sense.

http3/server.go
ohoushyar: Yes, sure.

http3/server.go
ohoushyar: Of course.

ohoushyar:
>  I just went through the H2 package, and I'm a bit confused now:

Indeed it's confusing.

> It makes sense to reset the connection level deadlines. We can skip that skip, there's no way to set a connection-level deadline in QUIC anyway. However, there's an idle timeout. The idle timeout timer is reset every time we receive a packet from the peer (and in some other corner cases). So it sounds like it would make sense to set it to the read timeout, doesn't it?
> 
> H2 only resets the read deadline _after_ the header has been read, so we probably should respect the `ReadHeaderTimeout`?

I think it make sense. H2 as well using H1 `IdleTimeout` or `ReadTimeout` as `IdleTimeout` and disarm connection level `ReadDeadline`. I agree that we can remove stream level `ReadDeadline` and let QUIC takes care of it.

> H2 doesn't use a write deadline (via `SetWriteDeadline`), but sets a timer that calls `onWriteTimeout` (see https://github.com/golang/net/blob/118fecf932d8fea033f3b7aca66fba22cfdd517d/http2/server.go#L1763-L1767). This function then sends an internal error. Do we need to do the same?

It doesn't, but I can see that you've already implemented using deadline in `stream.Write`. Do you see any advantage of using the timer like the way H2 does over the current Write implementation?

ohoushyar:
Hi @marten-seemann,

Hope you're doing well. I've done some more research about the idle timeout cause it was a little unclear to me, specially in the case that the ReadTimeout is shorter than WriteTimeout and there is no IdleTimeout and there is only one stream. The stream request had been processed and now server is working on handler, however the handler runtime is longer than read timeout but it's withing write timeout, QUIC will use read and close the connection even before the write timeout is exceeded. So apparently we cannot just use read timeout. Imagine we set read timeout and it is used as idle ... :) (probably I didn't clearly understand H2 process)

It's tricky the way H2 handles IdleTimeout. IdleTimeout is either H1.IdleTimeout or H2.ReadTimeout and it's a connection level property.

At the beginning of the serve, an idleTimer will be set. Then it will be stopped on processHeaders (it make sense as it relies on write timer now). So as long as there are active streams on a connection it will remain stopped. Once all streams are closed, it will be reactivated/reset again.

Now on H3, we should probably do the same thing. We set the quic.MaxIdleTimeout to server IdleTimeout or ReadTimeout at the beginning and start listening. When the first stream started, we should set it to zero or default value which is 30s (not sure if it's possible though). Then after all streams are closed, we set it back to server value timeout (idle or read).

Does it make sense? What do you think? (I hope I could articulate my point properly)

ohoushyar:
Hi @marten-seemann,
 
We don't need to do any of the stuff I said in my last comment if we don't use `ReadTimeout` as `IdleTimeout`. I believe it make more sense to just set stream `ReadDeadline` and then at the end reset it. Also from code point of view, it looks cleaner.

Thanks again for all feedback.

http3/server.go
ohoushyar: If we want to remove it we have to disable QUIC idleTimeout timer after parsing header. Let's keep it as it is.

http3/server.go
ohoushyar: I saw that it's been used like [`_ = str.SetWriteDeadline(...)`](https://github.com/lucas-clemente/quic-go/blob/4ae1a1350368c2e21f4213b71faf492f487e14bf/stream.go#L139) as it would never return err  but I thought maybe that would be the future improvement to add error. Nevertheless, I changed it to internal error.

I can remove the error handling though ...

marten-seemann:
Hi ohoushyar, yes, that makes sense, let's use the QUIC idle timeout. For the stream deadline, should we use timed function like H2 (see https://github.com/lucas-clemente/quic-go/pull/2350#pullrequestreview-378975942)?

ohoushyar:
The answer is Yes. We have to send the error to client instead of just closing the stream:
`
Get "https://127.0.0.1:6121/demo/tile": EOF
exit status 1
`

I'll update the code. Thanks for bringing it up again.

ohoushyar:
Hi @marten-seemann, I have done quite a bit of research on the way H1 and H2 deal with write timeout/deadline. Long story short, I think we can proceed with using deadline. I tested different scenarios, using writeTimeout, TimeoutHandler on both H1 and H2. I saw the difference but couldn't really find the advantage of using timer. Using timer seems to be a significant change, but again it's up to you guys. You're more experienced than me and I'm here to help. I really enjoy working on this code.

Thanks,
Omid.

PR #2351, opened 2020-02-14
Do not ignore loss timer

zverevm:
In _setLossDetectionTimer_ the first assignment to _h.alarm_ is redundant, since it is going to be overwritten.

### DIFF
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -372,6 +372,7 @@ func (h *sentPacketHandler) setLossDetectionTimer() {
 	if lossTime, _ := h.getEarliestLossTimeAndSpace(); !lossTime.IsZero() {
 		// Early retransmit timer or time loss detection.
 		h.alarm = lossTime
+		return
 	}
 
 	// Cancel the alarm if no packets are outstanding


zverevm:
@googlebot I signed it!

marten-seemann:
Thank you for finding this!

I don't think we need the comparison with the PTO timer is necessary though. It should be enough to return from the function when the loss timer is set, as shown in the pseudo code here: https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-setting-the-loss-detection-

Can you update the PR?

marten-seemann:
Also, it would be really helpful to have a test case for this.

zverevm:
I think I just messed up the PR... Sorry, I am quit new to git. Already fixing it.

marten-seemann:
@zverevm Something seems wrong with the PR. Can you rebase and make sure that there's only a single commit in the PR?

marten-seemann:
That looks a lot better now :)

Would you mind adding a test case for this? The most useful test here would be one that fails without your fix and doesn't fail with your fix.

zverevm:
I came up with this not through experimentation, but by looking where should I insert new code for my experiments. This is why I don't have any log or circumstances under which the impact of this change can be seen.

As stated in the [recovery draft](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#section-5.2.1-9), the time threshold for early retransmit is expected to both expire earlier than the PTO and be less likely to spuriously retransmit data. This change takees into account the earliest loss time for

Analyzing the code (and the draft), these are the conditions in which the impact of this change will be clearly seen:
 - sentPacketHandler.rttStats.smoothedRTT >= sentPacketHandler.rttStats.latestRTT
 - sentPacketHandler.rttStats.smoothedRTT * ackhandler.timeThreshold > protocol.TimerGranularity
 - sentPacketHandler.rttStats.smoothedRTT > 0
 - 4 * sentPacketHandler.rttStats.meanDeviation > protocol.TimerGranularity
 - sentPacketHandler.ptoCount >= 1
 - Not receiving an ACK for enough time for triggering both timeouts.

**With this change**: Only the lost packet is retransmitted, earlier than with PTO.
**Without this change**: Lost packet will be retransmitted later, either as one of the 2 probe packets, or after the 2 probe packets as a response to the stimulated ACK.

In [session.go:527](https://github.com/lucas-clemente/quic-go/blob/master/session.go#L527), in the run loop, we are checking the timeout defined by `sentPacketHandler.alarm`. If this alarm points to a moment in the past, then `sentPacketHandler.OnLossDetectionTimeout()` calls `sentPacketHandler.onVerifiedLossDetectionTimeout()`, which detects if this timeout is due to early retransmit timeout or PTO and acts accordingly. If we never set early retransmit timeout, we will only be checking PTO.

According to [recovery draft](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#time-threshold), time threshold for early retransmit is calculated as
`max(kTimeThreshold * max(smoothed_rtt, latest_rtt), kGranularity)`

Assuming `smoothed_rtt >= latest_rtt`, and
`kTimeThreshold /* 9.0 / 8 */ * smoothed_rtt > kGranularity`, loss time is `kTimeThreshold * smoothed_rtt`.

Looking at the [pseudocode](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-setting-the-loss-detection-), we see how PTO timeout is defined. Assuming
`smoothed_rtt > 0 && 4 * rttvar > kGranularity`, PTO timeout is
`(smoothed_rtt + 4 * rttvar + max_ack_delay) * (2 ^ pto_count)`.

After 1 PTO timeout, the next PTO timeout will be `(smoothed_rtt + (more things)) * 2`, which is bigger than early retransmission timeout, `9.0 / 8 * smoothed_rtt`.

If were asking me to modify sent_packet_handler_test.go, it will take some time... :)

PR #2360, opened 2020-02-21
remove unused parameter from SentPacketHandler.ReceivedAck

marten-seemann:

Issue #1862, opened 2019-04-11
implement the amplification protection


marten-seemann:
This will require a major refactoring how we send packets. Furthermore, this also has implications how we retransmit packets, since unblocking the server requires the client so sometimes send fully padded packets.

See https://github.com/quicwg/base-drafts/pull/2590

PR #2362, opened 2020-02-21
implement the client side for the amplification protection

marten-seemann:
See #1862. Depends on #2360.

As a client, we need to set a timer to send

- an Initial packet, even if all Initial data has already been acknowledged
- a Handshake packet, if we've received a Handshake packet, but don't have any Handshake data to send (yet)


marten-seemann:
I messed up merging this :( Had to revert and will open a new PR.

marten-seemann:
Actually this was ok. Should have merged #2360 first, but it doesn't really matter.

PR #2371, opened 2020-02-23
Make http3.client.Close() succeed if session was not started

lorenzosaino:
Invoking `http3.client.Close()` before `client.dial()` is invoked causes a segmentation fault. That occurs because, in this circumstance, invoking `client.Close()` results in invoking `client.session.CloseWithError(...)` while `client.session` is `nil`.

This pull request changes the behavior of `http3.client.Close()` to return `nil` if `client.session` is `nil` and adds an associated test case.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -127,6 +127,9 @@ func (c *client) setupSession() error {
 }
 
 func (c *client) Close() error {
+	if c.session == nil {
+		return nil
+	}
 	return c.session.CloseWithError(quic.ErrorCode(errorNoError), "")
 }
 
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -146,6 +146,12 @@ var _ = Describe("Client", func() {
 		Expect(err).To(MatchError(testErr))
 	})
 
+	It("closes correctly if session was not created", func() {
+		client = newClient("localhost:1337", nil, &roundTripperOpts{}, nil, nil)
+		err := client.Close()
+		Expect(err).ToNot(HaveOccurred())
+	})
+
 	Context("Doing requests", func() {
 		var (
 			request *http.Request


marten-seemann:
@lorenzosaino How can this happen?

lorenzosaino:
@marten-seemann that happens for example if you create an `http.Client` with `http3.RoundTripper` as transport, try to issue a request to an unreachable destination and then you try to close the roundtripper. Here is a reproducer:

```go
package main

import (
	"fmt"
	"net/http"
	"time"

)

func main() {
	url := "https://1.2.3.4" // Unreachable

	transport := &http3.RoundTripper{
		QuicConfig: &quic.Config{},
	}

	client := http.Client{
		Timeout:   time.Second,
		Transport: transport,
	}

	req, err := http.NewRequest("GET", url, nil)
	if err != nil {
		panic(err)
	}

	// Issue request
	response, err := client.Do(req)
	if err != nil {
		fmt.Printf("Error issuing request: %s\n", err)
	}

	// Close body (irrelevant for this example)
	if response != nil {
		fmt.Println("Closing response body")
		response.Body.Close()
	}

	// Crashes without patch
	transport.Close()

	fmt.Println("Done")
}
```

marten-seemann:
This seems to be a consequence of https://github.com/lucas-clemente/quic-go/issues/765 then. Would you agree that this would be the proper fix?

lorenzosaino:
Thanks a lot for the feedback and for pointing out to that issue. Yes, fixing #765 would fix a more general problem and would prevent this specific issue too. I'd be more than happy to contribute a fix to #765 if that's welcome.

One question I have though is: would you rather have a fix for #765 in lieu of this fix or do you think there would be value in applying both?

I am asking because (if I understand correctly) while a fix for #765 would prevent clients with failed connections to be kept at the roundtripper (and then attempting to close them), the issue remains that closing a newly instantiated client will result in a segfault rather than being a no-op as one would intuitively expect. Or am I missing something?

marten-seemann:
> Yes, fixing #765 would fix a more general problem and would prevent this specific issue too. I'd be more than happy to contribute a fix to #765 if that's welcome.

A PR would be very much appreciated. The logic behind #765 is kind of involved, please have a look at the HTTP/2 standard library implementation, they had to solve a very similar problem for HTTP/2.

> I am asking because (if I understand correctly) while a fix for #765 would prevent clients with failed connections to be kept at the roundtripper (and then attempting to close them), the issue remains that closing a newly instantiated client will result in a segfault rather than being a no-op as one would intuitively expect.

You mean you create the client, and immediately close it (before the quic.Dial has returned)? That would make sense then.

lorenzosaino:
> A PR would be very much appreciated. The logic behind #765 is kind of involved, please have a look at the HTTP/2 standard library implementation, they had to solve a very similar problem for HTTP/2.

Thanks a lot for the pointer. I will look into the HTTP/2 standard library and work on a fix.

> You mean you create the client, and immediately close it (before the quic.Dial has returned)? That would make sense then.

Yes, this fix prevents segfaults when an `http3.Client` is created and closed before a dial successfully returns. Here are two example cases where this happens.

Close client before attempting a dial:
```go
client = newClient("localhost:1337", nil, &roundTripperOpts{}, nil, nil)
err := client.Close()  // segfaults
```

Attempt dial with unsupported HTTP scheme:
```go
client = newClient("localhost:1337", nil, &roundTripperOpts{}, nil, nil)
req, _ := http.NewRequest("GET", "http://www.example.com", nil) // 'http', not 'https'
client.RoundTrip(req)  // returns error
err := client.Close()  // segfaults
```

PR #2385, opened 2020-02-25
run the internal and http3 tests with race detector on Travis

marten-seemann:
Depends on #2380.


marten-seemann:
@lucas-clemente Please have another look. I had to make a lot of changes to this PR.
Note the trick using `go:linkname` in the handshake tests to access unexported functions in the qtls package without any external dependencies.

PR #2396, opened 2020-03-02
fix ChaCha20 header protection

marten-seemann:


lucas-clemente:
No test? ;)

marten-seemann:
I confirmed interop with picoquic, and thinking about adding a dedicated interop runner test case using ChaCha.

PR #2408, opened 2020-03-05
update GolangCI-Lint to v1.23.8

marten-seemann:


internal/handshake/crypto_setup_test.go
```diff
@@ -54,6 +54,8 @@ func (s *stream) Write(b []byte) (int, error) {
 var _ = Describe("Crypto Setup TLS", func() {
 	var clientConf, serverConf *tls.Config
 
+	// unparam incorrectly complains that the first argument is never used.
+	//nolint:unparam
```
marten-seemann: False positives are annoying...

internal/handshake/crypto_setup_test.go
```diff
@@ -54,6 +54,8 @@ func (s *stream) Write(b []byte) (int, error) {
...
 	initStreams := func() (chan chunk, *stream /* initial */, *stream /* handshake */) {
 		chunkChan := make(chan chunk, 100)
```
lucas-clemente: The syntax here looks weird to me, maybe that's why the linteer is complaining?

PR #2438, opened 2020-03-25
pass the role (client or server) to the GetLogWriter callback

marten-seemann:
@lucas-clemente Since this is a breaking API change, we'll have to bump the version number to v0.16 with the next release, right?

### DIFF
--- a/client.go
+++ b/client.go
@@ -180,7 +180,7 @@ func dialContext(
 
 	var qlogger qlog.Tracer
 	if c.config.GetLogWriter != nil {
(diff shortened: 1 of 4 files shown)


lucas-clemente:
Ack on the version bump. If you want to avoid that you could rename the function as part of this (maybe `LogWriter` instead of `GetLogWriter`?)

marten-seemann:
Thinking a bit more about this, we probably don't need to do this at all.
Both `quic.Listen` and `quic.Dial` take a `quic.Config`, so the caller will already know which role is being logged.

PR #2442, opened 2020-03-27
Pluggable congestion control & example

tobyxdd:
This PR fulfills the feature request #776 with a new API `session.SetCongestion` and some other minor changes here and there. Users can now implement and use their own CC on a per session basis.

A simple example is provided in `example/custom_congestion`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,95 @@ fuzzing/*/corpus/
 !fuzzing/frames/single-frame*
 !fuzzing/frames/multiple-frame*
 !fuzzing/header/header*
+
+# Created by https://www.gitignore.io/api/intellij+all
+# Edit at https://www.gitignore.io/?templates=intellij+all
+
+### Intellij+all ###
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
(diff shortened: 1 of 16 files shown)


tobyxdd:
@googlebot I signed it!

marten-seemann:
Thank you for this PR.

I'm very hesitant to merge anything going in this direction, for the reason outlined in https://github.com/lucas-clemente/quic-go/issues/776#issuecomment-319321926: The current congestion controller interface is well suited for Reno and Cubic, but probably won't work well for other congestion controllers like BBR.

Furthermore, I don't think setting it on the session is a good solution, since you'd always miss the handshake phase of the connection.

tobyxdd:
Thanks for the response! I can fully understand your concerns. But even if SendAlgorithm interface changes in the future, I don't think it will affect how this new API works. Perhaps we can add a warning `(Warning: This API should not be considered stable and might change soon)` as some other APIs already have?

The current code uses default Cubic CC for the handshake phase of a connection. Maybe it's better to let users pass a factory-style func `newSendAlgorithm()` via Config?

marten-seemann:
That would still be a breaking API change, since every implementation of this interface would have to change.

> The current code uses default Cubic CC for the handshake phase of a connection. Maybe it's better to let users pass a factory-style func newSendAlgorithm() via Config?

That would definitely be cleaner, although it wouldn't solve the first problem.

tobyxdd:
Thanks. So it seems this PR can't be merged either way, should I close it?

PR #2443, opened 2020-03-27
use a buffered writer for http3 requests and responses

marten-seemann:


http3/request_writer.go
```diff
@@ -37,22 +38,47 @@ func newRequestWriter(logger utils.Logger) *requestWriter {
...
+		}
 		str.Close()
 		return nil
 	}
 
 	// send the request body asynchronously
 	go func() {
-		if err := w.sendRequestBody(req.Body, str); err != nil {
+		defer req.Body.Close()
+		b := make([]byte, 8*1024)
```
lucas-clemente, line 58: Nit: make a constant?

marten-seemann: Done.

PR #2445, opened 2020-03-27
qlog the packet_received event before any frame-specific events

marten-seemann:
Otherwise the qvis will look very messy.

### DIFF
--- a/session.go
+++ b/session.go
@@ -889,8 +889,12 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
 		if s.traceCallback != nil || s.qlogger != nil {
 			frames = append(frames, frame)
 		}
-		if err := s.handleFrame(frame, packet.encryptionLevel); err != nil {
-			return err
+		// Only process frames now if we're not logging.
+		// If we're logging, we need to make sure that the packet_received event is logged first.
+		if s.qlogger == nil {
+			if err := s.handleFrame(frame, packet.encryptionLevel); err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -889,8 +889,12 @@ func (s *session) handleUnpackedPacket(packet *unpackedPacket, rcvTime time.Time
 		if s.traceCallback != nil || s.qlogger != nil {
 			frames = append(frames, frame)
 		}
-		if err := s.handleFrame(frame, packet.encryptionLevel); err != nil {
-			return err
+		// Only process frames now if we're not logging.
+		// If we're logging, we need to make sure that the packet_received event is logged first.
+		if s.qlogger == nil {
```
lucas-clemente, line 894: This looks weird to me. Either there's a reason why we want to abort early (e.g. DOS), and then that shouldn't be affected by whether we're logging or not. Or, there's no reason, and then the below can be the default path?

marten-seemann: The `packet_received` event contains a list of all frames. I want to log that event **before** logging any events that were triggered by that packet (e.g. key updates, timer updates etc.).

lucas-clemente: Why do we immediately handle the frames in the non-logged path?

marten-seemann: To avoid allocating a slice of frames. If we want to live without allocations, we could pre-allocate a slice (and save it as a member variable to the session). Too bad Go doesn’t support static variables.

lucas-clemente: Hm, fair enough. It's somewhat bug-prone for enabling a logger to change the behavior of the code, as this can be very non-obvious when debugging. But this sounds like an acceptable exception.

(FWIW, a static wouldn't help here because this is concurrent code.)

marten-seemann: > (FWIW, a static wouldn't help here because this is concurrent code.)

It's not. Packets are processed sequentially in the session.

PR #2446, opened 2020-03-27
Revert "use Go 1.14's embedded interfaces, update CIs"

Stebalien:
This reverts commit c88a69034d6b6d288a01578f5256e90aca95428a.

Unfortunately, go 1.14's new scheduler and timer changes have a few bugs that need to be ironed out before it's ready for production use.

See:
* https://github.com/libp2p/go-libp2p/issues/858
* https://github.com/golang/go/issues/38119

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 executors:
   test:
     docker:
-      - image: "circleci/golang:1.14"
+      - image: "circleci/golang:1.13"
     environment:
       runrace: true
   interop:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ group: travis_latest
 language: go
 
 go:
-  - "1.14.x"
+  - "1.13.x"
 
 # first part of the GOARCH workaround
 # setting the GOARCH directly doesn't work, since the value will be overwritten later
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ If you want to use quic-go as a library in other projects, please consider using
 
 ## Guides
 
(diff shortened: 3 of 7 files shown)


Stebalien:
Looks like we'll also need to downgrade qtls.

marten-seemann:
@Stebalien Looking at the discussion in https://github.com/golang/go/issues/38119, is my understanding correct that we don't need to do this downgrade as long as we compile everything with Go 1.14.2?

Stebalien:
I'd still like to downgrade.

Unfortunately, it's unclear when 1.14.2 will be released (and the go-ipfs release is already late). Furthermore, I expect these won't be the last bugs in 1.14.2 as it has quite a few changes to critical parts of the go scheduler.

marten-seemann:
This is a situation that Go Modules seems not well equipped to handle.

The issue here is that in Go 1.14, crypto/tls made some significant changes regarding the certificate selection, e.g. by adding additional fields and functions to the `tls.ClientHelloInfo` (which I then transferred over to qtls). If we want to downgrade to Go 1.13, we also have to downgrade qtls to an earlier version. Which means that users of quic-go who use Go 1.14 will find those new features broken / not implemented.

As far as I can see, there's no way of telling Go to use version *x* of a dependency when built with Go 1.13 and version *y* if built with Go 1.14. Or am I missing something?

Stebalien:
You'd either have to release a qtls version with go 1.13 support using build flags, or downgrade. We can also do this in a set of separate branches.

lucas-clemente:
Can you just downgrade quic-go to the previous version in libp2p?

marten-seemann:
Not really. We need the qlogs to evaluate and debug transfers, and that only added in the last few weeks.

PR #2465, opened 2020-04-05
make sure that qlog events are logged in order

### LINKED ISSUE #2464: concurrent qlogging causes unordered qlogs
In the `cryptoSetup`, we qlog the key update events in `SetReadKey()` and `SetWriteKey()`. Those functions are called from qtls, which is ultimately driven by `RunHandshake()`, which is running in a separate go routine.

marten-seemann:
Fixes #2464.

### DIFF
--- a/qlog/qlog.go
+++ b/qlog/qlog.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/lucas-clemente/quic-go/internal/congestion"
@@ -40,6 +41,8 @@ type Tracer interface {
 }
 
 type tracer struct {
+	mutex sync.Mutex
+
 	w             io.WriteCloser
 	odcid         protocol.ConnectionID
 	perspective   protocol.Perspective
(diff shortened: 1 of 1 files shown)


qlog/qlog.go
```diff
@@ -139,13 +142,15 @@ func (t *tracer) StartedConnection(local, remote net.Addr, version protocol.Vers
 	if !ok {
 		return
 	}
+	t.mutex.Lock()
```
lucas-clemente, line 145: If you're doing this every time you call recordEvent, why not move the lock/unlock into that function?

marten-seemann: The thing this mutex protects is the call to `time.Now()`. We need to ensure that events are ordered when we write them to the qlog.

lucas-clemente: Then why not do the time.Now() within recordEvent?

marten-seemann: 1. Some events produce two log entries. Would be nice to have the same time stamp for the them.
2. `SetLossTimer` uses the difference between the timestamp of the event and the deadline.

lucas-clemente: sgtm

PR #2467, opened 2020-04-06
optimize initialization of the Initial AEAD

marten-seemann:
An attacker can make use initialize the Initial AEAD for every Initial packet it sends. This code path should therefore be as efficient as possible.

The standard library uses `cryptobyte.Builder` in `hkdfExpandLabel`. This costs quite a bit of performance. Using an optimized implementation speeds up the initialization of the AEAD used for the Initial encryption level by ~15%.

Using `qtls.HkdfExpandLabel`:
```
BenchmarkNewInitialAEAD-16        719976             16067 ns/op           10488 B/op        171 allocs/op
```

Using the optimized `hkdfExpandLabel`:
```
BenchmarkNewInitialAEAD-16        827632             13754 ns/op            8712 B/op        115 allocs/op
```

### DIFF
--- a/internal/handshake/hkdf.go
+++ b/internal/handshake/hkdf.go
@@ -1,0 +1,29 @@
+package handshake
+
+import (
+	"crypto"
+	"encoding/binary"
+
+	"golang.org/x/crypto/hkdf"
+)
+
+// hkdfExpandLabel HKDF expands a label.
+// Since this implementation avoids using a cryptobyte.Builder, it is about 15% faster than the
+// hkdfExpandLabel in the standard library.
+func hkdfExpandLabel(hash crypto.Hash, secret, context []byte, label string, length int) []byte {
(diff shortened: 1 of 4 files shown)


internal/handshake/hkdf.go
```diff
@@ -0,0 +1,29 @@
+package handshake
+
+import (
+	"crypto"
+	"encoding/binary"
+
+	"golang.org/x/crypto/hkdf"
+)
+
+// hkdfExpandLabel HKDF expands a label.
```
lucas-clemente, line 10: Why not change qtls's implementation instead?

marten-seemann: That would be an option. Just harder to test, since I'd like to run it against the standard library HKDF-Expand-Label. Unfortunately, it's not exported though.

PR #2468, opened 2020-04-06
don't create a new session for invalid Initial packets

marten-seemann:
@ptrd reported the following error: When we receive an Initial packet with a corrupted SCID, we'd still create a new session (and drop the Initial packet, since it doesn't get past the AEAD). However, the session is still created, and once we receive (the retransmission of) a valid Initial packet, we'd then use the corrupted SCID to reply to it.

This PR implements a check before creating the session. Basically, we create the AEAD in the server and decrypt the packet there. Only if it decrypts, we create the session. This is based on the realization that creating the AEAD is an order of magnitude more expensive than an AEAD operation (#2467 doesn't change this significantly):
```
BenchmarkAEADCreation-16    	  720339	     17222 ns/op	   10488 B/op	     171 allocs/op
BenchmarkAEADOpen-16        	13196452	       923 ns/op	    2696 B/op	       2 allocs/op
```

The sad thing is that we'll still decrypt the packet twice (and we have to copy all the packet data into a new buffer, to avoid touching the packet, which will have to be decrypted later). That means we're taking a (small) performance hit in the common case, in order to catch a problem in a corner case.

### DIFF
--- a/internal/handshake/crypto_setup.go
+++ b/internal/handshake/crypto_setup.go
@@ -138,10 +138,12 @@ func NewCryptoSetupClient(
 	qlogger qlog.Tracer,
 	logger utils.Logger,
 ) (CryptoSetup, <-chan *wire.TransportParameters /* ClientHello written. Receive nil for non-0-RTT */) {
+	initialSealer, initialOpener := NewInitialAEAD(connID, protocol.PerspectiveClient)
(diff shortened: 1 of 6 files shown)


marten-seemann:
#2477 should be a better solution for this.

PR #2492, opened 2020-04-14
don't handle version negotiation packets concurrently

marten-seemann:


marten-seemann:
I think the point of this was to not block the loop that's reading from the packet conn.

PR #2498, opened 2020-04-16
use the TIMESCALE_FACTOR in the qlog tests

marten-seemann:


lucas-clemente:
At some point we should probably DI a clock :/

PR #2504, opened 2020-04-18
make it more likely that a STREAM frame is bundled with the FIN

### LINKED ISSUE #2435: bundle STREAM frame and FIN
The common pattern of
```go
str.Write(b)
str.Close()
```
should (most of the time, at least) result in the generation of a single STREAM frame.

The problem here is that `str.Write(b)` doesn't return until all of `b` has been dequeued from the stream, i.e. until the STREAM frame has been generated. In most cases, this leads to packing of a packet, and the FIN will end up in a separate (empty) STREAM frame, which is sent out in a separate packet.

Things get complicated due to the contract of the `io.Writer`. We can only use `b` before returning `Write`, after this `b` belongs to the caller. What we could do is copy `b` into a buffer in the stream and return `Write` immediately (only blocking `Write` if the buffer has not been sent out, to provide back-pressure). Then `Close` has a good chance of being called before the packet is packed, so the FIN will end up in the same packet (and ideally, as a FIN bit on the data-carrying STREAM frame). However, this means that we'll do another copy of the data, so this is a suboptimal solution performance-wise.

marten-seemann:
Fixes #2435.

Currently, on `Stream.Write()`, we save the slice in `dataForWriting`, and pop STREAM frames from that slice. `Write` only returns after all the data has been popped in STREAM frames.

This PR changes this logic: As soon as `dataForWriting` becomes smaller than a certain size `x`, it's copied to a STREAM frame, which can the be popped the next time we assemble a packet. This allows us to return `Write()` when all data but `x` bytes have been sent out. When the user now calls `Close()`, this is much more likely to happen *before* we popped that last STREAM frame, allowing us to set the FIN bit on that frame (and avoiding sending out an empty STREAM frame with FIN set in a separate packet).

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -46,7 +46,8 @@ type sendStream struct {
 	finSent           bool // set when a STREAM_FRAME with FIN bit has been sent
 	completed         bool // set when this stream has been reported to the streamSender as completed
 
-	dataForWriting []byte
+	dataForWriting []byte // during a Write() call, this slice is the part of p that still needs to be sent out
+	nextFrame      *wire.StreamFrame
 
 	writeChan chan struct{}
 	deadline  time.Time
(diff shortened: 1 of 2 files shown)


send_stream.go
```diff
@@ -108,27 +109,46 @@ func (s *sendStream) Write(p []byte) (int, error) {
 		notifiedSender bool
 	)
 	for {
-		bytesWritten = len(p) - len(s.dataForWriting)
-		deadline := s.deadline
-		if !deadline.IsZero() {
-			if !time.Now().Before(deadline) {
-				s.dataForWriting = nil
-				return bytesWritten, errDeadline
+		var copied bool
```
lucas-clemente, line 112: We should probably document the packing strategy somewhere in the code – even with the PR description it's quite hard to follow this.

PR #2512, opened 2020-04-30
Allow setting MTU on Session

MagicalTux:
This adds `SetMaxPacketSize(packetSize uint64)` to the Session interface, and allows setting outgoing maximum packet size in bytes.

See https://github.com/lucas-clemente/quic-go/issues/2511 for details.

Note that this sets the size of UDP payloads, the actual packet will also include the UDP header (8 bytes), IP header (40 bytes for IPv6, 20 bytes or more for IPv4), and possibly ethernet header (18 bytes).

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -187,6 +187,9 @@ type Session interface {
 	// It blocks until the handshake completes.
 	// Warning: This API should not be considered stable and might change soon.
 	ConnectionState() ConnectionState
+	// SetMaxPacketSize allows setting the maximum packet size for outgoing
+	// packets, only use if you know what you are doing.
+	SetMaxPacketSize(packetSize uint64)
 }
 
 // An EarlySession is a session that is handshaking.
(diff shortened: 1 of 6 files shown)


marten-seemann:
The PR in its current form is racy.

To be honest, I don't think that the benefits of setting the MTU are high enough to justify another addition to our public API. PMTUD seems to be the better solution here.

MagicalTux:
PMTUD would be welcome addition to quic-go, and would solve https://github.com/lucas-clemente/quic-go/issues/94 which has been open since May 2016. I do not however the time nor the resources to dig that deep into `quic-go` in order to implement it, and it would be counter productive in my case. I already know the MTU when initializing the QUIC session, so I would want any QUIC PMTUD disabled with the ability to pass the value I already have measured.

I wouldn't mind if this API wasn't part of the public Session interface, but instead could be tested and accessed from outside, that'd work just fine for me.

Issue #2520, opened 2020-05-04
Make tests pass with the race detector enabled


Stebalien:
Running the tests with the race detector enabled reveals several races but it's unclear which ones are just test issues and which ones are issues with the actual code.

marten-seemann:
This only applies to the `quic` package. Most (and with #2385 all) other packages can be run with race detector.

Most importantly, we run the integration tests with race detector on CI: https://github.com/lucas-clemente/quic-go/blob/c0b6d4e141d6dca43023a8c9ae0f437bb943f182/.circleci/config.yml#L45-L47

Running the `quic` tests with race detector would be super nice, but unfortunately a lot of work, since we didn't pay attention to make the tests race-free back when we started the project a few years ago...

Stebalien:
I'd at least try running them with the race detector and see what it spits out. I saw several issues that looked real.

PR #2524, opened 2020-05-04
fix buffer use after it was released when sending an INVALID_TOKEN error

marten-seemann:
Related to #2520.

### DIFF
--- a/server.go
+++ b/server.go
@@ -292,7 +292,7 @@ func (s *baseServer) handlePacket(p *receivedPacket) {
 	}
 }
 
-func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* was the packet handled */ {
+func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* should the buffer be released */ {
 	// If we're creating a new session, the packet will be passed to the session.
 	// The header will then be parsed again.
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -292,7 +292,7 @@ func (s *baseServer) handlePacket(p *receivedPacket) {
 	}
 }
 
-func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* was the packet handled */ {
+func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* should the buffer be released */ {
```
lucas-clemente, line 295: This seems quite brittle, similar to manual mem management in C. Is there a way to have all free-ing always be done by a single function? If not, maybe it's worth copying the buf contents in those cases? I think we could take a small perf hit over the risk of serious security issues…?

Is there a way we can avoid similar bugs in the future, make sure we always release buffers properly? With tests we probably can't guarantee full coverage.

marten-seemann: The race detector will help us with that. In fact, that's how I found this issue.
When you use a buffer that you already released, and this buffer is used again, this is a race condition. Even if the detection is not 100% guaranteed, at least the test will be flaky...

PR #2536, opened 2020-05-11
implement the 3x amplification limit

### LINKED ISSUE #1862: implement the amplification protection

marten-seemann:
Fixes #1862.

### DIFF
--- a/integrationtests/self/self_suite_test.go
+++ b/integrationtests/self/self_suite_test.go
@@ -3,19 +3,23 @@ package self_test
 import (
 	"bufio"
 	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
 	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
 	"flag"
 	"fmt"
 	"io"
 	"log"
-	"math/rand"
+	"math/big"
+	mrand "math/rand"
 	"os"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/lucas-clemente/quic-go"
-
-	"github.com/lucas-clemente/quic-go/internal/testdata"
 	"github.com/lucas-clemente/quic-go/internal/utils"
 
 	. "github.com/onsi/ginkgo"
@@ -23,19 +27,6 @@ import (
 )
 
 const alpn = "quic-go integration tests"
-
-func getTLSConfig() *tls.Config {
-	conf := testdata.GetTLSConfig()
-	conf.NextProtos = []string{alpn}
-	return conf
-}
-
-func getTLSClientConfig() *tls.Config {
-	return &tls.Config{
-		RootCAs:    testdata.GetRootCA(),
-		NextProtos: []string{alpn},
-	}
-}
 
 const (
 	dataLen     = 500 * 1024       // 500 KB
@@ -93,6 +84,11 @@ var (
 	logBufOnce  sync.Once
 	logBuf      *syncedBuffer
 	enableQlog  bool
+
+	caPrivateKey   *rsa.PrivateKey
+	ca             *x509.Certificate
+	leafPrivateKey *rsa.PrivateKey
+	leafCert       *x509.Certificate
 )
 
 // read the logfile command line flag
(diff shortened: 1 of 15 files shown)


packet_packer.go
```diff
@@ -342,37 +342,45 @@ func (p *packetPacker) PackCoalescedPacket() (*coalescedPacket, error) {
 	return packet, nil
 }
 
-func (p *packetPacker) packCoalescedPacket(buffer *packetBuffer) (*coalescedPacket, error) {
+func (p *packetPacker) packCoalescedPacket(buffer *packetBuffer, maxPacketSize protocol.ByteCount) (*coalescedPacket, error) {
+	maxPacketSize = utils.MinByteCount(maxPacketSize, p.maxPacketSize)
+	if p.perspective == protocol.PerspectiveClient {
+		maxPacketSize = protocol.MinInitialPacketSize
```
lucas-clemente, line 348: So as client we're always ignoring the parameter? Should we make that clearer, e.g. maxPacketSizeForServer?

internal/ackhandler/sent_packet_handler.go
```diff
@@ -49,8 +49,12 @@ type sentPacketHandler struct {
 	handshakePackets *packetNumberSpace
 	appDataPackets   *packetNumberSpace
 
-	peerCompletedAddressValidation bool
-	handshakeComplete              bool
+	peerCompletedAddressValidation bool // always true for the server
```
lucas-clemente, line 52: Can you make the difference between `peerAddressValidated` and `peerCompletedAddressValidation` clearer, either through better naming or documentation?

internal/ackhandler/sent_packet_handler.go
```diff
@@ -683,6 +703,16 @@ func (h *sentPacketHandler) ShouldSendNumPackets() int {
...
 }
 
+func (h *sentPacketHandler) AmplificationWindow() protocol.ByteCount {
+	if h.peerAddressValidated {
+		return protocol.MaxByteCount
+	}
+	if h.bytesSent >= 3*h.bytesReceived {
+		return 0
+	}
+	return 3*h.bytesReceived - h.bytesSent
```
lucas-clemente, line 713: Should the magic number be a constant?

session.go
```diff
@@ -1427,7 +1428,7 @@ func (s *session) sendPacket() (bool, error) {
 
 	if !s.handshakeConfirmed {
 		now := time.Now()
-		packet, err := s.packer.PackCoalescedPacket(protocol.MaxByteCount)
+		packet, err := s.packer.PackCoalescedPacket(s.sentPacketHandler.AmplificationWindow())
```
lucas-clemente, line 1431: Is there a risk for busy-looping here, if we exit the send loop when it's time to send an ACK? Or is this the same as if we're congestion limited, and we'll deal with that gracefully?

integrationtests/self/self_suite_test.go
```diff
@@ -116,58 +136,110 @@ func generateCA() error {
...
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		BasicConstraintsValid: true,
+	}
+
+	lastCA := ca
+	lastCAPrivKey := caPrivateKey
+	certs := make([]*x509.Certificate, chainLen)
+	for i := 0; i < chainLen; i++ {
+		privKey, err := rsa.GenerateKey(rand.Reader, 2048)
```
lucas-clemente, line 197: I imagine this will increase the running time of our integration tests quite a bit. Should we reuse the same key maybe?

marten-seemann: Right, there's no advantage using different keys here.

packet_packer.go
marten-seemann: Only for *coalesced* packet sent by the client. This doesn't apply for normal packets.

internal/ackhandler/sent_packet_handler.go
marten-seemann: Done.

internal/ackhandler/sent_packet_handler.go

session.go
marten-seemann: This should be handled the same way as being congestion limited.

PR #2543, opened 2020-05-13
make sure that ACK frames are bundled with data

### LINKED ISSUE #2539: ACKs are not bundled any more
It looks like we're not bundling ACK frames any more when sending data, but we're waiting for 2 (or 10) ack-eliciting packets to be received before sending one.

marten-seemann:
Fixes #2539.


internal/ackhandler/received_packet_history.go
```diff
@@ -22,25 +22,26 @@ func newReceivedPacketHistory() *receivedPacketHistory {
 }
 
 // ReceivedPacket registers a packet with PacketNumber p and updates the ranges
-func (h *receivedPacketHistory) ReceivedPacket(p protocol.PacketNumber) {
+func (h *receivedPacketHistory) ReceivedPacket(p protocol.PacketNumber) bool /* is a new packet (and not a duplicate / delayed packet) */ {
```
lucas-clemente, line 25: Maybe return an enum instead? Bools as return values can be hard to understand when the function name doesn't make it clear.

internal/ackhandler/received_packet_handler.go
```diff
@@ -117,15 +117,15 @@ func (h *receivedPacketHandler) GetAckFrame(encLevel protocol.EncryptionLevel) *
 	switch encLevel {
 	case protocol.EncryptionInitial:
 		if h.initialPackets != nil {
-			ack = h.initialPackets.GetAckFrame()
+			ack = h.initialPackets.GetAckFrame(true)
```
lucas-clemente: Idem, this is quite hard to read without context :)

internal/ackhandler/received_packet_history.go
marten-seemann: How would that work? There are no enums in Go. Do you want to define a type alias for a bool?

```go
type isNewType bool

const (
    isNew isNewType = true
    isNotNew isNewType = false
)
```

PR #2553, opened 2020-05-21
use struct{} instead of bool in map just as a set

### LINKED ISSUE #2548: use struct{} instead of bool in map just as a set
current implementation windowUpdateQueue struct in window_update_queue.go.
```
type windowUpdateQueue struct {
	mutex sync.Mutex

	queue      map[protocol.StreamID]bool // used as a set
	queuedConn bool                       // connection-level window update

	streamGetter       streamGetter
	connFlowController flowcontrol.ConnectionFlowController
	callback           func(wire.Frame)
}
```
the variable `queue` use map  is just as a set,  the value field of `map` use type `bool`. Howerver, golang source use the type `struct{}` as the value of `map` when use the `map` just as a set, the reason is that `struct{}` in golang don't  consume  the memery. so i think it should uses `struct{}` instead of `bool` to reduce memery consumption, especially for server that can store large of StreamID.

I find the variable `streamsToDelete` was used in this way today (steams_map_incoming_generic.go for code generator).
```
type incomingItemsMap struct {
	mutex         sync.RWMutex
	newStreamChan chan struct{}

	streams map[protocol.StreamNum]item
	// When a stream is deleted before it was accepted, we can't delete it immediately.
	// We need to wait until the application accepts it, and delete it immediately then.
	streamsToDelete map[protocol.StreamNum]struct{} // used as a set

	nextStreamToAccept protocol.StreamNum // the next stream that will be returned by AcceptStream()
	nextStreamToOpen   protocol.StreamNum // the highest stream that the peer openend
	maxStream          protocol.StreamNum // the highest stream that the peer is allowed to open
	maxNumStreams      uint64             // maximum number of streams

	newStream        func(protocol.StreamNum) item
	queueMaxStreamID func(*wire.MaxStreamsFrame)
	// streamNumToID    func(protocol.StreamNum) protocol.StreamID // only used for generating errors

	closeErr error
}
```

mw-jn:
Fixes #2548.

### DIFF
--- a/crypto_stream.go
+++ b/crypto_stream.go
@@ -1,7 +1,6 @@
 package quic
 
 import (
-	"errors"
 	"fmt"
 	"io"
 
@@ -81,7 +80,7 @@ func (s *cryptoStreamImpl) GetCryptoData() []byte {
 
 func (s *cryptoStreamImpl) Finish() error {
 	if s.queue.HasMoreData() {
-		return errors.New("encryption level changed, but crypto stream has more data to read")
+		return qerr.NewError(qerr.ProtocolViolation, "encryption level changed, but crypto stream has more data to read")
 	}
 	s.finished = true
 	return nil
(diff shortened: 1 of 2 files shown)


marten-seemann:
@mw-jn Can you please run gofmt and amend the commit?

marten-seemann:
@mw-jn Looks like something went wrong here...

PR #2554, opened 2020-05-21
update the protobuf library to google.golang.org/protobuf/proto

marten-seemann:
The linter was complaining about deprecation of `github.com/golang/protobuf/proto`.


marten-seemann:
Merging without review, so that we can continue without the CI bugging us about it all the time.

PR #2567, opened 2020-05-25
authenticate connection IDs

### LINKED ISSUE #2529: authenticate connection IDs
See https://github.com/quicwg/base-drafts/pull/3499. This is currently in consensus call.

We need to implement this in a backwards-compatible way, i.e. require the TP only when on `> draft-27`. We MAY send it for draft-27, since unknown TPs are ignored anyway. Whatever is easier here.

marten-seemann:
Fixes #2529.

### DIFF
--- a/session.go
+++ b/session.go
@@ -274,6 +274,7 @@ var newSession = func(
 		StatelessResetToken:             &statelessResetToken,
 		OriginalDestinationConnectionID: origDestConnID,
 		ActiveConnectionIDLimit:         protocol.MaxActiveConnectionIDs,
+		InitialSourceConnectionID:       srcConnID,
 	}
 	if s.qlogger != nil {
 		s.qlogger.SentTransportParameters(params)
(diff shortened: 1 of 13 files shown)


session.go
```diff
@@ -1266,27 +1273,63 @@ func (s *session) dropEncryptionLevel(encLevel protocol.EncryptionLevel) {
...
 		s.qlogger.ReceivedTransportParameters(params)
 	}
 
-	// check the Retry token
-	if s.perspective == protocol.PerspectiveClient && !params.OriginalConnectionID.Equal(s.origDestConnID) {
-		s.closeLocal(qerr.NewError(qerr.TransportParameterError, fmt.Sprintf("expected original_connection_id to equal %s, is %s", s.origDestConnID, params.OriginalConnectionID)))
-		return
+	// check the initial_source_connection_id
+	if !params.InitialSourceConnectionID.Equal(s.handshakeDestConnID) {
+		return qerr.NewError(qerr.TransportParameterError, fmt.Sprintf("expected initial_source_connection_id to equal %s, is %s", s.handshakeDestConnID, params.InitialSourceConnectionID))
```
lucas-clemente, line 1307: Could exposing the expected id leak info to an attacker?

marten-seemann: Not really. The connection ID has already been used on the Initial packet, so it's known to a MITM.

PR #2584, opened 2020-06-02
fix closing of the http.Request.Body

### LINKED ISSUE #2551: http3 post request with body failed
this problem origin publish in caddy issue， but i think it is a quic go problems，
more info：https://github.com/caddyserver/caddy/issues/3429

post request has body so can not return response. you can reference this link:
https://github.com/caddyserver/caddy/issues/3429

i am a co-worker with @majc08149 , i find problem reason and code. But I do not know how to modify? @mholt

I paste problem code ：
file->net/http/transfer.go:371

if t.BodyCloser != nil { if err := t.BodyCloser.Close(); err != nil { return err } }
the function BodyCloser.Close() call quic-go->http3->body.go Close() like :
`
func (r *body) Close() error {
// quic.Stream.Close() closes the write side, not the read side
if r.isRequest {
return r.str.Close()
}
r.requestDone()
r.str.CancelRead(quic.ErrorCode(errorRequestCanceled))
return nil
}

`

this call quic-go->stream.go function 👍
func (s *stream) Close() error { if err := s.sendStream.Close(); err != nil { return err } return nil }

the function s.sendStream.Close() will call func (s *sendStream) Close() error , this function call
s.ctxCancel() to cancel context.
so the request reverseproxy will be canceled.

please help me how to fix this bug?

marten-seemann:
Fixes #2551.

On the server side, the `http.Request` is consumed by the HTTP handler. The HTTP handler may close the body (it doesn't have to though). In any case, closing the stream is the wrong thing to do, since that closes the write side of the stream. All we want to do is cancel the stream (if the
EOF hasn't been read yet).

On the client side (for the `http.Response`) the same logic applies: When the application calls `Body.Close()`, all we want to do is cancel the stream (if the EOF hasn't been read yet).

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -11,8 +11,6 @@ import (
 type body struct {
 	str quic.Stream
 
-	isRequest bool
-
 	// only set for the http.Response
 	// The channel is closed when the user is done with this response:
 	// either when Read() errors, or when Close() is called.
@@ -30,7 +28,6 @@ func newRequestBody(str quic.Stream, onFrameError func()) *body {
 	return &body{
 		str:          str,
 		onFrameError: onFrameError,
-		isRequest:    true,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


mholt:
Ah, you figured it out!! That's amazing! Would you like me to test as well? Let me know if I can do anything to help!

PR #2626, opened 2020-06-24
remove buffering of HTTP requests

### LINKED ISSUE #2620: HTTP/3 request timeout with io.PipeReader body
With an `http.Request` like this:

```go
pipeR, pipeW := io.Pipe()
relayReq, err := http.NewRequestWithContext(ctx, "POST", "https://127.0.0.1:8888/v1", pipeR)
resp, err := rt.RoundTrip(relayReq)
[...]
io.Copy(pipeW, os.Stdin)
```

The request fails with the following error.

```
$ echo TEST | go run .
2020/06/21 xx:xx:xx Read on stream 0 canceled with error code 268
```

But if the body in `http.NewRequestWithContext` is a `strings.Reader` or `nil`, the request would successfully complete.

The request body seems to be [sent asynchronously](https://github.com/lucas-clemente/quic-go/blob/c956ca4447b3945920219db6dc79cbc613761970/http3/request_writer.go#L57). Why would the RoundTrip call get blocked and eventually time out?

Example code:

```go
package main

import (
	"context"
	"crypto/tls"
	"io"
	"log"
	"net/http"
	"os"
	"time"

)

func main() {
	if err := dial(); err != nil {
		log.Fatal(err)
	}
}

func dial() error {
	http.Handle("/", http.NotFoundHandler())

	server := http3.Server{
		Server: &http.Server{
			Addr: "127.0.0.1:8888",
		},
	}
	go server.ListenAndServeTLS("c", "k")

	rt := http3.RoundTripper{
		DisableCompression: true,
		TLSClientConfig: &tls.Config{
			InsecureSkipVerify: true,
		},
	}

	pipeR, pipeW := io.Pipe()

	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()

	relayReq, err := http.NewRequestWithContext(ctx, "POST", "https://127.0.0.1:8888/v1", pipeR)
	if err != nil {
		return err
	}

	resp, err := rt.RoundTrip(relayReq)
	if err != nil {
		return err
	}

	errCh := make(chan error)
	go func() {
		io.Copy(pipeW, os.Stdin)
		pipeW.Close()
		errCh <- nil
	}()
	go func() {
		io.Copy(os.Stdout, resp.Body)
		resp.Body.Close()
		errCh <- nil
	}()
	<-errCh
	return nil
}
```

marten-seemann:
Fixes #2620.

Only the `http.ResponseWriter` implements the `http.Flusher` interface.
We need to make sure that writes to the `Request.Body` are actually sent out.

### DIFF
--- a/http3/request_writer.go
+++ b/http3/request_writer.go
@@ -1,7 +1,6 @@
 package http3
 
 import (
-	"bufio"
 	"bytes"
 	"fmt"
 	"io"
@@ -40,16 +39,15 @@ func newRequestWriter(logger utils.Logger) *requestWriter {
 }
 
 func (w *requestWriter) WriteRequest(str quic.Stream, req *http.Request, gzip bool) error {
-	wr := bufio.NewWriter(str)
-
-	if err := w.writeHeaders(wr, req, gzip); err != nil {
+	buf := &bytes.Buffer{}
+	if err := w.writeHeaders(buf, req, gzip); err != nil {
+		return err
+	}
+	if _, err := str.Write(buf.Bytes()); err != nil {
 		return err
 	}
 	// TODO: add support for trailers
 	if req.Body == nil {
-		if err := wr.Flush(); err != nil {
-			return err
-		}
 		str.Close()
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


http3/request_writer.go
```diff
@@ -70,20 +68,15 @@ func (w *requestWriter) WriteRequest(str quic.Stream, req *http.Request, gzip bo
 			}
```
lucas-clemente, line 68: In line 61 above I think there's a bug: io.Reader allows returning n > 0 together with io.EOF in one, so we should only break if err == io.EOF && n == 0, and also fix the CancelWrite handling.

(GH doesn't allow me to comment on a line that's not part of this commit 🙄

marten-seemann: Good catch! That's a separate bug though, I opened #2641 to track it.

PR #2637, opened 2020-06-29
introduce an assertion in the server

marten-seemann:


server.go
```diff
@@ -322,7 +322,7 @@ func (s *baseServer) handlePacketImpl(p *receivedPacket) bool /* should the buff
 	}
 	// Short header packets should never end up here in the first place
 	if !hdr.IsLongHeader {
-		return false
+		panic("misrouted packet")
```
lucas-clemente, line 325: Do we want a few more debug details?

PR #2638, opened 2020-06-29
split the qlog package into a logging and a qlog package, use a tracer interface in the quic.Config

marten-seemann:
This is the first step towards a more general logging approach, that will ultimately allow us to run both qlog and a metrics collector.


logging/interface.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"net"
+	"time"
+
+	"github.com/lucas-clemente/quic-go/internal/congestion"
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/wire"
+)
+
+type Tracer interface {
```
marten-seemann, line 14: Any idea for a better name?

logging/interface.go
```diff
@@ -0,0 +1,43 @@
...
+	TracerForServer(odcid protocol.ConnectionID) ConnectionTracer
+	TracerForClient(odcid protocol.ConnectionID) ConnectionTracer
+}
```
marten-seemann, line 17: We'll need to add another method here for logging events not related to any particular connection (e.g. stateless resets sent, Version Negotiation packets sent, packets that get dropped before they are sent to a connection, etc.).
Any suggestion for a name?

logging/interface.go
```diff
@@ -0,0 +1,43 @@
...
+
+// A ConnectionTracer records events.
+type ConnectionTracer interface {
```
marten-seemann, line 20: To make this useful, we'll have to make sure that all function parameters don't use any internal types. For most types, we can get away with defining type alias, e.g.
```go
type MaxDataFrame = wire.MaxDataFrame
```
Not sure if the fact the `MaxDataFrame` uses a `protocol.ByteCount` (which is also an internal type) will cause any problems here.

For some types, we might want to redefine the type. For example, we don't want to pass the contents of STREAM and CRYPTO frames to the logger:
```go
type StreamFrame struct {
    Offset protocol.ByteCount
    StreamID protocol.StreamID
    Length int
}
```

Suggestion:
We'll create a new package *logging/internal/to_logging*, that defines conversion for these types:
```go
func ConvertFrame(*wire.Frame) *logging.Frame
```

lucas-clemente: SGTM. Maybe we can reuse this conversion in the `LogFrame` function?

logging/interface.go
lucas-clemente: Why a single method, and not one per event, like in the ConnectionTracer?

logging/interface.go
lucas-clemente: Not really. Logger maybe, but that's already taken.

logging/interface.go
lucas-clemente: Actually I wonder whether the Logger can eventually be yet another tracer… Probably not?

marten-seemann: That would be an option. We could then define a `String() string` method on the `logging.Frame`, and get rid of the massive `switch` statement in `LogFrame`.

lucas-clemente: Maybe you don't even need that String() and can just use the fmt print?

marten-seemann: That would print the contents of STREAM and CRYPTO frames.

logging/interface.go
marten-seemann: That makes sense. Good thing about it, we don't need to come up with another name (`NonConnectionTracker`?).

logging/interface.go
lucas-clemente: But the LogFrame wouldn't include those, and instead have useful metadata (len, etc)? I guess the downside would be that we don't get to format fields usefully (hex vs dec e.g.), but not sure whether that's worth the complexity?

PR #2646, opened 2020-07-06
add a metrics package using OpenCensus, trace connections

marten-seemann:


metrics/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	stats.RecordWithTags(
+		context.Background(),
+		[]tag.Mutator{perspectiveTag, ipVersionTag},
+		connections.M(1),
+	)
+}
+
+func (t *connTracer) ClosedConnection(logging.CloseReason)                     {}
```
lanzafame, line 93: I know that opencensus doesn't technically have the concept of a gauge but I am pretty sure that you can just do a `connection.M(-1)`. So we should be able to model `activeConnections` and not just the total number of connections that have been opened since the program started. There may be some issues with this approach if the metrics collector restarts and the process doesn't but I don't think so.

marten-seemann: I'm a bit confused now. I had thought of this counter to be plotted as something like `rate(connections[30s])`, so it would measure the rate with which we're dialing and accepting new connections. Maybe `connections` is not the right name then?

If I understand correctly, what you're proposing is an active connection counter. I think I can introduce some counter variables into the `tracer`, then we don't have to worry about the metrics collector restarting.

Does that make sense to you?

metrics/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/lucas-clemente/quic-go/logging"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+// Measures
+var (
+	connections = stats.Int64("connections", "number of QUIC connections", stats.UnitDimensionless)
```
lucas-clemente, line 17: Should we have a quic prefix here, so this doesn't cause conflicts? Something like `/quic/connections`? (Not sure what the standard way of doing this is with opencensus.)

metrics/metrics.go
lucas-clemente: If you need a gauge, a better way of doing that could be keeping track of opened connects and of closed connections (with the error code) separately. That allows you to implement a gauge as one of multiple possible downstream reports.

metrics/metrics.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+// Tags
+var (
+	keyPerspective, _ = tag.NewKey("perspective")
```
lucas-clemente, line 22: Not sure about this, but another option here could be encoding server vs client in the counter name:

- /quic/server/connections
- /quic/server/var-that-only-makes-sense-for-servers
- /quic/client/connections
- /quic/client/var-that-only-makes-sense-for-clients

marten-seemann: Let's keep it as is for now. I'm just beginning to explore what we want to export, so this will probably change a couple of times anyway as we define more metrics.

metrics/metrics.go
marten-seemann: I went with `/quic-go/connections` for now. Happy to change this if @lanzafame has another suggestion.

PR #2651, opened 2020-07-06
enforce that a connection ID is not retired in a packet that uses that connection ID

marten-seemann:
The draft says:
> The sequence number specified in a RETIRE_CONNECTION_ID frame MUST NOT refer to the Destination Connection ID field of the packet in which the frame is contained. The peer MAY treat this as a connection error of type FRAME_ENCODING_ERROR.

### DIFF
--- a/conn_id_generator.go
+++ b/conn_id_generator.go
@@ -67,7 +67,7 @@ func (m *connIDGenerator) SetMaxActiveConnIDs(limit uint64) error {
 	return nil
 }
 
-func (m *connIDGenerator) Retire(seq uint64) error {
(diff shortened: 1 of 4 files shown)


conn_id_generator.go
```diff
@@ -76,6 +76,9 @@ func (m *connIDGenerator) Retire(seq uint64) error {
 	if !ok {
 		return nil
 	}
+	if connID.Equal(sentWithDestConnID) {
+		return qerr.NewError(qerr.FrameEncodingError, fmt.Sprintf("tried to retire connection ID %d (%s), which was used as the Destination Connection ID on this packet", seq, connID))
```
marten-seemann, line 80: This should be a PROTOCOL_VIOLATION, see https://github.com/quicwg/base-drafts/issues/3860.

PR #2656, opened 2020-07-08
don't block sendQueue.Send() if the runloop already exited.

### LINKED ISSUE #2655: Deadlock in session shutdown / sendQueue.Send()
It appears possible to wind up stuck in sendQueue.Send() after sendQueue.Run() has exited, preventing Session shutdown from exiting. 

Here is a boiled-down test case from the situation I saw it in (a failed handshake). 
https://gist.github.com/ltucker/0b433b775982176de4cfd8a540984d16

ltucker:
This can lead to a deadlock where session.shutdown() never exits
because it is blocked on a Send() but the sendQueue has exited due to
a write error.

Fixes #2655 but needs consideration for other consequences ...

### DIFF
--- a/send_queue_test.go
+++ b/send_queue_test.go
@@ -1,6 +1,8 @@
 package quic
 
 import (
+	"errors"
+
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


send_queue_test.go
```diff
@@ -68,6 +68,28 @@ var _ = Describe("Send Queue", func() {
...
 	})
 
+	It("does not block pending send after the queue has stopped running", func() {
+		done := make(chan struct{})
+		go func() {
+			defer GinkgoRecover()
+			q.Run()
+			close(done)
+		}()
+		q.Close()
```
marten-seemann, line 78: Can we use a `PacketConn` that errors here, instead of calling close? That more closely mirrors the actual issue that you discovered, as the session only calls `Close` after calls to `Send` have returned.

ltucker: @marten-seemann Thanks, I updated this to have the `MockConnection` trigger an error on `Write` (how the queue sees that error) -- I'm not sure rigging in an actual `PacketConn` would add much in this context, but let me know.

marten-seemann: Thank you, that's actually what I meant. Should have been more precise.

PR #2663, opened 2020-07-09
add integration tests using faulty packet conns

### LINKED ISSUE #2659: test erroring connections
We should probably have an integration test that makes the `PacketConn` error at a random time, and check that
1. the peer whose connection errored returns (that would have caught #2655).
2. the other peer runs into a handshake / idle timeout (just as a sanity check).

marten-seemann:
Fixes #2659. Depends on #2656.

This would have caught #2655. @ltucker, is it realistic to assume that if a connection fails, both `ReadFrom` and `WriteTo` will return an error?

### DIFF
--- a/integrationtests/self/timeout_test.go
+++ b/integrationtests/self/timeout_test.go
@@ -3,6 +3,9 @@ package self_test
 import (
 	"context"
 	"fmt"
+	"io"
+	"io/ioutil"
+	mrand "math/rand"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ltucker:
@marten-seemann oh sorry missed this, I think this is usually the case relatively quickly in practice...  I don't think golang technically claims anything about this property of the interface.  I can imagine these error-ing at slightly different times due to differences in deadlines or something, but may not be worth the fidgety-ness / buy much.

PR #2667, opened 2020-07-09
put the session in the packet handler map directly (for client sessions)

marten-seemann:
As of #2640, Version Negotiation packets are now handled in the session. This allows us to put the `session` in the packet handler map directly, without the need to route packets through the `client` first.


marten-seemann:
Increasing test coverage by deleting code. There's no better way to do it :)

PR #2676, opened 2020-07-16
add a metric for closed connections

marten-seemann:
I'm not really happy with this PR. This seems overly complicated (and converting a bool to string is not pretty either).

The reason this is so complicated is that there are many reasons a QUIC connection can be closed:
1. It can run into a (handshake or post-handshake) timeout
2. It can receive a stateless reset because the peer lost state
3. The application can close the connection, in which case we want to log which side did it and which error code was used
4. The transport can close the connection, in which case we also want to log which side did it and which error code was used

@lanzafame Does this look ok to you?

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"time"
 
@@ -17,6 +18,7 @@ var (
 	connections = stats.Int64("quic-go/connections", "number of QUIC connections", stats.UnitDimensionless)
 	lostPackets = stats.Int64("quic-go/lost-packets", "number of packets declared lost", stats.UnitDimensionless)
 	sentPackets = stats.Int64("quic-go/sent-packets", "number of packets sent", stats.UnitDimensionless)
+	closes      = stats.Int64("quic-go/close", "number of connections closed", stats.UnitDimensionless)
 )
 
 // Tags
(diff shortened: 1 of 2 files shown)


metrics/types.go
```diff
@@ -65,3 +65,16 @@ func (t packetType) String() string {
...
+type timeoutReason logging.TimeoutReason
+
+func (r timeoutReason) String() string {
+	switch logging.TimeoutReason(r) {
+	case logging.TimeoutReasonHandshake:
+		return "handshake_timeout"
+	case logging.TimeoutReasonIdle:
+		return "idle_timeout"
+	default:
+		panic("unknown timeout reason")
```
lanzafame, line 78: Don't panic, just return `"unknown"`. Metrics is never a good enough reason to crash a program.

metrics/metrics.go
```diff
@@ -106,7 +117,33 @@ func (t *connTracer) StartedConnection(local, _ net.Addr, _ logging.VersionNumbe
 	)
 }
 
-func (t *connTracer) ClosedConnection(logging.CloseReason)                     {}
+func (t *connTracer) ClosedConnection(r logging.CloseReason) {
+	var tags []tag.Mutator
```
lucas-clemente, line 121: I wonder whether this could be simplified by doing more of this processing during the CloseReason creation already. For example the type of reason, or whether it's remote or not could be set when creating the `CloseReason`, rather than interpreting errors later on. Essentially, make `CloseReason` POD?

metrics/types.go
marten-seemann: I tend to agree, although if we don't panic, those bugs will probably go unnoticed for a long time (until someone looks at the logs and sees a bunch of `unknown`s there). Anyway, this is a larger change, since we currently panic both in the `logging` and in the `qlog` package.

metrics/metrics.go
marten-seemann: Yes, `CloseReason` is a mess. I'd like to clean this up, but I'm not really sure how.

PR #2680, opened 2020-07-22
remove the SSLR (slow start large reduction) experiment

marten-seemann:
We apparently copied that code over from Chromium's Cubic implementation, but we certainly don't need it.


internal/congestion/cubic_sender.go
```diff
@@ -317,8 +317,3 @@ func (c *cubicSender) OnConnectionMigration() {
 	c.slowstartThreshold = c.initialMaxCongestionWindow
 	c.maxCongestionWindow = c.initialMaxCongestionWindow
 }
-
-// SetSlowStartLargeReduction allows enabling the SSLR experiment
-func (c *cubicSender) SetSlowStartLargeReduction(enabled bool) {
-	c.slowStartLargeReduction = enabled
```
lucas-clemente, line 323: We can probably remove the member & other references then?

marten-seemann: Oh nice, that removes another 10 LOC or so.

PR #2694, opened 2020-07-30
only count intentional delays towards the ack delay

### LINKED ISSUE #2639: Crash when qlogging a Version Negotiation packet
See https://github.com/ipfs/go-ipfs/issues/7512.

It looks like `qlog.Tracer.ReceivedVersionNegotiationPacket()` is called after `qlog.Tracer.Export()` has been called.
This could happen when the session runs into a handshake timeout, and then receives the Version Negotiation packet at just the right moment.

We should handle Version Negotiation packets in the session, so we don't have a race condition between `Export()` and other logging calls.

### LINKED ISSUE #2693: only count intentional delays for the ack_delay
See https://github.com/quicwg/base-drafts/pull/3971.

Specifically, ack_delay currently is set to the time difference between reading a packet from the socket and sending the ACK. However, queuing delays as well as frame processing delays should be considered to be part of the path delay.

marten-seemann:
Fixes #2693.

### DIFF
--- a/session.go
+++ b/session.go
@@ -1055,7 +1055,7 @@ func (s *session) handleUnpackedPacket(
 		}
 	}
 
-	return s.receivedPacketHandler.ReceivedPacket(packet.packetNumber, packet.encryptionLevel, rcvTime, isAckEliciting)
+	return s.receivedPacketHandler.ReceivedPacket(packet.packetNumber, packet.encryptionLevel, time.Now(), isAckEliciting)
 }
 
 func (s *session) handleFrame(f wire.Frame, encLevel protocol.EncryptionLevel, destConnID protocol.ConnectionID) error {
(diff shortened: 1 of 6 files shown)


lucas-clemente (approved):

session.go
```diff
@@ -1055,7 +1055,7 @@ func (s *session) handleUnpackedPacket(
 		}
 	}
 
-	return s.receivedPacketHandler.ReceivedPacket(packet.packetNumber, packet.encryptionLevel, rcvTime, isAckEliciting)
+	return s.receivedPacketHandler.ReceivedPacket(packet.packetNumber, packet.encryptionLevel, time.Now(), isAckEliciting)
```
marten-seemann, line 1058: This is not correct, RFC 9000 says:
> A packet might be held in the OS kernel or elsewhere on the host before being processed.  An endpoint MUST NOT include delays that it does not control when populating the ACK Delay field in an ACK frame. However, endpoints SHOULD include buffering delays caused by unavailability of decryption keys, since these delays can be large and are likely to be non-repeating.

PR #2710, opened 2020-08-11
never send a value larger than 2^60 in MAX_STREAMS frames

### LINKED ISSUE #2709: Confusing (incorrect?) behavior when both peers close a stream
Firstly, thank you @lucas-clemente for creating such a useful library and sharing it with the world.

I'm new to quic, and have stumbled upon something that may or may not be a bug. It might just need more documentation.

In short, if both peers `Close()` a stream, after the second or third stream they `Close()`, the client will receive the error:
```
FRAME_ENCODING_ERROR: 1152921504606846977 exceeds the maximum stream count
```
from the goroutine that is calling `AcceptStream()` on the `quic.Session`.

I've written a minimal program demonstrating the issue: https://gist.github.com/arashpayan/7c25dc8c26d3c0ea6ea43ace273fc09c

I've tested this issue against 0.17.3 and the HEAD of master as of this writing (https://github.com/lucas-clemente/quic-go/commit/d7c2169c55bd0ab0ce4724bcab270a6ec902534f). Simply run the sample with `go run main.go` and then tap 'enter' to make the "server" send a message to the "client". You'll notice the above error after the 4th stream/message is sent. To avoid the error, you can comment out the line where the client is closing the stream (`main.go:92`).

The number of successful streams before receiving this error is a function of the `MaxIncomingStreams` value passed in the `quic.Config` struct to `quic.DialAddr()`. In my program, I've been passing 2^60 for the maximum number streams, which lets me get away with 1 successful stream before receiving the above error. If I set the max to 2^60-1, I can get away with 2 streams. 2^60-2 gets me 3 streams and so on.

Anyway, I'm not sure if this is expected behavior, a bug, or something else.

Thanks.

marten-seemann:
Fixes #2709.

### DIFF
--- a/streams_map_incoming_bidi.go
+++ b/streams_map_incoming_bidi.go
@@ -159,12 +159,15 @@ func (m *incomingBidiStreamsMap) deleteStream(num protocol.StreamNum) error {
 	delete(m.streams, num)
 	// queue a MAX_STREAM_ID frame, giving the peer the option to open a new stream
 	if m.maxNumStreams > uint64(len(m.streams)) {
-		numNewStreams := m.maxNumStreams - uint64(len(m.streams))
-		m.maxStream = m.nextStreamToOpen + protocol.StreamNum(numNewStreams) - 1
-		m.queueMaxStreamID(&wire.MaxStreamsFrame{
-			Type:         protocol.StreamTypeBidi,
-			MaxStreamNum: m.maxStream,
-		})
+		maxStream := m.nextStreamToOpen + protocol.StreamNum(m.maxNumStreams-uint64(len(m.streams))) - 1
+		// Never send a value larger than protocol.MaxStreamCount.
+		if maxStream <= protocol.MaxStreamCount {
+			m.maxStream = maxStream
+			m.queueMaxStreamID(&wire.MaxStreamsFrame{
+				Type:         protocol.StreamTypeBidi,
+				MaxStreamNum: m.maxStream,
+			})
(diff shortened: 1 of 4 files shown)


streams_map_incoming_bidi.go
```diff
@@ -159,12 +159,15 @@ func (m *incomingBidiStreamsMap) deleteStream(num protocol.StreamNum) error {
...
 	if m.maxNumStreams > uint64(len(m.streams)) {
-		numNewStreams := m.maxNumStreams - uint64(len(m.streams))
-		m.maxStream = m.nextStreamToOpen + protocol.StreamNum(numNewStreams) - 1
-		m.queueMaxStreamID(&wire.MaxStreamsFrame{
-			Type:         protocol.StreamTypeBidi,
-			MaxStreamNum: m.maxStream,
-		})
+		maxStream := m.nextStreamToOpen + protocol.StreamNum(m.maxNumStreams-uint64(len(m.streams))) - 1
+		// Never send a value larger than protocol.MaxStreamCount.
+		if maxStream <= protocol.MaxStreamCount {
```
lucas-clemente, line 164: In the false case, should we send a capped maxStreamId frame, error, …? Silently ignoring this seems fragile?

marten-seemann: There's no error here. It's just that we're running out of stream IDs (which are limit by the way they're encoded... to 2^60 per stream type).

PR #2723, opened 2020-08-18
regenerate the testdata certificate with SAN instead of CommonName

### LINKED ISSUE #2721: generate certificates that don't set the CommonName
Apparently the [testdata certificates](https://github.com/lucas-clemente/quic-go/tree/master/internal/testdata) use the CommonName, which has been deprecated for 20 years (thanks for that, OpenSSL...).
That's why the PR adding Go 1.15 support (#2720) sets the `x509ignoreCN=0` GODEBUG flag to ignore that check. Obviously, this is just a temporary workaround, and we should figure out how to generate a valid certificate.

marten-seemann:
Fixes #2721. Follow-up to #2720.

As required by Go 1.15.

I also added the script used to generate the certificate, in case we need to regenerate it again at some point in the future. Maybe it's just me, but I don't find the openssl API not really intuitive...


marten-seemann:
Well that's not very helpful...

image

PR #2729, opened 2020-08-20
add the exhaustive linter, replace panics by return values in logging stringers

marten-seemann:
Depends on #2728.

A while back, @lanzafame suggested to replace the `panic`s in the `Stringer`s of logging types by "unknown" return values. Of course, he was right with that suggestion. Logging is not a good enough reason to panic.

Turns out that there's an even better solution for this: the [exhaustive](https://github.com/nishanths/exhaustive) linter, which even is part of the golangci-lint suite. This linter allows us to check that `switch` statement for enum types cover all values defined for that enum.

There are a few cases in our code where we deliberately don't do an exhaustive check, and I've added a `nolint:exhaustive` statement there. This is fine, at least it forces us to be explicit about this. In fact, activating this linter actually caught a bug in the packet packer (#2728).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,6 +20,7 @@ linters:
     - asciicheck
     - deadcode
     - depguard
+    - exhaustive
     - exportloopref
     - goconst
     - goimports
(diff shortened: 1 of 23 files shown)


lucas-clemente (approved):
GH is showing tons of lint errors. Do you want to fix those now or later?

PR #2741, opened 2020-09-04
read ECN bits and send ECN counters in ACK frames

marten-seemann:


marten-seemann:
Note the following API design problem that this PR brings:

Users pass a `net.PacketConn` to `quic.Listen` and `quic.Dial`. Currently, we'd use `ReadFrom` to read packets, but in order to read the ECN bits we'lll have to use `ReadMsgUDP` instead of `ReadFrom` on the `net.UDPConn`.
The problem is now that users might expect that quic-go will use the `ReadFrom` function of the packet conn. The question is now under which conditions to enable ECN support. Several options come to mind:
1. Only use ECN if the packet conn is a `*net.UDPConn`. This is probably the safest option, as there’s no way a user could have wrapped methods on this struct. However, it’s also the most inflexible, as users might want to wrap the UDP connection for a variety of reasons.
2. Define a new interface `type quic.ECNCapableConn interface { net.PacketConn; ReadMsgUDP(…); WriteMsgUDP(…); SyscallConn(…); }`, and do an interface assertion. This API is more flexible, but it requires more care from users: Directly embedding a packet conn `type wrappedConn struct { net.UDPConn }` would make quic-go bypass `ReadFrom`, which might be surprising. To prevent this, users would have to indirectly embed (is there any better term for this?) `type wrappedConn struct { conn net.UDPConn}` and then explicitly define all functions on `wrappedConn` such that it fulfills `net.PacketConn`.

Currently, this PR implements option 1, but I'm leaning towards implementing option 2 (with appropriate documentation of the pitfall described above).

lucas-clemente (approved):

PR #2749, opened 2020-09-07
fix incorrect documentation of sentPacketHistory.Iterate

marten-seemann:
We've been using this in https://github.com/lucas-clemente/quic-go/blob/bed802aee5732c8fe373bf4f71e789bea2f7f11f/internal/ackhandler/sent_packet_handler.go#L158-L166

I thought this was not a safe thing to do, but it turns out to be. Therefore, we can revise the documentation, and add a test case to make sure we don't introduce a regression at some point.


marten-seemann:
Turns out the documentation was right, and the test case I added is wrong. We need to test that we actually iterate over all packets instead.

PR #2761, opened 2020-09-08
support the CONNECT method

### LINKED ISSUE #2748: Support CONNECT method
CONNECT method has no :path. 

The current implementation requires a non-empty path somehow:

https://github.com/lucas-clemente/quic-go/blob/bed802aee5732c8fe373bf4f71e789bea2f7f11f/http3/request.go#L40-L47

Also url.ParseRequestURI will return an error on empty path.

klzgrad:
Fix #2748.

### DIFF
--- a/http3/request.go
+++ b/http3/request.go
@@ -37,7 +37,7 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
 		httpHeaders.Set("Cookie", strings.Join(httpHeaders["Cookie"], "; "))
 	}
 
-	isConnect := method == "CONNECT"
+	isConnect := method == http.MethodConnect
 	if isConnect {
 		if path != "" || authority == "" {
 			return nil, errors.New(":path must be empty and :authority must not be empty")
@@ -46,15 +46,15 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
 		return nil, errors.New(":path, :authority and :method must not be empty")
 	}
 
-	var url_ *url.URL
+	var u *url.URL
 	var requestURI string
 	var err error
 
 	if isConnect {
-		url_ = &url.URL{Host: authority}
+		u = &url.URL{Host: authority}
 		requestURI = authority
 	} else {
-		url_, err = url.ParseRequestURI(path)
+		u, err = url.ParseRequestURI(path)
 		if err != nil {
 			return nil, err
 		}
@@ -71,7 +71,7 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
 
 	return &http.Request{
 		Method:        method,
-		URL:           url_,
+		URL:           u,
 		Proto:         "HTTP/3",
 		ProtoMajor:    3,
 		ProtoMinor:    0,
--- a/http3/request.go
+++ b/http3/request.go
@@ -37,13 +37,28 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
 		httpHeaders.Set("Cookie", strings.Join(httpHeaders["Cookie"], "; "))
 	}
 
-	if len(path) == 0 || len(authority) == 0 || len(method) == 0 {
+	isConnect := method == "CONNECT"
+	if isConnect {
+		if path != "" || authority == "" {
+			return nil, errors.New(":path must be empty and :authority must not be empty")
+		}
+	} else if len(path) == 0 || len(authority) == 0 || len(method) == 0 {
 		return nil, errors.New(":path, :authority and :method must not be empty")
 	}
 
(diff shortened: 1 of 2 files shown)


klzgrad:
@googlebot I signed it!

klzgrad:
I don't understand this warning, "string `CONNECT` has 3 occurrences, make it a constant (goconst)." Isn't there only one occurrence?

http3/request.go
```diff
@@ -37,13 +37,28 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
 		httpHeaders.Set("Cookie", strings.Join(httpHeaders["Cookie"], "; "))
 	}
 
-	if len(path) == 0 || len(authority) == 0 || len(method) == 0 {
+	isConnect := method == "CONNECT"
```
marten-seemann, line 40: Can you use `http.MethodConnect` instead?

http3/request.go
```diff
@@ -37,13 +37,28 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
...
+			return nil, errors.New(":path must be empty and :authority must not be empty")
+		}
+	} else if len(path) == 0 || len(authority) == 0 || len(method) == 0 {
 		return nil, errors.New(":path, :authority and :method must not be empty")
 	}
 
-	u, err := url.ParseRequestURI(path)
-	if err != nil {
-		return nil, err
+	var url_ *url.URL
```
marten-seemann, line 49: Nit: Not a fan of a variable name followed by a _. Can you leave this at `u`?

http3/request_test.go
```diff
@@ -81,6 +81,16 @@ var _ = Describe("Request", func() {
...
 	})
 
+	It("handles CONNECT method", func() {
+		headers := []qpack.HeaderField{
+			{Name: ":authority", Value: "quic.clemente.io"},
+			{Name: ":method", Value: "CONNECT"},
+		}
+		req, err := requestFromHeaders(headers)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(req.Method).To(Equal("CONNECT"))
```
marten-seemann, line 91: Better: `http.MethodConnect`.

http3/request_test.go
```diff
@@ -81,6 +81,16 @@ var _ = Describe("Request", func() {
 		}))
 	})
 
+	It("handles CONNECT method", func() {
+		headers := []qpack.HeaderField{
+			{Name: ":authority", Value: "quic.clemente.io"},
+			{Name: ":method", Value: "CONNECT"},
+		}
+		req, err := requestFromHeaders(headers)
```
marten-seemann, line 89: Should we test for the `req.RequestURI` field?

marten-seemann (commented):
Thank you @klzgrad! This looks very good to me, I just found two very minor nits (see comments).

I also have a few questions from reading the spec:
> * The ":authority" pseudo-header field contains the host and port to connect to (equivalent to the authority-form of the request-target of CONNECT requests; see Section 3.2.3 of [HTTP11])

We're currently just copying over the `:authority` field. Am I right to assume that the upstream proxy implementation will perform a check that this field actually contains a host and port?

Furthermore, I'm wondering about this paragraph:
> Once the CONNECT method has completed, only DATA frames are permitted to be sent on the stream. Extension frames MAY be used if specifically permitted by the definition of the extension. Receipt of any other frame type MUST be treated as a connection error of type H3_FRAME_UNEXPECTED.

Do we need to check for this?

klzgrad:
> Am I right to assume that the upstream proxy implementation will perform a check that this field actually contains a host and port?

caddyserver/forwardproxy does this. In abstract, we are passing a host and port in a single authority string to the upstream proxy, and to actually perform the connection the proxy always has to first parse the string into a host and port. Checking its validity seems more suitable in that place.

> only DATA frames are permitted to be sent on the stream

Currently checking of this part is done in this code,

https://github.com/lucas-clemente/quic-go/blob/c5a132f158532ee8fbc0442f7c258c929fd77aab/http3/body.go#L50-L72

Not strictly conforming, as it merely ignores HEADERS frame, but the non-conformance is on the same level with regular HTTP requests. As in https://tools.ietf.org/id/draft-ietf-quic-http-29.html#section-4.1-6:

> Receipt of an invalid sequence of frames MUST be treated as a connection error of type H3_FRAME_UNEXPECTED (Section 8). In particular, a DATA frame before any HEADERS frame, or a HEADERS or DATA frame after the trailing HEADERS frame is considered invalid.

I hope fixing all that won't be required in this PR.

marten-seemann:
That makes sense. We can fix that in a separate issue.

marten-seemann (approved):
LGTM now. Thank you @klzgrad!

PR #2773, opened 2020-09-11
close the connection when the client sends a TLS version < TLS 1.3

marten-seemann:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25494.

In this fuzz case, the client sent a valid ClientHello that offered a TLS version different from TLS 1.3. We need to make sure that the server closes the connection in this case.


marten-seemann:
Merging this to get confirmation from OSS-Fuzz as quickly as possible.

PR #2789, opened 2020-09-15
set the ECN bits on outgoing packets

marten-seemann:
Not sure if we already want to merge this. Currently, it will just increase the size of ACK frames, as we're not yet using the ECN signal.


lucas-clemente (approved):
Probably still worth exercising it?

marten-seemann:
#3999 is tracking actual ECN support.

Issue #2586, opened 2020-06-04
Transport Performance in the wild


vyzo:
I wrote a couple of simple programs to test libp2p transport performance, and specifically compare QUIC to TCP (see https://github.com/vyzo/libp2p-perf-test).

Running a server in a linode in NJ and a client in a linode in Frankfurt, I have observed the following:
- QUIC is almost 3x slower than TCP
- Adding more streams slows down QUIC transfers even further.

Specifically, I used a 1GiB file with random data, with the timings below.

Transferring the file 3 times over TCP-vs-QUIC:
```
root@li1494-172:~# for x in {1..3}; do ./go/bin/test-client /ip4/50.116.48.114/tcp/4001/p2p/QmUgxn8vaVgVDnSxM6qhQzyeXTYXesjJM7iG9TomZjpPcR; done
2020/06/04 18:23:39 Connecting to QmUgxn8vaVgVDnSxM6qhQzyeXTYXesjJM7iG9TomZjpPcR
2020/06/04 18:23:39 Connected; requesting data...
2020/06/04 18:23:39 Transfering data...
2020/06/04 18:24:09 Received 1073741824 bytes in 30.149238941s
2020/06/04 18:24:09 Connecting to QmUgxn8vaVgVDnSxM6qhQzyeXTYXesjJM7iG9TomZjpPcR
2020/06/04 18:24:10 Connected; requesting data...
2020/06/04 18:24:10 Transfering data...
2020/06/04 18:24:47 Received 1073741824 bytes in 37.456968339s
2020/06/04 18:24:48 Connecting to QmUgxn8vaVgVDnSxM6qhQzyeXTYXesjJM7iG9TomZjpPcR
2020/06/04 18:24:48 Connected; requesting data...
2020/06/04 18:24:48 Transfering data...
2020/06/04 18:25:17 Received 1073741824 bytes in 29.308343925s
root@li1494-172:~# for x in {1..3}; do ./go/bin/test-client /ip4/50.116.48.114/udp/4001/quic/p2p/QmUgxn8vaVgVDnSxM6qhQzyeXTYXesjJM7iG9TomZjpPcR; done
2020/06/04 18:25:32 Connecting to QmUgxn8vaVgVDnSxM6qhQzyeXTYXesjJM7iG9TomZjpPcR
2020/06/04 18:25:32 Connected; requesting data...
2020/06/04 18:25:32 Transfering data...
[...]

PR #2791, opened 2020-09-16
increase UDP receive buffer size

### LINKED ISSUE #2255: investigate modifying the kernel's receive buffer size
It turns out that the Linux kernel's UDP receive buffer is too small by default (around 200k). In order to support higher bandwidths, we need a significantly higher value.

It would be good if we could set this on a per-connection basis. Not sure if Go allows us to do this.

marten-seemann:
Fixes #2255. Ref #2586.

... and emit a log message if we can't do so.

### DIFF
--- a/conn_windows.go
+++ b/conn_windows.go
@@ -7,3 +7,7 @@ import "net"
 func newConn(c net.PacketConn) (connection, error) {
 	return &basicConn{PacketConn: c}, nil
 }
+
+func inspectReadBuffer(net.PacketConn) (int, error) {
+	return 0, nil
+}
--- a/internal/protocol/params.go
+++ b/internal/protocol/params.go
@@ -2,6 +2,9 @@ package protocol
 
 import "time"
 
+// DesiredReceiveBufferSize is the kernel UDP receive buffer size that we'd like to use.
+const DesiredReceiveBufferSize = (1 << 20) * 2 // 2 MB
+
 // MaxPacketSizeIPv4 is the maximum packet size that we use for sending IPv4 packets.
 const MaxPacketSizeIPv4 = 1252
 
--- a/packet_handler_map.go
+++ b/packet_handler_map.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"hash"
+	"log"
 	"net"
 	"sync"
 	"time"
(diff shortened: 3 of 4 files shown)


packet_handler_map.go
```diff
@@ -53,13 +54,48 @@ type packetHandlerMap struct {
...
 
+func setReceiveBuffer(c net.PacketConn, logger utils.Logger) {
+	conn, ok := c.(interface{ SetReadBuffer(int) error })
+	if !ok {
+		logger.Debugf("Connection doesn't allow setting of receive buffer size")
+		return
+	}
+	size, err := inspectReadBuffer(c)
+	if err != nil {
+		log.Printf("Failed to determine receive buffer size: %s", err)
```
marten-seemann, line 65: I'm using `log` here, so that users will actually see the log message (even if QUIC_GO_LOG_LEVEL is not set).

internal/protocol/params.go
```diff
@@ -2,6 +2,9 @@ package protocol
 
 import "time"
 
+// DesiredReceiveBufferSize is the kernel UDP receive buffer size that we'd like to use.
+const DesiredReceiveBufferSize = (1 << 20) * 2 // 2 MB
```
marten-seemann, line 6: Not sure if this is what we really need, we might need to adjust this constant after more extensive benchmarking.

packet_handler_map.go
lucas-clemente: Why would this always need to log? What are the failure modes here that require user action?

We support [custom pconns](https://godoc.org/github.com/lucas-clemente/quic-go#DialContext), which don't need to implement `SyscallConn()`, and could benignly trigger this?

packet_handler_map.go
```diff
@@ -53,13 +54,48 @@ type packetHandlerMap struct {
...
+		return
+	}
+	if newSize == size {
+		log.Printf("Failed to determine receive buffer size: %s", err)
+		return
+	}
+	if newSize < protocol.DesiredReceiveBufferSize {
+		log.Printf("Failed to sufficiently increase receive buffer size. Was: %d kb, wanted: %d kb, got: %d kb.", size/1024, protocol.DesiredReceiveBufferSize/1024, newSize/1024)
```
lucas-clemente, line 85: kiB? ;)

packet_handler_map.go
marten-seemann: We do, but I'm assuming that all pconns at some point wrap a UDP conn. QUIC v1 is only defined to work over UDP anyway.

packet_handler_map.go
marten-seemann: Yes, you're right.

marten-seemann:
Merging this PR for now. We can still change the log messages later, if they turn out to be too annoying.

conn_windows.go
```diff
@@ -7,3 +7,7 @@ import "net"
 func newConn(c net.PacketConn) (connection, error) {
 	return &basicConn{PacketConn: c}, nil
 }
+
+func inspectReadBuffer(net.PacketConn) (int, error) {
+	return 0, nil
```
dswarbrick, line 12: Any particular reason why golang.org/x/sys wasn't used? `GetsockoptInt` should work for Windows in that package.

marten-seemann: @dswarbrick Not sure I understand. Can you elaborate?

dswarbrick: It appears that the `inspectReadBuffer` function here is a no-op on Windows, presumably because `syscall.GetsockoptInt` in the standard library is not implemented for Windows:
```go
func GetsockoptInt(fd Handle, level, opt int) (int, error) { return -1, EWINDOWS }
```
Since the syscall package is frozen, that functionality is not likely to make it into the standard library. However, it _is_ implemented in the golang.org/x/sys/windows package, as described in the link I posted.

marten-seemann: @dswarbrick Thanks for pointing this out, I wasn't aware of the syscall package being frozen.
Would you mind sending us a PR that adds this functionality on Windows?

PR #2793, opened 2020-09-21
remove support for the ECN test case

marten-seemann:
The interop runner test case requires us to proactively set ECN bits, which we currently don't do (see #2789).


marten-seemann:
Codecov is going crazy again. No idea why it thinks that this would reduce code coverage...

PR #2794, opened 2020-09-21
use certificates from /certs directory for the server

marten-seemann:
See https://github.com/marten-seemann/quic-interop-runner/pull/194.


marten-seemann:
Thanks again to Codecov for a completely bogus coverage report...

lucas-clemente (approved):

PR #2798, opened 2020-09-22
remove unneeded mutex in the client

### LINKED ISSUE #2717: Possible data race: A field in a structure is sometimes protected by Mutex, but sometimes unprotected.
A field in a structure is sometimes protected by Mutex, but sometimes unprotected.
This may lead to data race.
`s.currentFrame` is not protected by `s.mutex` on L168 in `receiveStream.readImpl()`. But for all the other places, it is protected by `s.mutex`.
https://github.com/lucas-clemente/quic-go/blob/d7c2169c55bd0ab0ce4724bcab270a6ec902534f/receive_stream.go#L166-L168
 `c.destConnID` and `c.version` are not protected by c.mutex on L245 in `client.dial()`. But for all the other places, it is protected by c.mutex.
https://github.com/lucas-clemente/quic-go/blob/d7c2169c55bd0ab0ce4724bcab270a6ec902534f/client.go#L244-L245

I found these inconsistencies through static analysis. Not sure if data race can actually happen in production.

I think `s.currentFrame` is more likely to have issues. All the calls to `s.dequeueNextFrame()` (which writes to `s.currentFrame`) are in the critical section, but since there is one read to `s.currentFrame` not in the critical section, I think it may have data race.

egonelbre:
Fixes #2717.

For s.currentFrame, it seems unlock/lock could be previously used, when it was using separate frame type, however, I don't think it's safe as mentioned in the issue.

In the dial, it's only calling recursively itself the mutex seems unnecessary. I think it was added because of the `go func` probably causing a race, however that doesn't fix the underlying potential race.

Note, I'm not sure, whether these changes are correct.

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net"
 	"strings"
-	"sync"
 
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 	"github.com/lucas-clemente/quic-go/internal/utils"
@@ -15,8 +14,6 @@ import (
 )
 
 type client struct {
-	mutex sync.Mutex
-
 	conn sendConn
 	// If the client is created with DialAddr, we create a packet conn.
 	// If it is started with Dial, we take a packet conn as a parameter.
@@ -252,7 +249,6 @@ func (c *client) dial(ctx context.Context) error {
 		c.tracer.StartedConnection(c.conn.LocalAddr(), c.conn.RemoteAddr(), c.version, c.srcConnID, c.destConnID)
 	}
 
-	c.mutex.Lock()
 	c.session = newClientSession(
 		c.conn,
 		c.packetHandlers,
@@ -268,7 +264,6 @@ func (c *client) dial(ctx context.Context) error {
 		c.logger,
 		c.version,
 	)
-	c.mutex.Unlock()
 	c.packetHandlers.Add(c.srcConnID, c.session)
 
 	errorChan := make(chan error, 1)


receive_stream.go
```diff
@@ -163,13 +163,10 @@ func (s *receiveStream) readImpl(p []byte) (bool /*stream completed */, int, err
 			return false, bytesRead, fmt.Errorf("BUG: readPosInFrame (%d) > frame.DataLen (%d) in stream.Read", s.readPosInFrame, len(s.currentFrame))
 		}
 
-		s.mutex.Unlock()
-
```
marten-seemann, line 167: As I have described in https://github.com/lucas-clemente/quic-go/issues/2717#issuecomment-696692143, there's no need to protect `currentFrame` by a mutex.

The unlocking before making the copy here is deliberate, as copying O(1000) bytes might take a while, during which we don't want to block other Go routines that might be waiting on the mutex.

marten-seemann (changes requested):
See comment.

receive_stream.go
egonelbre: Ah, so the mutex is protecting something else?

marten-seemann: Yes, it's protecting the 4 bools here:
https://github.com/lucas-clemente/quic-go/blob/27448c1fa520eddeda78cde02566d8b219963ba7/receive_stream.go#L43-L46

For example, the stream can get reset while the application is reading from it (either when `CancelRead` is called, or a RESET_STREAM frame is received). Similar for closing of the stream.

egonelbre: Anyways, I dropped that particular change from this change.

egonelbre: Note, I'm not completely convinced about the perf implications. A copy of 1000 bytes could end up being faster than a mutex unlock/lock; but I haven't looked done any detailed analysis... however, if it's needed for fairness, then it does make sense.

marten-seemann: My argument was based on performance considerations alone, not on fairness.
If copying is indeed faster than unlocking and locking the mutex, then this would be a fine change.

marten-seemann: If you want to write a micro-benchmark for this, we'd really appreciate that!
Please open a new issue / PR for this, and let's keep this PR for the changes to the `client`.

egonelbre: What's the most likely call to be also waiting on the mutex? I'm still getting familiar with the codebase. I only noticed CloseRemote/SetReadDeadline/CancelRead, but those don't seem like things that are called frequently.

marten-seemann: You're right about that. The other calls are mostly just called once in the lifetime of a stream (maybe a few times if we receive a retransmission for a RESET_STREAM frame). The only thing that's called more often would be `SetReadDeadline`, but that should also be relatively infrequent.

marten-seemann (approved):
LGTM now. It looks like we don't need the mutex in the client any more, so this is a nice simplification.
Thank you @egonelbre!

receive_stream.go
egonelbre: In that case, there wouldn't be contention on the mutex and probably there wouldn't be a significant difference in perf - i.e. it can use mutex fast-path. But, yeah, the only reason I can see to keep the mutex is to get the information from those methods earlier into the loop -- e.g. if there are bunch of frames in the queue, then I think it would need to exhaust them or fill `p`.

Either way, I'm convinced there isn't a significant problem with it.

PR #2802, opened 2020-09-25
fix deadlock in crypto setup when it is closed while handling a message

### LINKED ISSUE #2737: Long security handshake/leaked goroutine
I noticed this goroutine on a go-ipfs node, it seems suspicious to me that after 29 minutes there's a handshake still going on but I also am not familiar with the quic-go internals.

@marten-seemann does this look like a bug, or is it potentially valid behavior?  

```
goroutine 66760407 [chan send, 29m0s]:
github.com/lucas-clemente/quic-go/internal/handshake.(*cryptoSetup).SendAlert [0xc105b86780, 0xc057f36f2a]
	pkg/mod/github.com/lucas-clemente/quic-go@v0.16.2/internal/handshake/crypto_setup.go:689
github.com/marten-seemann/qtls.(*Conn).sendAlert [0xc02220b880, 0x2a, 0x0, 0x0]
	pkg/mod/github.com/marten-seemann/qtls@v0.9.1/conn.go:838
github.com/marten-seemann/qtls.(*Conn).verifyServerCertificate [0xc02220b880, 0xc0415cedc0, 0x1, 0x1, 0x35d, 0x0]
	pkg/mod/github.com/marten-seemann/qtls@v0.9.1/handshake_client.go:877
github.com/marten-seemann/qtls.(*clientHandshakeStateTLS13).readServerCertificate [0xc015a19e20, 0x0, 0x0]
	pkg/mod/github.com/marten-seemann/qtls@v0.9.1/handshake_client_tls13.go:464
github.com/marten-seemann/qtls.(*clientHandshakeStateTLS13).handshake [0xc015a19e20, 0xc03a782700, 0x0]
	pkg/mod/github.com/marten-seemann/qtls@v0.9.1/handshake_client_tls13.go:89
github.com/marten-seemann/qtls.(*Conn).clientHandshake [0xc02220b880, 0x0, 0x0]
	pkg/mod/github.com/marten-seemann/qtls@v0.9.1/handshake_client.go:240
github.com/marten-seemann/qtls.(*Conn).Handshake [0xc02220b880, 0x0, 0x0]
	pkg/mod/github.com/marten-seemann/qtls@v0.9.1/conn.go:1395
github.com/lucas-clemente/quic-go/internal/handshake.(*cryptoSetup).RunHandshake.func1 [0xc105b86780, 0xc063a3cfc0, 0xc113900a20]
	pkg/mod/github.com/lucas-clemente/quic-go@v0.16.2/internal/handshake/crypto_setup.go:266
created by github.com/lucas-clemente/quic-go/internal/handshake. [*cryptoSetup).RunHandshak]
	pkg/mod/github.com/lucas-clemente/quic-go@v0.16.2/internal/handshake/crypto_setup.go:264
```

marten-seemann:
Fixes #2737.

@egonelbre's analysis (option B) in https://github.com/lucas-clemente/quic-go/issues/2737#issuecomment-698514991 was right.

Unfortunately, I didn't manage to write an integration test that would trigger the bug, so I'm not 100% sure that this fix actually works under all circumstances.

### DIFF
--- a/internal/handshake/crypto_setup.go
+++ b/internal/handshake/crypto_setup.go
@@ -299,7 +299,6 @@ func (h *cryptoSetup) RunHandshake() {
 		h.mutex.Unlock()
 		h.runner.OnHandshakeComplete()
 	case <-h.closeChan:
-		close(h.messageChan)
 		// wait until the Handshake() go routine has returned
 		<-h.handshakeDone
 	case alert := <-h.alertChan:
@@ -345,6 +344,8 @@ readLoop:
 		case <-h.isReadingHandshakeMessage:
 			break readLoop
 		case <-h.handshakeDone:
+			break readLoop
+		case <-h.closeChan:
 			break readLoop
 		}
 	}
@@ -504,13 +505,18 @@ func (h *cryptoSetup) ReadHandshakeMessage() ([]byte, error) {
 	if !h.readFirstHandshakeMessage {
 		h.readFirstHandshakeMessage = true
 	} else {
-		h.isReadingHandshakeMessage <- struct{}{}
+		select {
+		case h.isReadingHandshakeMessage <- struct{}{}:
+		case <-h.closeChan:
+			return nil, errors.New("error while handling the handshake message")
+		}
 	}
-	msg, ok := <-h.messageChan
-	if !ok {
+	select {
+	case msg := <-h.messageChan:
+		return msg, nil
+	case <-h.closeChan:
 		return nil, errors.New("error while handling the handshake message")
 	}
-	return msg, nil
 }
 
 func (h *cryptoSetup) SetReadKey(encLevel qtls.EncryptionLevel, suite *qtls.CipherSuiteTLS13, trafficSecret []byte) {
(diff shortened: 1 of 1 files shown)


egonelbre:
I often try to separate the concurrency bits and channel handling into a separate struct, such that it can be separately tested... however seems bit difficult here.

If you could hook into `h.conn.Handshake` somehow, then you could call `SendAlert` from it. And concurrently call `Close`.

internal/handshake/crypto_setup.go
```diff
@@ -506,11 +507,12 @@ func (h *cryptoSetup) ReadHandshakeMessage() ([]byte, error) {
 	} else {
```
egonelbre, line 507: On line 488, is there a possibility the `go func() { <-h.isReadingHandshakeMessage` gets stuck somehow?

marten-seemann:
@egonelbre Yes, that would be nice, but the way the TLS library is structured doesn't allow that.
Ideally, the library would expose a `HandleHandshakeMessage() (response []byte, err error)` method. Instead, we need to call `Handshake`, which basically does the following:
```go
func Handshake() {
    for {
          msg := cryptosetup.ReadHandshakeMessage()
          if err := handleMsg(msg); err != nil {
               cryptosetup.SendAlert()
               break
          }
          if handshake complete {
               break
          }
    }
}

internal/handshake/crypto_setup.go
marten-seemann: Good catch. That one should probably be protected by a `select` statement as well.

lucas-clemente (approved):

PR #2814, opened 2020-10-05
client: Add DialEarlyContext and DialAddrEarlyContext API

### LINKED ISSUE #2812: Allow passing in context when using 0RTT
Hi there,

My team is trying to utilize the 0RTT feature from QUIC.  We use context cancellation in our code. However, I have noticed that `quic-go` provides `DialAddrContext` but no `DialAddrEarlyContext`. Is it possible that we can add another method for `DialAddrEarly` that can take in a context?

VinozzZ:
Fixes #2812.

### DIFF
--- a/client.go
+++ b/client.go
@@ -68,7 +68,18 @@ func DialAddrEarly(
 	tlsConf *tls.Config,
 	config *Config,
 ) (EarlySession, error) {
-	sess, err := dialAddrContext(context.Background(), addr, tlsConf, config, true)
+	return DialAddrEarlyContext(context.Background(), addr, tlsConf, config)
+}
+
+// DialAddrEarlyContext establishes a new 0-RTT QUIC connection to a server using provided context.
+// See DialAddrEarly for details
+func DialAddrEarlyContext(
+	ctx context.Context,
+	addr string,
+	tlsConf *tls.Config,
+	config *Config,
+) (EarlySession, error) {
+	sess, err := dialAddrContext(ctx, addr, tlsConf, config, true)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


marten-seemann (changes requested):
Thank you! This lgtm, just one suggestion:

Can you use `DialAddrEarlyContext` in `DialAddrEarly`, in order to avoid code duplication?
```go
func DialAddrEarly(...) {
     return DialAddrEarlyContext(context.Background(), ....)
}

File: client.go [hunk 1]
```diff
@@ -68,7 +68,23 @@ func DialAddrEarly(
 	tlsConf *tls.Config,
 	config *Config,
 ) (EarlySession, error) {
-	sess, err := dialAddrContext(context.Background(), addr, tlsConf, config, true)
+	sess, err := DialAddrEarlyContext(context.Background(), addr, tlsConf, config)
+	if err != nil {
+		return nil, err
+	}
+	utils.Logger.WithPrefix(utils.DefaultLogger, "client").Debugf("Returning early session")
+	return sess, nil
```
marten-seemann, line 76: You already have that log statement in `DialAddrEarlyContext`. `DialAddrEarly` can be as simple as just a `return DialAddrEarlyContext(context.Background(), ...)`.

marten-seemann (changes requested):

client.go
VinozzZ: My bad. Just fixed it

client.go
```diff
@@ -68,7 +68,23 @@ func DialAddrEarly(
 	tlsConf *tls.Config,
 	config *Config,
 ) (EarlySession, error) {
-	sess, err := dialAddrContext(context.Background(), addr, tlsConf, config, true)
+	sess, err := DialAddrEarlyContext(context.Background(), addr, tlsConf, config)
```
marten-seemann, line 71:
```go
return DialAddrEarlyContext(context.Background(), addr, tlsConf, config)
```

marten-seemann (changes requested):

marten-seemann (approved):
Thank you!

PR #2830, opened 2020-10-20
introduce a quic.UDPConn interface

marten-seemann:
This interface can be used by applications that want to make sure that the packet conn they pass to `Dial` and `Listen` actually makes use of all UDP-specific optimizations.

Related: https://github.com/libp2p/go-libp2p-quic-transport/issues/178.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -17,18 +17,17 @@ type connection interface {
 	io.Closer
 }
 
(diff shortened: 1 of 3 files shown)


lucas-clemente (approved):

marten-seemann:
We won't need this (for now at least), as https://github.com/libp2p/go-libp2p-quic-transport/issues/178 was fixed.

PR #2844, opened 2020-10-28
upgrade method to convert http client to http3

### LINKED ISSUE #2190: Add http3.ConfigureTransport
Upgrading the connection makes it hard to keep the same configuration as before. I believe a `http3.ConfigureTransport` method should be added, just like the one in [`http2.ConfigureTransport`]

This is based on the comment in https://github.com/tsenart/vegeta/pull/453/files#r339342957 by @tsenart 

> Any chance we can "translate" already set configuration in an existing Transport to this http3.RoundTripper? Akin to what http2.ConfigureTransport(tr) does. Example: Not loosing existing TLSClientConfig.

This shouldn't be too hard to implement unless I am missing something.

[`http2.ConfigureTransport`]: https://godoc.org/golang.org/x/net/http2#ConfigureTransport

tmcnicol:
[This](https://github.com/tsenart/vegeta/pull/453/files#r339342957) is the mention of copying the TLS config. If we do a clone will this prevent `tls.Config.SetSessionTicketKeys` from being used?

Sorry about the delayed, life got in the way and I let this slip. 

Fixes #2190.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -285,3 +285,14 @@ func (c *client) doRequest(
 
 	return res, requestError{}
 }
+
+// ConfigureTransports converts an net/http HTTP/1 Transport to a http3.RoundTripper
+// Uses the original TLS config, if present, relying on the clone created on a
+// new connection.
+func ConfigureTransports(t1 *http.Transport) *RoundTripper {
+	roundtripper := &RoundTripper{
+		TLSClientConfig: t1.TLSClientConfig,
+		QuicConfig:      &quic.Config{},
+	}
+	return roundtripper
+}
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -171,6 +171,23 @@ var _ = Describe("Client", func() {
 		Expect(client.Close()).To(Succeed())
 	})
 
+	Context("Configuring an http1 client", func() {
+		It("uses an empty tls config if none exists", func() {
+			t1 := &http.Transport{}
+			t2 := ConfigureTransports(t1)
+			Expect(t2.TLSClientConfig).To(BeNil())
+		})
+
+		It("uses tls configuration from the original transport", func() {
+			tlsConf := &tls.Config{ServerName: "foo.bar"}
+			t1 := &http.Transport{
+				TLSClientConfig: tlsConf,
+			}
+			t2 := ConfigureTransports(t1)
+			Expect(t2.TLSClientConfig).To(Equal(tlsConf))
+		})
+	})
+
 	Context("Doing requests", func() {
 		var (
 			request *http.Request


http3/client.go
```diff
@@ -273,3 +273,24 @@ func (c *client) doRequest(
...
+	transport, ok := client.Transport.(*http.Transport)
+	if !ok {
+		return errors.New("could not extract transport from client, must be of type *http.Transport")
+	}
+	tlsConf := transport.TLSClientConfig
+	if tlsConf == nil {
+		tlsConf = &tls.Config{}
+	} else {
+		tlsConf = tlsConf.Clone()
+	}
```
marten-seemann, line 289: Why is it necessary to clone the `tls.Config` here?

marten-seemann:
Is this supposed to be a fix for #2190?

http3/client.go
tmcnicol: It was requested in the initial request, https://github.com/tsenart/vegeta/pull/453/files#r339342957 

I am happy to remove it as I don't fully understand the nuance involved in copying the TLS config. Is it acceptable to use a fresh one every time.

tmcnicol:
> Is this supposed to be a fix for #2190?
Yes

http3/client.go
marten-seemann: We already clone the config here, before we modify it: https://github.com/lucas-clemente/quic-go/blob/b034f1ac7c9b66ad5814e747d43d7c7915310ff3/http3/client.go#L59-L70

I don't think we need another `Clone` here, unless you want to modify fields.

marten-seemann:
Why is it called `UpgradeClient` and not `ConfigureTransport`?

tmcnicol:
> Why is it called `UpgradeClient` and not `ConfigureTransport`?

I did this because it takes in an `http.Client` instead of a transport and I didn't want to confuse the two functions. I didn't find a way to make it work similarly to `ConfigureTransport`.

marten-seemann:
What's the problem with `ConfigureTransport`?

tmcnicol:
The issue I ran into is that the http 1 to 2 upgrade are upgrade functions on the same transport. As this roundtripper is a drop in replacement for the entire Transport the arguments differ
```
// http2 
http2.ConfigureTransport(*http.Transport)

// http3
http3.ConfigureTransport(*http.Client)
```
I thought that giving them the same name would be misleading. Perhaps `ConfigureClient` would be a better name?

marten-seemann:
Why not `http3.ConfigureTransport(*http.Transport) *http3.Transport`?

tmcnicol:
That would work, something in line with this? https://github.com/golang/net/blob/master/http2/transport.go#L167-L169

marten-seemann:
That looks reasonable.

marten-seemann:
@tmcnicol Are you still planning to update this PR?

tmcnicol:
Any further comments?

http3/client.go
```diff
@@ -285,3 +285,14 @@ func (c *client) doRequest(
...
 	return res, requestError{}
 }
+
+// ConfigureTransports converts an net/http HTTP/1 Transport to a http3.RoundTripper
+// Uses the original TLS config, if present, relying on the clone created on a
+// new connection.
+func ConfigureTransports(t1 *http.Transport) *RoundTripper {
+	roundtripper := &RoundTripper{
+		TLSClientConfig: t1.TLSClientConfig,
+		QuicConfig:      &quic.Config{},
```
marten-seemann, line 295: You probably don't need to set an empty config here.

marten-seemann (commented):
I'm not sure if we actually need this change. After all, all that is does is setting the `TLSClientConfig` from the `http.Transport` in the `http3.Roundtripper`.

This can be done in a single line by a user of this library.

PR #2851, opened 2020-10-31
wait until sessions have been handled when closing the HTTP/3 server

marten-seemann:


lucas-clemente (approved):

marten-seemann:
This PR is wrong. It would remove the listener after the first session has been handled.

PR #2863, opened 2020-11-06
qlog tokens in NEW_TOKEN frames, Retry packets and Initial packets

marten-seemann:


qlog/frame.go
```diff
@@ -130,8 +130,7 @@ func marshalCryptoFrame(enc *gojay.Encoder, f *logging.CryptoFrame) {
 
 func marshalNewTokenFrame(enc *gojay.Encoder, f *logging.NewTokenFrame) {
 	enc.StringKey("frame_type", "new_token")
-	enc.IntKey("length", len(f.Token))
-	enc.StringKey("token", fmt.Sprintf("%x", f.Token))
+	enc.ObjectKey("token", &token{Raw: f.Token})
```
lucas-clemente, line 133: how sensitive is the token, should it appear in logs?

lucas-clemente (approved):

qlog/frame.go
marten-seemann: Not very sensitive. All you can do with is initiate a new connection with a lower risk of receiving a Retry.
Also, you could correlate two connections.

PR #2866, opened 2020-11-07
Split http3.ListenAndServe function in two parts to allow use of certificates which are not file based

je4:
The http3.ListenAndServe function works only with certificate and key files. There's no easy way to use certificates which are already loaded. Therefor, the function should be splitted in two functions, whereas the only task of http3.ListenAndServe is to create a tls.Config structure from cert and key files and call a second function http3.ListenAndServeCfg which does the work.


je4:
@googlebot I signed it!

je4:
@googlebot I fixed it.

je4:
i will try again with the cla...

PR #2867, opened 2020-11-07
http3.ListenAndServe() splitted in two functions

je4:
to allow the use of certificates which are not file based, http3.ListenAndServe is splitted in two functions.
http3.ListenAndServe opens the certificate and key files and creates a tls.Config file. This will be the parameter for the second function  http3.ListenAndServeCfg which does the work...


marten-seemann (changes requested):
Thank you for the PR!
We try to keep our API as close as possible to the API of the `net/http` package of the standard library. As far as I can see, there's no `http.ListenAndServeCfg`, which is why I'm inclined to not merge this PR.

PR #2869, opened 2020-11-08
remove unnecessary mutex locking in the stream flow controller

marten-seemann:


internal/flowcontrol/base_flow_controller.go
```diff
@@ -60,10 +60,8 @@ func (c *baseFlowController) sendWindowSize() protocol.ByteCount {
 	return c.sendWindow - c.bytesSent
 }
 
+// needs to be called with locked mutex
```
lucas-clemente, line 63: It shouldn't be a public fn then, no?

lucas-clemente (commented):
GH is showing some unused failures?

internal/flowcontrol/base_flow_controller.go
marten-seemann: That probably applies to all function of the `baseFlowController`.

marten-seemann: Done.

PR #2871, opened 2020-11-11
only trace dropped 0-RTT packets when a tracer is set

marten-seemann:


marten-seemann:
@lucas-clemente Merging without review, as I need to get the new version out.

PR #2876, opened 2020-11-15
use PADDING frames to pad packets

### LINKED ISSUE #2875: padding of Server Initials fails if the certifictate is very small
If the certificate is very small, the first packet will contain an Initial, Handshake and 1-RTT packet. This will cause a problem cause by padding the Initial, which we introduced in https://github.com/lucas-clemente/quic-go/issues/2840.

In this case, we can't pad by appending 0s to the end of the packet, as the 1-RTT packet doesn't have a length field and always extends to the end of the packet. This will lead to the first packet ending up undecryptable for the client.

Unfortunately, the congestion controller on the server will notice this: As the the first packet is never acknowledged, the server will declare it lost, and start the connection in recovery. This is **bad** for throughput.

marten-seemann:
Fixes #2875.

This is a non-trivial rewrite of how we pack packets. Now we compose the payload of an entire datagram (i.e. multiple coalesced packets), before serializing anything. This allows us to add PADDING frames to the Initial packet.

In order for this to work properly, we need to make sure encode the Length field in the Long Header in 2 bytes (even if the actual length would have in 1 byte).

### DIFF
--- a/internal/utils/varint.go
+++ b/internal/utils/varint.go
@@ -83,6 +83,34 @@ func WriteVarInt(b *bytes.Buffer, i uint64) {
 	}
 }
 
+// WriteVarIntWithLen writes a number in the QUIC varint format, with the desired length.
+func WriteVarIntWithLen(b *bytes.Buffer, i uint64, length protocol.ByteCount) {
+	if length != 1 && length != 2 && length != 4 && length != 8 {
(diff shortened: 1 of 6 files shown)


marten-seemann:
Merging, as we need to release a new version asap.

PR #2885, opened 2020-11-19
use a sync.Pool for ACK frames

marten-seemann:
ACKs are one of the most common frames, and so far they cost 2 allocations: one for the frame itself, and one for ACK range slice.

### DIFF
--- a/internal/wire/pool.go
+++ b/internal/wire/pool.go
@@ -6,19 +6,24 @@ import (
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 )
 
-var pool sync.Pool
+var streamFramePool, ackFramePool sync.Pool
 
 func init() {
-	pool.New = func() interface{} {
+	streamFramePool.New = func() interface{} {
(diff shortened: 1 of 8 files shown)


internal/wire/pool.go
```diff
@@ -27,7 +32,20 @@ func putStreamFrame(f *StreamFrame) {
...
-		panic("wire.PutStreamFrame called with packet of wrong size!")
+		panic("wire.putStreamFrame called with frame of wrong size!")
+	}
+	streamFramePool.Put(f)
+}
+
+// GetAckFrame gets an ACK frame from the pool.
+// It is the callers responsibility to fill *all* of the fields of the returned ACK frame.
+func GetAckFrame() *AckFrame {
+	return ackFramePool.Get().(*AckFrame)
```
lucas-clemente, line 43: How expensive would it be to zero out all fields here, just to be on the safe side?

(similar for the stream frame)

lucas-clemente (approved):

PR #2888, opened 2020-11-20
cross compile quic-go for all platforms / architectures

marten-seemann:
Add a cross-compilation GitHub Actions test. Thanks to @mvdan for his input on the script!

The problem here is that the test runs for ~10 minutes. We might want to limit this to PRs and commits to master.

### DIFF
--- a/.github/workflows/cross-compile.yml
+++ b/.github/workflows/cross-compile.yml
@@ -1,0 +1,19 @@
+on: [push, pull_request]
+jobs:
+  crosscompile:
+    strategy:
+      matrix:
+        go: [ "1.14.x", "1.15.x" ]
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


mvdan (approved):
Nice :)

.github/workflows/cross-compile.yml
```diff
@@ -0,0 +1,19 @@
+on: [push, pull_request]
+jobs:
+  crosscompile:
+    strategy:
+      matrix:
+        go: [ "1.14", "1.15" ]
```
mvdan, line 6: beware that this is 1.14.0, not the latest 1.14.x. You should use `[ "1.14.x", "1.15.x" ]`.

lucas-clemente (approved):

PR #2889, opened 2020-11-20
optimize control message parsing

marten-seemann:
`syscall.ParseSocketControlMessage` allocates, as it returns a `[]syscall.SocketControlMessage`. To avoid allocations, we can parse the information we're looking for while parsing the control messages.

The speedup is significant, at the cost of pulling yet another `unsafe` operation into the project.
```
❯ go test -bench . -benchmem -run Bench 
goos: darwin
goarch: amd64
pkg: github.com/lucas-clemente/quic-go
BenchmarkSyscallParseControlMessage-16          14987308                72.7 ns/op            48 B/op             1 allocs/op
BenchmarkParseControlMessage-16                 163042294                7.21 ns/op            0 B/op             0 allocs/op
PASS
ok      github.com/lucas-clemente/quic-go       3.232s
```

### DIFF
--- a/cmsg.go
+++ b/cmsg.go
@@ -1,0 +1,46 @@
+// +build darwin linux
+
+package quic
+
+import (
+	"syscall"
+	"unsafe"
+
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+)
+
+// ecnFromControlMessage parses b as an array of socket control messages.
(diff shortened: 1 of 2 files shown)


lucas-clemente:
Do the 70ns really make a big enough difference at the macro level that this is worth the added complexity? I'd consider the bar for `unsafe` rather high :)

PR #2895, opened 2020-11-21
index connections directly in the multiplexer

### LINKED ISSUE #2597: allow applications to use SO_REUSEPORT
SO_REUSEPORT allows us to create multiple listeners for a single address / port.

I created a simple script to test what's going on: https://gist.github.com/marten-seemann/42199c0b8462c1a5748c6b1c1aca22c4

It looks like the kernel will deterministically route packets coming from a certain remote address to a single listener. That means that this won't show up as additional reordering (see https://github.com/lucas-clemente/quic-go/issues/2586#issuecomment-641171225).

This means that reuseport could help us improve performance when handling *multiple* connections. It won't help with single-connection performance, since in that case only a single socket is used. Further benchmarking is needed to evaluate how big the performance gain is for quic-go in specific, but https://blog.cloudflare.com/how-to-receive-a-million-packets/ suggest that this could be worth doing.

marten-seemann:
This PR needs some careful review.

Currently, the `multiplexer` takes care of deduplicating `net.PacketConns` that are passed to quic-go via `Dial` and `Listen`. We need to do that because we cannot have more than one `ReadFrom` go-routine per connection. We currently deduplicate by looking at the `LocalAddr()`. This makes it possible for users to use the following pattern:
```go
conn, _ := net.ListenUDP(...)
quic.Listen(conn, ...)
quic.Dial(conn, "host1", ...)
quic.Dial(conn, "host2", ...)
```
As the `multiplexer` detects that it's the same `conn` used here, it will only start a single go routine that calls `ReadFrom` on that connection.
The problem with this approach is that it won't work with SO_REUSEPORT, as connections will have the same `LocalAddr`.

As an alternative approch, it would also be possible to compare the `conn` directly, instead of just asking if they have a matching `LocalAddr()`. This approach would work with wrapped connections:
```go
type wrappedConn { net.PacketConn }
c, _ := net.ListenUDP(...)
conn := &wrappedConn{PacketConn: c}
quic.Listen(conn, ...)
quic.Dial(conn, "host1", ...)
```
However, it will fail when used in the following way:
```go
type wrappedConn { net.PacketConn }
c, _ := net.ListenUDP(...)
quic.Listen(&wrappedConn{PacketConn: c}, ...)
quic.Dial(&wrappedConn{PacketConn: c}, "host1", ...)
```
In fact, that's what Caddy does, and what made us introduce the current logic in #2111.

The question now is, what does a reasonable API for quic-go look like? Should we look at the `LocalAddr()` (or any other function on the interface), or should we *require* users to never wrap the same connection into different objects?

If we decide to merge this PR, we'd have to ask @mholt to change the logic in Caddy, and [not create](https://github.com/caddyserver/caddy/blob/96058538f01bf3fc4f2f7c25dbb2338f058f46c7/listeners.go#L75-L92) a new `wrappedConn` for the same underlying `net.PacketConn`.

The big advantage is that the approach here now allows users to use SO_REUSEPORT (that's why this PR fixes #2597): Connections using SO_REUSEPORT will naturally have the same `LocalAddr()`, so with the current approach of identifying connections, they would be detected as duplicated, although they really aren't.

An alternative to this PR would be to define another interface that connections can implement.
```go
type PacketConnWithIdent interface {
    net.PacketConn
    Identifier() string
}
```
Applications that use SO_REUSEPORT would then have to implement the `Identifier() string` method on the `net.PacketConn` they pass to quic-go.

@lucas-clemente WDYT?

### DIFF
--- a/mock_multiplexer_test.go
+++ b/mock_multiplexer_test.go
@@ -51,7 +51,7 @@ func (mr *MockMultiplexerMockRecorder) AddConn(arg0, arg1, arg2, arg3 interface{
 }
 
 // RemoveConn mocks base method
-func (m *MockMultiplexer) RemoveConn(arg0 indexableConn) error {
+func (m *MockMultiplexer) RemoveConn(arg0 net.PacketConn) error {
 	m.ctrl.T.Helper()
 	ret := m.ctrl.Call(m, "RemoveConn", arg0)
 	ret0, _ := ret[0].(error)
--- a/multiplexer.go
+++ b/multiplexer.go
@@ -15,13 +15,9 @@ var (
 	connMuxer     multiplexer
 )
 
-type indexableConn interface {
-	LocalAddr() net.Addr
-}
-
 type multiplexer interface {
 	AddConn(c net.PacketConn, connIDLen int, statelessResetKey []byte, tracer logging.Tracer) (packetHandlerManager, error)
-	RemoveConn(indexableConn) error
+	RemoveConn(conn net.PacketConn) error
 }
 
 type connManager struct {
@@ -36,7 +32,7 @@ type connManager struct {
 type connMultiplexer struct {
 	mutex sync.Mutex
 
-	conns                   map[string] /* LocalAddr().String() */ connManager
+	conns                   map[net.PacketConn]connManager
 	newPacketHandlerManager func(net.PacketConn, int, []byte, logging.Tracer, utils.Logger) (packetHandlerManager, error) // so it can be replaced in the tests
 
 	logger utils.Logger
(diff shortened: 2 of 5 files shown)


mholt:
Thanks for the ping.

Obviously my vote is going to be for an approach that doesn't break Caddy. :) We need to wrap the packetconn so that we can impose our own `Close()` behavior. We can't share the wrapper value (i.e. we can't wrap it only once and then reuse that value for both Listen and Dial) because each instance of the wrapper has its own state, shared by the same underlying packetconn:

https://github.com/caddyserver/caddy/blob/96058538f01bf3fc4f2f7c25dbb2338f058f46c7/listeners.go#L205-L231

So that's why we make a new one even when we are re-using the underlying listener:

https://github.com/caddyserver/caddy/blob/96058538f01bf3fc4f2f7c25dbb2338f058f46c7/listeners.go#L79

marten-seemann:
Closing this PR due to the considerations outlined in https://github.com/lucas-clemente/quic-go/issues/2597#issuecomment-740624756.

PR #2896, opened 2020-11-21
set the UDP receive buffer size on Windows

### LINKED ISSUE #2837: use golang.org/x/sys/windows for syscalls on Windows
It appears that the `inspectReadBuffer` function here is a no-op on Windows, presumably because `syscall.GetsockoptInt` in the standard library is not implemented for Windows:
```go
func GetsockoptInt(fd Handle, level, opt int) (int, error) { return -1, EWINDOWS }
```
Since the syscall package is frozen, that functionality is not likely to make it into the standard library. However, it _is_ implemented in the golang.org/x/sys/windows package, as described in the link I posted.

_Originally posted by @dswarbrick in https://github.com/lucas-clemente/quic-go/pull/2791#discussion_r511426208_

marten-seemann:
Fixes #2837.

### DIFF
--- a/conn_generic.go
+++ b/conn_generic.go
@@ -1,4 +1,4 @@
-// +build !darwin,!linux
+// +build !darwin,!linux,!windows
 
 package quic
 
--- a/conn_windows.go
+++ b/conn_windows.go
@@ -1,0 +1,37 @@
+// +build windows
+
+package quic
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"syscall"
+
+	"golang.org/x/sys/windows"
+)
+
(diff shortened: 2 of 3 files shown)


mvdan (approved):
SGTM, but you should really not use the syscall package directly on linux/darwin and other unixes. Just like you use x/sys/windows for windows, use x/sys/unix for those, which is actively maintained.

The only time it makes sense to use syscall is for std APIs which are stuck with syscall forever, since x/sys/... came after.

lucas-clemente (approved):

marten-seemann:
Good point. Will do so in a separate PR.

PR #2903, opened 2020-11-23
fix decoding of packet numbers in different packet number spaces

### LINKED ISSUE #2902: unpacker is unaware of packet number spaces
This could lead us to incorrectly derive the packet number of different packet number spaces. Especially relevant if the peer greases the starting packet number.

marten-seemann:
Fixes #2902.

This would also justify a new release.


mvdan (approved):
I think I understand the issue and the idea behind the fix, so SGTM. Thanks for the commit message. Though this change is relatively large and I'm not familiar enough with the codebase to fully say that I think it's correct.

PR #2905, opened 2020-11-26
various improvements to the packet number generator

### LINKED ISSUE #398: reduce frequency of skipped packets on long-lived connections
Currently we're randomly skipping 1 out of 500 packet numbers, and expecting the client to NACK the skipped packet. This is intended to prevent an Optimistic ACK attack.
By NACKing the skipped packet, the client effectively proves the ownership of the IP address it's claiming to have. This is something that's more relevant at the beginning of the connection, and we could increase the frequency in a long-lived connection.

Note that after rebinding of the address, i.e. when the client does a connection migration, the frequency should be reset to the initial value. See https://github.com/quicwg/base-drafts/issues/161.

marten-seemann:
Fixes #398.

We skip packet numbers to defend against the [Optimistic ACK attack](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#name-optimistic-ack-attack).

1. Reduce calls to `crypto/rand.Read`. Instead, use a `math/rand.Rand`, initialized with cryptographic random.
2. Only skip packet numbers in the application data packet number space. Initial and Handshake packets are only used for very few packets during the connection, so there was only a small chance we'd skip a PN there anyway.
3. Implement an exponential backoff for skipping packet numbers, starting at an expected frequency of once per 256 packets, going up to once per 128*1024 packets.

A good way to think about this is by looking at the packet numbers skipped per max congestion windows (10000 packets). The current algorithm skips (on average) one PN in 500, which leads to 20 skipped packet numbers per cwnd (and therefore, per RTT). As a consequence, the ACK frames will have roughly 20 ACK ranges (assuming no packet loss). Clearly, this is a waste.

### DIFF
--- a/internal/ackhandler/packet_number_generator.go
+++ b/internal/ackhandler/packet_number_generator.go
@@ -6,6 +6,7 @@ import (
 	mrand "math/rand"
 
 	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/utils"
 )
 
 type packetNumberGenerator interface {
@@ -37,8 +38,9 @@ func (p *sequentialPacketNumberGenerator) Pop() protocol.PacketNumber {
 // it randomly skips a packet number every averagePeriod packets (on average).
 // It is guaranteed to never skip two consecutive packet numbers.
 type skippingPacketNumberGenerator struct {
-	rand          *mrand.Rand
-	averagePeriod protocol.PacketNumber
+	rand      *mrand.Rand
+	period    protocol.PacketNumber
+	maxPeriod protocol.PacketNumber
 
 	next       protocol.PacketNumber
 	nextToSkip protocol.PacketNumber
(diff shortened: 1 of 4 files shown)


lucas-clemente (commented):
LGTM, but two questions:

- Why isn't the optimistic ack attack an issue for long-lived connections? Sure, it requires more upfront investment from an attacker, but then they can still max out the server's connection "for free"?
- Using a mathematical PRNG seeded with cryptographic random means an attacker could (in principle) reconstruct the seed, and avoid the defense. Not an issue if we don't worry about long-lived connections.

marten-seemann:
> Why isn't the optimistic ack attack an issue for long-lived connections? Sure, it requires more upfront investment from an attacker, but then they can still max out the server's connection "for free"?

It is, but we don't need a check every `n` packets. Note that we never stop skipping packet numbers, we just do it less and less frequently. By using exponential backoff, we make sure that the peer received about as many packets as it can expect to receive before the next skip as it has already successfully acknowledged. I believe that this is "good enough" defense against an attack that is not that costly on our side.

> Using a mathematical PRNG seeded with cryptographic random means an attacker could (in principle) reconstruct the seed, and avoid the defense. Not an issue if we don't worry about long-lived connections.

Unfortunately, I don't have precise numbers here how many numbers you can receive from `math/rand.Rand` before it becomes feasible to reconstruct the internal state. My intuition is *many*, and I'm assuming that's good enough to make this attack impractical in practice.

PR #2906, opened 2020-11-27
disable key updates when using HTTP/3 to avoid breaking Chrome 87

### LINKED ISSUE #2901: Chrome 87 fails on key updates
Chrome 87 with `--enable-quic --quic-version=h3-29` downloads a large file from quic-go's example server on localhost (`sudo ./main -bind 127.0.0.1:443 -www /tmp -qlog` with some certificate and /etc/hosts setup).

After the first several megabytes the connection stalls and the server reports:
```
server Peer closed session with error: NO_ERROR: 85:Network blackhole detected
server could not flush to stream: NO_ERROR: 85:Network blackhole detected
```

The qlog is quite large so I won't attach it. I believe this is now easy to reproduce.

marten-seemann:
Fixes #2901.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"net/http"
 	"runtime"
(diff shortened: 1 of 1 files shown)


mvdan (approved):
SGTM. Do we want a TODO to remember to remove this at some point later, once Chrome 87 is no longer supported/used anymore?

lucas-clemente (approved):

marten-seemann:
@mvdan Opened a new issue instead: #2926.

PR #2907, opened 2020-11-27
Fix request tls

mpx:


http3/server.go
```diff
@@ -277,6 +278,9 @@ func (s *Server) handleRequest(sess quic.Session, str quic.Stream, decoder *qpac
 	req.RemoteAddr = sess.RemoteAddr().String()
 	req.Body = newRequestBody(str, onFrameError)
 
+	cs := qtls.ToTLSConnectionState(sess.ConnectionState())
```
marten-seemann, line 281: `ConnectionState()` blocks until the handshake completes, so this would delay responses to 0-RTT requests, wouldn't it?

mpx: Ah, that's awkward. It would prevent 0-RTT requests.

I suspect it makes more sense for `ConnectionState()` to not block successful 0-RTT connections. The `tls.ConnectionState` fields should all be resolved by then. I need this to validate/use `tls.ConnectionState.ServerName`. The other fields are useful too. There is currently no way to do this with the `http3` implementation.

I haven't stepped through the code to see how difficult it is to change `ConnectionState()` tho'

marten-seemann: The code is in the [qtls repo](https://github.com/marten-seemann/qtls-go1-15), basically the same logic as in `crypto/tls`, if you want to give it a try.

PR #2909, opened 2020-11-28
add a generic Debug() function to the connection tracer

marten-seemann:
This might not be the cleanest API, but we really need a way to get custom events into qlog / qvis, to facilitate debugging.


integrationtests/self/key_update_test.go
```diff
@@ -89,6 +89,7 @@ func (t *connTracer) DroppedKey(logging.KeyPhase)
 func (t *connTracer) SetLossTimer(logging.TimerType, logging.EncryptionLevel, time.Time) {}
 func (t *connTracer) LossTimerExpired(logging.TimerType, logging.EncryptionLevel)        {}
 func (t *connTracer) LossTimerCanceled()                                                 {}
+func (t *connTracer) Log(string, string)                                                 {}
```
lucas-clemente, line 92: Maybe give it a more descriptive name if it's meant as "hack"? `Debug`?

lucas-clemente (approved):

integrationtests/self/key_update_test.go
marten-seemann: Good point. Done.

PR #2923, opened 2020-12-02
improve the warning about the UDP receive buffer size

marten-seemann:
1. only print it once
2. link to the wiki page: https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size

### DIFF
--- a/packet_handler_map.go
+++ b/packet_handler_map.go
@@ -4,6 +4,7 @@ import (
 	"crypto/hmac"
 	"crypto/rand"
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"hash"
 	"log"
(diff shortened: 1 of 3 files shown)


packet_handler_map.go
```diff
@@ -54,48 +55,50 @@ type packetHandlerMap struct {
...
 	c net.PacketConn,
 	connIDLen int,
 	statelessResetKey []byte,
 	tracer logging.Tracer,
 	logger utils.Logger,
 ) (packetHandlerManager, error) {
-	setReceiveBuffer(c, logger)
+	if err := setReceiveBuffer(c, logger); err != nil {
+		receiveBufferWarningOnce.Do(func() {
+			log.Printf("%s. See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.\n", err)
```
mvdan, line 99: nit: logs already end in newlines, so you don't need the trailing newline

packet_handler_map_test.go
```diff
@@ -63,6 +66,7 @@ var _ = Describe("Packet Handler Map", func() {
 	})
 
 	JustBeforeEach(func() {
+		log.SetOutput(ioutil.Discard)
```
mvdan, line 69: This is not thread-safe, so I would honestly do it at the start of TestMain and not worry about setting it back to anything else. I assume none of the tests should log directly to stderr, anyway.

mvdan (approved):
I assume a test for this would be overkill :) SGTM modulo two nits.

PR #2927, opened 2020-12-04
use golang.org/x/sys/unix instead of syscall

marten-seemann:
As suggested by @mvdan in https://github.com/lucas-clemente/quic-go/pull/2896#pullrequestreview-544133159. An excellent suggestion, as it allows us to get rid of two OS-specific files.

### DIFF
--- a/conn_ecn.go
+++ b/conn_ecn.go
@@ -9,6 +9,8 @@ import (
 	"syscall"
 	"time"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 	"github.com/lucas-clemente/quic-go/internal/utils"
 )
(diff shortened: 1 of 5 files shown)


mvdan (approved):

conn_ecn.go
```diff
@@ -29,7 +31,7 @@ func inspectReadBuffer(c net.PacketConn) (int, error) {
 	var size int
 	var serr error
 	if err := rawConn.Control(func(fd uintptr) {
-		size, serr = syscall.GetsockoptInt(int(fd), syscall.SOL_SOCKET, syscall.SO_RCVBUF)
+		size, serr = unix.GetsockoptInt(int(fd), syscall.SOL_SOCKET, syscall.SO_RCVBUF)
```
dswarbrick, line 34: These constants exist in the unix package also, for consistency.
```suggestion
		size, serr = unix.GetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_RCVBUF)
```

marten-seemann: Thank you, missed that one!

marten-seemann:
Looks like I was too optimistic here.
On Linux, the `Type` of the TOS control message is `unix.IP_TOS`, whereas on Darwin it's `unix.IP_RECVTOS` 🤦🏻‍♂️:
https://github.com/lucas-clemente/quic-go/blob/5a0ce242654c153128ca8679e8fe19fc3be966d3/conn_ecn.go#L96-L105

That means we'll still need OS-specific files to set the `msgTypeIPTOS` constant.

PR #2940, opened 2020-12-15
use cryptographic random for determining skipped packet numbers

marten-seemann:
As @lucas-clemente suggested in https://github.com/lucas-clemente/quic-go/pull/2905#pullrequestreview-551638802.

### DIFF
--- a/internal/ackhandler/packet_number_generator.go
+++ b/internal/ackhandler/packet_number_generator.go
@@ -3,7 +3,6 @@ package ackhandler
 import (
 	"crypto/rand"
 	"encoding/binary"
-	mrand "math/rand"
 
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 	"github.com/lucas-clemente/quic-go/internal/utils"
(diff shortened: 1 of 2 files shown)


internal/ackhandler/packet_number_generator.go
```diff
@@ -34,25 +33,45 @@ func (p *sequentialPacketNumberGenerator) Pop() protocol.PacketNumber {
...
+}
+
+// copied from the standard library math/rand implementation of Int63n
+func (r *rng) Int31n(n int32) int32 {
+	if n&(n-1) == 0 { // n is power of two, can mask
+		return r.Int31() & (n - 1)
+	}
+	max := int32((1 << 31) - 1 - (1<<31)%uint32(n))
+	v := r.Int31()
+	for v > max {
```
lucas-clemente, line 52: Maybe increase the fallback by powers of two? That would allow you to drop this loop / skip the loop in crypto/rand.Int()?

marten-seemann: Not sure I understand your comment. tbh, I just copied this from the `math/rand` implementation of `Int63n`. `max` should be very close to `math.MaxInt32`, so this loop is very unlikely to be executed.

PR #2949, opened 2020-12-21
implement HTTP/3 control stream handling

marten-seemann:
We'll need this if we ever want to read the SETTINGS frame. This frame is used to negotiate HTTP/3 extensions.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -2,6 +2,7 @@ package http3
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -109,7 +110,7 @@ func (c *client) dial() error {
 		return err
 	}
 
-	// run the sesssion setup using 0-RTT data
+	// send the SETTINGs frame, using 0-RTT data, if possible
 	go func() {
 		if err := c.setupSession(); err != nil {
 			c.logger.Debugf("Setting up session failed: %s", err)
@@ -117,6 +118,7 @@ func (c *client) dial() error {
 		}
 	}()
 
+	go c.handleUnidirectionalStreams()
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


http3/server.go
```diff
@@ -254,6 +258,33 @@ func (s *Server) handleConn(sess quic.EarlySession) {
 	}
 }
 
+func (s *Server) handleUnidirectionalStreams(sess quic.EarlySession) {
+	for {
+		str, err := sess.AcceptUniStream(context.Background())
+		if err != nil {
+			return
```
lucas-clemente, line 266: Can/should we somehow handle the error? Log at least?

lucas-clemente: Same for the other exit points in this fn.

http3/client.go
```diff
@@ -127,15 +129,45 @@ func (c *client) setupSession() error {
...
 	buf := &bytes.Buffer{}
-	// write the type byte
-	buf.Write([]byte{0x0})
+	utils.WriteVarInt(buf, streamTypeControlStream)
 	// send the SETTINGS frame
 	(&settingsFrame{}).Write(buf)
-	if _, err := str.Write(buf.Bytes()); err != nil {
-		return err
-	}
+	_, err = str.Write(buf.Bytes())
```
lucas-clemente, line 135: Ignoring short writes here?

http3/client.go
```diff
@@ -127,15 +129,45 @@ func (c *client) setupSession() error {
...
+	return err
+}
 
-	return nil
+func (c *client) handleUnidirectionalStreams() {
+	for {
+		str, err := c.session.AcceptUniStream(context.Background())
+		if err != nil {
+			return
```
lucas-clemente, line 143: Handle / log the error?

lucas-clemente: Same below

lucas-clemente (approved):

http3/client.go
marten-seemann: Not ignoring it. But there's no temporary `Write` errors.

Issue #2933, opened 2020-12-11
Go 1.16: use net.ErrClosed


marten-seemann:
From the docs:
> ErrClosed is the error returned by an I/O call on a network connection that has already been closed, or that is closed by
another goroutine before the I/O is completed. This may be wrapped in another error, and should normally be tested using errors.Is(err, net.ErrClosed).

PR #2953, opened 2020-12-23
add support for Go 1.16

marten-seemann:
@mvdan FYI.

As always, the most important change here is the rebase of the qtls changes on top the new crypto/tls.
Regarding new features in Go 1.16, the only thing I'm currently aware of is using `net.ErrClosed` (see #2933).

### DIFF
--- a/.github/workflows/cross-compile.yml
+++ b/.github/workflows/cross-compile.yml
@@ -3,13 +3,14 @@ jobs:
   crosscompile:
     strategy:
       matrix:
-        go: [ "1.14.x", "1.15.x" ]
+        go: [ "1.14.x", "1.15.x", "1.16.0-beta1" ]
     runs-on: ubuntu-latest
     name: "Cross Compilation (Go ${{matrix.go}})"
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
+          stable: !contains(${{ matrix.go }}, "beta") && !contains(${{ matrix.go }}, "rc")
           go-version: ${{ matrix.go }}
       - name: Install build utils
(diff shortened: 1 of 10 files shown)


.github/workflows/cross-compile.yml
```diff
@@ -3,13 +3,14 @@ jobs:
...
       matrix:
-        go: [ "1.14.x", "1.15.x" ]
+        go: [ "1.14.x", "1.15.x", "1.16.0-beta1" ]
     runs-on: ubuntu-latest
     name: "Cross Compilation (Go ${{matrix.go}})"
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
+          stable: contains(${{ matrix.go }}, "beta") || contains(${{ matrix.go }}, "rc")
```
mvdan, line 13: should this not be "unstable" or "prerelease"?

mvdan (approved):
Can't really comment on the TLS code, but the rest looks good.

lucas-clemente (approved):

.github/workflows/cross-compile.yml
marten-seemann: Not sure what you mean, but I think I go the condition here wrong (need to negate it).

mvdan: Yeah, that's what I mean. The logic you have is the opposite of stable, so I was suggesting names.

marten-seemann: `stable` is defined by the setup-go, so we can't change that. I agree with you that it would be better if a boolean option with a default would default to `false`, but maybe that's a too Go-y way of thinking.

PR #2961, opened 2020-12-27
Fill http.Request.TLS with TLS ConnectionState

### LINKED ISSUE #2879: Cannot access TLS ConnectionState from a http.Request
In http3/server.go some information is copied from the quic.Session to the request in the method handleRequest. This allows one to access the data, e.g. RemoteAddr, in the ServeHTTP handler.

Unfortunately the request.TLS is not populated this way, so there is no way to access the quic.Session.ConnectionState() from a request. This means that there is no way to access a client certificate from a request handler.

A simple fix is to populate the request.TLS with data from the quic session in handleRequest.

skirsten:
This sets the `http.Request.TLS` to the `quic.ConnectionState` obtained from the QUIC session.

Inspired by https://github.com/mpx/quic-go/commit/b37e51a814a9698ba86c1f405b3f4c14f26d291d

closes #2879

### DIFF
--- a/http3/request.go
+++ b/http3/request.go
@@ -1,7 +1,6 @@
 package http3
 
 import (
-	"crypto/tls"
 	"errors"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


skirsten:
@googlebot I signed it!

marten-seemann (commented):
`ConnectionState` blocks until the TLS handshake is complete. This means that we wouldn't be able to handle 0-RTT requests any more, right?

skirsten:
I have not considered that yet. In my testing (integrated into caddy) everything was working fine, but I'm unsure if I also tested the 0-RTT resumption.

PR #2962, opened 2020-12-29
add a http3.RoundTripOpt to skip the request scheme check

marten-seemann:
Otherwise, we'll only be able to issue https requests. This is what we usually want, but for MASQUE support, the URL will be of the form masque://example.org.

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -61,11 +61,12 @@ type RoundTripper struct {
 
 // RoundTripOpt are options for the Transport.RoundTripOpt method.
 type RoundTripOpt struct {
-	// OnlyCachedConn controls whether the RoundTripper may
-	// create a new QUIC connection. If set true and
-	// no cached connection is available, RoundTrip
-	// will return ErrNoCachedConn.
+	// OnlyCachedConn controls whether the RoundTripper may create a new QUIC connection.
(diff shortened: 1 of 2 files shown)


http3/roundtrip.go
```diff
@@ -66,6 +66,8 @@ type RoundTripOpt struct {
 	// no cached connection is available, RoundTrip
 	// will return ErrNoCachedConn.
 	OnlyCachedConn bool
+	// SkipSchemeCheck controls whether we check if the scheme is https.
+	SkipSchemeCheck bool
```
lucas-clemente, line 70: How does Go's HTTP package deal with this? Does it check the scheme?

(In any case, the docs could be a bit more exhaustive – e.g. mention masque as the motivating use case.)

lucas-clemente (approved):

http3/roundtrip.go
marten-seemann: It does, see https://github.com/golang/go/blob/520f3b72db7befab2028d9a47376267cf2d274a9/src/net/http/transport.go#L516 and
https://github.com/golang/go/blob/520f3b72db7befab2028d9a47376267cf2d274a9/src/net/http/transport.go#L546-L549. Only "http" and "https" are allowed. Good point on expanding the comment, will do.

PR #2963, opened 2020-12-29
expose methods to send and receive datagrams on the H3 response writer

marten-seemann:
Needed for MASQUE.

### DIFF
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -441,7 +441,7 @@ var _ = Describe("Client", func() {
 
 		It("returns a response", func() {
 			rspBuf := &bytes.Buffer{}
-			rw := newResponseWriter(rspBuf, utils.DefaultLogger)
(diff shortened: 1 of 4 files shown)


lucas-clemente (approved):

marten-seemann:
This PR is incomplete. H3 Datagrams come with a flow ID, and we should make sure we properly parse it.

ydnar:
@marten-seemann thanks for this PR. I incorporated a variation of it here: https://github.com/alta/quic-go/commit/f84dd058a9bf87d5453b68f9c51a9611a1d0b44a

I think there needs to be something that wraps `quic.EarlySession` to track individual H3 sessions so datagrams and client streams are correctly routed to the correct H3 session. Thankfully the H3 datagram and WebTransport spec specify that an H3 datagram frame is prefixed with the stream ID of the control stream for each H3 session.

PR #2965, opened 2020-12-30
make it possible to send and receive datagrams for a HTTP/3 client

marten-seemann:
Depends on #2963.

Most users will just need to call `http3.RoundTripper.RoundTrip()`. For users that need access to the underlying QUIC connection, they can get a particular client (which owns one QUIC session) using the newly added `GetRoundTripperForRequest`. This will be heavily used by MASQUE.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -61,7 +61,10 @@ type client struct {
 	logger utils.Logger
 }
 
-var _ http.RoundTripper = &client{}
+var (
+	_ http.RoundTripper = &client{}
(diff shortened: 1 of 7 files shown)


lucas-clemente (approved):

marten-seemann:
This PR is incomplete. H3 Datagrams come with a flow ID, and we should make sure we properly parse it.

ydnar:
@marten-seemann what’s the plan (if any) to resurrect this functionality?

thanks!

PR #2971, opened 2020-12-31
stop generating new packets when the send queue is full

### LINKED ISSUE #2941: run loop is blocked by sending packets
`WriteTo` comes with no guarantees when it will unblock. During the time we're blocked on sending a packet, we can't receive any new packets, thus the `receivedPackets` channel can fill up and start dropping new incoming packets.

marten-seemann:
Fixes #2941. Depends on #2980.

We shouldn't block the `run` loop in the session if `WriteTo` is too slow to send out packets. If we do that, we won't be able to receive any incoming packets, and the `receivedPackets` channel might overflow, leading to avoidable packet loss.


session.go
```diff
@@ -563,8 +566,9 @@ runLoop:
 			// We do all the interesting stuff after the switch statement, so
 			// nothing to see here.
 		case <-s.sendingScheduled:
-			// We do all the interesting stuff after the switch statement, so
-			// nothing to see here.
+		// We do all the interesting stuff after the switch statement, so
+		// nothing to see here.
+		case <-sendQueueAvailable:
```
lucas-clemente, line 571: Do we need to drain `sendQueueAvailable` if we end up taking another path in this switch, and then enter the if checks below? Otherwise it could be that we incorrectly assume there's space available even if there isn't, right?

send_queue.go
```diff
@@ -1,29 +1,55 @@
...
+	WouldBlock() bool
+	Available() <-chan struct{}
+	Close()
+}
+
 type sendQueue struct {
 	queue       chan *packetBuffer
 	closeCalled chan struct{} // runStopped when Close() is called
 	runStopped  chan struct{} // runStopped when the run loop returns
+	available   chan struct{}
```
lucas-clemente, line 15: This is a bit difficult to follow, and tightly coupled to run loop internals (e.g. if you use it wrong it's inherently racy, see below).

I wonder whether there's a simpler way of achieving similar behavior? Not sure if it's simpler, but possibly a nextPacket member in the session that we try to send in the session's switch statement?

(If we decide to stick with this, we should probably document the reasoning behind this API.)

session.go
marten-seemann: The `select` can unblock for multiple reasons (sending scheduled, receiving a packet, etc.), so we always have to check if there's room in the send queue before sending a packet: https://github.com/lucas-clemente/quic-go/blob/b4ffb97831546fc8336ed5f0c37ed6b6c8c40246/session.go#L621-L629

send_queue.go
marten-seemann: I'm not sure I understand why it's inherently race. It's true that the `sendQueue` can only be used from a single Go routine (which we guarantee by just running it from the session), but as long as you adhere to this, there shouldn't be any race here.

session.go
marten-seemann: I think I misunderstood your comment. Draining the channel makes sense if `WouldBlock()` returns `true`.

marten-seemann: Actually, that would be racy (the send queue could have unblocked between the calls to `WouldBlock` and Available`.

PR #2973, opened 2021-01-01
publicize QUIC varint reading and writing

marten-seemann:
Packages outside of quic-go might want to make use of QUIC's varint format. We should expose the respective functions, so they don't have to reimplement this.

Specifically, MASQUE uses varints for the [Stream Chunk](https://www.ietf.org/archive/id/draft-ietf-masque-connect-udp-02.html#name-stream-chunks) encoding.

### DIFF
--- a/framer.go
+++ b/framer.go
@@ -5,8 +5,8 @@ import (
 
 	"github.com/lucas-clemente/quic-go/internal/ackhandler"
 	"github.com/lucas-clemente/quic-go/internal/protocol"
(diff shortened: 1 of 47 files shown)


lucas-clemente (approved):
WDYT about s/VarInt//g in the function names in the package? `quickvarint.ReadVarInt` reads a bit redundant.

marten-seemann:
That's what the second commit does.

marten-seemann:
But it looks like I forgot `WriteVarInt`. Fixed now.

PR #2983, opened 2021-01-04
only run gofumpt on .go files in pre-commit hook

marten-seemann:
Editing `README.md` otherwise produces a warning like this:
```
README.md:1:1: illegal character U+0023 '#'
```

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -13,8 +13,8 @@ for f in $(git diff --cached --name-only); do
 	fi
 done
 
-# Check that all files are properly gofumpt-ed.
-output=$(gofumpt -d $(git diff --cached --name-only))
+# Check that all Go files are properly gofumpt-ed.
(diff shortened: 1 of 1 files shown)


mvdan (approved):
Huh, I would have thought that `*.go` would be enough for git. Is three stars any different?

marten-seemann:
@mvdan You're right. The 3 stars are supposed to do recursive matching, however...
```
 git diff --cached --name-only -- '*.go'
```
already matches recursively, whereas 
```
 git diff --cached --name-only -- *.go
```
does not. By the way,
```
 git diff --cached --name-only -- ***.go
```
doesn't match recursively either, but
```
git diff --cached --name-only -- **/*.go
```
does.

Please remember this for the next time you use git to filter through files ;)

mvdan:
Those are different :) I always use `'*.go'`, since that literally just passes that string to git, meaning "any indexed file which has a go extension". Without the quoting, then the shell just tries to expand to the Go files in the current directory, which is entirely different.

PR #2989, opened 2021-01-07
fix response content length

### LINKED ISSUE #2398: ContentLength equals zero with response body
I'm currently trying to use v0.14.4 of quic-go. When a HTTP3 response is returned, the `ContentLength` is set to zero but the response body has valid data. I'm getting an error later while processing the response body because the `ContentLength` is 0. I was previously using v0.10.2 without error and mostly just upgraded library version in my go.mod file without really changing the code base. I was expecting for the `ContentLength` to be anything except for zero (i.e -1 or 6016). The response body is a gob encoded JWE object. `KeepAlive` is set to false in the quic config for both the server and the client. The `reqDoneClose` field is set to true so I'm assuming the connection is done. I'm not really sure how to troubleshoot more and any help would be appreciated.

Example of debug output from the response:
```
[DEBUG]HTTP Response:
&{Status:200 OK StatusCode:200 Proto:HTTP/3 ProtoMajor:3 ProtoMinor:0 Header:map[Content-Type:[application/octet-stream]] Body:0xc000410040 ContentLength:0 TransferEncoding:[] Close:false Uncompressed:false Trailer:map[] Request:<nil> TLS:<nil>}
[DEBUG]ContentLength: 0
[DEBUG]Body: &{str:0xc0001961e0 isRequest:false reqDone:0xc00002e360 reqDoneClosed:true onFrameError:0x914e30 bytesRemainingInFrame:0}
[DEBUG]Body Length(read): 6016
```

bachue:
Fixes #2398.

Now http.Response.ContentLength is always 0 even the response header includes content-length, this patch is to fix it.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -327,6 +327,15 @@ func (c *client) doRequest(
 		res.Uncompressed = true
 	} else {
 		res.Body = respBody
+		if contentLengthStr := res.Header.Get("Content-Length"); contentLengthStr != "" {
+			if contentLength, err := strconv.ParseInt(contentLengthStr, 10, 64); err == nil {
+				res.ContentLength = contentLength
+			}
+		} else {
+			if isEnded, _ := respBody.ended(); !isEnded {
+				res.ContentLength = -1
+			}
+		}
 	}
 
 	return res, requestError{}
--- a/http3/body.go
+++ b/http3/body.go
@@ -82,6 +82,17 @@ func (r *body) readImpl(b []byte) (int, error) {
 	return n, err
 }
 
+func (r *body) ended() (bool, error) {
+	_, err := r.readImpl(make([]byte, 0))
+	if err == nil {
+		return false, nil
+	} else if err == io.EOF {
+		return true, nil
+	} else {
+		return false, err
+	}
+}
+
 func (r *body) requestDone() {
 	if r.reqDoneClosed || r.reqDone == nil {
 		return
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -440,7 +440,37 @@ var _ = Describe("Client", func() {
 			Expect(decodeHeader(buf)).To(HaveKeyWithValue(":method", "GET"))
 		})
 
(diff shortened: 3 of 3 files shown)


bachue:
@googlebot I signed it!

marten-seemann:
Is this supposed to be a fix for #2398?

Can you add tests please?

bachue:
@marten-seemann Done

http3/client.go
```diff
@@ -327,6 +327,11 @@ func (c *client) doRequest(
 		res.Uncompressed = true
 	} else {
 		res.Body = respBody
+		if contentLengthStr := res.Header.Get("Content-Length"); contentLengthStr != "" {
```
marten-seemann, line 330: Do we need to set it to -1 if there's no Content-Length header?
See https://github.com/golang/net/blob/4c5254603344ea4a8ae4bed7e296a9588303e14f/http2/transport.go#L2007-L2018.

bachue: @marten-seemann I think you're right

marten-seemann (commented):
If I understand correctly, `-1` should only be set if there's actually a response body, right? Otherwise, this should be `0`.

Can you please
1. fix this and add test cases, both for the 0 and the -1 case. Please create new `It`s for this.
2. squash all your commits into one.

marten-seemann:
@bachue Are you still planning to update this PR?

bachue:
@marten-seemann Working on it

bachue:
@marten-seemann Done

marten-seemann (changes requested):
Why do we need `body.ended()`. Why is it not enough to use logic analogous to https://github.com/golang/net/blob/4c5254603344ea4a8ae4bed7e296a9588303e14f/http2/transport.go#L2007-L2018?

marten-seemann:
@bachue Are you still working on this?

marten-seemann:
Closing due to inactivity. Too bad, we were so close.

PR #2993, opened 2021-01-11
allow access to the underlying quic.Stream from a http.ResponseWriter

### LINKED ISSUE #1492: Expose underlying data stream in h2quic
# Background

When using HTTP/2, `http.Request` and `http.ResponseWriter` do not implement `http.Hijacker`.  This makes sense in the context of TCP because the stream multiplexing happens outside of the wire protocol; the call to `Hijack` would return a TCP connection over which multiple request/response cycles travel, not just the one being hijacked.  As such, it is often said that [HTTP/2 and Hijack are fundamentally incompatible](https://github.com/golang/go/issues/14797#issuecomment-196103814).

In the context of QUIC, this is no longer true.  Each request/response cycle is mapped to its own OSI Layer-4 stream, which can be manipulated independently of others.

#### As such, HTTP/2 over QUIC is not fundamentally incompatible with `Hijack`-like semantics.

# Proposal

I would like to propose a mechanism for exposing the `quic.Stream` that underpins HTTP/2 over QUIC, similar to the `Hijack` semantics in `net/http`.

The use-cases for this feature are mostly the same ones as for WebSocket.  Nevertheless there are a few important design considerations:

1. **This is NOT a proposal to implement WebSocket.**  The intent is not to expose message-oriented semantics, nor any of the special features proposed by the protocol.  The WebSocket spec contains too many corner-cases that map onto QUIC and HTTP/2 quite poorly.  Rather, the idea is to expose the `quic.Stream` directly, enabling bidirectional streaming.
2. **This is NOT a proposal to implement `http.Hijacker`**.  Implementing `http.Hijacker` sends a strong message that this library can be used interchangeably with standard transports.  This is likely to break things that don't expect separate streams for headers and data, e.g. `gorilla/websocket`.  I propose instead to make the data stream available to those who are **intentionally** looking for it, while making it clear this "thing" behaves differently from `Hijack`.

Instead of implementing `Hijacker`, I propose the following:

```go
type DataStreamer interface {
    DataStream() quic.Stream
}
```

I have to admit to a small bit of pleasure that the interface is so simple.  Used in conjunction with `http.ResponseWriter`, application developers can both send response headers **and** enjoy bidirectional streaming with a familiar interface.

Consider the following example, which shows a homegrown protocol for negotiating such a connection upgrade:

```go
func handler(w http.ResponseWriter, r *http.Request) {
    if r.Header.Get("Connection") != "Upgrade" || r.Header.Get("Upgrade") != "QUIC-Stream" {
        http.Error(w, "raw QUIC stream required", http.StatusUpgradeRequired)
        return
    }

    w.WriteHeader(http.StatusSwitchingProtocols)  // this writes to the header stream

    s := w.(h2quic.DataStreamer).DataStream()
    useStream(s)
}
```

# Final note

@lucas-clemente, @marten-seemann and contributors:  I wanted to open a discussion before submitting a PR.  I already have this working on a private branch, which I have been happily using for several months, so I'm happy to submit this implementation for review if you're okay with this in principle.

Cheers,
Louis

marten-seemann:
Fixes #1492.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -3,21 +3,33 @@ package http3
 import (
 	"bufio"
 	"bytes"
-	"io"
 	"net/http"
 	"strconv"
 	"strings"
 
+	"github.com/lucas-clemente/quic-go"
 	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/marten-seemann/qpack"
 )
 
-type responseWriter struct {
-	stream *bufio.Writer
+// DataStreamer lets the caller take over the stream. After a call to DataStream
+// the HTTP server library will not do anything else with the connection.
+//
+// It becomes the caller's responsibility to manage and close the stream.
+//
+// After a call to DataStream, the original Request.Body must not be used.
+type DataStreamer interface {
+	DataStream() quic.Stream
+}
 
-	header        http.Header
-	status        int // status code passed to WriteHeader
-	headerWritten bool
+type responseWriter struct {
+	stream         quic.Stream // needed for DataStream()
+	bufferedStream *bufio.Writer
+
+	header         http.Header
+	status         int // status code passed to WriteHeader
+	headerWritten  bool
+	dataStreamUsed bool // set when DataSteam() is called
 
 	logger utils.Logger
 }
@@ -25,13 +37,15 @@ type responseWriter struct {
 var (
 	_ http.ResponseWriter = &responseWriter{}
 	_ http.Flusher        = &responseWriter{}
+	_ DataStreamer        = &responseWriter{}
 )
 
-func newResponseWriter(stream io.Writer, logger utils.Logger) *responseWriter {
+func newResponseWriter(stream quic.Stream, logger utils.Logger) *responseWriter {
 	return &responseWriter{
-		header: http.Header{},
-		stream: bufio.NewWriter(stream),
-		logger: logger,
+		header:         http.Header{},
+		stream:         stream,
+		bufferedStream: bufio.NewWriter(stream),
+		logger:         logger,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


http3/response_writer.go
```diff
@@ -3,35 +3,49 @@ package http3
...
 	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/marten-seemann/qpack"
 )
 
+// DataStreamer lets the caller take over the stream. After a call to DataStreamer
+// the HTTP server library will not do anything else with the connection.
+//
+// It becomes the caller's responsibility to manage and close the connection.
+//
+// After a call to DataStreamer, the original Request.Body must not be used.
```
lucas-clemente, line 20: -r

http3/response_writer.go
```diff
@@ -3,35 +3,49 @@ package http3
...
 	"net/http"
 	"strconv"
 	"strings"
 
+	"github.com/lucas-clemente/quic-go"
 	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/marten-seemann/qpack"
 )
 
+// DataStreamer lets the caller take over the stream. After a call to DataStreamer
```
lucas-clemente, line 15: -r

http3/response_writer.go
```diff
@@ -3,35 +3,49 @@ package http3
...
+// the HTTP server library will not do anything else with the connection.
```
lucas-clemente, line 16: is connection the best term here?

lucas-clemente (approved):

http3/response_writer.go
marten-seemann: That was copied from `http.Hijacker` ;)
Of course it's the stream here.

PR #3003, opened 2021-01-17
immediately complete streams that were canceled, drop retransmissions

### LINKED ISSUE #3002: SendStream doesn't complete if canceled when it has inflight frames or queued retransmissions
`cancelWriteImpl` does not call `onStreamCompleted` if there is inflight frames for the stream. If any of those inflight frames is lost then `queueRetransmission` is called. It does not check for completion but queues the frame for retransmission. That retransmission never happens because `popStreamFrame` checks if `cancelWrite` flag has been set and returns early. `onStreamCompleted` is never called.

I think `queueRetransmission` should check `cancelWrite` flag and not add the frame to retransmission queue if set. Then it would check for completion. `cancelWriteImpl` should also clear the retransmission queue before checking for completion.

marten-seemann:
Fixes #3002.

@rip-create-your-acccount Would be great if you could review this PR!

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -346,6 +346,10 @@ func (s *sendStream) frameAcked(f wire.Frame) {
 	f.(*wire.StreamFrame).PutBack()
 
 	s.mutex.Lock()
+	if s.canceledWrite {
+		s.mutex.Unlock()
+		return
+	}
 	s.numOutstandingFrames--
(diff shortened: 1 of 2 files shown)


rip-create-your-account:
Looks good to me. I agree that the immediate completion is better.

PR #3020, opened 2021-01-23
make sure the server is stopped before closing all server sessions

### LINKED ISSUE #3019: New session is left to linger if created while (*packetHandlerMap).CloseServer is executing
This happens when `(*baseServer).Close` calls `(*packetHandlerMap).CloseServer` before `(*baseServer).createNewSession` calls `(*packetHandlerMap).AddWithConnID`.

`(*packetHandlerMap).CloseServer` will shutdown the active sessions while `(*packetHandlerMap).AddWithConnID` blocks on the mutex. Eventually `(*packetHandlerMap).AddWithConnID` runs and adds the Session successfully. The Session then ends in `(*baseServer).handleNewSession` where it waits until handshake/idle timeout. `(*baseServer).Close` does nothing additional that could cause this Session to close.

I didn't find easy way to write a test to validate this so this is only based on reading the code.

marten-seemann:
Fixes #3019.

@rip-create-your-account Do you think this fixes the bug that you discovered?

### DIFF
--- a/server.go
+++ b/server.go
@@ -274,25 +274,27 @@ func (s *baseServer) accept(ctx context.Context) (quicSession, error) {
 
 // Close the server
 func (s *baseServer) Close() error {
-	s.sessionHandler.CloseServer()
 	s.mutex.Lock()
-	defer s.mutex.Unlock()
 	if s.closed {
+		s.mutex.Unlock()
 		return nil
 	}
 	if s.serverError == nil {
 		s.serverError = errors.New("server closed")
 	}
-	var err error
 	// If the server was started with ListenAddr, we created the packet conn.
 	// We need to close it in order to make the go routine reading from that conn return.
-	if s.createdPacketConn {
-		err = s.sessionHandler.Destroy()
-	}
+	createdPacketConn := s.createdPacketConn
 	s.closed = true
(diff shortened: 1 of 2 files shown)


lucas-clemente (approved):

server.go
```diff
@@ -274,23 +274,26 @@ func (s *baseServer) accept(ctx context.Context) (quicSession, error) {
...
 	s.closed = true
 	close(s.errorChan)
+	s.mutex.Unlock()
+
+	s.sessionHandler.CloseServer()
+	var err error
+	if createdPacketConn {
+		err = s.sessionHandler.Destroy()
+	}
 	<-s.running
```
rip-create-your-account, line 297: I think `CloseServer()` could still be called before `(*baseServer).run` finishes processing a packet and calls `AddWithConnID` with the same end result.

I think receiving `<-s.running` is necessary before calling `CloseServer()` to ensure that `(*baseServer).run` is not processing a packet. 
```golang
	<-s.running
	s.sessionHandler.CloseServer()
  var err error
	if createdPacketConn {
		err = s.sessionHandler.Destroy()
	}
```

marten-seemann: Right. Thank you!

PR #3031, opened 2021-02-05
add support for QUIC draft-34

marten-seemann:
QUIC draft-34 will (if everything goes according to plan at the IETF) become QUIC v1.

draft-34 uses a different salt for Initial packet encryption, and it changes the TLS extension code point for the QUIC transport parameter extension (yay!).

We're not allowed to deploy this version on the public internet yet. I added a note for `SupportedVersions`, so we don't forget that when cutting the next release.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -32,6 +32,7 @@ var (
 const (
 	nextProtoH3Draft29      = "h3-29"
 	nextProtoH3Draft32      = "h3-32"
+	nextProtoH3Draft34      = "h3-34"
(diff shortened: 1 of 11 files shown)


lucas-clemente (approved):

marten-seemann:
I got this wrong in part. draft-34 is draft-34, not version 1. *Maybe* (after IESG review), there'll be a version 1, which will be identical to draft-34 though.

I updated the PR accordingly.

PR #3033, opened 2021-02-07
qlog the quic-go version

marten-seemann:
Using `runtime/debug.ReadBuildInfo()` only works when quic-go is used as a library, so that's *almost* good enough for us. For the interop runner image, we can manually set the version string using `ldflags`.

Thanks to @mvdan for pointing me towards `BuildInfo`!

### DIFF
--- a/qlog/qlog.go
+++ b/qlog/qlog.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log"
 	"net"
+	"runtime/debug"
 	"sync"
 	"time"
 
@@ -16,6 +17,24 @@ import (
 
 	"github.com/francoispqt/gojay"
 )
+
+var quicGoVersion = "(devel)"
+
+func init() {
+	info, ok := debug.ReadBuildInfo()
+	if !ok { // no build info available. This happens when quic-go is not used as a library.
+		return
+	}
+	for _, d := range info.Deps {
+		if d.Path == "github.com/lucas-clemente/quic-go" {
+			quicGoVersion = d.Version
+			if d.Replace != nil {
+				quicGoVersion += " (replaced)"
+			}
+			break
+		}
+	}
+}
 
 const eventChanSize = 50
 
--- a/qlog/qlog.go
+++ b/qlog/qlog.go
@@ -18,9 +18,15 @@ import (
 	"github.com/francoispqt/gojay"
 )
 
+// Setting of this only works when quic-go is used as a library.
+// When building a binary from this repository, the version can be set using the following go build flag:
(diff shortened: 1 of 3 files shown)


qlog/qlog.go
```diff
@@ -17,6 +18,27 @@ import (
...
+func init() {
+	info, ok := debug.ReadBuildInfo()
+	if !ok { // no build info available. This happens when quic-go is not used as a library.
+		return
+	}
+	for _, d := range info.Deps {
+		if d.Path == "github.com/lucas-clemente/quic-go" {
+			quicGoVersion = d.Version
+			if d.Replace != nil {
+				quicGoVersion += " (replaced)"
```
mvdan, line 35: if the module that replaces quic-go has a version, you could just show its version. one could replace quic-go with an older version of itself to force a downgrade, for example.

qlog/qlog.go
```diff
@@ -17,6 +18,27 @@ import (
...
+// When building a binary from this repository, the version can be set using the following go build flag:
+// -ldflags="-X github.com/lucas-clemente/quic-go/qlog.quicGoVersion=foobar"
+var quicGoVersion = "(devel)"
+
+func init() {
+	info, ok := debug.ReadBuildInfo()
+	if !ok { // no build info available. This happens when quic-go is not used as a library.
+		return
+	}
+	for _, d := range info.Deps {
```
mvdan, line 31: you could also check the case where quic-go is the main module, and has a version.

this will be rather niche, but it could happen with e.g. `go test github.com/lucas-clemente/quic-go/...` from another module.

qlog/qlog.go
```diff
@@ -17,6 +18,27 @@ import (
 	"github.com/francoispqt/gojay"
 )
 
+// Setting of this only works when quic-go is used as a library.
+// When building a binary from this repository, the version can be set using the following go build flag:
+// -ldflags="-X github.com/lucas-clemente/quic-go/qlog.quicGoVersion=foobar"
+var quicGoVersion = "(devel)"
+
+func init() {
```
mvdan, line 26: this func should be a no-op if quickGoVersion is not (devel). otherwise you can overwrite the explicit ldflags value with the implicit runtime/debug version.

qlog/qlog.go
marten-seemann: I'm not sure if that's a case that we really need to cover.

qlog/qlog.go
marten-seemann: Good point. Added.

qlog/qlog.go
marten-seemann: Fixed.

lucas-clemente (approved):

PR #3039, opened 2021-02-15
fix error message when the UDP receive buffer size can't be increased

marten-seemann:
As reported by @mholt in https://github.com/caddyserver/caddy/pull/3999#issuecomment-778427020.


mholt (approved):
:+1:

PR #3051, opened 2021-02-24
refactor merge packet number ranges

q191201771:
Since we iterate ranges from back to front, it is possible to merge two ranges only if extended at the beginning of a range.

1. It's unnecessary to do a merge-check when extended at the end.
2. `// if the two ranges were not merge, we're done here`, this comment is not matched with the original source code, we're done here because the range was extended(no matter merge or not).


marten-seemann (approved):
Nice!!! I really like this, it makes the code much simpler!
Thank you for the contribution!

PR #3058, opened 2021-03-03
fix issuing of connection IDs when dialing a 0-RTT connections

### LINKED ISSUE #3057: Client issues too many CIDs when client resumes 0-RTT session and client set (*Config).ConnectionIDLength
For me the following diff is enough to make the 0-RTT tests fail. The server sends `CONNECTION_ID_LIMIT_ERROR`.

``` diff
diff --git a/integrationtests/self/zero_rtt_test.go b/integrationtests/self/zero_rtt_test.go
index 3ab95efb..c98eca6f 100644
--- a/integrationtests/self/zero_rtt_test.go
+++ b/integrationtests/self/zero_rtt_test.go
@@ -93,7 +93,11 @@ var _ = Describe("0-RTT", func() {
                                sess, err := quic.DialAddrEarly(
                                        fmt.Sprintf("localhost:%d", proxyPort),
                                        clientConf,
-                                       getQuicConfig(&quic.Config{Versions: []protocol.VersionNumber{version}}),
+                                       getQuicConfig(
+                                               &quic.Config{
+                                                       Versions:           []protocol.VersionNumber{version},
+                                                       ConnectionIDLength: 4,
+                                               }),
                                )
                                str, err := sess.OpenUniStream()
```

It seems that `(*connIDGenerator).SetMaxActiveConnIDs` is first called for the resumed 0-RTT params, then for a second time when the params from the server are received. Every time that function is called it issues new CIDs.

marten-seemann:
Fixes #3057.

When dialing a 0-RTT connection, the client first restores the transport parameters from the original connection, and then applies the transport parameters provided by the server on the new connections.

@rip-create-your-account Do you think this fix is correct?

### DIFF
--- a/conn_id_generator.go
+++ b/conn_id_generator.go
@@ -63,7 +63,7 @@ func (m *connIDGenerator) SetMaxActiveConnIDs(limit uint64) error {
 	// transport parameter.
 	// We currently don't send the preferred_address transport parameter,
 	// so we can issue (limit - 1) connection IDs.
-	for i := uint64(1); i < utils.MinUint64(limit, protocol.MaxIssuedConnectionIDs); i++ {
+	for i := uint64(len(m.activeSrcConnIDs)); i < utils.MinUint64(limit, protocol.MaxIssuedConnectionIDs); i++ {
 		if err := m.issueNewConnID(); err != nil {
 			return err
 		}
--- a/conn_id_generator_test.go
+++ b/conn_id_generator_test.go
@@ -79,6 +79,37 @@ var _ = Describe("Connection ID Generator", func() {
 		Expect(queuedFrames).To(HaveLen(protocol.MaxIssuedConnectionIDs - 1))
 	})
 
+	// SetMaxActiveConnIDs is called twice when we dialing a 0-RTT connection:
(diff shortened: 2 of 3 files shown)


rip-create-your-account:
What happens if the `(*TransportParameters).ActiveConnectionIDLimit` for the new connection is less than that of the 0-RTT params?  [(*TransportParameters).ValidFor0RTT](https://github.com/lucas-clemente/quic-go/blob/2c45f2b11db0ceb1e6d8bc9cd668f27f5dde8b00/internal/wire/transport_parameters.go#L443) doesn't check that field and I don't see server validating it elsewhere. So if the server were to reduce its `ActiveConnectionIDLimit` then same issue would occur as the client would issue CIDs according to the higher limit of the 0-RTT params. Should server reject 0-RTT in such case?

marten-seemann:
Good catch! That’s definitely an oversight, it’s even in the spec: https://www.ietf.org/archive/id/draft-ietf-quic-transport-34.html#name-values-of-transport-paramet

I’ll create a separate PR for that.

rip-create-your-account:
In that case this PR looks like good to me

lucas-clemente (approved):

PR #3060, opened 2021-03-03
only accept 0-RTT it the active_connection_id_limit didn't change

marten-seemann:
As discovered by @rip-create-your-account in https://github.com/lucas-clemente/quic-go/pull/3058#issuecomment-789689206, we need to check that the `active_connection_id_limit` transport parameter didn't change when accepting 0-RTT.

### DIFF
--- a/internal/wire/transport_parameter_test.go
+++ b/internal/wire/transport_parameter_test.go
@@ -482,7 +482,7 @@ var _ = Describe("Transport Parameters", func() {
 		})
 
 		Context("rejects the parameters if they changed", func() {
-			var p *TransportParameters
+			var p TransportParameters
 			params := &TransportParameters{
 				InitialMaxStreamDataBidiLocal:  1,
 				InitialMaxStreamDataBidiRemote: 2,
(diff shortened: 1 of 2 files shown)


rip-create-your-account:
There's a tricky bug here which occurs when the client has already sent the `NewConnectionIDFrames` based on the 0-RTT `(*TransportParameters).ActiveConnectionIDLimit`. If the server has reduced its limit it will with this PR reject the 0-RTT handshake. When the client is notified that the 0-RTT handshake was rejected, [it will drop the 0-RTT](https://github.com/lucas-clemente/quic-go/blob/543d1446900dbc0fd9ad5d0fa521ac1b1b071cfc/internal/handshake/crypto_setup.go#L492) packets. This means that it [will end up retransmitting](https://github.com/lucas-clemente/quic-go/blob/543d1446900dbc0fd9ad5d0fa521ac1b1b071cfc/internal/ackhandler/sent_packet_handler.go#L180) the `NewConnectionIDFrames` of the 0-RTT packets, which is too many of them.

marten-seemann:
@rip-create-your-account That's https://github.com/lucas-clemente/quic-go/issues/2067.

PR #3062, opened 2021-03-03
fix retry key and nonce for draft-34

marten-seemann:
Fortunately, we haven't cut any release that officially supports draft-34 yet.


marten-seemann:
The interop runner is happy with these changes.

PR #3066, opened 2021-03-04
require the application to handle 0-RTT rejection

### LINKED ISSUE #2067: invalidate sent data when 0-RTT is rejected
## Background

When 0-RTT is rejected, this invalidates all application data sent in 0-RTT packets.

[Section 4.2.10 of RFC 8446](https://tools.ietf.org/html/rfc8446#section-4.2.10) says:
> If the server rejects the "early_data" extension, the client application MAY opt to retransmit the Application Data previously sent in early data once the handshake has been completed.  Note that automatic retransmission of early data could result in incorrect assumptions regarding the status of the connection.  For instance, when the negotiated connection selects a different ALPN protocol from what was used for the early data, an application might need to construct different messages.  Similarly, if early data assumes anything about the connection state, it might be sent in error after the handshake completes.

For example, it might be the case that the original connection used application protocol `X`, but the resumed connection negotiates the use of application protocol `Y`. 0-RTT data is sent under the assumption that protocol `X` is used. Obviously, one can't just retransmit that data and assume that it's valid for `Y`.

In QUIC, we also check if the transport parameters used on the old connection are valid. For example, on the original connection the server might have allowed 10 concurrent streams, but that limit was reduced in the mean time and now only allows 5 concurrent streams. A blind retransmission of 0-RTT data might lead to a protocol violation if the client opened more than 5 streams to send 0-RTT data.

## Current Implementation

When dialing a 0-RTT-enabled connection (using `DialEarly`), quic-go returns an `EarlySession`:
https://github.com/lucas-clemente/quic-go/blob/9b627ac93d5e74268118d6c099265de6eb9dee44/interface.go#L199-L210

This is just a normal `Session` with an additional method that allows applications to determine when the handshake completes. When 0-RTT is rejected, we currently just retransmit this data in 1-RTT packets, ignoring the problems described in the previous section.

## Proposal

When 0-RTT is rejected, we return a `quic.Err0RTTRejected` from all calls on the session and on streams. An application that wishes to use 0-RTT needs to handle this error gracefully.

We extend the `EarlySession` interface:
```go
type EarlySession interface {
    Session
    HandshakeCompleted() context.Context
    // If 0-RTT is rejected, the old session becomes invalid.
    // All calls to OpenStream and AcceptStream will return an Err0RTTRejected,
    // as will all calls to Read and Write on streams opened on that session.
    // In that case, NextSession has to be called to obtain the new session.
    // It is invalid to call this method before a Err0RTTRejected has been returned.
    NextSession() Session
}
```

As soon as this error occurs, the application needs to obtain a new session using `NextSession()`. This returns the session established during the handshake. It's the application's responsibility to reopen streams and retransmit data (if it wishes to do so) on this new session.

Internally, we won't allocate a new `session`, but we'll use the call to `NextSession` to reset application state: 
* all flow-control state is reset to its initial state
* the streams map to its initial state, thereby invalidating all streams that might have been opened during 0-RTT
* we remove all 0-RTT packets from the packet history and from the bytes in flight

We *don't* reset any congestion controller state. Having an RTT estimate is valuable and completely independent of the outcome of the cryptographic handshake.

marten-seemann:
Fixes #2067. Depends on #3073.

Implementing the proposal made in this issue, without the wrapper for the easy / common case.

### DIFF
--- a/framer.go
+++ b/framer.go
@@ -1,6 +1,7 @@
 package quic
 
 import (
+	"errors"
 	"sync"
 
 	"github.com/lucas-clemente/quic-go/internal/ackhandler"
@@ -17,6 +18,8 @@ type framer interface {
 
 	AddActiveStream(protocol.StreamID)
 	AppendStreamFrames([]ackhandler.Frame, protocol.ByteCount) ([]ackhandler.Frame, protocol.ByteCount)
+
+	Handle0RTTRejection() error
 }
 
 type framerI struct {
@@ -139,4 +142,30 @@ func (f *framerI) AppendStreamFrames(frames []ackhandler.Frame, maxLen protocol.
 		length += lastFrame.Length(f.version) - lastFrameLen
 	}
 	return frames, length
+}
+
+func (f *framerI) Handle0RTTRejection() error {
+	f.mutex.Lock()
+	defer f.mutex.Unlock()
+
+	f.controlFrameMutex.Lock()
+	f.streamQueue = f.streamQueue[:0]
+	for id := range f.activeStreams {
+		delete(f.activeStreams, id)
+	}
+	var j int
+	for i, frame := range f.controlFrames {
+		switch frame.(type) {
+		case *wire.MaxDataFrame, *wire.MaxStreamDataFrame, *wire.MaxStreamsFrame:
+			return errors.New("didn't expect MAX_DATA / MAX_STREAM_DATA / MAX_STREAMS frame to be sent in 0-RTT")
+		case *wire.DataBlockedFrame, *wire.StreamDataBlockedFrame, *wire.StreamsBlockedFrame:
+			continue
+		default:
+			f.controlFrames[j] = f.controlFrames[i]
+			j++
+		}
+	}
+	f.controlFrames = f.controlFrames[:j]
+	f.controlFrameMutex.Unlock()
+	return nil
 }
--- a/internal/flowcontrol/connection_flow_controller.go
+++ b/internal/flowcontrol/connection_flow_controller.go
@@ -1,6 +1,7 @@
 package flowcontrol
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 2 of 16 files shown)


framer.go
```diff
@@ -140,3 +142,27 @@ func (f *framerI) AppendStreamFrames(frames []ackhandler.Frame, maxLen protocol.
...
+	f.controlFrameMutex.Lock()
+	f.streamQueue = f.streamQueue[:0]
+	for id := range f.activeStreams {
+		delete(f.activeStreams, id)
+	}
+	var j int
+	for i, frame := range f.controlFrames {
+		switch frame.(type) {
+		case *wire.MaxDataFrame, *wire.MaxStreamDataFrame, *wire.MaxStreamsFrame:
+			panic("didn't expect this frame to be sent in 0-RTT")
```
lucas-clemente, line 157: Return an error? Seems like a rather complex condition for a server-crashing assert to me :)

(Then also defer the unlocks.)

internal/flowcontrol/connection_flow_controller.go
```diff
@@ -86,3 +86,16 @@ func (c *connectionFlowController) EnsureMinimumWindowSize(inc protocol.ByteCoun
...
 	c.mutex.Unlock()
 }
+
+// The flow controller is reset when 0-RTT is rejected.
+// All stream data is invalidated, it's if we had never opened a stream and never sent any data.
+// At that point, we only have sent stream data, but we didn't have the keys to open 1-RTT keys yet.
+func (c *connectionFlowController) Reset() {
+	c.mutex.Lock()
+	if c.bytesRead > 0 || c.highestReceived > 0 || !c.epochStartTime.IsZero() {
+		panic("flow controller reset after reading data")
```
lucas-clemente, line 96: Idem

streams_map.go
```diff
@@ -53,6 +58,8 @@ type streamsMap struct {
 	outgoingUniStreams  *outgoingUniStreamsMap
 	incomingBidiStreams *incomingBidiStreamsMap
 	incomingUniStreams  *incomingUniStreamsMap
+
+	resetErr utils.AtomicError // can be nil, otherwise must be an error
```
lucas-clemente, line 62: I find it hard to understand which parts of the streamsMap are expected to be thread safe, probably some comments would help. 

That said, this looks weird to me: The error is set from `ResetFor0RTT` and read from the `Open*` / `Accept*` fns. So those are called from different threads? But then, they read the maps, which are set from `initMaps`, called from  `ResetFor0RTT`? Why does the error need to be protected, but not those maps?

marten-seemann: `Open` and `Accept` is called by the application. When 0-RTT is reset, these calls should return `Err0RTTRejected`, until the application calls `NextSession`.
`initMaps` is called after the error is set, so we don't need introduce a mutex for `Open` and `Accept`. Note that `ResetFor0RTT` also closes the maps, which makes them return an `Err0RTTRejected` if `Open` or `Accept` is called concurrently.

marten-seemann:
@lucas-clemente I introduced a mutex in the `streams_map`. PTAL.

kaiyuhou:
Hi, I noticed the correct http3 client would not handle the "0-RTT rejected" correctly. The http3 client will only raise the error to the user if it received this error and will not resend the request. 

Or do I have the wrong expectation for the http3 client? Should It be defined as an application? If it shouldn't be considered as an application, as an end-user when I received the "0-RTT rejected" error from the http3 client, should I manually call the "Do" function again? 

In this case, will the first data packet be the "1-RTT" case, or actually the "2-RTT" case?

Issue #2607, opened 2020-06-15
use batch read / writes, if available


marten-seemann:
There's a [ReadBatch](https://godoc.org/golang.org/x/net/ipv4#PacketConn.ReadBatch) and a [WriteBatch]([WriteBatch](https://godoc.org/golang.org/x/net/ipv4#PacketConn.WriteBatch) on the `ipv4.PacketConn`. The equivalent functions are available on the `ipv6.PacketConn`.

We need to evaluate if we can either
* convert a `net.PacketConn` passed in by the the user to an `ipv4.PacketConn` or an `ipv6.PacketConn`, depending on the address, or otherwise
* ask users to pass in an `ipv{4,6}.PacketConn`, and then do an interface assertion, and use batched read / writes if possible

marten-seemann:
Also see https://github.com/lucas-clemente/quic-go/issues/2586#issuecomment-642334615.

bt90:
With https://github.com/lucas-clemente/quic-go/pull/3142 merged, what's the state of batch writes?

marten-seemann:
> With #3142 merged, what's the state of batch writes?

It's not really clear if one would want to batch writes on a connection basis or aggregated over multiple (all?) connections. More benchmarking work would be needed to determine the best solution.

bt90:
The main goal is to avoid syscall overhead IMHO. Go seems to use sendmmsg which allows different destinations.

marten-seemann:
That's correct, it's just not clear to me which of the two options will give better performance.

bt90:
Probably also depends on if go can use GSO for UDP and if that works with different destinations.

sendmmsg+GSO should be the goal as outlined by Cloudflares blog post:

https://blog.cloudflare.com/accelerating-udp-packet-transmission-for-quic/

PR #3067, opened 2021-03-04
Use the correct source IP when binding multiple IPs

### LINKED ISSUE #1736: Incorrect source ip address in outgoing UDP datagrams
Hello guys, I've stumbled upon an issue while implementing a test server that listens on all public interfaces of a host.  

When the server is listening on all network interfaces e.g. `0.0.0.0` (assuming server host has more than one public interface) outgoing datagrams are delivered with the wrong source ip address since UDP is connectionless.

The linux kernel will try to guess  the correct source ip for the outgoing datagrams (based on heuristics) but that is not reliable.

As described in [this](https://blog.powerdns.com/2012/10/08/on-binding-datagram-udp-sockets-to-the-any-addresses/) blog post from the PowerDNS guys, the solution is to use `IP_PKTINFO` on Linux (other unixes also have this  feature but with a different API) to read the dest ip from the control message of the incoming datagram with `recvmsg()` and use  `sendmsg()`  to set the source ip of the outgoing datagram.

I've also looked at the source code of the QUIC implementation in chromium and it seems that they are already doing this but I'm not 100% sure (more like 80%) if it's used for the same purpose.

Chromium code reference: https://cs.chromium.org/chromium/src/net/third_party/quic/platform/impl/quic_socket_utils.cc?type=cs&sq=package:chromium&g=0&l=100

I've got a  working POC with `quic-go` based on the `gquic` branch,  here is the commit : https://github.com/ironsteel/quic-go/commit/582e8a1bcefbf6de3e79f6823252f756b6eabb19

So question number 1: Would you guys accept it if I manage to do it in a more portable and sane way?

and question number 2: Right now the POC is using the concrete `net.UDPConn` struct (with ugly type asserts) but I was thinking of a different implementation of the `connection` interface in the the `go-quic` code and possibly an extension to the `connection` interface. Also I saw that the code does not always use the `Read()` and `Write()` methods of the connection but sometimes uses the wrapped `net.PacketConn` directly to `Read()` and `Write()` UDP datagrams on the wire.

So what are your thoughts on this?

rs:
When the server is listening on multiple interfaces or interfaces with multiple IPs, the outgoing datagrams are sometime delivered with the wrong source IP address.

In order to fix that, each quic connection needs to extract the destination IP (and optionally interface id) of the received datagrams, and set it as source IP (and interface) on the sent datagrams.

On most platforms, this can be done using ancillary data with recvmsg() and sendmsg(). Some of the machinery for this is already there for ECN, this change extends it to read the destination IP info and write it to the outgoing packets.

Fix #1736

Note that this change is highly untested at this point.


marten-seemann:
Thank you for this PR! Before I do an in-depth review, 3 questions:
1. Do you see any problems when using batch reads and writes (see #2607)? Can we just use the serialized `pktinfo` in the `ipv{4,6}.Message.OOB`?
2. Is there anything we can do about the FreeBSD special casing? We can't test FreeBSD on GitHub Actions (and most other CIs), and I've made the painful experience that everything that's not tested regularly will end up broken rather sooner than later.
3. What's the reason you're not using the functions in https://github.com/golang/net/blob/e18ecbb05110/ipv4/control_pktinfo.go for encoding / decoding the `pktinfo`? Is it because they're unexported?

rs:
1. I see no issue as OOB seems to be supported with this API
2. I guess we could use x/net/ipv{4,6}
3. I did not want to add a dep and I had hard time following the code of the lib regarding control messages.

marten-seemann:
1. Great!
2. + 3. Does that allow us to get rid of the special case for FreeBSD? If so, that would be a big win in terms of testing of this change.

rs:
> * / 3. Does that allow us to get rid of the special case for FreeBSD? If so, that would be a big win in terms of testing of this change.

Yes but x/net does not support ECN so we will have to use two different ways to deal with OOB. It means we will be parsing OOB twice.

send_conn.go
```diff
@@ -13,22 +13,56 @@ type sendConn interface {
...
 
-func newSendConn(c net.PacketConn, remote net.Addr) sendConn {
-	return &sconn{PacketConn: c, remoteAddr: remote}
+func newSendConn(c connection, remote net.Addr, info *packetInfo) sendConn {
+	return &sconn{connection: c, remoteAddr: remote, info: info}
 }
 
 func (c *sconn) Write(p []byte) error {
-	_, err := c.PacketConn.WriteTo(p, c.remoteAddr)
+	_, err := c.WritePacket(p, c.remoteAddr, c.info)
```
marten-seemann, line 29: Would it make sense to serialize the `packetInfo` once, instead of doing it for every `WritePacket` call? This would save a lot of garbage, wouldn't it?

send_conn.go
```diff
@@ -13,22 +13,56 @@ type sendConn interface {
...
+		if udpAddr, ok := addr.(*net.UDPAddr); ok {
+			addrCopy := *udpAddr
+			addrCopy.IP = c.info.addr
+			addr = &addrCopy
+		}
+	}
+	return addr
+}
+
+type spconn struct {
```
marten-seemann, line 49: I'm not sure I understand the logic here. As far as I can see, the `spconn` is used by the client, whereas the server would use the `sconn`. That means we assume that the client's kernel will choose the right interface to send the packet from. Is my reasoning correct?

Can we achieve this by using the same struct though? `WritePacket` also works if the `packetInfo` is `nil`, doesn't it?

marten-seemann (commented):
This looks pretty good to me. I left a few comments on the PR, please have a look.

Test-wise, it would be nice if we could have a unit test, along the lines of the ECN test: https://github.com/lucas-clemente/quic-go/blob/9dcb56b76f21241216e90009f3ed26d82c6c163f/conn_ecn_test.go#L56-L74. Making the unit test platform independent is a good way to test that all the platform-dependent code  behaves as expected.

conn_oob.go
```diff
@@ -0,0 +1,213 @@
...
+	// }
+	hdrLen := 16
+	if ip4 := info.addr.To4(); ip4 != nil {
+		// struct in_pktinfo {
+		// 	unsigned int   ipi_ifindex;  /* Interface index */
+		// 	struct in_addr ipi_spec_dst; /* Local address */
+		// 	struct in_addr ipi_addr;     /* Header Destination
+		// 									address */
+		// };
+		oob := make([]byte, hdrLen+12)
```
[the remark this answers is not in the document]
rs: Good catch

send_conn.go
rs: Yes, that's a good point.

send_conn.go
rs: Your reasoning is correct, although, using the same struct for client would require the client code to wrap the `net.PacketConn` to make it a `connection` in all the client code too.

send_conn.go
rs: We also probably don't need all this machinery in case the server does not listen on ANY.

On freebsd, there is also the limitation that it does not seem to work if ANY v4+v6 is bound, you need to create v4 and v6 listeners separately.

conn_oob.go
```diff
@@ -0,0 +1,247 @@
...
+		// 	struct in_addr ipi_spec_dst; /* Local address */
+		// 	struct in_addr ipi_addr;     /* Header Destination address */
+		// };
+		msgLen := 12
+		if runtime.GOOS == "freebsd" {
+			msgLen = 4
+		}
+		cmsglen := cmsgLen(msgLen)
+		oob := make([]byte, cmsglen)
+		cmsg := (*syscall.Cmsghdr)(unsafe.Pointer(&oob[0]))
```
marten-seemann, line 208: Why do you need unsafe here? As far as I can see, you’re setting all fields of the struct manually afterwards.

rs: Basically, the struct's storage is `oob`. Setting fields in the struct will fill the right bytes in the `oob` slice without having to do handle per platform alignment.

marten-seemann: Makes sense. I missed this because Godoc doesn't show that the `Cmsghdr` differs by architecture.

marten-seemann:
Looks like the cross-compilation test is failing on some architectures. Can you please check?
As far as I can see, we're also still missing a unit test, right?

rs:
Yes that’s next on my list. I need to learn this testing framework first.

marten-seemann:
@rs I'm happy to help if you run into any problem, please let me know.
This is a really valuable PR, and I'd like to get it merged as soon as possible.

rs:
@marten-seemann I'm not sure why the keepalive test fails since I rebased.

marten-seemann:
Unit tests LGTM. Don't worry about the timeout test, it's been flaky recently. Will investigate it soon.

Can you please rebase & squash your PR?

rs:
Done

marten-seemann (approved):
@rs LGTM now. Thank you for your contribution!

PR #3074, opened 2021-03-08
fix race condition when receiving 0-RTT packets

### LINKED ISSUE #2629: Server receives 0-RTT packets, but does not deliver them to QUIC session.
If 0-RTT packet arrives after an Inital packet, it is buffered in zeroRTTQueue and never recovered. 0-RTT packets cannot be sent before Initial in a new connection.

Initial and 0-RTT packets are handled with [baseServer.handlePacketImpl(*receivedPacket)](https://github.com/lucas-clemente/quic-go/blob/master/server.go#L315) method. If the packet is Inital, [another method](https://github.com/lucas-clemente/quic-go/blob/master/server.go#L359) is called which starts the session, [checks](https://github.com/lucas-clemente/quic-go/blob/master/server.go#L429) if there are any 0-RTT packets in the buffer, and passes them to the new session. When 0-RTT packet arrives (after the Initial one), it will be handled by the new session if it is ready. If the 0-RTT is sent right after the Initial, the session is not ready yet, and the packet is handled by baseServer.handlePacketImpl, which [buffers it](https://github.com/lucas-clemente/quic-go/blob/master/server.go#L338). Once the session is created, new packets are handled with session.handlePacketImpl, and baseServer.handlePacketImpl, which triggers 0-RTT packets transfer to the session, is never called again. 

My team has discovered this issue with an experiment based on the code in quic-go/integrationtest/self/zero_rtt_test.go . The test does not fail, because it check if the client has sent the 0-RTT packet, not if the other endpoint has received it ([line 104](https://github.com/lucas-clemente/quic-go/blob/master/integrationtests/self/zero_rtt_test.go#L104)).

A possible solution is to pass *zeroRTTQueue to the session, and recover 0-RTT packets there.

marten-seemann:
Fixes #2629. Depends #3066.

By moving the logic for queueing 0-RTT packets into the packet handler map, we can avoid the race condition between receiving 0-RTT packets and the creation of new sessions.


packet_handler_map.go
```diff
@@ -331,6 +384,38 @@ func (h *packetHandlerMap) handlePacket(p *receivedPacket) {
...
+		}
+		h.numZeroRTTEntries++
+		queue := &zeroRTTQueue{queue: make([]*receivedPacket, 0, 8)}
+		h.handlers[string(connID)] = packetHandlerMapEntry{
+			packetHandler: queue,
+			is0RTTQueue:   true,
+		}
+		queue.retireTimer = time.AfterFunc(h.zeroRTTQueueDuration, func() {
+			var deleted bool
+			h.mutex.Lock()
```
lucas-clemente, line 399: defer unlock?

lucas-clemente (approved):
lgtm modulo comments in dependee PR

PR #3081, opened 2021-03-09
add an internal queue to signal that a datagram frame has been dequeued

### LINKED ISSUE #3075: Deadlock in datagramQueue
I am trying to send a lot of datagrams and I think there may be a deadlock in the datagram queue. After reading the code, my guess is, that after calling `SendMessage`, `AddAndWait` triggers `scheduleSending` and then blocks until the frame has been dequeued. But if the `runLoop` tries to empty the queue before the frame is actually enqueued, `AddAndWait` will block until the connection is closed.

I am not sure how to write a proper test case for this, so I used a variation of the echo example to reproduce it:
https://gist.github.com/mengelbart/cab8d851e6ad101b37a7db9f632292c6

After a few seconds, it blocks until the connection times out. If I try to delay the `hasData`/`scheduleSending` until the frame is enqueued it seems to work fine.

mengelbart:
Fixes #3075

### DIFF
--- a/datagram_queue.go
+++ b/datagram_queue.go
@@ -15,14 +15,17 @@ type datagramQueue struct {
 
 	hasData func()
 
+	dequeued chan struct{}
+
 	logger utils.Logger
 }
 
 func newDatagramQueue(hasData func(), logger utils.Logger) *datagramQueue {
 	return &datagramQueue{
 		hasData:   hasData,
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):

marten-seemann:
Thank you @mengelbart!

Issue #3061, opened 2021-03-03
relax the requirement on transport parameters when accepting a 0-RTT connection


marten-seemann:
We currently check that the parameters stayed exactly the same:
https://github.com/lucas-clemente/quic-go/blob/2c45f2b11db0ceb1e6d8bc9cd668f27f5dde8b00/internal/wire/transport_parameters.go#L443-L450

This is stricter than necessary. If we increased the parameters, there's no harm in resuming the connection.
I don't think this will get used a lot in practice, but we should make sure that at least there isn't any bug hiding behind this implicit assumption of equality here. This will allow us to create a bunch of nice integration tests.

PR #3083, opened 2021-03-10
make the initial stream / connection flow control windows configurable

### LINKED ISSUE #88: think about server parameter values
e.g. `ReceiveConnectionFlowControlWindow`, `ReceiveStreamFlowControlWindow`, `MaxIdleConnectionStateLifetime`

marten-seemann:
This is primarily to make 0-RTT better testable when we relax the resumption condition, see #3061.

As we're shifting more responsibility for setting these values to the application, we could say this fixes #88.

Should we rename the config options? The current ones are a bit of a mouthful. Suggestion:
`Initial{Stream, Connection}FlowControlWindow` => `Initial{Stream, Connection}ReceiveWindow`
`MaxReceive{Stream, Connection}FlowControlWindow` => `Max{Stream, Connection}ReceiveWindow`

### DIFF
--- a/config.go
+++ b/config.go
@@ -71,6 +71,10 @@ func populateConfig(config *Config) *Config {
 	if config.MaxIdleTimeout != 0 {
 		idleTimeout = config.MaxIdleTimeout
 	}
+	initialStreamFlowControlWindow := config.InitialStreamFlowControlWindow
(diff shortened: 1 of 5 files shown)


lucas-clemente (approved):
Renaming sgtm.

PR #3108, opened 2021-03-21
avoid type confusion between protocol.PacketType and logging.PacketType

marten-seemann:
The enums have completely different meanings. `protocol.PacketType` only defines long header types, whereas `logging.PacketType` defines all different types of QUIC packets (including short header packets).

This led to a bug in metrics implementation in libp2p: There, we were doing something along those lines:
```go
func (m *tracer) getEncLevel(hdr *logging.ExtendedHeader) string {
	if hdr.Type == logging.PacketType0RTT {
		return "0-RTT"
	}
        // ....
}
```

This leads to wrong results, as `hdr.Type` is a `protocol.PacketType` and not a `logging.PacketType`. It just compiled because we defined `logging.PacketType` as an alias for `protocol.PacketType`. With this PR, the code snippet above will fail to compile.

### DIFF
--- a/internal/mocks/logging/connection_tracer.go
+++ b/internal/mocks/logging/connection_tracer.go
@@ -40,7 +40,7 @@ func (m *MockConnectionTracer) EXPECT() *MockConnectionTracerMockRecorder {
 }
 
 // BufferedPacket mocks base method.
-func (m *MockConnectionTracer) BufferedPacket(arg0 protocol.PacketType) {
+func (m *MockConnectionTracer) BufferedPacket(arg0 logging.PacketType) {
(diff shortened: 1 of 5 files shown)


lucas-clemente (approved):
IIRC we initially did this because the protocol pkg is internal, so you wouldn't be able to use APIs with `PacketType`s. Is that not an issue anymore?

marten-seemann:
That's not an issue when we set an alias, as we do for many other types.
The issue here is that the values defined for the `logging.PacketType` and the `protocol.PacketType` are different. For example, the `protocol.PacketType` doesn't define a type for short header packets.

PR #3102, opened 2021-03-20
use a tracer to make the packetization test more useful

marten-seemann:
Fixes #3065.

This will hopefully fix the flakiness of this test.

PR #3121, opened 2021-03-23
only start PMTUD after handshake confirmation

marten-seemann:
Our PMTUD packets are never coalesced packets. Since we try packing coalesced packets until the handshake is confirmed, it makes sense to start PMTUD then.

This mismatch in expectation can also lead to busy-looping: we might set a timer to send a PMTUD packet, but then we never send one, because the handshake isn't confirmed yet.


lucas-clemente (approved):

marten-seemann:
This was unmerged due to the mess in #3102. Will replace with a new one.

Issue #3122, opened 2021-03-23
another PMTUD panic


marten-seemann:
This time, it seems to be unrelated to the initial maximum packet size.

```
panic: congestion BUG: decreased max datagram size from 1352 to 1302

goroutine 134676 [running]:
github.com/lucas-clemente/quic-go/internal/congestion.(*cubicSender).SetMaxDatagramSize(0xc0053c4840, 0x516)
        src/lucas-clemente/quic-go/internal/congestion/cubic_sender.go:308 +0x126
github.com/lucas-clemente/quic-go/internal/ackhandler.(*sentPacketHandler).SetMaxDatagramSize(0xc002e3b790, 0x516)
        src/lucas-clemente/quic-go/internal/ackhandler/sent_packet_handler.go:714 +0x3e
github.com/lucas-clemente/quic-go.(*session).handleHandshakeConfirmed.func1(0x516)
        src/lucas-clemente/quic-go/session.go:829 +0x4b
github.com/lucas-clemente/quic-go.(*mtuFinder).GetPing.func2(0x25e2960, 0x3324bf8)
        src/lucas-clemente/quic-go/mtu_discoverer.go:81 +0x3a
github.com/lucas-clemente/quic-go/internal/ackhandler.(*sentPacketHandler).detectAndRemoveAckedPackets(0xc002e3b790, 0xc008361700, 0x2eb66b5904, 0x32f15a0, 0x8762ae66e, 0x372c963f73, 0xc00e9ba15308e3e2, 0xc007543b90)
        src/lucas-clemente/quic-go/internal/ackhandler/sent_packet_handler.go:398 +0x2ba
github.com/lucas-clemente/quic-go/internal/ackhandler.(*sentPacketHandler).ReceivedAck(0xc002e3b790, 0xc008361700, 0xc0071a8404, 0xc00e9ba153084f22, 0x372c95aad1, 0x32f15a0, 0x1, 0x1)
        src/lucas-clemente/quic-go/internal/ackhandler/sent_packet_handler.go:293 +0xc5
github.com/lucas-clemente/quic-go.(*session).handleAckFrame(0xc003822c00, 0xc008361700, 0x25e2604, 0xc008361700, 0x0)
        src/lucas-clemente/quic-go/session.go:1364 +0x77
[...]

PR #3124, opened 2021-03-25
return an error when the max datagram size is reduced

marten-seemann:
There's no need to panic here. This seems to be more brittle than expected, see #3122.
An error will show up in the logs as well.


marten-seemann:
It's a lot of work to drag this `error` return value on `OnAcked` through the entire code base. Worth it?

PR #3132, opened 2021-03-29
various amplification limit fixes

### LINKED ISSUE #3053: investigate recent loss recovery change
https://github.com/quicwg/base-drafts/pull/4831/

marten-seemann:
Fixes #3053.

1. When becoming amplification limited, we don't need to set a loss recovery timer. All we can do at that point is wait for a packet from the client to arrive.
2. Once that packet arrives, we need to restart the loss recovery timer.

### DIFF
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -194,7 +194,11 @@ func (h *sentPacketHandler) dropPackets(encLevel protocol.EncryptionLevel) {
 }
 
 func (h *sentPacketHandler) ReceivedBytes(n protocol.ByteCount) {
+	wasAmplificationLimit := h.isAmplificationLimited()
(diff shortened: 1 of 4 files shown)


internal/ackhandler/sent_packet_handler.go
```diff
@@ -490,20 +496,34 @@ func (h *sentPacketHandler) setLossDetectionTimer() {
...
+			}
 		}
 		return
 	}
 
 	// PTO alarm
 	ptoTime, encLevel := h.getPTOTimeAndSpace()
 	h.alarm = ptoTime
-	if h.tracer != nil && h.alarm != oldAlarm {
+	if h.tracer != nil { //&& h.alarm != oldAlarm {
```
lucas-clemente, line 526: Nit

lucas-clemente (approved):

PR #3142, opened 2021-04-05
use batched reads

marten-seemann:


conn_oob.go
```diff
@@ -94,23 +113,41 @@ func newConn(c OOBCapablePacketConn) (*oobConn, error) {
 			return nil, errors.New("activating packet info failed for both IPv4 and IPv6")
 		}
 	}
-	return &oobConn{
+	oobConn := &oobConn{
 		OOBCapablePacketConn: c,
-		oobBuffer:            make([]byte, 128),
-	}, nil
+		batchConn:            ipv4.NewPacketConn(c),
+		messages:             make([]ipv4.Message, batchSize),
```
lucas-clemente, line 119: Maybe it's a bit less error-prone if you keep a `[batchSize]ipv4.Message` array and a `currentMessages` slice over that as members? That could allow you to get rid of the `readPos` tracking…

lucas-clemente (approved):

PR #3146, opened 2021-04-14
make it possible to associate a ConnectionTracer with a Session

### LINKED ISSUE #3145: find an idiomatic way to associate a tracer with a connection
`quic.Listen` and `quic.Dial` both take a `quic.Config`. The `Config` has a [`logging.Tracer`](https://pkg.go.dev/github.com/lucas-clemente/quic-go@v0.20.1/logging#Tracer), which, if set, allows the application to return a `logging.ConnectionTracer`. Over the lifetime of the connection, quic-go emits a series of events, which can be traced by the connection tracer.

This is very helpful for generating metrics, and go-libp2p-quic-transport [uses](https://github.com/libp2p/go-libp2p-quic-transport/blob/master/tracer_metrics.go) it for this purpose.
The problem is, that it would be nice to associate application-layer information with that tracer. For example, it would be really helpful to also log the peer ID and the `AgentVersion`. This is not possible at the moment, since it's not possible to associate a tracer created by the `logging.Tracer.TracerForConnection` with the connection returned from `quic.Listen` or `quic.Dial`.

I'm wondering if there's an idiomatic way to solve this. @mvdan, do you have any idea?

marten-seemann:
Fixes #3145.

This PR implements a combination of the options discussed on the issue:
* `logging.Tracer.NewConnectionTracer` now takes a `context.Context`, which has values associated with it.
* Both this context as well as the context returned from `Session.Context` have a context with a `SessionTracingKey` value (which is a uint64).
* For outgoing connections, the context passed to `NewConnectionTracer` is derived from the context passed to `Dial`. This allows us (for example) to associate the peer ID we were dialing even if the handshake fails (and no `Session` is returned).

@mvdan, how does this look to you?

### DIFF
--- a/client.go
+++ b/client.go
@@ -203,13 +203,17 @@ func dialContext(
 	}
 	c.packetHandlers = packetHandlers
 
+	c.tracingID = nextSessionTracingID()
 	if c.config.Tracer != nil {
-		c.tracer = c.config.Tracer.TracerForConnection(protocol.PerspectiveClient, c.destConnID)
+		c.tracer = c.config.Tracer.TracerForConnection(
+			context.WithValue(ctx, SessionTracingKey, c.tracingID),
+			protocol.PerspectiveClient,
+			c.destConnID,
+		)
 	}
 	if c.tracer != nil {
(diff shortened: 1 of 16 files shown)


mvdan:
SGTM. You could always refine this if users find it's not enough.

PR #3147, opened 2021-04-14
remove the .editorconfig

marten-seemann:


mvdan (approved):
Yes please :) Should make reviews easier.

PR #3148, opened 2021-04-14
declare the handshake confirmed when receiving an ACK for a 1-RTT packet

### LINKED ISSUE #3144: SentPacketHandler.OnLossDetectionTimeout sometimes called repeatedly during handshake
I noticed irregular 100% CPU usage in a handshake test. Running the below command reproduces the issue for me roughly in 1/20 runs.

`ginkgo --focus "Handshake drop tests with QUIC version TLS dev version \(WIP\) when not using Retry using a long certificate chain: false client speaks first establishes a connection when 1/3 of the packets are lost in both direction"  --untilItFails`

I pinpointed the issue to `SentPacketHandler.OnLossDetectionTimeout` being called repeatedly. Logging the internal state during that call I noticed that when this happens `h.handshakeConfirmed` is always false and `h.appDataPackets.lossTime` is always set and always has the same value. From this I figured why the repeated calls are being made. The [check](https://github.com/lucas-clemente/quic-go/blob/3138a45fded18b7c15c12424ea662127b4b80c9f/internal/ackhandler/sent_packet_handler.go#L602) in `SentPacketHandler.OnLossDetectionTimeout` due to `h.handshakeConfirmed=false` stops `onVerifiedLossDetectionTimeout` from being called even though the `h.appDataPackets.lossTime` was reached. Then the loss detection timer is set to the exact same `h.appDataPackets.lossTime` value and the session wakes up for it, repeatedly.

marten-seemann:
... on the client side. Both the receipt of HANDSHAKE_DONE and the receipt of an ACK for a 1-RTT packet are sufficient for declaring confirmation of the handshake.

Fixes #3144. @rip-create-your-account

### DIFF
--- a/internal/ackhandler/interfaces.go
+++ b/internal/ackhandler/interfaces.go
@@ -27,7 +27,7 @@ type Packet struct {
 type SentPacketHandler interface {
 	// SentPacket may modify the packet
 	SentPacket(packet *Packet)
-	ReceivedAck(ackFrame *wire.AckFrame, encLevel protocol.EncryptionLevel, recvTime time.Time) error
+	ReceivedAck(ackFrame *wire.AckFrame, encLevel protocol.EncryptionLevel, recvTime time.Time) (bool /* 1-RTT packet acked */, error)
 	ReceivedBytes(protocol.ByteCount)
 	DropPackets(protocol.EncryptionLevel)
 	ResetForRetry() error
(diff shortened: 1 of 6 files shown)


rip-create-your-account:
SentPacketHandler stores only ack-eliciting packets so an opportunity to confirm the handshake could be missed. Or is there a reason to not confirm handshake with packets that are not ack-eliciting? The suggested implementation allows it.

Also with the current implementation, I think the comments on the SentPacketHandler interface like `/* contained 1-RTT packet */` should be `/* contained 1-RTT ack-eliciting packet */` just to be exact.

lucas-clemente (approved):

marten-seemann:
> Or is there a reason to not confirm handshake with packets that are not ack-eliciting?

@rip-create-your-account we don't save them in the packet history.

PR #3155, opened 2021-04-20
set the don't fragment (DF) bit on Windows

tobyxdd:


marten-seemann:
Is there any way we can test this? I realize testing that DF is actually set might be difficult, but at least we should be able to pass a net.UDPConn to newConn and make sure it doesn’t return an error.

tobyxdd:
Not sure what happened but it doesn't look related to my changes

marten-seemann (approved):
I haven't checked Wireshark to see if the DF bit is actually set. @tobyxdd, have you?

tobyxdd:
Yes I did

tobyxdd:
image
image

without vs with on Windows

marten-seemann (approved):
Thank you @tobyxdd!

marten-seemann:
@tobyxdd Have you checked if this also works on IPv6? I was just working on an equivalent PR for OSX, and it looks like there are different syscalls for IPv4 and IPv6.

marten-seemann:
Or am I misunderstanding something here? IPv6 shouldn't allow any packet fragmentation anyway, but then why is there a `unix.IPV6_DONTFRAG` (see https://pkg.go.dev/golang.org/x/sys/unix).

tobyxdd:
Just realized that IPv6 does have fragmentation (as an extension header), but only at the source, not routers.

https://en.wikipedia.org/wiki/IPv6_packet#Fragment
https://tools.ietf.org/html/rfc3542
https://blog.apnic.net/2016/05/19/fragmenting-ipv6/

tobyxdd:
So yes I guess we do need to set DF for IPv6 as well (or get local MTU from the system?)

marten-seemann:
> So yes I guess we do need to set DF for IPv6 as well (or get local MTU from the system?)

I'm not sure. It depends on what happens when we try to send a too large packet. We'd need some way to feed that information back into the MTU discovery code. Getting the local MTU from the system and initializing the MTU discoverer with that seems like the safer option, but I'm not sure it's needed in practice, as our maximum packet size is below 1500 bytes anyway.

PR #3157, opened 2021-04-24
Remove a h3 client from the pool when dead

### LINKED ISSUE #765: handle closed clients in http3.RoundTripper

rs:
When the session of a client becomes invalid, all subsequent requests
with that client will fail. The RoundTripper's connection pool is not
aware of that and keep using it, which creates a dead end for the
impacted hostname.

This change make sure clients are removed from the pool when in
permanent error. A proper change should probably also retry the requests
on connection errors, but it would require a larger rearchitecture of
how the RoundTripper works. With this, the RoundTripper gives a chance
to the caller to handle errors and retry, which makes at least the
RoundTripper usable.

Fixes #765

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"strconv"
 	"sync"
+	"sync/atomic"
 
 	"github.com/lucas-clemente/quic-go"
 	"github.com/lucas-clemente/quic-go/internal/protocol"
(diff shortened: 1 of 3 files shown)


rs:
The unit tests fails and I don't understand why. The unit test framework used makes it very hard to debug.

marten-seemann:
Why can't we make this analogous to the H2 client?

rs:
That would be great but would be a large change.

brotaxt:
Is there any chance this will get merged someday?

PR #3158, opened 2021-04-25
Expose the methods to create a H3 client and server handler for a low-level library stream  

costinm:
This is similar with the methods exposed by the x/net/http2 library, adapted for quic. Right now it is not
possible to mix use of quic and quic/http3 - this PR provides one way.

I'm using this for some experiments with datagrams and MASQUE - opening the pull request go get 
feedback, if there is interest I can improve it. 

It is also possible to add extra methods to allow the H3 library to dispatch non-http3 low-level streams and 
datagrams to a handler.


marten-seemann:
Interesting. I started playing around with MASQUE a while back: https://github.com/marten-seemann/masque-go. My approach was to expose the datagram receive / send functions on the HTTP/3 connection, not pass a QUIC connection to HTTP/3.

> It is also possible to add extra methods to allow the H3 library to dispatch non-http3 low-level streams and
datagrams to a handler.

Can you elaborate on this?

costinm:
Masque is one way to use QUIC - HTTP/3 centric. It would be great for the http3 library to expose methods to receive/send datagrams, possibly based on a 'flow ID' or app dispatcher. Also would be great if the library exposed methods to for server to 
open streams - which is not allowed in HTTP/3, but very useful when layering other protocols over. 

But at the same time it will be useful to also expose methods similar with the H2 library, letting apps use the low-level QUIC 
( or maybe another implementation of the interfaces defined there). I'm still evaluating the performance of tunneling over 'raw QUIC' versus tunneling over 'H3' - there is some extra overhead, buffer allocations, extra framing for data - and the RoundTrip interface is quite inconvenient and heavily oriented towards HTTP semantics. In particular around dealing with the request body, or sending a FIN (CloseWrite).

costinm:
For the record: I ended up using the low level interface. 

Forwarding streams using the go HTTP interface is not ideal - I'll try to see if the low level qpack can be used, but it's just too much impedance mismatch and overhead.

I think this PR - or a better variant - may still be useful in quic-go, at least for parity with h2 library - but not for my original reasons.

PR #3159, opened 2021-04-26
rework error return values

### LINKED ISSUE #2441: Extract the ErrorCode
I am unable to extract the error code, because the QuicError is hidden within the internal package. Are there any plans to allow access to user defined application errors?

marten-seemann:
Fixes #2441.

This is a massive refactor changing the way we return QUIC errors. Calls to the session (and to streams) can now different errors, which can be checked with `errors.Is` / `errors.As`:
* `quic.TransportError`: for errors triggered by the QUIC transport (in many cases a misbehaving peer)
* `quic.ApplicationError`: for errors triggered by the application running on top of QUIC
* `quic.IdleTimeoutError`: when the peer goes away unexpectedly (this is a `net.Error` timeout error)
* `quic.HandshakeTimeoutError`: when the cryptographic handshake takes too long  (this is a `net.Error` timeout error)
* `quic.StatelessResetError`: when we receive a stateless reset  (this is a `net.Error` temporary error)
* `quic.VersionNegotiationError`: returned by the client, when there's no version overlap between the peers

In addition to that, streams that are canceled return a:
* `quic.StreamError`: both for local and remote cancellations

@mvdan, what do you think about this? Do the API changes here make sense to you?

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -1,0 +1,30 @@
+package quic
+
+import "github.com/lucas-clemente/quic-go/internal/qerr"
+
+type (
+	TransportError       = qerr.TransportError
+	ApplicationError     = qerr.ApplicationError
+	TransportErrorCode   = qerr.TransportErrorCode
+	ApplicationErrorCode = qerr.ApplicationErrorCode
+)
+
+const (
+	NoError                   = qerr.NoError
+	InternalError             = qerr.InternalError
+	ConnectionRefused         = qerr.ConnectionRefused
+	FlowControlError          = qerr.FlowControlError
+	StreamLimitError          = qerr.StreamLimitError
+	StreamStateError          = qerr.StreamStateError
+	FinalSizeError            = qerr.FinalSizeError
+	FrameEncodingError        = qerr.FrameEncodingError
+	TransportParameterError   = qerr.TransportParameterError
+	ConnectionIDLimitError    = qerr.ConnectionIDLimitError
+	ProtocolViolation         = qerr.ProtocolViolation
+	InvalidToken              = qerr.InvalidToken
+	ApplicationErrorErrorCode = qerr.ApplicationErrorErrorCode
+	CryptoBufferExceeded      = qerr.CryptoBufferExceeded
+	KeyUpdateError            = qerr.KeyUpdateError
+	AEADLimitReached          = qerr.AEADLimitReached
+	NoViablePathError         = qerr.NoViablePathError
+)
--- a/errors.go
+++ b/errors.go
@@ -1,10 +1,16 @@
 package quic
 
-import "github.com/lucas-clemente/quic-go/internal/qerr"
+import (
+	"github.com/lucas-clemente/quic-go/internal/qerr"
+)
 
 type (
-	TransportError       = qerr.TransportError
-	ApplicationError     = qerr.ApplicationError
+	TransportError          = qerr.TransportError
+	ApplicationError        = qerr.ApplicationError
+	VersionNegotiationError = qerr.VersionNegotiationError
+)
+
+type (
 	TransportErrorCode   = qerr.TransportErrorCode
 	ApplicationErrorCode = qerr.ApplicationErrorCode
 )
--- a/errors.go
+++ b/errors.go
@@ -11,6 +11,8 @@ type (
 	ApplicationError        = qerr.ApplicationError
 	VersionNegotiationError = qerr.VersionNegotiationError
 	StatelessResetError     = qerr.StatelessResetError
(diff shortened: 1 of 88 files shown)


errors.go
```diff
@@ -0,0 +1,58 @@
+package quic
+
+import (
+	"fmt"
+
+	"github.com/lucas-clemente/quic-go/internal/qerr"
+)
+
+type (
+	TransportError          = qerr.TransportError
```
marten-seemann, line 10: Now this is a bit weird. We have a
1. `TransportError`: the type of a transport error
2. `TransportErrorCode`: the type of the error code for transport errors (a `uint64`)
3. `TransportErrorErrorCode`: the error code the the TRANSPORT_ERROR (see `0xa` in https://www.ietf.org/archive/id/draft-ietf-quic-transport-34.html#name-transport-error-codes)

internal/qerr/quic_error.go
```diff
@@ -89,34 +49,86 @@ func (e *QuicError) Error() string {
...
 }
+
+func (e *VersionNegotiationError) Is(target error) bool {
+	_, ok := target.(*VersionNegotiationError)
+	return ok
+}
+
+// A StatelessResetError occurs when we receive a stateless reset.
+type StatelessResetError struct {
+	Token protocol.StatelessResetToken
```
marten-seemann, line 119: It's nice to have all errors in the `qerr` package, but this means that we have to export the `Token` here. There's nothing wrong with that, but it's a piece of information that has no value to the user, so it would be nicer if it wasn't exported.

session.go
```diff
@@ -1431,45 +1444,61 @@ func (s *session) shutdown() {
...
 	}
 
-	var quicErr *qerr.QuicError
-	var ok bool
-	if quicErr, ok = closeErr.err.(*qerr.QuicError); !ok {
-		quicErr = qerr.ToQuicError(closeErr.err)
+	if !errors.Is(e, qerr.ErrIdleTimeout) && !errors.Is(e, qerr.ErrHandshakeTimeout) &&
+		!errors.Is(e, &StatelessResetError{}) && !errors.Is(e, &VersionNegotiationError{}) &&
+		!errors.Is(e, &errCloseForRecreating{}) &&
+		!errors.Is(e, &qerr.ApplicationError{}) && !errors.Is(e, &qerr.TransportError{}) {
```
marten-seemann, line 1469: 🤢

errors.go
mvdan: >     2\. `TransportErrorCode`

I guess just `ErrorCode` is not an option?



> `TransportErrorErrorCode`: the error code the the TRANSPORT_ERROR (see `0xa` in https://www.ietf.org/archive/id/draft-ietf-quic-transport-34.html#name-transport-error-codes)

not sure I follow; that page shows PROTOCOL_VIOLATION for 0xa. Could you not use ProtocolViolation?

internal/qerr/quic_error.go
mvdan: since this is an internal package, does that matter?

session.go
mvdan: Hmm, can we not use a type switch here? I get that `As` is meant to peek at the error's layers, but I think `Is` isn't meant to do that - so just doing a type switch on the top-level error might be fine.

errors.go
marten-seemann: Sorry, I meant 0xc.

internal/qerr/quic_error.go
marten-seemann: Yes, because we're exporting it in the `quic` package: `type StatelessResetError = qerr.StatelessResetError`

mvdan:
> @mvdan, what do you think about this? Do the API changes here make sense to you?

From the PR description you wrote, it sounds fine.

session.go
marten-seemann: A type switch with 7 empty cases and one default case that has some logic? Maybe that's at least a little bit nicer?

mvdan: Indeed, I'd find that nicer. You can even group the cases, like:

```
switch e.(type) {
case T1, T2, T3...:
default:
    // your logic
}
```

lucas-clemente (approved):
Can you include your PR description as doc for the errors in the main package?

session.go
marten-seemann: > I get that As is meant to peek at the error's layers, but I think Is isn't meant to do that 

I think it is. From the docs: 
> Is reports whether any error in err's chain matches target.

Now we shouldn't encounter any wrapped errors here (at least not with the current code), but I guess it's still safer to use `errors.Is`.

PR #3163, opened 2021-05-01
use net.ErrClosed (for Go 1.16)

### LINKED ISSUE #2933: Go 1.16: use net.ErrClosed

marten-seemann:
Fixes #2933.

@mvdan, can I ask for your review here?


marten-seemann:
Merging, as we need to cut a release.

PR #3171, opened 2021-05-08
Fix compilation with gccgo

### LINKED ISSUE #3168: Fails to compile with gccgo
When I try to compile github.com/syncthing/syncthing, which uses quic-go as a dependency, with gccgo I came accross the following error:

```
% CGO_ENABLED=1 go run build.go -no-upgrade
Notice: Next generation GUI will not be built; see --with-next-gen-gui.
......
# github.com/lucas-clemente/quic-go/internal/qtls
/tmp/ccw2vu1Y.s: Assembler messages:
/tmp/ccw2vu1Y.s:5968: Error: junk `.cipherSuiteTLS13ByID ' after expression
/tmp/ccw2vu1Y.s:5968: Error: invalid operands (*UND* and *UND* sections) for `/'
/tmp/ccw2vu1Y.s:5968: Error: invalid operands (*UND* and *UND* sections) for `/'
/tmp/ccw2vu1Y.s:5968: Error: can't resolve `L0' {*ABS* section} - `go1' {*UND* section}
exit status 2
exit status 1
```

As mentioned in #2883, the problem may be with `go:linkname`. It seems to me that gccgo is looking for mangled function names here.

After manually mangling the function name (` '.'->'_0' '/'->'_1' '-'->'_x2d'`)
```go
//go:linkname cipherSuiteTLS13ByID "github_0com_1marten_x2dseemann_1qtls_x2dgo1_x2d15.cipherSuiteTLS13ByID"
```
I managed to compile the package and the resulted binary seems working fine.

Maybe a new conditional file with `// +build gccgo` can solve this better?

My `go version`:
```
go version go1.16.3 gccgo (Gentoo 11.1.0 p1) 11.1.0 linux/amd64
```

My `go env`:
```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/<username redacted>/.cache/go-build"
GOENV="/home/<username redacted>/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/<username redacted>/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/<username redacted>/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/libexec/gcc/x86_64-pc-linux-gnu/11.1.0"
GOVCS=""
GOVERSION="go1.16.3 gccgo (Gentoo 11.1.0 p1) 11.1.0"
GCCGO="/usr/x86_64-pc-linux-gnu/gcc-bin/11.1.0/gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build577597224=/tmp/go-build -gno-record-gcc-switches -funwind-tables"
```

OceanS2000:
gccgo handles `//go:linkname` directive differently from gc compiler. This PR adds conditional compiled code to make the package can be built with gccgo.

As for CI, a preliminary workflow for build test is added. However, currently `ubuntu-latest` only ships `gccgo` package that implements go 1.14, which is too old. Ubuntu 21.04 has the updated `gccgo` that implements go 1.16, but we have to wait for GitHub.

Fixes #3168.

### DIFF
--- a/.github/workflows/gccgo.yml
+++ b/.github/workflows/gccgo.yml
@@ -1,0 +1,14 @@
+on: [push, pull_request]
+jobs:
+  gccgo:
+    runs-on: ubuntu-latest
+    name: "gccgo Build"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install build utils
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib gccgo
+      - run: go-10 version
+      - name: Run build
+        run: go-10 build example/main.go
--- a/internal/qtls/gccgo115.go
+++ b/internal/qtls/gccgo115.go
@@ -1,0 +1,26 @@
+// +build go1.15
+// +build !go1.16
+// +build gccgo
+
+package qtls
+
+import (
+	"unsafe"
+
+	qtls "github.com/marten-seemann/qtls-go1-15"
+)
+
+//go:linkname cipherSuiteTLS13ByID "github_0com_1marten_x2dseemann_1qtls_x2dgo1_x2d15.cipherSuiteTLS13ByID"
+func cipherSuiteTLS13ByID(id uint16) *cipherSuiteTLS13
+
+// CipherSuiteTLS13ByID gets a TLS 1.3 cipher suite.
+func CipherSuiteTLS13ByID(id uint16) *CipherSuiteTLS13 {
+	val := cipherSuiteTLS13ByID(id)
+	cs := (*cipherSuiteTLS13)(unsafe.Pointer(val))
+	return &qtls.CipherSuiteTLS13{
+		ID:     cs.ID,
+		KeyLen: cs.KeyLen,
+		AEAD:   cs.AEAD,
+		Hash:   cs.Hash,
+	}
+}
--- a/internal/qtls/gc115.go
+++ b/internal/qtls/gc115.go
@@ -1,0 +1,26 @@
+// +build go1.15
+// +build !go1.16
+// +build gc
+
+package qtls
+
+import (
+	"unsafe"
+
+	qtls "github.com/marten-seemann/qtls-go1-15"
+)
+
+//go:linkname cipherSuiteTLS13ByID github.com/marten-seemann/qtls-go1-15.cipherSuiteTLS13ByID
+func cipherSuiteTLS13ByID(id uint16) *cipherSuiteTLS13
+
+// CipherSuiteTLS13ByID gets a TLS 1.3 cipher suite.
+func CipherSuiteTLS13ByID(id uint16) *CipherSuiteTLS13 {
+	val := cipherSuiteTLS13ByID(id)
+	cs := (*cipherSuiteTLS13)(unsafe.Pointer(val))
+	return &qtls.CipherSuiteTLS13{
+		ID:     cs.ID,
+		KeyLen: cs.KeyLen,
+		AEAD:   cs.AEAD,
+		Hash:   cs.Hash,
+	}
+}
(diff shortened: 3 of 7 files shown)


.github/workflows/gccgo.yml
```diff
@@ -0,0 +1,14 @@
...
+  gccgo:
+    runs-on: ubuntu-latest
+    name: "gccgo Build"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install build utils
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib gccgo
+      - run: go-10 version
```
marten-seemann, line 12: What is `go-10`?

internal/qtls/gccgo115.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+type cipherSuiteTLS13 struct {
+	ID     uint16
+	KeyLen int
+	AEAD   func(key, fixedNonce []byte) cipher.AEAD
+	Hash   crypto.Hash
+}
+
+//go:linkname cipherSuiteTLS13ByID "github_0com_1marten_x2dseemann_1qtls_x2dgo1_x2d15.cipherSuiteTLS13ByID"
```
marten-seemann, line 103: This is the only change to this file, isn't it? Can we split things up into separate files such that we don't need to duplicate the definition of all the constants and functions in this file?

.github/workflows/gccgo.yml
OceanS2000: In order to avoid conflict between default go compiler and gcc go compiler, gccgo package from Ubuntu has a suffix of corresponding gcc version, and it is `10` in this case meaning it comes with gcc 10. AFAIK, Debian uses the same convention, whereas Arch disallows  gccgo and default go compiler to be installed at the same time.

OceanS2000: However gcc 10 cannot compile go 1.15 properly, so we may need upstream GitHub base image to update.

OceanS2000:
I still need to figure out how to test it on CI. Moreover, it seems that gcc 10 supports go 1.14 and gcc 11 supports go 1.16; therefore there will be no need for a `gccgo115.go` file. Mark as draft for now.

marten-seemann:
Does [setup-go](https://github.com/actions/setup-go/) install gccgo?

OceanS2000:
> Does [setup-go](https://github.com/actions/setup-go/) install gccgo?

I don’t believe so, as https://github.com/actions/go-versions/blob/main/versions-manifest.json does not mention any gccgo version.

.github/workflows/gccgo.yml
marten-seemann: Makes sense. The Go versions available via `apt` are always unpredictable and usually outdated.
Can we download the code from golang.org?

marten-seemann:
Any progress with this PR?

marten-seemann:
Closing due to age.

PR #3172, opened 2021-05-09
Benchmark to reproduce slowness

AudriusButkevicius:
This is in relation to https://github.com/syncthing/syncthing/issues/7636

I've wrote a minimal reproducer that reproduces this issue.
I cannot understand why a wrapper connection as implemented below would have such a dramatic impact on quic's performance.
It seems to have close to no effect on standard UDPs performance.

There is a `direct: false` flag on line 66 (https://github.com/lucas-clemente/quic-go/pull/3172/files#diff-0e0b4af59631d95ab56a75a2017c5e35e77b75354a8c85f22ca2ce83c1042ce1R66) that controls whether packet delivery happens directly, or via a channel.
Seems that packet delivery via a channel cuts the performance by 20x.

### DIFF
--- a/benchmark/benchmark_wrapped_test.go
+++ b/benchmark/benchmark_wrapped_test.go
@@ -1,0 +1,100 @@
+package benchmark
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"math/rand"
+	"net"
+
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
I'm reusing the same buffer between two routines, end up corrupting the data... 🤦

PR #3182, opened 2021-05-17
add support for Go 1.17 (tip)

marten-seemann:
@mvdan Ready for Go 1.17 beta1! All we have to do is rebase the qtls repository onto the beta1 Go commit. Assuming that there won't be any significant changes to crypto/tls, that should be trivial.


mvdan (approved):
As soon as 1.17 is released you'll be able to remove 1.15, too.

marten-seemann:
> As soon as 1.17 is released you'll be able to remove 1.15, too.

That's the plan. And as soon as the beta is out, we can also test this in CI.

PR #3183, opened 2021-05-17
Fill `TLS` field in http request

v-zubko:
Right now `TLS` field not filled and that make impossible to identify clients in handlers.


v-zubko:
@googlebot I signed it!

Issue #3194, opened 2021-06-03
go mod vendor fails with quic-go v0.21.0


ShangguanHong:
Using v0.21.0 of quic-go, the following error occurs when executing the **go mod vendor** command
"github.com\lucas-clemente\quic-go@v0.21.0\internal\qtls\go118.go:3:1: expected 'package', found "quic-go doesn't build on Go 1.18 yet."
It works the same in both windows and linux environments.
Check out the "github.com\lucas-clemente\quic-go@v0.21.0\internal\qtls\go118.go" file with the following code
![image](https://user-images.githubusercontent.com/37330396/120570336-771dd580-c44a-11eb-970f-c1f389cd3d5a.png)
Is this some new way of writing?

marten-seemann:
This prevents quic-go from being built with Go 1.18, as we don't support that version yet.
I'm a bit surprised that `go mod vendor` tries to compile that file, but I guess it's trying to vendor dependencies regardless of the build flags.

@mvdan, do you have any idea?

ameshkov:
Same issue here. Probably, it'd be better to replace it with something like `function init() { panic('xxx') }` to avoid breaking mod vendor.

mvdan:
Ah, commands `go mod vendor` will need to parse import statements from all Go files to tell what packages need to be vendored. This will ignore build tags by definition, because one wants to vendor all dependencies across all build configurations.

@marten-seemann I think you need to provide valid syntax, then, but pulling imports doesn't require typechecking, so you can still use a typechecking error rather than a syntax error:

```
$ cat f.go
package p

var _ int = "quic-go doesn't build on Go 1.18 yet."
$ go build f.go
# command-line-arguments
./f.go:3:5: cannot use "quic-go doesn't build on Go 1.18 yet." (type untyped string) as type int in assignment
$ go mod vendor
go: no dependencies to vendor
$ gofmt f.go
package p

var _ int = "quic-go doesn't build on Go 1.18 yet."
```

marten-seemann:
```go
// +build go1.18

package qtls

"quic-go doesn't build on Go 1.18 yet."
```

should be good enough. `go mod vendor` only seems to care about the `package` (and the `import`) part of the file.

mvdan:
I don't think that prints the string when building, though. I arrived at my example above with that objective :)

marten-seemann:
You're right. 
```go
var "quic-go doesn't build on Go 1.18 yet."
```
[...]

PR #3202, opened 2021-06-13
add a CI test that go mod vendor works

marten-seemann:
CI will fail, as #3195 is not merged yet.
Makes sure that a failure as reported in #3194 won't happen again in the future.

@mvdan, WDYT?


integrationtests/gomodvendor/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/lucas-clemente/quic-go/integrationtets/gomodvendor
```
mvdan, line 1: best to not make this module public/importable, to prevent confusion. you could name it `module test`, as that's a module path prefix reserved specifically for this kind of thing.

mvdan (approved):
SGTM. You could also go the route of creating the module with `go mod init` and `go mod edit` in CI, to perhaps have fewer files around, but it would essentially be the same.

integrationtests/gomodvendor/go.mod
marten-seemann: Good point. Didn't know that this was an option :)

lucas-clemente (approved):

PR #3215, opened 2021-06-27
fix incorrect usage of errors.Is

marten-seemann:
`errors.Is` is supposed to used for equality of errors, not for type assertions. That's what `errors.As` is there for.

cc @mvdan


internal/qerr/errors_go116.go
```diff
@@ -6,50 +6,9 @@ import (
...
-		return true
-	}
-	return target == net.ErrClosed
-}
+func (e *TransportError) Is(target error) bool          { return target == net.ErrClosed }
+func (e *ApplicationError) Is(target error) bool        { return target == net.ErrClosed }
+func (e *IdleTimeoutError) Is(target error) bool        { return target == net.ErrClosed }
+func (e *HandshakeTimeoutError) Is(target error) bool   { return target == net.ErrClosed }
+func (e *VersionNegotiationError) Is(target error) bool { return target == net.ErrClosed }
+func (e *StatelessResetError) Is(target error) bool     { return target == net.ErrClosed }
```
mvdan, line 14: :+1: not sure how I didn't catch the type assertions in the last review :)

session.go
```diff
@@ -1480,14 +1475,19 @@ func (s *session) handleCloseError(closeErr *closeError) {
 		}()
 	}
 
+	statelessResetErr := &StatelessResetError{}
+	versionNegotiationErr := &VersionNegotiationError{}
+	recreateErr := &errCloseForRecreating{}
+	applicationErr := &qerr.ApplicationError{}
+	transportErr := &qerr.TransportError{}
```
mvdan, line 1482: nit: for all these where you declare the error pointers separately, you don't need to actually initialize an error:

```
var someErr *SomeType
if errors.As(err, &someErr) { ... }
```

You do need the `&SomeType{}` for `if stmt; cond { ... }`, because that form doesn't accept a `var name Type`.

mvdan (approved):

session.go
marten-seemann: Fixed.

PR #3218, opened 2021-06-28
return error on SendMessage when session is closed

mengelbart:
Returns the error from `AddAndWait` to the user of `SendMessage` e.g. when the session is closed.


marten-seemann (approved):

marten-seemann:
Thank you!

PR #3226, opened 2021-07-08
[draft] support WebTransport

### LINKED ISSUE #3191: WebTransport Support
Hi, I try to connect http3 server writed by quic-go.
And I setup https by tls certification files.
When I try to use WebTransport in web:
```
let trans = new WebTransport('https://rtc1.live.corp.xxxx.com:443/live/2000.flv');
```

But it fail and console log is:
```
Failed to establish a connection to https://rtc1.live.corp.xxxx.com/live/2000.flv: net::ERR_QUIC_PROTOCOL_ERROR.QUIC_TLS_CERTIFICATE_UNKNOWN (TLS handshake failure (ENCRYPTION_HANDSHAKE) 46: certificate unknown).
```

but I write http3 client by quicgo and use the http3 client to connect http3 server by quic-go, It work fine.

My chrome version is  91.0.4472.77.

Please help.

ydnar:
Submitting this PR in draft form for early feedback (if you’re interested).

TODO:

- [ ] Datagram support
- [ ] WebTransport client
- [ ] Some variation of `webtransport.Upgrade(rw http.ResponseWriter, req *http.Request)`
- [ ] Unbreak existing tests
- [ ] Tests for new functionality
- [ ] Integration tests with headless Chromium

Fixes #3191

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -135,7 +135,10 @@ func (c *client) setupSession() error {
 	buf := &bytes.Buffer{}
 	quicvarint.Write(buf, streamTypeControlStream)
 	// send the SETTINGS frame
-	(&settingsFrame{Datagram: c.opts.EnableDatagram}).Write(buf)
+	(&settingsFrame{
+		Datagram:     c.opts.EnableDatagram,
+		WebTransport: c.opts.EnableWebTransport,
+	}).Write(buf)
 	_, err = str.Write(buf.Bytes())
 	return err
 }
--- a/http3/client.go
+++ b/http3/client.go
@@ -39,6 +39,7 @@ var dialAddr = quic.DialAddrEarly
 type roundTripperOpts struct {
 	DisableCompression bool
 	EnableDatagram     bool
+	EnableWebTransport bool
 	MaxHeaderBytes     int64
 }
 
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -429,7 +429,7 @@ var _ = Describe("Client", func() {
 			buf := &bytes.Buffer{}
 			rstr := mockquic.NewMockStream(mockCtrl)
 			rstr.EXPECT().Write(gomock.Any()).Do(buf.Write).AnyTimes()
-			rw := newResponseWriter(rstr, utils.DefaultLogger)
+			rw := newResponseWriter(nil, rstr, utils.DefaultLogger)
 			rw.WriteHeader(status)
 			rw.Flush()
 			return buf.Bytes()
@@ -715,7 +715,7 @@ var _ = Describe("Client", func() {
 				buf := &bytes.Buffer{}
 				rstr := mockquic.NewMockStream(mockCtrl)
 				rstr.EXPECT().Write(gomock.Any()).Do(buf.Write).AnyTimes()
-				rw := newResponseWriter(rstr, utils.DefaultLogger)
+				rw := newResponseWriter(sess, rstr, utils.DefaultLogger)
 				rw.Header().Set("Content-Encoding", "gzip")
 				gz := gzip.NewWriter(rw)
 				gz.Write([]byte("gzipped response"))
@@ -741,7 +741,7 @@ var _ = Describe("Client", func() {
 				buf := &bytes.Buffer{}
 				rstr := mockquic.NewMockStream(mockCtrl)
 				rstr.EXPECT().Write(gomock.Any()).Do(buf.Write).AnyTimes()
-				rw := newResponseWriter(rstr, utils.DefaultLogger)
+				rw := newResponseWriter(sess, rstr, utils.DefaultLogger)
 				rw.Write([]byte("not gzipped"))
 				rw.Flush()
 				str.EXPECT().Write(gomock.Any()).AnyTimes().DoAndReturn(func(p []byte) (int, error) { return len(p), nil })
--- a/http3/error_codes.go
+++ b/http3/error_codes.go
@@ -26,6 +26,9 @@ const (
 	errorMessageError         errorCode = 0x10e
 	errorConnectError         errorCode = 0x10f
 	errorVersionFallback      errorCode = 0x110
+
+	// https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-01.html#section-7.5
+	errorWebTransportBufferedStreamRejected errorCode = 0x3994bd84
 )
 
 func (e errorCode) String() string {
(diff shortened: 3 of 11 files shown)


ydnar:
Basic server: https://gist.github.com/ydnar/f73e1806c50c83f90f9b46e13bf751fc

ydnar:
Given that the WebTransport draft is still in flux, I’m hesitant to couple the WebTransport implementation too tightly with the existing `http3` package.

I can think of three ways to implement WT here:

1. Continue evolving the current implementation in this branch, where the `http3` package is WebTransport-aware, and handles (de)multiplexing streams and datagrams to the correct handler. A `webtransport` package in this repo, if any, would be a thin wrapper around `http3` interfaces that expose multiplexed streams and datagrams for a given H3 request/response session.
2. Do something along the lines of #3158, where say, a `webtransport.Server` wraps an `http3.Server`. The `webtransport.Server` would manage the QUIC listener and delegate handling non-WebTransport streams and datagrams to the `http3` package.
3. Add middleware or interceptor hooks to `http3.Server` to handle incoming QUIC sessions, streams, and datagrams. The hooks would “fall through” to the `http3` package for anything they don’t process. For instance, `http3.Server` could have an additional fields:

```go
	// Process incoming QUIC sessions.
	// If non-nil, the server will pass the newly-accepted session to the function.
	// It is the responsibility of the function to call (*Server).HandleSession()
	// to continue normal processing.
	// If the function returns an error, the QUIC connection will be closed.
	SessionHandler func(sess quic.EarlySession) error

	// Process incoming unidirectional HTTP/3 streams.
	// If non-nil, the server will parse the stream type and supply it to the function.
	// It is the responsibility of the function to call (*Server).HandleUniStream()
	// to continue normal processing.
	// If the function returns an error, the QUIC connection will be closed.
	UniStreamHandler func(sess quic.EarlySession, str quic.ReceiveStream, streamType uint64) error
```

@marten-seemann thoughts?

ydnar:
Here’s a rough design for where I landed with this. In the spirit of “smaller interfaces are better”:

### `http3.Server`

To enable HTTP/3 extensions, I propose adding a new field to `http3.Server`: `Requester`. If set, it allows callers to intercept an accepted `quic.EarlySession` and provide an `http3.Requester`. It would default to `http3.Accept`, a new exported func that would tie together the QUIC session with a QPACK decoder and some other state (see below).

```go
	// If set, the server will call Requester for each accepted QUIC session.
	// It is the responsibility of the function to return a valid Requester.
	// If Requester returns an error, the server will close the QUIC session.
	// If nil, http3.Open will be used.
	Requester func(quic.EarlySession, Settings) (Requester, error)
```

Second, add a flag to `Server` to enable the [extended CONNECT method](https://datatracker.ietf.org/doc/html/rfc8441) (necessary for WebTransport or WebSockets):

```go
	// Enable support for extended CONNECT method.
	// If set to true, the server will support CONNECT requests with a :path and :protocol header.
	EnableConnectProtocol bool
```

### `http3.Requester`

A `Requester` is responsible for providing HTTP requests to the server.

The default implementation wraps a QUIC session and handles H3 framing, request and response body streaming, and translation to/from `http` semantics. It accepts streams and datagrams, (de)multiplexing them to the relevant H3 request sessions.

The `http3` package would provide a default `Accept()` func to create a `Requester` from a QUIC session. It sets up initial state, opens the unidirectional control stream, and sends the H3 settings frame.

A WebTransport extension, for example, could override `(Server).Requester` to provide a WebTransport-aware `Requester`, and dispatch incoming WT streams and datagrams to the appropriate H3 session.

```go
// Requester represents a server-side HTTP/3 connection.
// Implementations may implement other interfaces.
type Requester interface {
	AcceptHTTP() (*http.Request, http.ResponseWriter, error)
	io.Closer
}

// Accept takes a QUIC session and HTTP/3 settings, and returns a Requester.
// It opens the control stream and sends the initial H3 settings frame,
// returning an error if either fail. The returned Requestor is ready to use.
func Accept(session quic.EarlySession, settings Settings) (Requester, error) {
	...
}
```

A `Requester` can be extended to support additional features, e.g. `interface Pusher { ... }`.

### `http3.Conn`

Internally, the default implementation of `Requester` sits on top of `http3.Conn`, which combines a `quic.EarlySession` with a QPACK handler and some other state. Both H3 client and server connections would use `Conn`.

It can be created from a quic.EarlySession via `http3.Open(quic.EarlySession, http3.Settings) (Conn, error)`.

Internally, a `Conn` holds:

- `quic.EarlySession`
- `qpack.Decoder`
- Control stream
- Peer control stream
- Settings
- Peer settings

```go
type Conn interface {
	AcceptStream(ctx) (http3.Stream, error)
	AcceptUniStream(ctx) (http3.ReceiveStream, error)
	OpenStream() (http3.SendStream, error)
	OpenUniStream() (http3.SendStream, error)

	ParseHeaders(http3.HeaderFrame) (http.Header, error)

	PeerSettings() (Settings, error)

	io.Closer
}
```

### Misc

- `http3.Settings` is a `map[uint64]uint64` with some helper methods.
- `http3.Frame` is `interface { FrameType() uint64 }`
- `http3.HeaderFrame` is a `Frame`, etc.

notedit:
Hi @ydnar great work.  I want to test this, is there a working demo now?

ydnar:
> Hi @ydnar great work. I want to test this, is there a working demo now?

Hi @notedit! It’s still quite a WIP. I’ve broken out some prerequisites into other PRs (#3233 and #3238) to enable HTTP/3 extensions to be built on top of the `http3` package, like MASQUE or WebTransport.

Ultimately, my goal is a `webtransport` package with a simple API that can upgrade an HTTP/3 client or server connection.

marten-seemann:
Sorry I haven't gotten around to reviewing this PR yet. Will do so next week.

ydnar:
> Sorry I haven't gotten around to reviewing this PR yet. Will do so next week.

No worries. This PR isn't ready for review. The other two PRs (#3233 and #3238), which add APIs for H3 extensions to use, are though.

ydnar:
Given that WebTransport depends in part and interacts with MASQUE and the H3 datagram drafts, I think it makes sense to implement these first. Especially since the WT spec needs time to align with the H3 datagram draft for multiplexing datagrams (stream ID vs quarter stream ID and context IDs).

See ietf-wg-webtrans/draft-ietf-webtrans-http3#54 for more context.

ydnar:
This PR has been replaced by #3256.

The new API is slightly different. Instead of type-asserting the `http.ResponseWriter`, get the `quic.Stream` and call `WebTransport()` to initialize the WebTransport session:

```go
	var wt http3.WebTransport
	if ds, ok := rw.(http3.DataStreamer); ok {
		if str, ok := ds.DataStream().(http3.RequestStream); ok {
			var err error
			wt, err = str.WebTransport()
			if err != nil {
				str.Close()
				return
			}
		}
	}
```

PR #3233, opened 2021-07-19
quicvarint: add Reader and Writer interfaces

ydnar:
This PR adds two interfaces to the `quicvarint` package: `Reader` and `Writer`, that are `io.Reader` + `io.ByteReader`, and `io.Writer` + `io.ByteWriter`, respectively, along with constructors that pass through the argument unchanged if it implements the necessary interfaces.

The interfaces to `quicvarint.Read` and `Write` are changed to accept a `Reader` and `Writer`, respectively. This change didn’t break tests in any package in this repo, nor require any changes in any package that depends on `quicvarint`.

Why? This is a small bit of support work for #3226, to ease construction of a package that processes H3 streams and datagrams outside of the `http3` package.

This PR also removes the existing `http3.byteReaderImpl` type, replacing it with `quicvarint.Reader`.


marten-seemann:
Thank you for this PR.
Do you have any benchmarks for this change? The thing that worries me the most are the calls to `Read` and `Write`, as these functions are called **a lot**. If I understand correctly, `NewReader` and `NewWriter` shouldn't allocate when passed a `bytes.Reader` or a `bytes.Writer`, respectively (which is how we commonly use it for parsing the header and the frames).
Given that the actual work done in `Read` and `Write` is quite trivial though, my fear is that the type assertions might show up in a benchmark.

ydnar:
> Do you have any benchmarks for this change? The thing that worries me the most are the calls to `Read` and `Write`, as these functions are called **a lot**. If I understand correctly, `NewReader` and `NewWriter` shouldn't allocate when passed a `bytes.Reader` or a `bytes.Writer`, respectively (which is how we commonly use it for parsing the header and the frames).
> Given that the actual work done in `Read` and `Write` is quite trivial though, my fear is that the type assertions might show up in a benchmark.

When I ran the existing benchmarks suite, it seemed on par. I can add benchmarks for `quicvarint.Read` and `Write` if that’d be helpful.

If the type assertion does negatively impact the benchmarks, what do you think about changing the method signatures to accept a `quicvarint.Reader` or `Writer` instead of an `io.Reader`?

marten-seemann:
> When I ran the existing benchmarks suite, it seemed on par. I can add benchmarks for quicvarint.Read and Write if that’d be helpful.

Our benchmark package is a few years old and frankly, it's not very useful any more. I was thinking of a microbenchmark just for the `Read` and `Write` function here.

> If the type assertion does negatively impact the benchmarks, what do you think about changing the method signatures to accept a quicvarint.Reader or Writer instead of an io.Reader?

I wouldn't be opposed to that. I assume it doesn't break any (quic-go) code, does it?

ydnar:
> > When I ran the existing benchmarks suite, it seemed on par. I can add benchmarks for quicvarint.Read and Write if that’d be helpful.
> 
> Our benchmark package is a few years old and frankly, it's not very useful any more. I was thinking of a microbenchmark just for the `Read` and `Write` function here.

Sounds good. Would you prefer a gomega/ginkgo or vanilla Go benchmark?

> > If the type assertion does negatively impact the benchmarks, what do you think about changing the method signatures to accept a quicvarint.Reader or Writer instead of an io.Reader?
> 
> I wouldn't be opposed to that. I assume it doesn't break any (quic-go) code, does it?

It didn’t, since every existing caller passes a `*bytes.Buffer` to `Write`, which implements `quicvarint.Writer` and `Reader`.

ydnar:
@marten-seemann looks like you’re right. The type assertion exacts a material penalty in the `Read` call, almost doubling its execution time:

```shell
❯ go test -bench . ./quicvarint
Running Suite: QUIC Varint Suite
================================
Random Seed: 1626839707
Will run 40 of 40 specs

••••••••••••••••••••••••••••••
------------------------------
• [MEASUREMENT]
Benchmarks
  with type assertion to quicvarint.Reader
    reading from a bytes.Reader

    Ran 100 samples:
    read:
      Fastest Time: 0.000s
      Slowest Time: 0.000s
      Average Time: 0.000s ± 0.000s
    read one (ns):
      Smallest: 17.913
       Largest: 64.634
       Average: 21.401 ± 5.843
    read all (ns):
      Smallest: 5732.000
       Largest: 20683.000
       Average: 6848.420 ± 1869.883
------------------------------
• [MEASUREMENT]
Benchmarks
  old Read(io.ByteReader) without type assertion
    reading from a bytes.Reader

    Ran 100 samples:
    read:
      Fastest Time: 0.000s
      Slowest Time: 0.000s
      Average Time: 0.000s ± 0.000s
    read one (ns):
      Smallest: 10.466
       Largest: 20.159
       Average: 11.693 ± 1.975
    read all (ns):
      Smallest: 3349.000
       Largest: 6451.000
       Average: 3741.860 ± 631.840
------------------------------
• [MEASUREMENT]
Benchmarks
  new Read(Reader) without type assertion
    reading from a bytes.Reader

    Ran 100 samples:
    read:
      Fastest Time: 0.000s
      Slowest Time: 0.000s
      Average Time: 0.000s ± 0.000s
    read one (ns):
      Smallest: 10.637
       Largest: 63.631
       Average: 11.613 ± 5.237
    read all (ns):
      Smallest: 3404.000
       Largest: 20362.000
       Average: 3716.000 ± 1675.817
------------------------------
••••••
Ran 40 of 40 Specs in 0.010 seconds
SUCCESS! -- 39 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
ok  	github.com/lucas-clemente/quic-go/quicvarint	0.145s
```

quicvarint/varint.go
```diff
@@ -16,9 +14,9 @@ const (
 	maxVarInt8 = 4611686018427387903
 )
 
-// Read reads a number in the QUIC varint format
-func Read(b io.ByteReader) (uint64, error) {
-	firstByte, err := b.ReadByte()
+// Read reads a number in the QUIC varint format from r.
+func Read(r Reader) (uint64, error) {
```
marten-seemann, line 18: What do we get out of changing the `io.ByteReader` to a `Reader` here? `io.ByteReader` is the smaller interface, so it would be preferable to use that one instead of the `Reader`, wouldn't it?

quicvarint/varint.go
```diff
@@ -28,53 +26,53 @@ func Read(b io.ByteReader) (uint64, error) {
...
 	if err != nil {
 		return 0, err
 	}
 	return uint64(b8) + uint64(b7)<<8 + uint64(b6)<<16 + uint64(b5)<<24 + uint64(b4)<<32 + uint64(b3)<<40 + uint64(b2)<<48 + uint64(b1)<<56, nil
 }
 
-// Write writes a number in the QUIC varint format
-func Write(b *bytes.Buffer, i uint64) {
+// Write writes i in the QUIC varint format to w.
+func Write(w Writer, i uint64) {
```
marten-seemann, line 67: Same here.

marten-seemann (commented):
Thank you for the benchmarks!

I have to admit I'm a bit confused now about the interfaces changes, see my comments on the code.

quicvarint/varint.go
ydnar: `Write` uses both `Write` and `WriteByte`.

quicvarint/varint.go
ydnar: I did this mostly for symmetry, and to mimic the `http3.byteReader` interface (https://github.com/lucas-clemente/quic-go/blob/master/http3/frames.go#L13-L16). But I see your point!

Curious: did you ever benchmark an implementation of `Read` that uses `io.Reader` instead of `io.ByteReader`?

ydnar: > Curious: did you ever benchmark an implementation of `Read` that uses `io.Reader` instead of `io.ByteReader`?

I went ahead and tried this. An implementation of `Read` that uses `io.Reader` is roughly twice as slow as the existing `ReadByte`-based implementation.

ydnar: @marten-seemann OK, so I’m happy to either revert `Read` to accept an `io.ByteReader`, or keep it as a `quicvarint.Reader`.

quicvarint/varint.go
marten-seemann: True, sorry, I missed that.

quicvarint/varint.go
marten-seemann: The Go philosophy says to use the smallest interface possible, so let's keep it an `io.ByteReader`.

ydnar:
Unrelated: do you plan to remove the CircleCI and Travis checks?

ydnar:
@marten-seemann I put up a PR (#3238) that builds on this PR.

Do you want to review or merge this one?

quicvarint/varint.go
```diff
@@ -83,35 +82,36 @@ func Write(b *bytes.Buffer, i uint64) {
...
 	l := Len(i)
 	if l == length {
-		Write(b, i)
+		Write(w, i)
 		return
 	}
 	if l > length {
 		panic(fmt.Sprintf("cannot encode %d in %d bytes", i, length))
 	}
+	qw := NewWriter(w)
```
marten-seemann, line 98: Why is `qw` needed here?

marten-seemann (approved):
One small comment, then this should be good to merge.
Can you please remove the benchmark file? Ginkgo has deprecated the benchmark functionality anyway.

quicvarint/varint.go
ydnar: It's not. Good catch!

ydnar:
> Can you please remove the benchmark file? Ginkgo has deprecated the benchmark functionality anyway.

Done. 

This was my first time using Ginkgo. I normally write vanilla Go tests. What drove the decision to use it?

marten-seemann (approved):

PR #3235, opened 2021-07-22
wrap PacketConn once

### LINKED ISSUE #3228: wrapConn called twice in server
While debugging #3227, I noticed that the server wraps the `net.PacketConn` twice into a `connection` interface, once in `(multiplexer).AddConn` and immediately after in `listen` before instantiating the `baseServer`:

https://github.com/lucas-clemente/quic-go/blob/master/server.go#L190-L201

```go
	sessionHandler, err := getMultiplexer().AddConn(conn, config.ConnectionIDLength, config.StatelessResetKey, config.Tracer)
	if err != nil {
		return nil, err
	}
	tokenGenerator, err := handshake.NewTokenGenerator(rand.Reader)
	if err != nil {
		return nil, err
	}
	c, err := wrapConn(conn)
	if err != nil {
		return nil, err
	}
```

It looks like this change was introduced in #3067 (cc @rs).

I tried a small change to `listen` that extracts the `connection` from the `packetHandlerMap` and it seems to work (although it doesn’t fix my issue with Chrome):

```go
	var c connection
	if m, ok := sessionHandler.(*packetHandlerMap); ok {
		c = m.conn
		utils.DefaultLogger.Debugf("extracted wrapped connection: %v", m.conn)
	} else {
		c, err = wrapConn(conn)
		if err != nil {
			return nil, err
		}
	}
```

I can make a PR if that’d be helpful.

ydnar:
Fixes #3228.

I discovered this when debugging issues with Chromium QUIC.

### DIFF
--- a/server.go
+++ b/server.go
@@ -195,10 +195,18 @@ func listen(conn net.PacketConn, tlsConf *tls.Config, config *Config, acceptEarl
 	if err != nil {
 		return nil, err
 	}
-	c, err := wrapConn(conn)
-	if err != nil {
-		return nil, err
+
+	// Extract the already wrapped connection if available
+	var c connection
+	if m, ok := sessionHandler.(*packetHandlerMap); ok {
+		c = m.conn
+	} else {
+		c, err = wrapConn(conn)
+		if err != nil {
+			return nil, err
+		}
 	}
+
 	s := &baseServer{
 		conn:                c,
 		tlsConf:             tlsConf,
--- a/server.go
+++ b/server.go
@@ -173,7 +173,7 @@ func ListenEarly(conn net.PacketConn, tlsConf *tls.Config, config *Config) (Earl
 	return &earlyServer{s}, nil
 }
 
-func listen(conn net.PacketConn, tlsConf *tls.Config, config *Config, acceptEarly bool) (*baseServer, error) {
+func listen(pconn net.PacketConn, tlsConf *tls.Config, config *Config, acceptEarly bool) (*baseServer, error) {
 	if tlsConf == nil {
 		return nil, errors.New("quic: tls.Config not set")
 	}
@@ -187,6 +187,11 @@ func listen(conn net.PacketConn, tlsConf *tls.Config, config *Config, acceptEarl
 		}
 	}
 
+	conn, err := wrapConn(pconn)
+	if err != nil {
+		return nil, err
+	}
+
 	sessionHandler, err := getMultiplexer().AddConn(conn, config.ConnectionIDLength, config.StatelessResetKey, config.Tracer)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


marten-seemann (commented):
I'm not sure I understand this PR. `sessionHandler` is always a `packetHandlerMap`, isn't it?
The problem here is _not_ that the same `net.PacketConn` is wrapped twice, but only that we wrap the same `net.UDPConn` twice (independently from each other).

ydnar:
Why do you want to wrap the UDPConn twice?

marten-seemann:
I don't, and I think we should fix #3228. I'm just saying that this PR doesn't fix the problem. The type assertion `sessionHandler.(*packetHandlerMap)` will always succeed.

ydnar:
> I don't, and I think we should fix #3228. I'm just saying that this PR doesn't fix the problem. The type assertion `sessionHandler.(*packetHandlerMap)` will always succeed.

OK, now I’m confused. It sounded like you *didn’t* think wrapping the `UDPConn` was a problem.

This PR does fix that, but maybe not in the most elegant way.

`newPacketHandlerMap` returns `packetHandlerManager`, which is an interface, rather than a concrete `* packetHandlerMap`. I think unless necessary for testing, a non-exported func should return a concrete type rather than an interface. The type assertion `sessionHandler.(*packetHandlerMap)` will succeed only because if the existing implementation.

How would you fix it?

ydnar:
Committed a different fix.

marten-seemann:
I don't think this fix is correct. As I said before, we're **not** double-wrapping the connection, we're just wrapping the same `net.UDPConn` twice, independent from each other.

ydnar:
I’m really confused. Do you *want* the same `net.UDPConn` wrapped twice?

marten-seemann:
It's not ideal, but it doesn't really hurt that much either.

server.go
```diff
@@ -187,20 +187,22 @@ func listen(conn net.PacketConn, tlsConf *tls.Config, config *Config, acceptEarl
 		}
 	}
 
-	sessionHandler, err := getMultiplexer().AddConn(conn, config.ConnectionIDLength, config.StatelessResetKey, config.Tracer)
+	conn, err := wrapConn(pconn)
```
ydnar, line 190: @marten-seemann this change ensures that `pconn` isn’t wrapped twice.

server.go
```diff
@@ -187,20 +187,22 @@ func listen(conn net.PacketConn, tlsConf *tls.Config, config *Config, acceptEarl
...
 	if err != nil {
 		return nil, err
 	}
-	tokenGenerator, err := handshake.NewTokenGenerator(rand.Reader)
+
+	sessionHandler, err := getMultiplexer().AddConn(conn, config.ConnectionIDLength, config.StatelessResetKey, config.Tracer)
```
ydnar, line 195: …since `AddConn` calls `wrapConn`, which will fall through because `conn` is already wrapped.

marten-seemann: ... which never happens, because `listen` is only called from `Listen{Addr}{Early}`. Or am I missing something?

ydnar: 1. `listen` calls `(*connMultiplexer).AddConn`, which calls `newPacketHandlerManager` (`newPacketHandlerMap`), which calls `wrapConn`.
2. `listen` also calls `wrapConn` itself on the same `net.UDPConn`.

That said, this doesn’t matter because `oobConn` isn’t thread-safe. That’s why I closed this PR.

PR #3236, opened 2021-07-22
Permit underlying conn to implement batch interface directly

AudriusButkevicius:
This is so that (ab)users of the package that do packet filtering to filter out stun packets before they reach quic, could still use ReadBatch.

In my implementation I simply read len(ms) packets instead of 1 off the buffer channel.

Doubt there is much gain in that, as it's all userspace already, but nice to be able to take this code path to pick up other related optimisations.

### DIFF
--- a/conn_oob.go
+++ b/conn_oob.go
@@ -115,6 +115,7 @@ func newConn(c OOBCapablePacketConn) (*oobConn, error) {
 	}
 
 	var bc batchConn
+	// Allows applications to pass in a connection that already satisfies batchConn interface
 	if ibc, ok := c.(batchConn); ok {
 		bc = ibc
 	} else {
--- a/conn_oob.go
+++ b/conn_oob.go
@@ -114,8 +114,10 @@ func newConn(c OOBCapablePacketConn) (*oobConn, error) {
 		}
 	}
 
+	// Allows callers to pass in a connection that already satisfies batchConn interface
+	// to make use of the optimisation. Otherwise, ipv4.NewPacketConn would unwrap the file descriptor 
+	// via SyscallConn(), and read it that way, which might not be what the caller wants.
 	var bc batchConn
-	// Allows applications to pass in a connection that already satisfies batchConn interface
 	if ibc, ok := c.(batchConn); ok {
 		bc = ibc
 	} else {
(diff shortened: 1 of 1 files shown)


conn_oob.go
```diff
@@ -113,9 +113,17 @@ func newConn(c OOBCapablePacketConn) (*oobConn, error) {
 			return nil, errors.New("activating packet info failed for both IPv4 and IPv6")
 		}
 	}
+
+	var bc batchConn
+	if ibc, ok := c.(batchConn); ok {
```
marten-seemann, line 118: Can you add a comment here?

AudriusButkevicius: Explaining why this would be desireable?

marten-seemann: Yeah, that applications may pass in a connection that implements `batchConn`. Just to make clear that this is a code path that's only taken if the application takes special care.

AudriusButkevicius, line 118:
```suggestion
	// Allows applications to pass in a connection that already satisfies batchConn interface
	if ibc, ok := c.(batchConn); ok {
```

AudriusButkevicius: Done, I think.
My main motive is that it would circumvent my ability to manage what is read, as ipv4.NetPacketConn ends up upwrapping the raw file descriptor and taking away any ability for me to control what is being read.

marten-seemann: SGTM.

marten-seemann (approved):

AudriusButkevicius:
Could I also ask that this would go out at the same time the batch read support goes out, which I assume should be the next release.

We'd like to get RFC9000 support in syncthing, and pick up batch reads (yet in my isolated localhost tests, I don't see any performance gains on Linux).

marten-seemann:
> Could I also ask that this would go out at the same time the batch read support goes out, which I assume should be the next release?

Yes of course.

Seems like gofmt is unhappy right now.

AudriusButkevicius:
Removed the spurious space, should be good now.

marten-seemann (approved):
Thank you!

AudriusButkevicius:
Do you guys have a release schedule? Or is it ad-hoc?

PR #3238, opened 2021-07-26
http3: add exported Settings type

ydnar:
Add a new exported `http3.Settings` type as a prerequisite for building HTTP/3 extensions like WebTransport (#3191). Builds on #3233.

- `Settings` replaces the existing `settingsFrame` type.
- New exported `SettingID` and `FrameType` types (both uint64)
- Adds support for the new `H3_DATAGRAM` [setting value](https://www.ietf.org/archive/id/draft-ietf-masque-h3-datagram-03.html#name-http-3-settings-parameter) (`0xffd276`).
- Retains compatibility with previous `H3_DATAGRAM` value (`0x276`) via a shim.

### DIFF
--- a/http3/frame.go
+++ b/http3/frame.go
@@ -1,0 +1,42 @@
+package http3
+
+import "fmt"
+
+const (
+	FrameTypeData          FrameType = 0x0
+	FrameTypeHeaders       FrameType = 0x1
+	FrameTypeCancelPush    FrameType = 0x3
+	FrameTypeSettings      FrameType = 0x4
+	FrameTypePushPromise   FrameType = 0x5
+	FrameTypeGoAway        FrameType = 0x7
+	FrameTypeMaxPushID     FrameType = 0xd
+	FrameTypeDuplicatePush FrameType = 0xe
+)
+
+// A FrameType represents an HTTP/3 frame type.
+// https://www.ietf.org/archive/id/draft-ietf-quic-http-34.html#name-frame-definitions
+type FrameType uint64
+
+// String returns the IETF registered name for t if available.
+func (t FrameType) String() string {
+	switch t {
+	case FrameTypeData:
+		return "DATA"
+	case FrameTypeHeaders:
+		return "HEADERS"
+	case FrameTypeCancelPush:
+		return "CANCEL_PUSH"
+	case FrameTypeSettings:
+		return "SETTINGS"
+	case FrameTypePushPromise:
+		return "PUSH_PROMISE"
+	case FrameTypeGoAway:
+		return "GO_AWAY"
+	case FrameTypeMaxPushID:
+		return "MAX_PUSH_ID"
+	case FrameTypeDuplicatePush:
+		return "DUPLICATE_PUSH"
+	default:
+		return fmt.Sprintf("H3 frame type 0x%x", uint64(t))
+	}
+}
--- a/http3/frame.go
+++ b/http3/frame.go
@@ -37,6 +37,6 @@ func (t FrameType) String() string {
 	case FrameTypeDuplicatePush:
 		return "DUPLICATE_PUSH"
 	default:
-		return fmt.Sprintf("0x%x", uint64(t))
+		return fmt.Sprintf("%#x", uint64(t))
 	}
 }
--- a/http3/frame.go
+++ b/http3/frame.go
@@ -3,14 +3,14 @@ package http3
 import "fmt"
 
 const (
-	FrameTypeData          FrameType = 0x0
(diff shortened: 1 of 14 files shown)


marten-seemann:
Can you rebase now that #3233 was merged?

http3/frame.go
```diff
@@ -0,0 +1,42 @@
...
+	case FrameTypePushPromise:
+		return "PUSH_PROMISE"
+	case FrameTypeGoAway:
+		return "GO_AWAY"
+	case FrameTypeMaxPushID:
+		return "MAX_PUSH_ID"
+	case FrameTypeDuplicatePush:
+		return "DUPLICATE_PUSH"
+	default:
+		return fmt.Sprintf("0x%x", uint64(t))
```
marten-seemann, line 40: This should have the same function:
```suggestion
		return fmt.Sprintf("%#x", uint64(t))
```

http3/settings.go
```diff
@@ -0,0 +1,126 @@
...
+	case SettingMaxFieldSectionSize:
+		return "MAX_FIELD_SECTION_SIZE"
+	case SettingQPACKBlockedStreams:
+		return "QPACK_BLOCKED_STREAMS"
+	case SettingDatagram:
+		return "H3_DATAGRAM"
+	case SettingDatagramDraft00:
+		return "H3_DATAGRAM (draft 00)"
+	default:
+		return fmt.Sprintf("0x%x", uint64(id))
```
marten-seemann, line 41: Same here.

http3/settings.go
```diff
@@ -0,0 +1,126 @@
...
+const (
+	SettingQPACKMaxTableCapacity SettingID = 0x01
+	SettingMaxFieldSectionSize   SettingID = 0x06
+	SettingQPACKBlockedStreams   SettingID = 0x07
+
+	// https://www.ietf.org/archive/id/draft-ietf-masque-h3-datagram-02.html#name-http-settings-parameter
+	SettingDatagram SettingID = 0xffd276
+
+	// https://datatracker.ietf.org/doc/draft-ietf-masque-h3-datagram/00/
+	SettingDatagramDraft00 SettingID = 0x276
```
marten-seemann, line 22: Do we need backwards compatibility here? IMO it would be sufficient to support one draft version.

http3/settings.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		id := SettingID(i)
+		val, err := quicvarint.Read(br)
+		if err != nil { // should not happen. We allocated the whole frame already.
+			return nil, err
+		}
+		if _, ok := s[id]; ok {
+			return nil, fmt.Errorf("duplicate setting: %d", id)
+		}
+		switch id {
```
marten-seemann, line 116: You'll need a nolint statement here to make the linter shut up.

marten-seemann (commented):
These changes mostly look fine to me on first sight, but I'm not sure I understand the motivation of this PR. Why do you want to export the `Settings` and the `FrameType`?

ydnar:
> These changes mostly look fine to me on first sight, but I'm not sure I understand the motivation of this PR. Why do you want to export the `Settings` and the `FrameType`?

The motivation for this PR is to start incrementally exposing a minimal API for packages outside of http3 to implement HTTP/3 extensions.

- I imagine a Settings field on http3.Server which a caller or an extension can specify specific settings.
- Similarly, an extension outside of this package would need to parse frames. Exporting FrameType is a step in that direction.

Would you prefer I bundle these changes into a larger PR that utilizes them?

http3/settings.go
ydnar: Chromium via quiche, still uses `0x276`: https://github.com/google/quiche/blob/2682b515465886008918b88d0c3dee48414caa4f/quic/core/http/http_constants.h#L42

```cpp
  // draft-ietf-masque-h3-datagram.
  SETTINGS_H3_DATAGRAM = 0x276,
```

marten-seemann:
> The motivation for this PR is to start incrementally exposing a minimal API for packages outside of http3 to implement HTTP/3 extensions.

That makes sense. It would be great if the API we’re exposing here is general enough that most HTTP/3 extensions can actually live outside of this package. It seems like DATAGRAM is essential enough to justify implementing it here, right?
I’m not sure if I like exposing the entire settings map to every extension. I guess it would be helpful to see how you want to use it.

ydnar:
Here's my WIP branch that implements http3.Conn as a foundation to (de)multiplex datagrams and eventually WT streams:

https://github.com/alta/quic-go/blob/http3/conn/http3/conn.go

marten-seemann:
Closing since we now have WebTransport support.

PR #3240, opened 2021-07-28
fix race when stream.Read and CancelRead are called concurrently

marten-seemann:
Not sure if this is an actual bug or a misuse of the API, but this caused CI in go-libp2p-swarm to fail.

Bubbling up this change would fix https://github.com/libp2p/go-libp2p/issues/1144.
cc @Stebalien


Stebalien (approved):
Yeah, I need to be able to call `Reset` while reading (the only way to abort a read). That will internally call `CancelRead`.

PR #3243, opened 2021-08-01
Expose convenience method for exposing TLS state

AudriusButkevicius:
We seem to craft this manually in syncthing, which seems to break quite frequently between the releases.
I know the API is marked as not stable, but if there is a helper method to get the struct we are after, makes our life much easier.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -2,6 +2,7 @@ package quic
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"io"
 	"net"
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Actually, seems I could do:
```
return q.Session.ConnectionState().TLS.ConnectionState
```
yet it's not clear to me if that is stable.

marten-seemann:
I'm not sure if I want to merge this. This is increasing the API surface of an API that, as you say, we change quite frequently.

marten-seemann:
I _think_ the current API should be fairly stable by now, but of course this very much depends on how we integrate qtls into quic-go. If I remember correctly, we haven't changed it in a while. What exactly are you struggling with?

AudriusButkevicius:
tls.ConnectionState was not exposed before. You used to have qtls.ConnectionState which was almost the same, so we used to create a new tls.ConnectionState copying fields from qtls.ConnectionState.

But tls.ConnectionState itself is not stable and keeps changing (new fields added), and the fields in qtls moved from state.Foo to state.ConnectionState().Foo etc.

PR #3254, opened 2021-08-09
add (Session).Perspective() method

ydnar:
Currently the only way to determine if a `quic.Session` is operating in client or server mode is to open or receive a stream, and get the stream ID. This PR exposes a `Perspective` method, which allows callers to determine if the session is operating on behalf of a server or a client.

- The existing `getPerspective` method is renamed to `Perspective`.
- `quic.Perspective` is a type alias for `protocol.Perspective`.
- Same for the `PerspectiveServer` and `PerspectiveClient` constants.

### DIFF
--- a/closed_session.go
+++ b/closed_session.go
@@ -89,7 +89,7 @@ func (s *closedLocalSession) destroy(error) {
 	})
 }
 
-func (s *closedLocalSession) getPerspective() protocol.Perspective {
+func (s *closedLocalSession) Perspective() Perspective {
 	return s.perspective
 }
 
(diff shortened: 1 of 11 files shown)


marten-seemann:
Why is this needed? A session that you dialed is a client-side session, a session that you accepted is a server-side session.

ydnar:
> Why is this needed? A session that you dialed is a client-side session, a session that you accepted is a server-side session.

I need it for [this branch](https://github.com/alta/quic-go/blob/http3/conn/http3/conn.go#L83-L84) where an `http3.Conn` is created from a `quic.EarlySession`, which would be used by both an H3 client and server. `http3.Conn` replaces some of the common guts from `http3.client` and `http3.Server`.

marten-seemann:
Wouldn’t it make more sense then to put this information in the http3.Conn?

ydnar:
> Wouldn’t it make more sense then to put this information in the http3.Conn?

The `http3` package needs to know if the underlying `quic.Session` is a server or client. Currently it determines that by looking at the stream ID of the control stream.

marten-seemann:
> The http3 package needs to know if the underlying quic.Session is a server or client.

The http3 package knows if it dialed the connection or if it accepted it, doesn't it?

ydnar:
How do you mean?

ydnar:
> Could this be refactored such that the `http3.Conn` is initialized close to where `quic.Listen` and `quic.Accept` is called?

#3256 implements `http3.Accept(quic.EarlyStream, Settings)` for servers and `http3.Open(quic.EarlyStream, Settings)` for clients, returning a `ServerConn` and `ClientConn`, respectively.

Internally, an `http3.Conn` behaves slightly differently based on the `(quic.Session).Perspective()`.

marten-seemann:
Closing since we now have WebTransport support.

PR #3266, opened 2021-09-07
List projects using quic-go

bt90:
Added a table with a few promiment projects which use quic-go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,19 @@ http.Client{
 }
 ```
 
+## Projects using quic-go
+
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
That's not a bad idea. What are the inclusion criteria here? Should we say something like "Listed are projects with more 1000 GitHub stars." to prevent discussions in the future who gets and who doesn't get included?

bt90:
That was at least the criteria i used. Btw it's quite hard to find those projects as Githubs code search doesn't allow to filter by project stars

PR #3268, opened 2021-09-08
enter the regular run loop if no undecryptable packet was processed

### LINKED ISSUE #3265: session: else if case is unreachable
https://github.com/lucas-clemente/quic-go/blob/ebcd98ed431baba6ab1c31e278fb80f23fdf1fd7/session.go#L593

This "else if" statement checks "!processedUndecryptablePacket" but it's not possible for the case to ever be true. processedUndecryptablePacket will always be false in that case - it can only be true when the first "if" case was taken, thus skipping the "else if"

This is a weird bit of code and I'm experiencing some cases where this particular "for" loop is infinitely spinning and taking CPU cycles, wondering if this bit of awkwardness points to possible merge mistakes in this area.

marten-seemann:
Fixes #3265.

### DIFF
--- a/session.go
+++ b/session.go
@@ -590,7 +590,9 @@ runLoop:
 				default:
 				}
 			}
-		} else if !processedUndecryptablePacket {
+		}
+		// If we processed any undecryptable packets, jump to the resetting of the timers directly.
+		if !processedUndecryptablePacket {
 			select {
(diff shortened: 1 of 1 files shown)


paralin:
Tested-by: Christian Stewart <christian@paral.in>

PR #3272, opened 2021-09-11
Allow use of custom port value in Alt-Svc header.

aaronriekenberg:
Fixes https://github.com/lucas-clemente/quic-go/issues/3262

### DIFF
--- a/example/main.go
+++ b/example/main.go
@@ -144,6 +144,7 @@ func main() {
 	flag.Var(&bs, "bind", "bind to")
 	www := flag.String("www", "", "www data")
 	tcp := flag.Bool("tcp", false, "also listen on TCP")
+	customAltSvcPort := flag.Uint("customAltSvcPort", 0, "use custom Alt-Svc header port value")
 	enableQlog := flag.Bool("qlog", false, "output a qlog (in the same directory)")
 	flag.Parse()
 
@@ -182,7 +183,12 @@ func main() {
 			var err error
 			if *tcp {
 				certFile, keyFile := testdata.GetCertificatePaths()
-				err = http3.ListenAndServe(bCap, certFile, keyFile, handler)
+				if *customAltSvcPort != 0 {
+					logger.Infof("using customAltSvcPort = %v", *customAltSvcPort)
+					err = http3.ListenAndServeWithCustomAltSvcPort(bCap, certFile, keyFile, handler, uint32(*customAltSvcPort))
+				} else {
+					err = http3.ListenAndServe(bCap, certFile, keyFile, handler)
+				}
 			} else {
 				server := http3.Server{
 					Server:     &http.Server{Handler: handler, Addr: bCap},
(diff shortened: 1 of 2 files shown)


aaronriekenberg:
@googlebot I signed it!

http3/server.go
```diff
@@ -530,7 +548,8 @@ func ListenAndServe(addr, certFile, keyFile string, handler http.Handler) error
 	}
 
 	quicServer := &Server{
-		Server: httpServer,
+		Server:           httpServer,
+		customAltSvcPort: customAltSvcPort,
```
aaronriekenberg, line 552: Another possible way to do this would be to just set `port` to `customAltSvcPort` here, and then `SetQuicHeaders` would find the port is already having a non-zero value.  Open to suggestions of what you think is the best way to do this. :)

http3/server.go
```diff
@@ -486,6 +493,17 @@ func ListenAndServeQUIC(addr, certFile, keyFile string, handler http.Handler) er
...
+}
+
+// ListenAndServeWithCustomAltSvcPort listens on the given network address for both, TLS and QUIC
+// connetions in parallel. It returns if one of the two returns an error.
+// http.DefaultServeMux is used when handler is nil.
+// The correct Alt-Svc headers for QUIC are set.
+// customAltSvcPort is used to override the default port value in the Alt-Svc response header.
+// This is useful when a Layer 4 firewall is redirecting UDP traffic and clients must use
+// a port different from the port the QUIC server itself is listening on.
+func ListenAndServeWithCustomAltSvcPort(addr, certFile, keyFile string, handler http.Handler, customAltSvcPort uint32) error {
```
marten-seemann, line 506: I don't think we need an extra function here. Users who need to change the port should just initialize an `http3.Server.`.

example/main.go
```diff
@@ -182,7 +183,12 @@ func main() {
 			var err error
 			if *tcp {
 				certFile, keyFile := testdata.GetCertificatePaths()
-				err = http3.ListenAndServe(bCap, certFile, keyFile, handler)
+				if *customAltSvcPort != 0 {
+					logger.Infof("using customAltSvcPort = %v", *customAltSvcPort)
+					err = http3.ListenAndServeWithCustomAltSvcPort(bCap, certFile, keyFile, handler, uint32(*customAltSvcPort))
+				} else {
+					err = http3.ListenAndServe(bCap, certFile, keyFile, handler)
+				}
```
marten-seemann, line 191: I don't think this is important enough to warrant complicating the example.

http3/server.go
```diff
@@ -92,7 +92,8 @@ type Server struct {
 	// See https://www.ietf.org/archive/id/draft-schinazi-masque-h3-datagram-02.html.
 	EnableDatagrams bool
 
-	port uint32 // used atomically
+	port             uint32 // used atomically
+	customAltSvcPort uint32 // custom port used for Alt-Svc response header
```
marten-seemann, line 96: I think it should be fine just to export `port` here.

marten-seemann (changes requested):

marten-seemann (commented):
Code looks good now. Can we have a test case please?

marten-seemann (approved):

PR #3289, opened 2021-10-15
don't send a RESET_STREAM frame on a stream that already sent the FIN

### LINKED ISSUE #3257: INTERNAL_ERROR while writing/reading to/from a stream
My application sporadically generates the following error while reading (or writing) to a `quic.Stream`:
`INTERNAL_ERROR: tried to delete unknown incoming stream 205766289`

The stream number is always different. The reading and writing sides of the stream both encounter the same error.

Are there any recommended steps I should take in my app to avoid such an error? I can't seem to find a pattern to the error occurring.

Thanks in advance.

marten-seemann:
Fixes #3257. @arashpayan, could you review this PR?

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -414,7 +414,7 @@ func (s *sendStream) CancelWrite(errorCode StreamErrorCode) {
 // must be called after locking the mutex
 func (s *sendStream) cancelWriteImpl(errorCode qerr.StreamErrorCode, writeErr error) {
 	s.mutex.Lock()
-	if s.canceledWrite {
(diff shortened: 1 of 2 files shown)


marten-seemann:
This PR is wrong. See https://datatracker.ietf.org/doc/html/rfc9000#section-3.1.
After the FIN has been sent, a RESET_STREAM can still be sent: The reason is that as soon as the stream is canceled, retransmissions of STREAM frames are also canceled, so the peer might not even have received the FIN.

PR #3290, opened 2021-10-16
don't unlock the receive stream mutex for copying from STREAM frames

### LINKED ISSUE #3257: INTERNAL_ERROR while writing/reading to/from a stream

marten-seemann:
Fixes #3257. @arashpayan can you review this PR please? I'm planning to cut a new quic-go release early next week, and would love to get this fix in.

The problem here was the mutex in the `receiveStream`:
https://github.com/lucas-clemente/quic-go/blob/a22a9eae30618491bf6c40ad20918650e2a2ab7b/receive_stream.go#L169-L175
We unlocked this mutex when copying data from the STREAM frame to the `b` slice of the `Read` call. If during this time `CancelRead` is called:
https://github.com/lucas-clemente/quic-go/blob/a22a9eae30618491bf6c40ad20918650e2a2ab7b/receive_stream.go#L200-L209
This would call `onStreamCompleted` for the first time. `onStreamCompleted` removes the stream from the streams map.

If the STREAM frame copied in the `Read` call was the last STREAM frame of the stream (i.e. the one carrying the FIN bit), we'd call `onStreamCompleted` again, trying to delete the previously deleted stream, leading to an INTERNAL_ERROR.

This PR also adds an integration test which is capable of reproducing the bug reliable within just a few repetitions.

### DIFF
--- a/integrationtests/self/cancelation_test.go
+++ b/integrationtests/self/cancelation_test.go
@@ -631,4 +631,92 @@ var _ = Describe("Stream Cancelations", func() {
 			Expect(server.Close()).To(Succeed())
 		})
 	})
+
+	It("doesn't run into any errors when streams are canceled all the time", func() {
+		const maxIncomingStreams = 1000
+		server, err := quic.ListenAddr(
+			"localhost:0",
+			getTLSConfig(),
+			getQuicConfig(&quic.Config{MaxIncomingStreams: maxIncomingStreams, MaxIdleTimeout: 10 * time.Second}),
+		)
+		Expect(err).ToNot(HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


arashpayan:
I'll be able to test this out either tomorrow or Monday, @marten-seemann.

PR #3317, opened 2022-01-14
add a AllowConnectionWindowIncrease config option

marten-seemann:
This option can be used to limit the memory consumption of a QUIC connection. When set, the application can decide if the connection flow control window is increased.

Note that this only applies to the connection flow control window, not to stream windows. However, stream windows are limited by the connection window (the sum of outstanding data on all streams must not exceed the connection window), so this actually does limit the memory usage.

cc @vyzo

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -270,6 +270,8 @@ type Config struct {
 	// to increase the connection flow control window.
 	// If set, the caller can prevent an increase of the window. Typically, it would do so to
 	// limit the memory usage.
+	// To avoid deadlocks, it is not valid to call other functions on the session or on streams
+	// in this callback.
 	AllowConnectionWindowIncrease func(sess Session, delta uint64) bool
 	// MaxIncomingStreams is the maximum number of concurrent bidirectional streams that a peer is allowed to open.
 	// Values above 2^60 are invalid.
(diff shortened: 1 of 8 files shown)


vyzo (approved):
looks nice and simple

interface.go
```diff
@@ -266,6 +266,11 @@ type Config struct {
 	// MaxConnectionReceiveWindow is the connection-level flow control window for receiving data.
 	// If this value is zero, it will default to 15 MB.
 	MaxConnectionReceiveWindow uint64
+	// AllowConnectionWindowIncrease is called every time the connection flow controller attempts
+	// to increase the connection flow control window.
+	// If set, the caller can prevent an increase of the window. Typically, it would do so to
+	// limit the memory usage.
+	AllowConnectionWindowIncrease func(sess Session, delta int) bool
```
lucas-clemente, line 273: Maybe [u]int64 to be consistent with other byte-valued vars in this struct?

lucas-clemente, line 273: When calling this, we hold at least one mutex (the one in the connection flow controller), and are within the session run loop. Is it safe to call all Session functions from here, or are there any restrictions we need to document?

lucas-clemente (approved):

interface.go
marten-seemann: Done.

marten-seemann: Yeah, you should call any functions on the Session, are any Stream. I added a comment.

PR #3326, opened 2022-01-28
Disable anti-amplification limit by address validation token

### LINKED ISSUE #3319: Address Validation Token does not disable Anti-Amplification Limit
Currently [ackhandler.sentPacketHandler](https://github.com/lucas-clemente/quic-go/blob/master/internal/ackhandler/sent_packet_handler.go) always waits for a handshake packet from the client to disable the anti-amplification limit.
But the anti-amplification limit could be disabled when an address validation token is provide by the client.
This would be beneficial for servers responding to 0-RTT requests.
See https://datatracker.ietf.org/doc/html/rfc9000#section-8.1

birneee:
Fixes #3319.


internal/ackhandler/ackhandler.go
```diff
@@ -6,7 +6,9 @@ import (
 	"github.com/lucas-clemente/quic-go/logging"
 )
 
-// NewAckHandler creates a new SentPacketHandler and a new ReceivedPacketHandler
+// NewAckHandler creates a new SentPacketHandler and a new ReceivedPacketHandler.
+// clientAddressValidated indicates weather the address was validated beforehand by an address validation token.
```
marten-seemann, line 10:
```suggestion
// clientAddressValidated indicates whether the address was validated beforehand by an address validation token.
```

internal/ackhandler/ackhandler.go
```diff
@@ -15,7 +17,8 @@ func NewAckHandler(
 	tracer logging.ConnectionTracer,
 	logger utils.Logger,
 	version protocol.VersionNumber,
+	clientAddressValidated bool,
```
marten-seemann, line 20: Nit: move this after the `rttStats`.

internal/ackhandler/sent_packet_handler.go
```diff
@@ -101,13 +101,17 @@ var (
...
+// clientAddressValidated immediately disables the amplification limit.
+// clientAddressValidated has no effect for a client.
 func newSentPacketHandler(
 	initialPN protocol.PacketNumber,
 	initialMaxDatagramSize protocol.ByteCount,
 	rttStats *utils.RTTStats,
 	pers protocol.Perspective,
 	tracer logging.ConnectionTracer,
 	logger utils.Logger,
+	clientAddressValidated bool,
```
marten-seemann, line 114: Here too.

session.go
```diff
@@ -249,6 +249,7 @@ var newSession = func(
 	tracingID uint64,
 	logger utils.Logger,
 	v protocol.VersionNumber,
+	clientAddressValidated bool,
```
marten-seemann, line 252: Nit: move this after `enable0RTT`.

internal/ackhandler/sent_packet_handler.go
```diff
@@ -101,13 +101,17 @@ var (
 	_ sentPacketTracker = &sentPacketHandler{}
 )
 
+// clientAddressValidated indicates weather the address was validated beforehand by an address validation token.
+// clientAddressValidated immediately disables the amplification limit.
+// clientAddressValidated has no effect for a client.
```
marten-seemann, line 106:
```suggestion
// If the address was validated, the amplification limit doesn't apply. It has no effect for a client.
```

marten-seemann (commented):
This looks pretty good already. Can you add a unit test in the ackhandler package?

birneee:
@marten-seemann please check again

internal/ackhandler/sent_packet_handler.go
```diff
@@ -101,10 +101,13 @@ var (
 	_ sentPacketTracker = &sentPacketHandler{}
 )
 
+// clientAddressValidated indicates weather the address was validated beforehand by an address validation token.
```
marten-seemann, line 104:
```suggestion
// clientAddressValidated indicates whether the address was validated beforehand by an address validation token.
```

internal/ackhandler/sent_packet_handler_test.go
```diff
@@ -944,6 +944,18 @@ var _ = Describe("SentPacketHandler", func() {
...
 
+	Context("amplification limit, for the server, with validated address", func() {
+		JustBeforeEach(func() {
+			rttStats := utils.NewRTTStats()
+			handler = newSentPacketHandler(42, protocol.InitialPacketSizeIPv4, rttStats, true, perspective, nil, utils.DefaultLogger)
+		})
+
+		It("do not limits the window", func() {
+			Expect(handler.SendMode()).To(Equal(SendAny))
+			Expect(handler.peerAddressValidated).To(Equal(true))
```
marten-seemann, line 955: It would be better to test the expected behavior than to make assertion on implementation details.

marten-seemann (changes requested):

marten-seemann:
@birneee Are you still working on this? Would love to get this merged!

birneee:
@marten-seemann i just worked on it :)

marten-seemann (approved):

marten-seemann (changes requested):
So, I messed up this review. Sorry for that!

I don't think it's correct to use the result of `AcceptToken`. Consider the following scenario: A server only wants to enable Retries when it's under load. Therefore, it will return `true` from the `AcceptToken` callback when it is not under load. However, that doesn't mean that we should trust the address and send an unlimited amount of data there.

birneee:
@marten-seemann you are right, because the AcceptToken can be configured. The defaultAcceptToken function however checks the validity. So we could either use a function similar to defaultAcceptToken by calling it directly or add a new config ValidateToken, where by default AcceptToken = ValidateToken. But I am not sure if ValidateToken should be configurable.

birneee:
@marten-seemann please check
I think, configuring token validity is another issue.

marten-seemann:
@birneee We just completely rewrote the Retry API in #3501 (see #3494 for design considerations). As the token validation logic now lives inside quic-go, this should make it possible to disable the anti-amplification limit when we successfully validated a token.

Do you want to rebase and modify your PR to take advantage of this? Fixing #3319 would be a super valuable contribution for everyone who actually wants to take full advantage of 0-RTT.

birneee:
@marten-seemann it is rebased now

marten-seemann (approved):
This looks very clean! Thank you @birneee!

Issue #3327, opened 2022-02-03
PMTUD may be implemented incorrectly


tobyxdd:
Hi,

We talked about the DF bit on Windows a while ago, and I submitted https://github.com/lucas-clemente/quic-go/pull/3155. But apparently it led to this unforeseen problem https://github.com/lucas-clemente/quic-go/issues/3273 so PMTUD ended up being disabled on Windows as a dirty fix. As I revisited this today I realized that perhaps there is something fundamentally wrong with the way quic-go currently implements PMTUD, and it's not really a Windows issue.

Previously we both assumed that the DF bit is enabled by default on Linux. But that's not true, as the man page says this:

![image](https://user-images.githubusercontent.com/1414953/152260803-1865538b-da59-473e-8118-cd0a67842116.png)

The default behavior on Linux (`IP_PMTUDISC_WANT`) is to only set DF bit when the packet is smaller than the path MTU, and fragment the packet otherwise. This implies that right now even if quic-go sends a packet larger than the path MTU, it will still be fragmented, sent, and received, nullifying PMTUD altogether. And if you explicitly set it to `IP_PMTUDISC_DO`, you will get a similar behavior as on Windows - errors when trying to send packets larger than the path MTU, instead of discarding them silently.

I wrote a demo and verified this on my Ubuntu 20.04 server, which has an MTU of 1500. When it sends packets smaller than 1500, the DF bit is set. When it sends packets larger than 1500, it fragments them and sends them successfully nonetheless.

![image](https://user-images.githubusercontent.com/1414953/152258873-d57f59fd-c430-4324-b6d2-e6c2b243af61.png)

tobyxdd:
I also wrote a custom PacketConn that adds some padding bytes to every packet to make it larger, and passed it to quic-go. The QUIC payload reaches MaxPacketBufferSize regardless of the MTU and the actual packet size going out.

marten-seemann:
Does setting the DF socket option resolve this?

tobyxdd:
That would lead to the same problem as in https://github.com/lucas-clemente/quic-go/issues/3273 - the kernel returns error instead of silently dropping the packet, at least on Linux. Haven't tried it on any other platform.

tobyxdd:
The error looks like this on Linux: `write udp [::]:52857->x.x.x.x:6003: sendto: message too long`

marten-seemann:
[...]

PR #3328, opened 2022-02-09
sendQueue: ignore "datagram too large" error

### LINKED ISSUE #3273: mtu_discoverer: INTERNAL_ERROR on Windows
With Path MTU Discovery enabled on windows, an error is thrown:
```
{"level":"error","ts":1631441093.781284,"caller":"peer/quic.go:65","msg":"accepting stream from peers","PeerID":11890175577191328473,"Initiator":true,"error":"INTERNAL_ERROR: write udp [::]:62885->[redacted]:1111: wsasendto: A message sent on a datagram socket was larger than the internal message buffer or some other network limit, or the buffer used to receive a datagram into was smaller than the datagram itself.","stacktrace":"github.com/zllovesuki/t/peer.(*QUIC).Start\n\tC:/Users/Ultium/code/t/peer/quic.go:65"}
```

`*quic.Config` is as follows:
```
&quic.Config{
	KeepAlive:               true,
	HandshakeIdleTimeout:    time.Second * 3,
	MaxIdleTimeout:          time.Second * 15,
}
```

Curiously, the INTERNAL_ERROR is not raised immediately. I was able to `AcceptStream(ctx)`/`OpenStream()` to do application layer handshake, then the quic session is passed to a different struct to do `AcceptStream(ctx)` in a loop. The error is then raised.

zllovesuki:
This commit introduces additional platform-dependent checking when the
kernel returns an error. Previously, the session is terminated when
PingFrame sends a discovery packet larger than the limit. With this
commit, the error is checked, and if it is "datagram too large", the
error is ignored.

Additionally,
1. This commit re-enables MTU discovery on Windows unless it
is disabled explicitly by user (Undo #3276),
2. Set IP_DONTFRAGMENT and IPV6_DONTFRAG with error checking on Windows,
   and
3. Set IP_MTU_DISCOVERY to PMTUDISC_DO for both IPv4 and IPv6 on Linux
   so that the kernel will return "message too long".

Fixes #3273 #3327

### DIFF
--- a/conn_windows.go
+++ b/conn_windows.go
@@ -9,12 +9,16 @@ import (
 	"net"
 	"syscall"
 
+	"github.com/lucas-clemente/quic-go/internal/utils"
 	"golang.org/x/sys/windows"
 )
 
 const (
-	disablePathMTUDiscovery = true
-	IP_DONTFRAGMENT         = 14
+	// same for both IPv4 and IPv6 on Windows
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IP_DONTFRAG.html
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IPV6_DONTFRAG.html
+	IP_DONTFRAGMENT = 14
+	IPV6_DONTFRAG   = 14
 )
 
 func newConn(c OOBCapablePacketConn) (connection, error) {
@@ -22,13 +26,22 @@ func newConn(c OOBCapablePacketConn) (connection, error) {
 	if err != nil {
 		return nil, fmt.Errorf("couldn't get syscall.RawConn: %w", err)
 	}
+	var errDFIPv4, errDFIPv6 error
 	if err := rawConn.Control(func(fd uintptr) {
-		// This should succeed if the connection is a IPv4 or a dual-stack connection.
-		// It will fail for IPv6 connections.
-		// TODO: properly handle error.
-		_ = windows.SetsockoptInt(windows.Handle(fd), windows.IPPROTO_IP, IP_DONTFRAGMENT, 1)
+		errDFIPv4 = windows.SetsockoptInt(windows.Handle(fd), windows.IPPROTO_IP, IP_DONTFRAGMENT, 1)
+		errDFIPv6 = windows.SetsockoptInt(windows.Handle(fd), windows.IPPROTO_IPV6, IPV6_DONTFRAG, 1)
 	}); err != nil {
 		return nil, err
+	}
+	switch {
+	case errDFIPv4 == nil && errDFIPv6 == nil:
+		utils.DefaultLogger.Debugf("Setting DF for IPv4 and IPv6.")
+	case errDFIPv4 == nil && errDFIPv6 != nil:
+		utils.DefaultLogger.Debugf("Setting DF for IPv4.")
+	case errDFIPv4 != nil && errDFIPv6 == nil:
+		utils.DefaultLogger.Debugf("Setting DF for IPv6.")
+	case errDFIPv4 != nil && errDFIPv6 != nil:
+		return nil, errors.New("setting Df failed for both IPv4 and IPv6")
 	}
 	return &basicConn{PacketConn: c}, nil
 }
--- a/err_size.go
+++ b/err_size.go
@@ -1,0 +1,9 @@
+//go:build !linux && !windows
+// +build !linux,!windows
+
+package quic
+
+func isMsgSizeErr(err error) bool {
+	// to be implemented for more specific platforms
+	return false
+}
--- a/send_queue.go
+++ b/send_queue.go
@@ -64,7 +64,13 @@ func (h *sendQueue) Run() error {
 			shouldClose = true
 		case p := <-h.queue:
 			if err := h.conn.Write(p.Data); err != nil {
-				return err
+				// This additional check enables:
+				// 1. Checking for "datagram too large" message from the kernel, as such,
+				// 2. Path MTU discovery,and
+				// 3. Eventual detection of loss PingFrame.
+				if !isMsgSizeErr(err) {
+					return err
+				}
 			}
 			p.Release()
 			select {
--- a/conn_generic.go
+++ b/conn_generic.go
@@ -5,8 +5,6 @@ package quic
 
 import "net"
 
-const disablePathMTUDiscovery = false
-
 func newConn(c net.PacketConn) (connection, error) {
 	return &basicConn{PacketConn: c}, nil
 }
--- a/conn_helper_darwin.go
+++ b/conn_helper_darwin.go
@@ -5,10 +5,7 @@ package quic
 
 import "golang.org/x/sys/unix"
 
-const (
-	msgTypeIPTOS            = unix.IP_RECVTOS
-	disablePathMTUDiscovery = false
-)
+const msgTypeIPTOS = unix.IP_RECVTOS
 
 const (
 	ipv4RECVPKTINFO = unix.IP_RECVPKTINFO
--- a/conn_helper_freebsd.go
+++ b/conn_helper_freebsd.go
@@ -6,8 +6,7 @@ package quic
 import "golang.org/x/sys/unix"
 
 const (
-	msgTypeIPTOS            = unix.IP_RECVTOS
-	disablePathMTUDiscovery = false
+	msgTypeIPTOS = unix.IP_RECVTOS
 )
 
 const (
--- a/conn_helper_linux.go
+++ b/conn_helper_linux.go
@@ -5,10 +5,7 @@ package quic
 
 import "golang.org/x/sys/unix"
 
-const (
-	msgTypeIPTOS            = unix.IP_TOS
-	disablePathMTUDiscovery = false
-)
+const msgTypeIPTOS = unix.IP_TOS
 
 const (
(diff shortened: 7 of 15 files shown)


zllovesuki:
Hooking into `onLost` and `onAcked` shows that MTU was able to increase with this commit.
```
		OnLost: func(wire.Frame) {
			f.probeInFlight = false
			f.max = size
			fmt.Printf("MTU DISCOVERY GONE: %+v\n", f)
		},
		OnAcked: func(wire.Frame) {
			f.probeInFlight = false
			f.current = size
			f.mtuIncreased(size)
			fmt.Printf("MTU DISCOVERY GOT: %+v\n", f)
		},
```

On a connection via WireGuard that causes MTU to be reduced:
```
MTU DISCOVERY GOT: &{lastProbeTime:{wall:13869129029832926620 ext:622940201 loc:0x171a600} probeInFlight:false mtuIncreased:0x102ede0 rttStats:0xc0000d43c0 current:1352 max:1452}
MTU DISCOVERY GONE: &{lastProbeTime:{wall:13869129030004107620 ext:794121201 loc:0x171a600} probeInFlight:false mtuIncreased:0x102ede0 rttStats:0xc0000d43c0 current:1352 max:1402}
MTU DISCOVERY GOT: &{lastProbeTime:{wall:13869129038891791436 ext:9018128601 loc:0x171a600} probeInFlight:false mtuIncreased:0x102ede0 rttStats:0xc0000d43c0 current:1377 max:1402}
MTU DISCOVERY GOT: &{lastProbeTime:{wall:13869129039070470136 ext:9196807301 loc:0x171a600} probeInFlight:false mtuIncreased:0x102ede0 rttStats:0xc0000d43c0 current:1389 max:1402}
```

On vanilla connection:
```
MTU DISCOVERY GOT: &{lastProbeTime:{wall:13869129101720445128 ext:470531601 loc:0x11ba600} probeInFlight:false mtuIncreased:0xacede0 rttStats:0xc00002c540 current:1352 max:1452}
MTU DISCOVERY GOT: &{lastProbeTime:{wall:13869129101906285828 ext:656372301 loc:0x11ba600} probeInFlight:false mtuIncreased:0xacede0 rttStats:0xc00002c540 current:1402 max:1452}
MTU DISCOVERY GOT: &{lastProbeTime:{wall:13869129102166036452 ext:842381101 loc:0x11ba600} probeInFlight:false mtuIncreased:0xacede0 rttStats:0xc00002c540 current:1427 max:1452}
MTU DISCOVERY GOT: &{lastProbeTime:{wall:13869129102354195652 ext:1030540301 loc:0x11ba600} probeInFlight:false mtuIncreased:0xacede0 rttStats:0xc00002c540 current:1439 max:1452}
```

err_size.go
```diff
@@ -0,0 +1 @@
+package quic
```
marten-seemann, line 1: We probably don't need this file?

send_queue.go
```diff
@@ -64,7 +64,9 @@ func (h *sendQueue) Run() error {
 			shouldClose = true
 		case p := <-h.queue:
 			if err := h.conn.Write(p.Data); err != nil {
-				return err
+				if !isMsgSizeErr(err) {
```
marten-seemann, line 67: Can you add a comment here, explaining
1. That this makes it possible to run MTU discovery
1. That we're returning MTU errors from the kernel (and that the kernel might act upon ICMP packets)
1. That the packet will be detected lost by loss recovery (eventually)

conn_windows.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	disablePathMTUDiscovery = true
+	disablePathMTUDiscovery = false
```
marten-seemann, line 16: We can remove this const (and the associated code path now), right? It's set to `false` on all architectures.

marten-seemann (approved):
This looks very clean. Thank you @zllovesuki!
Just a few nits.

err_size.go
zllovesuki: I think the `// +build unix` flag is not yet merged on Go upstream yet. I need to make some changes.

zllovesuki: ok, i moved _generic into this file instead.

conn_windows.go
zllovesuki: undo previous PR that added the checks.

zllovesuki:
Unfortunately I couldn't properly test the commit on Linux or FreeBSD as they are running in Hyper-V. If someone can help with testing those (e.g. vanilla vs via VPN so reduced MTU).

On illumos/amd64 it produces no regression (good).

tobyxdd:
Perhaps we should disable PMTUD for all platforms, except for those that we have an implementation that turns on DF bit? This PR only covers Windows and Linux. Does anyone know the default behavior and how to turn on DF on platforms like macOS, FreeBSD, etc.?

zllovesuki:
That is certainly an idea, however I wold _really_ prefer if we can add more support for those OSes. MTU of 1300 vs 1400, that's extra 10 packets to send for 1000 bytes of data (!). Literally making the internet a better place if we can make it more efficient.

tobyxdd:
@marten-seemann  care to comment on this?

I can confirm that this works fine on Linux & Windows.

tobyxdd:
https://gist.github.com/tobyxdd/1964cc8d559fb367010c80c1dafee1b1

This patch (based on the code of current PR) moves the set DF implementation to a separate file for each platform & the passed-in `PacketConn` no longer needs to be an `OOBCapablePacketConn` - it only needs to have `SyscallConn`.

This is useful when the `PacketConn` is not a `net.UDPConn` for whatever reason (e.g. in [my project](https://github.com/HyNetwork/hysteria) where there's an added layer of obfuscation on top of `UDPConn`)

marten-seemann (approved):
LGTM. Thank you @zllovesuki!

zllovesuki:
> [gist.github.com/tobyxdd/1964cc8d559fb367010c80c1dafee1b1](https://gist.github.com/tobyxdd/1964cc8d559fb367010c80c1dafee1b1)
> 
> This patch (based on the code of current PR) moves the set DF implementation to a separate file for each platform & the passed-in `PacketConn` no longer needs to be an `OOBCapablePacketConn` - it only needs to have `SyscallConn`.
> 
> This is useful when the `PacketConn` is not a `net.UDPConn` for whatever reason (e.g. in [my project](https://github.com/HyNetwork/hysteria) where there's an added layer of obfuscation on top of `UDPConn`)

Do you think we should abstract this part out? @marten-seemann

marten-seemann:
Yes, but let's do that in a follow-up PR. Makes reviewing easier.

zllovesuki:
/unrelated

I got my hands on an ppc64le machine to test (courtesy of folks at https://openpower.ic.unicamp.br/minicloud/), and it seems like quic-go works quite well on ppc64le as well (on Debian/ppc64el at least).

zllovesuki:
@tobyxdd would you mind submitting a new PR with your patch after this PR has been merged?

tobyxdd:
> If we are going to ignore the ICMP PTB message anyway, why not just set `IP(V6)_MTU_DISCOVER` to `IP_PMTUDISC_PROBE`? This disables IP fragmentation and allows you to send UDP packets exceeding the system cached PMTU without returning `-EMSGSIZE`.
> 
> `IP_PMTUDISC_PROBE` is supported by Linux and all versions of Windows that are still supported by Microsoft. Unless supporting outdated Windows systems like Windows 7 is a goal, `IP(V6)_MTU_DISCOVER` should be preferred over `IP(V6)_DONTFRAGMENT`.

https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options#windows-support-for-ip_proto-options

According to MS docs, neither `IP_PMTUDISC_PROBE` nor `IP_MTU_DISCOVER` is available on Windows

database64128:
@tobyxdd The link you referenced directly contradicts what you said. Just press Ctrl + F and search `IP_MTU_DISCOVER`.

If you have any Windows device around, open `"C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\shared\ws2ipdef.h"` and go to line 121. Everything I mentioned is right there.

I suggested using `IP_MTU_DISCOVER` on Windows, because recently I have been working on [this project of mine](https://github.com/database64128/swgp-go) and it requires disabling IP fragmentation. The code dealing with disabling IP fragmentation on Linux, Windows, macOS and FreeBSD is [here](https://github.com/database64128/swgp-go/tree/main/conn). I can confirm it works as intended on my Windows and Linux PC. A friend of mine confirmed that it also works on macOS.

tobyxdd:
> @tobyxdd The link you referenced directly contradicts what you said. Just press Ctrl + F and search `IP_MTU_DISCOVER`.
> 
> If you have any Windows device around, open `"C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\shared\ws2ipdef.h"` and go to line 121. Everything I mentioned is right there.
> 
> I suggested using `IP_MTU_DISCOVER` on Windows, because recently I have been working on [this project of mine](https://github.com/database64128/swgp-go) and it requires disabling IP fragmentation. The code dealing with disabling IP fragmentation on Linux, Windows, macOS and FreeBSD is [here](https://github.com/database64128/swgp-go/tree/main/conn). I can confirm it works as intended on my Windows and Linux PC. A friend of mine confirmed that it also works on macOS.

My bad. I did see `IP_MTU_DISCOVER`, but since it's not listed in the "Windows support for IP_PROTO options" chart, I thought it was unsupported. I'll give it a try later today, and possibly submit a new PR.

PR #3334, opened 2022-02-20
move set DF implementation to separate files & avoid the need for OOBCapablePacketConn

tobyxdd:


marten-seemann (commented):
This looks good to me. Can we merge the err_size_*.go files with the conn_df_*.go files?

You'll also need to run `gofumpt` (probably `gofmt` would be enough), CI is unhappy with the `import`s.

marten-seemann (approved):

PR #3349, opened 2022-03-17
Implement http3.Server.ServeListener

### LINKED ISSUE #3347: introduce a http3.Server.Serve
`http3.Server.Serve` would be the analogue of [`http.Server.Serve`](https://pkg.go.dev/net/http#Server.Serve), taking a `quic.EarlyListener` as a parameter.

The `http3.Server` would then accept connections from that listener until it is closed. Importantly, it would not close the listener when closing, thereby allowing one `http3.Server` being switched out for another one seamlessly.

See https://github.com/caddyserver/caddy/issues/4348#issuecomment-1068726696 for motivation.

renbou:
Fixes #3347

Added a `http3.Server.ServeListener` (`http3.Server.Serve` is already taken and serves `net.PacketConn`) method which is can be used to serve content on an already created `quic.EarlyListener` which works like `http.Server.Serve` in net/http.

Originally, however, it was planned for these listeners not to be closed using `http3.Server.Close` so that `http3.Server`s can be switched between without closing/starting another listener. It turned out that this is easier implemented externally (for example Caddy uses "fake close" wrappers around listeners), since there isn't an easy way to interrupt the `Accept` running in `http3.Server.serveImpl` which was run using an `quic.EarlyListener` without closing the listener.

Note that `net/http`'s `Server.Close` also closes the listener passed to `Server.Serve`, so the behavior is even more similar this way.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -180,10 +180,37 @@ func (s *Server) Serve(conn net.PacketConn) error {
 // and use it to construct a http3-friendly QUIC listener.
 // Closing the server does close the listener.
 func (s *Server) ServeListener(listener quic.EarlyListener) error {
-	return s.serveImpl(listener)
+	return s.serveImpl(func() (quic.EarlyListener, error) { return listener, nil })
 }
 
 func (s *Server) serveConn(tlsConf *tls.Config, conn net.PacketConn) error {
+	return s.serveImpl(func() (quic.EarlyListener, error) {
+		baseConf := ConfigureTLSConfig(tlsConf)
+		quicConf := s.QuicConfig
+		if quicConf == nil {
+			quicConf = &quic.Config{}
+		} else {
+			quicConf = s.QuicConfig.Clone()
+		}
+		if s.EnableDatagrams {
+			quicConf.EnableDatagrams = true
+		}
+
+		var ln quic.EarlyListener
+		var err error
+		if conn == nil {
+			ln, err = quicListenAddr(s.Addr, baseConf, quicConf)
+		} else {
+			ln, err = quicListen(conn, baseConf, quicConf)
+		}
+		if err != nil {
+			return nil, err
+		}
+		return ln, nil
+	})
+}
+
+func (s *Server) serveImpl(startListener func() (quic.EarlyListener, error)) error {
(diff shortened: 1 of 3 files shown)


http3/server.go
```diff
@@ -127,17 +165,52 @@ func (s *Server) ListenAndServeTLS(certFile, keyFile string) error {
...
+// Serve an existing QUIC listener.
+// Make sure you use http3.ConfigureTLSConfig to configure a tls.Config
+// and use it to construct a http3-friendly QUIC listener.
+// Closing the server does close the listener.
+func (s *Server) ServeListener(listener quic.EarlyListener) error {
+	return s.serveImpl(func() (quic.EarlyListener, error) { return listener, nil })
+}
+
+func (s *Server) serveConn(tlsConf *tls.Config, conn net.PacketConn) error {
+	return s.serveImpl(func() (quic.EarlyListener, error) {
```
marten-seemann, line 187: Is there any reason not to construct the `quic.EarlyListener` here, and pass it to `serveImpl` directly, instead of passing this massive constructor callback?

marten-seemann (commented):
This already looks very good to me. Thank you for working on this!

Two things:
1. See my comment below regarding avoiding the massive callback?
2. Can we have tests? ;) We definitely need a unit test in this package. Ideally, we'll also have an integration test in the `integrationtests` package, where we swap out the server (basically the thing that Caddy will do).

http3/server.go
renbou: I decided to use a callback here since we only construct the `quic.EarlyListener` after a couple of if's and other statements. Moving that code to a different method like `Server.preServe` which should be called here and in `Server.ServeListener` (basically everywhere before the actual `serveImpl`) also doesn't seem like the best way, at least to me. What do you think?

marten-seemann: Fair enough, let's keep it this way.

marten-seemann (commented):
LGTM so far, but there seems to be a race condition in one of the tests.

renbou:
The tests are fixed now, doesn't seem like the CircleCI fail is related to my changes

marten-seemann:
Thank you @renbou!

PR #3361, opened 2022-03-26
rename the Session to Connection

marten-seemann:
The term "session" is a remnant from gQUIC days. The QUIC RFC doesn't talk about "sessions", only about connections.

Note that this will be a quite disruptive change for downstream users.


lucas-clemente (approved):
Do we want any disambiguation from all the conn_* files? With this, we use "conn" for both classes of connections we have. (Maybe rename the conn_* stuff to rawConn, udpConn, sysConn?)

PR #3360, opened 2022-03-25
add support for serializing Extended CONNECT requests

marten-seemann:

PR #3362, opened 2022-03-27
extend the HTTP/3 API for WebTransport support

marten-seemann:
Builds on top of #3360. Only provides support for bidirectional streams at the moment.

To support WebTransport, we need to be able to "hijack" a stream after reading a HTTP/3 frame of type 0x41. The WebTransport implementation then takes over the stream and writes on the QUIC stream directly.

We also need to be able to "hijack" the connection: After we receive a WebTransport upgrade request from the client, we establish a WebTransport connection. On this connection, it is possible to open new streams (that's why we need the `http3.StreamCreator`).

A work-in-progress implementation of WebTransport has been started here: https://github.com/marten-seemann/webtransport-go.

cc @ydnar @adriancable

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -316,7 +316,7 @@ func (c *client) doRequest(
 			res.Header.Add(hf.Name, hf.Value)
 		}
 	}
-	respBody := newResponseBody(str, reqDone, func() {
+	respBody := newResponseBody(str, c.conn, reqDone, func() {
 		c.conn.CloseWithError(quic.ApplicationErrorCode(errorFrameUnexpected), "")
 	})
 
--- a/http3/client.go
+++ b/http3/client.go
@@ -42,6 +42,7 @@ type roundTripperOpts struct {
 	DisableCompression bool
 	EnableDatagram     bool
 	MaxHeaderBytes     int64
+	AdditionalSettings map[uint64]uint64
 }
 
 // client is a HTTP3 client doing requests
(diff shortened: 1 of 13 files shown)


adriancable:
@marten-seemann - this is great.

My implementation of webtransport in go (which your PR will, and should, replace) does implement all the other stream types including uni streams, so feel free to use any of my code if it's helpful.

https://github.com/adriancable/webtransport-go

marten-seemann:
Thank you @adriancable! The main reason that it doesn't have unidirectional stream support yet is that I first wanted to get the basic version running. There are no principal blockers.
We'll need yet another callback on the `http3.Server` and `http3.RoundTripper` that allows us to hijack unidirectional streams, as those are not marked as WebTransport streams by the frame type, but by the HTTP/3 stream type.

http3/client.go
```diff
@@ -136,6 +140,30 @@ func (c *client) setupConn() error {
...
+					return c.opts.StreamHijacker(ft, str)
+				})
+				if err == errHijacked {
+					return
+				}
+				if err != nil {
+					c.logger.Debugf("error handling stream: %s", err)
+				}
+				c.conn.CloseWithError(quic.ApplicationErrorCode(errorFrameUnexpected), "received HTTP/3 frame on bidirectional stream")
+			}
```
lucas-clemente, line 164: Do we need to return here to avoid an infinite loop?

lucas-clemente (approved):

PR #3367, opened 2022-04-02
stop using the deprecated net.Error.Temporary, update golangci-lint to v1.45.2

marten-seemann:


Mitumaru (commented):
Thank you very much

PR #3378, opened 2022-04-15
use IP_PMTUDISC_PROBE for setting the DF bit on Linux & Windows, add support for darwin & FreeBSD

tobyxdd:
Thank @database64128

I can confirm that this (still) works on Linux & Windows. But we probably need to have someone test it under macOS and FreeBSD.

### DIFF
--- a/sys_conn_df_windows.go
+++ b/sys_conn_df_windows.go
@@ -44,7 +44,7 @@ func setDF(rawConn syscall.RawConn) error {
 	case errDFIPv4 != nil && errDFIPv6 == nil:
 		utils.DefaultLogger.Debugf("Setting DF for IPv6.")
 	case errDFIPv4 != nil && errDFIPv6 != nil:
-		return errors.New("setting DF failed for both IPv4 and IPv6")
+		utils.DefaultLogger.Errorf("setting DF failed for both IPv4 and IPv6: errDFIPv4=%s, errDFIPv6=%s", errDFIPv4.Error(), errDFIPv6.Error())
 	}
 	return nil
 }
--- a/sys_conn_df_windows.go
+++ b/sys_conn_df_windows.go
@@ -13,17 +13,26 @@ import (
 
 const (
 	// same for both IPv4 and IPv6 on Windows
-	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IP_DONTFRAG.html
-	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IPV6_DONTFRAG.html
-	IP_DONTFRAGMENT = 14
-	IPV6_DONTFRAG   = 14
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IP_MTU_DISCOVER.html
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IPV6_MTU_DISCOVER.html
+	IP_MTU_DISCOVER   = 14
+	IPV6_MTU_DISCOVER = 14
+)
+
+// enum PMTUD_STATE from ws2ipdef.h
+const (
+	IP_PMTUDISC_NOT_SET = iota
+	IP_PMTUDISC_DO
+	IP_PMTUDISC_DONT
+	IP_PMTUDISC_PROBE
+	IP_PMTUDISC_MAX
 )
 
 func setDF(rawConn syscall.RawConn) error {
 	var errDFIPv4, errDFIPv6 error
 	if err := rawConn.Control(func(fd uintptr) {
-		errDFIPv4 = windows.SetsockoptInt(windows.Handle(fd), windows.IPPROTO_IP, IP_DONTFRAGMENT, 1)
-		errDFIPv6 = windows.SetsockoptInt(windows.Handle(fd), windows.IPPROTO_IPV6, IPV6_DONTFRAG, 1)
+		errDFIPv4 = windows.SetsockoptInt(windows.Handle(fd), windows.IPPROTO_IP, IP_MTU_DISCOVER, IP_PMTUDISC_PROBE)
+		errDFIPv6 = windows.SetsockoptInt(windows.Handle(fd), windows.IPPROTO_IPV6, IPV6_MTU_DISCOVER, IP_PMTUDISC_PROBE)
 	}); err != nil {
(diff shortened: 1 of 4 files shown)


database64128 (approved):
LGTM.

sys_conn_df_windows.go
```diff
@@ -13,17 +13,26 @@ import (
...
-	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IPV6_DONTFRAG.html
-	IP_DONTFRAGMENT = 14
-	IPV6_DONTFRAG   = 14
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IP_MTU_DISCOVER.html
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IPV6_MTU_DISCOVER.html
+	IP_MTU_DISCOVER   = 14
+	IPV6_MTU_DISCOVER = 14
+)
+
+// enum PMTUD_STATE from ws2ipdef.h
```
marten-seemann, line 22: Can we link to the source here?

marten-seemann (commented):
> But we probably need to have someone test it under macOS and FreeBSD.

What exactly do we need to do to test this? Is it sufficient to run a localhost <-> localhost transfer and run Wireshark to see if the DF bit is set?

sys_conn_df_windows.go
```diff
@@ -13,17 +13,26 @@ import (
...
 const (
 	// same for both IPv4 and IPv6 on Windows
-	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IP_DONTFRAG.html
-	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IPV6_DONTFRAG.html
-	IP_DONTFRAGMENT = 14
-	IPV6_DONTFRAG   = 14
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IP_MTU_DISCOVER.html
+	// https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/constant.IPV6_MTU_DISCOVER.html
+	IP_MTU_DISCOVER   = 14
+	IPV6_MTU_DISCOVER = 14
```
database64128, line 19: From `ws2ipdef.h`:

```c
#define IP_MTU_DISCOVER           71 // Set/get path MTU discover state.
```

```c
#define IPV6_MTU_DISCOVER           71 // Set/get path MTU discover state.
```

database64128 (changes requested):

sys_conn_df_windows.go
database64128: @marten-seemann `ws2ipdef.h` in Windows SDK is the source.

sys_conn_df_windows.go
tobyxdd: Oof. Sorry for the mistake

database64128 (approved):

tobyxdd:
> > But we probably need to have someone test it under macOS and FreeBSD.
> 
> What exactly do we need to do to test this? Is it sufficient to run a localhost <-> localhost transfer and run Wireshark to see if the DF bit is set?

I'm not sure if localhost works the same as real ones. I think the best way is to set up 2 machines (at least one of them should have a path MTU < MaxPacketBufferSize) and run some transfers. Use Wireshark to see the DF bit & packet size.

database64128:
> I'm not sure if localhost works the same as real ones. I think the best way is to set up 2 machines (at least one of them should have a path MTU < MaxPacketBufferSize) and run some transfers. Use Wireshark to see the DF bit & packet size.

Or we can just assume that it works, and deal with it in the future if someone found a problem and opened an issue. After all, the worst case scenario for macOS and FreeBSD is only like: These new socket options didn't work, and quic-go behaves exactly the same as before.

marten-seemann:
Just tested this PR between a OSX and an Ubuntu. None of the QUIC packets originating from OSX hat DF set, only the packets from Ubuntu did.

database64128:
I think it's better than nothing. At least merging this PR makes it easier for someone with macOS or FreeBSD experience to fix this in the future.

marten-seemann:
> I think it's better than nothing. At least merging this PR makes it easier for someone with macOS or FreeBSD experience to fix this in the future.

According to my test on OSX, it doesn't work on OSX. What does this PR achieve then?

database64128:
> What does this PR achieve then?

It also fixes the behavior on Linux and Windows by setting the correct `IP_MTU_DISCOVER` value.

tobyxdd:
> @database64128 @tobyxdd What shall we do with this PR?

I'll try to fix it on macOS. If it's not possible, I can drop macOS support, and at least we still added it for FreeBSD.

database64128:
> [return to IP_DONTFRAGMENT on Windows for compatibility](https://github.com/lucas-clemente/quic-go/pull/3378/commits/c3007f8c5d24ffbe693e2e76d302864517f6b21f)

`IP_PMTUDISC_PROBE` allows you to send UDP packets exceeding the system cached PMTU without returning `-EMSGSIZE`, which is the optimal behavior we want here. This is not the case with `IP_DONTFRAGMENT`. If you want compatibility with unsupported Windows versions, maybe simply fallback to `IP_DONTFRAGMENT` when setting `IP(V6)_MTU_DISCOVER` fails?

marten-seemann:
@tobyxdd Any updates?

tobyxdd:
> > [return to IP_DONTFRAGMENT on Windows for compatibility](https://github.com/lucas-clemente/quic-go/pull/3378/commits/c3007f8c5d24ffbe693e2e76d302864517f6b21f)
> 
> `IP_PMTUDISC_PROBE` allows you to send UDP packets exceeding the system cached PMTU without returning `-EMSGSIZE`, which is the optimal behavior we want here. This is not the case with `IP_DONTFRAGMENT`. If you want compatibility with unsupported Windows versions, maybe simply fallback to `IP_DONTFRAGMENT` when setting `IP(V6)_MTU_DISCOVER` fails?

I don't think the return value matters that much when we have `isMsgSizeErr` to check and ignore the error anyway.

Compatibility is a real concern as I got some [feedback](https://github.com/HyNetwork/hysteria/issues/327) immediately after switching to IP_MTU_DISCOVER in my fork.

marten-seemann:
I think this has been fixed in the mean time.

Issue #3376, opened 2022-04-13
INTERNAL_ERROR: tried to delete unknown outgoing stream


amulepeweichan:
I have a program where a server repeatedly opens streams to a client. Sometimes if the server closes the stream almost immediately after starting it, the client gets that error. For the client, Close() returns fine without error. That error is returned by the next call the client makes to AcceptStream(), and it's of type qerr.TransportError which is package internal so I'm not sure how to gracefully handle it.

marten-seemann:
That sounds like a quic-go bug. Which quic-go version are you using?

amulepeweichan:
I was using v0.26 but upgraded to v0.27 and got the same error. But it's maybe just me not understanding properly quic or this library. I figured more exactly what is triggering it. It started after I added io.Copy(ioutil.Discard, stream) after my calls to stream.Close to not get error about too many open streams (https://github.com/lucas-clemente/quic-go/issues/3371). But when the side calling OpenStream does that discard after close, the side that called AcceptStream sometimes gets this internal error on it's next call to AcceptStream. When only the side that called AcceptStream does the discard, everything works with no errors, neither too many streams nor this unknown outgoing stream error. However I'm a little worried that the side calling OpenStream if it doesn't read/discard the final EOF the stream stays around and memory usage will continually grow?

amulepeweichan:
https://gist.github.com/amulepeweichan/7a66a6c9b2637f9bb64cc43dd3d656f6
Here's a basic modified version of the quic-go example, it has client listen on tcp and forward the tcp connection to a new quic stream. In one terminal run:
`while true; do echo test | nc -N localhost 4040; done`
In another run that go program. Sometimes it will get stuck blocking on a connection, never receiving the close of the stream from the server, like:
```
$ go run echo.go 
2022/04/14 18:03:02 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
2022/04/14 18:03:02 Done reading from client
got stream:
test

closing
2022/04/14 18:03:02 Done reading from quic
2022/04/14 18:03:02 Done reading from client
got stream:
test

closing
```
If you understand why it blocks sometimes I'd appreciate it cause I have no idea, but anyway if it blocks just ctrl-C and rerun the go program and on the first try or after some tries you'll see the error:
```
$ go run echo.go 
[...]

PR #3380, opened 2022-04-15
protect against concurrent use of Stream.Read

marten-seemann:
Concurrent use of Read is not permitted (and doesn't make any sense), but sometimes people do it anyway.

In this case, it's easy and cheap enough to protect against misuse of the API, and means we don't have to debug failures that look like quic-go bugs but are really bugs in the user's code. See #3376 for an example.


AnomalRoil (approved):
Nice and easy solution, I like it.

lucas-clemente (approved):

PR #3381, opened 2022-04-15
protect against concurrent use of Stream.Write

marten-seemann:
Equivalent of #3380, but for the write side.


AnomalRoil (approved):
LGTM

lucas-clemente (approved):

PR #3393, opened 2022-04-25
Fix returning incorrect url.URL in HTTP/3 extended CONNECT method

hareku:
Currently, quic-go returns `url.URL.Path` that contains url query when a request is Extended CONNECT method.


marten-seemann:
Duplicate of #3388?

PR #3395, opened 2022-04-25
Expose quic server closed err

hareku:
To provide a way to check whether a returned error is expected or not from `quic.Listener` and `http3.Server`.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -255,9 +255,6 @@ func (s *Server) serveImpl(startListener func() (quic.EarlyListener, error)) err
 	for {
 		conn, err := ln.Accept(context.Background())
 		if err != nil {
-			if errors.Is(err, quic.ErrServerClosed) {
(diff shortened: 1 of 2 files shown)


http3/server.go
```diff
@@ -255,6 +255,9 @@ func (s *Server) serveImpl(startListener func() (quic.EarlyListener, error)) err
 	for {
 		conn, err := ln.Accept(context.Background())
 		if err != nil {
+			if errors.Is(err, quic.ErrServerClosed) {
+				return http.ErrServerClosed
```
marten-seemann, line 259: Why do you convert to the `http.ErrServerClosed` here? The HTTP server might not be closed at this point.

hareku: Yes, that's right. I reverted it, and there is no particular need to convert in order to reach the goal of this PR.

marten-seemann (approved):

PR #3397, opened 2022-05-01
do not embed http.Server in http3.Server

tie:
This change removes the embedded http.Server struct from http3.Server. It slightly changes some error behavior, in particular, it mandates TLSConfig for methods that create QUIC listeners.

Before this change, only Addr, TLSConfig, Handler and MaxHeaderBytes options were used from the http.Server. These are now defined directly in http3.Server with an improved documentation.

### DIFF
--- a/example/main.go
+++ b/example/main.go
@@ -185,7 +185,8 @@ func main() {
 				err = http3.ListenAndServe(bCap, certFile, keyFile, handler)
 			} else {
 				server := http3.Server{
-					Server:     &http.Server{Handler: handler, Addr: bCap},
+					Handler:    handler,
+					Addr:       bCap,
 					QuicConfig: quicConf,
 				}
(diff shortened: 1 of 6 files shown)


marten-seemann:
What's the motivation for this PR?

tie:
@marten-seemann IMO, the current implementation is misleading since most of the options in net/http.Server do not apply to http3, and if they do, they may have different meaning (e.g. http.Server.Addr defaults to ":http" for HTTP/1 and HTTP/2, but HTTP/3 uses ":https"). Before this change, a user had to read the implementation to find knobs that http3.Server supports (and uses from http.Server struct).

marten-seemann:
I think this makes sense. We don't really use most of the fields of the `http.Server`, so we probably shouldn't pretend that setting these values will make any difference.

marten-seemann (approved):

lucas-clemente (approved):

PR #3399, opened 2022-05-03
remove warning comments of stable implementation

hareku:
These were implemented 4 or 5 years ago, so they actually seem to be stable.

git blame: https://github.com/lucas-clemente/quic-go/blame/master/interface.go

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -127,7 +127,6 @@ type SendStream interface {
 	// The Context is canceled as soon as the write-side of the stream is closed.
 	// This happens when Close() or CancelWrite() is called, or when the peer
 	// cancels the read-side of their stream.
-	// Warning: This API should not be considered stable and might change soon.
 	Context() context.Context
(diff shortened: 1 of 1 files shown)


interface.go
```diff
@@ -184,11 +183,9 @@ type Connection interface {
 	// The error string will be sent to the peer.
 	CloseWithError(ApplicationErrorCode, string) error
 	// The context is cancelled when the connection is closed.
-	// Warning: This API should not be considered stable and might change soon.
 	Context() context.Context
 	// ConnectionState returns basic details about the QUIC connection.
 	// It blocks until the handshake completes.
-	// Warning: This API should not be considered stable and might change soon.
```
marten-seemann, line 191: Don't remove this one please.

marten-seemann (commented):
I'm ok with removing some of them, but the `ConnectionState` has undergone repeated change.

marten-seemann (approved):

PR #3401, opened 2022-05-05
allow hijacking quic.Stream

hareku:
Currently, there is no way to know if the hijacked stream is closed, so this PR makes hijackable quic.Stream.

Here is an example of webtransport server. The client side has a similar implementation.
```go
func (s *WebTransportServer) Upgrade(w http.ResponseWriter, r *http.Request) (*Conn, error) {
	// ....

	hijacker := w.(http3.Hijacker)
	qconn := hijacker.StreamCreator()
	ctx := hijacker.Stream().Context() // to notice the stream state for WebTransport session.
	return newWebTransportConn(ctx, qconn)
}
```

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -20,9 +20,10 @@ type StreamCreator interface {
 
 var _ StreamCreator = quic.Connection(nil)
 
-// A Hijacker allows hijacking of the stream creating part of a quic.Session from a http.Response.Body.
+// A Hijacker allows hijacking of the stream and stream creating part of a quic.Session from a http.Response.Body.
 // It is used by WebTransport to create WebTransport streams after a session has been established.
 type Hijacker interface {
+	Stream() quic.Stream
 	StreamCreator() StreamCreator
 }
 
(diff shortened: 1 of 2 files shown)


hareku:
I think adding `StreamContext()` instead of `Stream()` may also be good.

marten-seemann:
Why is this needed? There's no way the stream can be closed at that point.

hareku:
> Why is this needed? There's no way the stream can be closed at that point.

I think if the client side sends STOP_SENDING or RESET_STREAM frame, the stream can be closed. That's right?

marten-seemann:
Sending a STOP_SENDING doesn't terminate the WebTransport session, you need to actually close the stream.

hareku:
Of course, the stream will be closed after receiving RESET_STREAM.

Returning to the subject, this PR is intended to provide context.Context for here.
https://github.com/marten-seemann/webtransport-go/blob/dd506a6913d45f308548c985d4537b828ad00b0f/conn.go#L54
I would appreciate it if you could tell me how you planned to implement here.

marten-seemann:
I think we've resolved this problem by now.

PR #3404, opened 2022-05-10
optimization quicvarint.Read(): make the function more short

dchaofei:
Hello, I've read the [RFC](https://www.rfc-editor.org/rfc/rfc9000.html#sample-varint) and I think this will make quicvarint. The Read () function is more elegant and shorter, so I contributed the code.

If the code have what problem, can at any time to reply me, thank you.

### DIFF
--- a/quicvarint/varint.go
+++ b/quicvarint/varint.go
@@ -28,46 +28,19 @@ func Read(r io.ByteReader) (uint64, error) {
 		return 0, err
 	}
 	// the first two bits of the first byte encode the length
-	len := 1 << ((firstByte & 0xc0) >> 6)
-	b1 := firstByte & (0xff - 0xc0)
-	if len == 1 {
-		return uint64(b1), nil
+	length := 1 << ((firstByte & 0xc0) >> 6)
+	v := uint64(firstByte & (0xff - 0xc0))
+
+	// the first byte has been read
+	length--
+	for ; length > 0; length-- {
+		nextByte, err := r.ReadByte()
+		if err != nil {
+			return 0, err
+		}
+		v = (v << 8) + uint64(nextByte)
 	}
-	b2, err := r.ReadByte()
-	if err != nil {
-		return 0, err
-	}
-	if len == 2 {
-		return uint64(b2) + uint64(b1)<<8, nil
-	}
-	b3, err := r.ReadByte()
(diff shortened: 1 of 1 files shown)


marten-seemann:
Have you benchmarked this? I suspect that spawning a loop will be slower, but I'm happy to be convinced otherwise.

marten-seemann:
That's because your benchmark code is allocating, which is shadowing the actual effect of parsing. Better benchmark:
```go
var values [][]byte

var Sink uint64

func init() {
	for i := 0; i < 1000; i++ {
		var max int64
		// Tilt the distribution towards smaller lengths. They are more common.
		switch i % 9 {
		case 0, 1, 2:
			max = maxVarInt1
		case 3, 4, 5:
			max = maxVarInt2
		case 6, 7:
			max = maxVarInt4
		case 8:
			max = maxVarInt8
		}
		b := &bytes.Buffer{}
		Write(b, uint64(rand.Int63n(max)))
		values = append(values, b.Bytes())
	}
}

func BenchmarkReadUseFor(b *testing.B) {
	var v uint64
	r := bytes.NewReader(nil)
	for i := 0; i < b.N; i++ {
		r.Reset(values[i%len(values)])
		var err error
		v, err = ReadUseFor(r)
		if err != nil {
			b.Fatal(err)
		}
	}
	Sink = v
}

func BenchmarkReadNormal(b *testing.B) {
	var v uint64
	r := bytes.NewReader(nil)
	for i := 0; i < b.N; i++ {
		r.Reset(values[i%len(values)])
		var err error
		v, err = Read(r)
		if err != nil {
			b.Fatal(err)
		}
	}
	Sink = v
}
```

With that, the result is consistent:
```
❯ go test -bench . -run Benchmark -benchmem
goos: darwin
goarch: arm64
pkg: github.com/lucas-clemente/quic-go/quicvarint
BenchmarkReadUseFor-10          146965569                8.087 ns/op           0 B/op          0 allocs/op
BenchmarkReadNormal-10          156249838                7.681 ns/op           0 B/op          0 allocs/op
PASS
ok      github.com/lucas-clemente/quic-go/quicvarint    4.125s
```

marten-seemann:
Closing since the current version is more performant than the proposed change.

Issue #3405, opened 2022-05-10
'(*client).handleUnidirectionalStreams' can hang on `AcceptStream` for pinned up connections


upsampled:
I am trying to experiment with leaving a QUIC connection pinned up for as long as possible and running multiple http3 requests over it. Pretty easy to implement for a client:

```go
return &http3.RoundTripper{
                ...
		Dial: func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error) {
			return PinnedUpConn, nil
		},
		...
	}
```
The main issue I am having is a ton of orphaned AcceptStream

```go
func (c *client) dial(ctx context.Context) error {
...
   go c.handleUnidirectionalStreams()
...
}
...
func (c *client) handleUnidirectionalStreams() {
...
		str, err := c.conn.AcceptUniStream(context.Background())
```
Because a `context.Background` is sent to Accept, I cannot stop the accept goroutine unless a stream requests comes in or I close the connection.

marten-seemann:
What is a "squint connection" and what does it mean to be "pinned up"?

upsampled:
@marten-seemann sorry that was a typo, i meant QUIC connection.

'Pinned up' in this context means to keep alive for as long as possible. This is desirable if establishing the connection to the server is difficult (IE: [client and server behind a NAT](https://tailscale.com/blog/how-nat-traversal-works/)).

upsampled:
To better explain the #PR3406 ,

The `handleUniStream` eventually ends up getting to `(*incomingUniStreamsMap) AcceptStream(context.Context)` and blocking:

```go
func (m *incomingUniStreamsMap) AcceptStream(ctx context.Context) (receiveStreamI, error) {
...

	for {
		num = m.nextStreamToAccept
		if m.closeErr != nil {
			m.mutex.Unlock()
			return nil, m.closeErr
		}
		var ok bool
		entry, ok = m.streams[num]
		if ok {
			break
		}
		m.mutex.Unlock()
		select {
		case <-ctx.Done():
			return nil, ctx.Err()
		case <-m.newStreamChan:
		}
		m.mutex.Lock()
	}
...
}
```
There are essentially three exit conditions from here:

1. We already have a stream with the expected ID
2. `Connection.CloseWithError` is called - `m.closeErr` i set and `m.newStreamChan` is closed
3. The context is canceled

I need to investigate more as why condition 1 is not being met (will update ticket).

[...]

PR #3406, opened 2022-05-10
Allow the Accept*Streams functions to canceled by the `http.Request` context

upsampled:
Regarding #3405 

This change will allow the Accept*Stream to canceled by `http.Request` context. This will allow for QUIC connections to be used for many http3 requests without experiencing a goroutine leak.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -123,9 +123,9 @@ func (c *client) dial(ctx context.Context) error {
 	}()
 
 	if c.opts.StreamHijacker != nil {
-		go c.handleBidirectionalStreams()
+		go c.handleBidirectionalStreams(ctx)
 	}
-	go c.handleUnidirectionalStreams()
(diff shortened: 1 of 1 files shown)


http3/client.go
```diff
@@ -123,9 +123,9 @@ func (c *client) dial(ctx context.Context) error {
 	}()
 
 	if c.opts.StreamHijacker != nil {
-		go c.handleBidirectionalStreams()
+		go c.handleBidirectionalStreams(ctx)
 	}
-	go c.handleUnidirectionalStreams()
+	go c.handleUnidirectionalStreams(ctx)
```
marten-seemann, line 128: That's not correct. We can't stop unidirectional stream handling when the request is canceled.

marten-seemann (changes requested):

upsampled:
issue was with using multiple clients on the same connection

PR #3410, opened 2022-05-12
remove error return value from ComposeVersionNegotiation

dchaofei:
Fixes: https://github.com/lucas-clemente/quic-go/issues/3408

ComposeVersionNegotiation() function will never return an error, but the return value defines the error type.


marten-seemann (approved):
Thank you!

PR #3416, opened 2022-05-15
add a Config.AcceptConnection function to block connection attempts

Jessica-Xamaters:
Good day,

I'm trying to set it up for my wordpress page but somehow it doesn't work for me. The site is called https://www.xamaters.com.
I don't know if I'm asking on the right thread but it found a quic cloud.
I try to set it up via the plugin but also unsuccessfully.

I'm not writing me off either, so I'm asking here.
I apologize in advance if I opened the wrong thread. :)
Thank you :)


Jessica-Xamaters:
Hello,

why closed?

PR #3420, opened 2022-05-19
don't ignore errors that occur when the TLS ClientHello is generated

### LINKED ISSUE #3419: not every crypto/tls error generates an alert
Errors that occur during `makeClientHello` don't generate an alert, and are currently silently ignored by the crypto setup: https://github.com/lucas-clemente/quic-go/blob/60bbe92194a5d4d90bfb3f142cbab69b9cb51561/internal/handshake/crypto_setup.go#L302-L323

We should return the error the caller instead.

This bug is easy to reproduce by not setting an empty string in `qtls.Config.NextProtos`, but there are numerous other error paths.

marten-seemann:
Fixes #3419.

### DIFF
--- a/internal/handshake/crypto_setup.go
+++ b/internal/handshake/crypto_setup.go
@@ -113,7 +113,8 @@ type cryptoSetup struct {
 
 	zeroRTTParameters      *wire.TransportParameters
 	clientHelloWritten     bool
-	clientHelloWrittenChan chan<- *wire.TransportParameters
+	clientHelloWrittenChan chan struct{} // is closed as soon as the ClientHello is written
+	zeroRTTParametersChan  chan<- *wire.TransportParameters
 
 	rttStats *utils.RTTStats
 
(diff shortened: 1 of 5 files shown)


internal/handshake/crypto_setup.go
```diff
@@ -113,7 +113,8 @@ type cryptoSetup struct {
 
 	zeroRTTParameters      *wire.TransportParameters
 	clientHelloWritten     bool
-	clientHelloWrittenChan chan *wire.TransportParameters
+	clientHelloWrittenChan chan struct{} // is closed as soon as the ClientHello is written
```
xieyuschen, line 116: Is there some reasons to change the type from `chan *wire.TransportParameters` from `chan struct{}`?

marten-seemann: It was not changed. The channel was renamed to `zeroRTTParametersChan`, and a new `clientHelloWrittenChan` was added.

xieyuschen: > It was not changed. The channel was renamed to `zeroRTTParametersChan`, and a new `clientHelloWrittenChan` was added.

Ohh, thanks for pointing it out.

lucas-clemente (approved):

marten-seemann:
Wrong merge method. Will revert and re-merge.

PR #3422, opened 2022-05-19
add readme for example and remove the reference to internal code

### LINKED ISSUE #3421: Enhance the example
Currently, our example needs to be enhanced as quic is new to users and they do need detailed examples to startup. However, our examples have the following shortages:
- No documents and it cannot run with no flag set.  
This will hurt the new user as they cannot try by simple `go run` command. If they are quite interesting about it, they need to read the codes carefully.  
-  The example uses the code inside the `internal` folder. 
Our example helps users who aims to use our quic-go library. If we reference the internal folder, users are confused when they import our library.

xieyuschen:
Enhance our example by following changes, fixes #3421.
- Add readme with details
- Remove the references to the code inside the `internal` folder.


marten-seemann:
@xieyuschen, thank you for this PR. Unfortunately, we can't merge this contribution. The documentation changes are not up to our standards language-wise, I'd have to rewrite every single sentence if I wanted to accept this change.

xieyuschen:
> @xieyuschen, thank you for this PR. Unfortunately, we can't merge this contribution. The documentation changes are not up to our standards language-wise, I'd have to rewrite every single sentence if I wanted to accept this change.

Thanks for your feedback:)

xieyuschen:
> @xieyuschen, thank you for this PR. Unfortunately, we can't merge this contribution. The documentation changes are not up to our standards language-wise, I'd have to rewrite every single sentence if I wanted to accept this change.

So if the documents is messy now, how about the code?
I think that it's better for user to run our example without any flags.

(I would like to enhance the documents also, is it acceptable once it's up to the level wish?)

marten-seemann:
What's the problem? The flags are all explained:
```
❯ go run example/main.go -h
Usage of *****/main:
  -bind value
    	bind to
  -qlog
    	output a qlog (in the same directory)
  -tcp
    	also listen on TCP
  -v	verbose
  -www string
    	www data

xieyuschen:
> What's the problem? The flags are all explained:
> 
> ```
> 
> ❯ go run example/main.go -h
> 
> Usage of *****/main:
> 
>   -bind value
> 
>     	bind to
> 
>   -qlog
> 
>     	output a qlog (in the same directory)
> 
>   -tcp
> 
>     	also listen on TCP
> 
>   -v	verbose
> 
>   -www string
> 
>     	www data


> The flags are all explained

Agree with you, but shall we provide a default value if the user would like to run without set flag?
Such as a default host like(https://localhost:port) both on client and server side.

PR #3427, opened 2022-05-21
Add readme and remove internal reference

### LINKED ISSUE #3421: Enhance the example

xieyuschen:
Sorry for that the last document is lower that the language-wish, so I revised it and made a new edition.
Fixed #3421 
Thanks.


marten-seemann:
I still don't understand what motivates this change.

xieyuschen:
> I still don't understand what motivates this change.

My motivation is that make the example easier for the user so they can try the quic-go library and needn't to find some blogs to use our example.

For your confusion, I think it might be caused by the gap in our experience.
For me, as a newer to this library example, I spent some times to figure out where the cert files are stored, how to make a request by the client in example, how to set the destination, use `localhost:port` or `https://localhost:port` as the request destination.
For you with quite a lot of experiences, especially on the project here so it's quite clear to you.

Thanks for you to review this PR. Regards

xieyuschen:
> I still don't understand what motivates this change.

My motivation is that make the example easier for the user so they can try the quic-go library and needn't to find some blogs to use our example.

For your confusion, I think it might be caused by the gap in our experience.
For me, as a newer to this library example, I spent some times to figure out where the cert files are stored, how to make a request by the client in example, how to set the destination, use localhost:port or https://localhost:port as the request destination.
For you with quite a lot of experiences, especially on the project here so it's quite clear to you.

Thanks for you to review this PR. Regards

marten-seemann:
I agree that some things can be changed, but I disagree about the path chosen here.

I appreciate that you want to improve things, but adding a README generated by some automated translation tool isn't helpful. I'd have to completely rewrite it just to bring it up to a decent linguistic standard.

xieyuschen:
> I agree that some things can be changed, but I disagree about the path chosen here.
> 
> I appreciate that you want to improve things, but adding a README generated by some automated translation tool isn't helpful. I'd have to completely rewrite it just to bring it up to a decent linguistic standard.

Ok get it. It does need a person with decent English ability. Sorry for not so good at English.

PR #3428, opened 2022-05-22
make it possible to parse a varint at the end of a reader

marten-seemann:
An `io.Reader` can read into the buffer and return the `io.EOF` in the same call. In fact, that's how the `quic.Stream` is implemented.
In that case, we shouldn't return an error from `quicvarint.Read`, otherwise the caller won't be able to parse a varint sent just before a stream was closed.

### DIFF
--- a/quicvarint/io.go
+++ b/quicvarint/io.go
@@ -31,7 +31,10 @@ func NewReader(r io.Reader) Reader {
 
 func (r *byteReader) ReadByte() (byte, error) {
 	var b [1]byte
-	_, err := r.Reader.Read(b[:])
+	n, err := r.Reader.Read(b[:])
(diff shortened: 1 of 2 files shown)


quicvarint/io.go
```diff
@@ -31,7 +31,10 @@ func NewReader(r io.Reader) Reader {
 
 func (r *byteReader) ReadByte() (byte, error) {
 	var b [1]byte
-	_, err := r.Reader.Read(b[:])
+	n, err := r.Reader.Read(b[:])
```
MarcoPolo, line 34: This is obvious and good as is, but as a general note `ReadAtLeast` follows the same semantics here (not suggesting we change this here).

MarcoPolo (approved):
Looks good!

PR #3429, opened 2022-05-23
pass frame / stream type parsing errors to the hijacker callbacks

marten-seemann:
When a stream is reset, we might not have received the frame / stream type yet. The callback might be able to identify if it was a stream intended for that application by analyzing the stream reset error.

This will allow us to resolve https://github.com/marten-seemann/go-libp2p-webtransport/issues/3. The WebTransport implementation can determine if the stream reset error code is in the [WebTransport error code range](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-02.html#name-http-3-error-code-registrat).

@ydnar @hareku Thoughts?

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -43,8 +43,8 @@ type roundTripperOpts struct {
 	EnableDatagram     bool
 	MaxHeaderBytes     int64
 	AdditionalSettings map[uint64]uint64
-	StreamHijacker     func(FrameType, quic.Connection, quic.Stream) (hijacked bool, err error)
-	UniStreamHijacker  func(StreamType, quic.Connection, quic.ReceiveStream) (hijacked bool)
+	StreamHijacker     func(FrameType, quic.Connection, quic.Stream, error) (hijacked bool, err error)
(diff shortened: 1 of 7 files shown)


MarcoPolo (approved):
Granted, I'm not familiar with this codebase, this change seems relatively straightforward and it makes sense that a hijacker would have more context than the general library.

One question not specific to this PR but in general is that it seems like there are `UniStreamHijacker` tests but no matching `StreamHijacker` tests. Any specific reason?

hareku:
> One question not specific to this PR but in general is that it seems like there are UniStreamHijacker tests but no matching StreamHijacker tests. Any specific reason?

Please use https://github.com/lucas-clemente/quic-go/pull/3391 if you like.

marten-seemann:
@hareku Thank you. I had actually missed your PR. I've merged it now, and rebased this PR.

@MarcoPolo Thanks for pointing this is out in the review. This actually helped uncover a bug, which I fixed in 5cb2e8265cd2832b0d9b961ca48645c1ebd993b3.

PR #3439, opened 2022-05-31
Change "HTTP/3" to "HTTP/3.0".

aaronriekenberg:
See https://github.com/caddyserver/caddy/issues/4819 for context.

go's `http.ParseHTTPVersion` is unable to parse `HTTP/3`.  This causes issues when using caddy as a reverse proxy to a go fastcgi server using `net/http/fcgi`.

Found instances of "HTTP/3" to change using grep:

```
grep -r '"HTTP/3"' *
```


mholt:
Very nice, thanks! When should we update Caddy to get this patch?

marten-seemann:
@mholt Sorry for the late reply. We'll be releasing a new minor version (v0.28.0) in the next couple of days.

mholt:
No worries, and that's great, thanks!

PR #3441, opened 2022-06-04
add a OnStreamDone callback to quic.Config

marten-seemann:
In WebTransport we need to keep a map of streams belonging to one WebTransport session (multiple WebTransport sessions can share the same QUIC connection). This is a requirement, since we need to close all streams when a WebTransport session is closed.

Since we don't want to keep an unbounded map, we want to remove closed streams from that map. Unfortunately, this is not trivial. One way to implement this would be to observe when the user reads the EOF or a stream reset error from the stream (for the stream's read side), and when the stream's write side context is canceled (consuming one Go routine per stream). In addition to the huge Go routine resource requirement, this feels very brittle.

The alternative is the `quic.Config.OnStreamDone` callback. It is called as soon as the stream is removed from quic-go's internal streams map. In WebTransport, we'd use this callback to clean up finished streams from the the webtransport-go streams map.

I'm not in love with this solution, but I think it's cleaner than the approach described above. Happy for suggestions for alternatives.

### DIFF
--- a/config.go
+++ b/config.go
@@ -113,6 +113,7 @@ func populateConfig(config *Config) *Config {
 		AllowConnectionWindowIncrease:    config.AllowConnectionWindowIncrease,
 		MaxIncomingStreams:               maxIncomingStreams,
 		MaxIncomingUniStreams:            maxIncomingUniStreams,
+		OnStreamDone:                     config.OnStreamDone,
 		ConnectionIDLength:               config.ConnectionIDLength,
 		StatelessResetKey:                config.StatelessResetKey,
 		TokenStore:                       config.TokenStore,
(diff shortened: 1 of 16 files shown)


lucas-clemente:
Maybe wrapping the conn & streams could be an alternative? That would allow you to intercept the close error without requiring a coroutine?

lucas-clemente:
You'd see that when wrapping the Read() method. That's assuming the application will read data reasonably quickly. Not sure if that's a valid assumption?

marten-seemann:
We're now tracking the error code in WebTransport: https://github.com/marten-seemann/webtransport-go/pull/10.

PR #3442, opened 2022-06-04
Ability to strictly prioritize streams, defaulting to round-robin

kixelated:
Only applies when the congestion window is full. This gives applications a way to specify which streams should be sent first under poor network conditions.

### DIFF
--- a/framer.go
+++ b/framer.go
@@ -2,6 +2,7 @@ package quic
 
 import (
 	"errors"
+	"sort"
 	"sync"
 
 	"github.com/lucas-clemente/quic-go/internal/ackhandler"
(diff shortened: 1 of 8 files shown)


kixelated:
I didn't actually meant to submit a PR upstream; just to my fork. Oh well.

`Priority` could be replaced with `Order` or `Precedence`, seeing as it's an often overloaded term. There's a sort that could be removed with a bit of a refactoring but I suppose it doesn't matter.

marten-seemann:
> Only applies when the congestion window is full.

Why?

kixelated:
> > Only applies when the congestion window is full.
> 
> Why?

Or when rate limited by a pacer.

I'm currently working on sample code for [a new live video protocol](https://www.ietf.org/archive/id/draft-lcurley-warp-00.html). The idea is that we prioritize newer media during congestion, skipping over old media instead of buffering. This either requires cancelling streams when we're network bound or more elegantly, prioritizing streams. The lowest priority streams will be starved until the network recovers or they are cancelled.

This code review adds the ability to specify which streams take strict priority instead of using round-robin. A future implementation might use weights instead.

PR #3444, opened 2022-06-06
make the keep alive interval configurable

nmldiegues:
So far the `KeepAlive` property for `quic.Config` allowed only to enable Ping send on inactive `quic.Connection`. That would happen half-way through the `MaxIdleTimeout`.

With this change, we make the configuration more flexible, by turning that property into `KeepAlivePeriod`, which allows to define the period for sending Ping frame since last activity for that `quic.Connection`.

The value used is the smaller one among:
- `MaxIdleTimeout/2`
- `KeepAlivePeriod`
- `MaxKeepAliveInterval`

Context: We're using `quic-go` in https://github.com/cloudflare/cloudflared as a substrate for proxying multiplexed application traffic (e.g. TCP/UDP traffic).

Why: We've found situations where our `quic.Connection`s were closed due to no network activity and we believe they should not have, for which having more pings could have potentially have helped during the idle period.

This should not change behaviour for anyone not caring about keep alive.
For those that defined the property, this is a breaking change API-wise (albeit a trivial one to fix). If that's worrisome, we can keep the existing `KeepAlive` flag and add the new one on the side (where `KeepAlive` being true defaults to half the idle period unless the `KeepAlivePeriod` new flag is configured).

### DIFF
--- a/config.go
+++ b/config.go
@@ -105,7 +105,7 @@ func populateConfig(config *Config) *Config {
 		HandshakeIdleTimeout:             handshakeIdleTimeout,
 		MaxIdleTimeout:                   idleTimeout,
 		AcceptToken:                      config.AcceptToken,
-		KeepAlive:                        config.KeepAlive,
+		KeepAlivePeriod:                  config.KeepAlivePeriod,
 		InitialStreamReceiveWindow:       initialStreamReceiveWindow,
 		MaxStreamReceiveWindow:           maxStreamReceiveWindow,
 		InitialConnectionReceiveWindow:   initialConnectionReceiveWindow,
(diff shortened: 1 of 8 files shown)


marten-seemann:
> Why: We've found situations where our quic.Connections were closed due to no network activity and we believe they should not have, for which having more pings could have potentially have helped during the idle period.

This is interesting. What was the idle timeout value in those cases, and what keep alive are you planning to use instead?

This PR lgtm, but CI seems to be unhappy.

nmldiegues:
> This is interesting. What was the idle timeout value in those cases, and what keep alive are you planning to use instead?

We have 15 sec idle timeout (so Ping at 7.5sec of inactivity).
The plan is to make it 5 sec idle timeout, and ping every 1sec of inactivity.


> This PR lgtm, but CI seems to be unhappy.

Ups, I ran the integration tests, but forgot to run the unit tests. Found the culprit and pushed fix.

PR #3448, opened 2022-06-14
only set DF for sockets that can handle it

### LINKED ISSUE #3445: setsockopt fails to set IP_MTU_DISCOVER in SOCK_SEQPACKET
**Problem:**
Set up a QUIC client that dials to a server via a Unix Sequenced Packet socket (SOCK_SEQPACKET)
That triggers this error: https://github.com/lucas-clemente/quic-go/blob/master/sys_conn_df_linux.go#L32
`setting DF failed for both IPv4 and IPv6`

**Reproduce:**
Run this command to set up a SOCK_SEQPACKET: `socat -v UNIX-LISTEN:/tmp/uds,socktype=5 -`

Then run this main:
```
package main

import (
	"crypto/tls"
	"net"

)

func main() {
	remoteAddr := &net.UnixAddr{
		Net:  "unixpacket",
		Name: "/tmp/uds",
	}
	conn, err := net.DialUnix("unixpacket", nil, remoteAddr)
	if err != nil {
		panic(err)
	}

	tlsClientConfig := &tls.Config{
		InsecureSkipVerify: true,
		NextProtos:         []string{"h3"},
	}
	quicConfig := &quic.Config{}
	_, err = quic.DialEarly(conn, conn.LocalAddr(), "example.com", tlsClientConfig, quicConfig)
	if err != nil {
		panic(err)
	}
}
```

That will fail with the above mentioned error.

This used to work in v0.24 but stopped working ever since: https://github.com/lucas-clemente/quic-go/pull/3328/files#diff-446dcd03f343ca7249e090d11da194b71c2524d2c2ed49540c9e2085c1236291

Should we not fail hard in that case?

nmldiegues:
Fixes #3445

### DIFF
--- a/sys_conn.go
+++ b/sys_conn.go
@@ -30,9 +30,13 @@ func wrapConn(pc net.PacketConn) (rawConn, error) {
 		if err != nil {
 			return nil, err
 		}
-		err = setDF(rawConn)
-		if err != nil {
-			return nil, err
+
+		if _, ok := pc.LocalAddr().(*net.UDPAddr); ok {
+			// Only set DF on sockets that we expect to be able to handle that configuration.
+			err = setDF(rawConn)
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
 	c, ok := pc.(OOBCapablePacketConn)


marten-seemann (approved):
Thank you!

PR #3452, opened 2022-06-17
add support for providing a custom Connection ID generator via Config

joliveirinha:
This work makes it possible for servers or clients to control how
ConnectionIDs are generated, which in turn will force peers in the
connection to use those ConnectionIDs as destination connection IDs  when sending packets.

This is useful for scenarios where we want to perform some kind
selection on the QUIC packets at the L4 level.


marten-seemann (commented):
Is this PR sufficient so that one could implement [QUIC-LB](https://datatracker.ietf.org/doc/html/draft-ietf-quic-load-balancers) using this config option?

mock_connection_id_generator_test.go
```diff
@@ -0,0 +1,15 @@
+package quic
```
marten-seemann, line 1: By convention, the `mock_xxx_test.go` files are all auto-generated and use gomock. Can you move this to the test files please?

joliveirinha: yes. missed that. moved.

joliveirinha:
> Is this PR sufficient so that one could implement [QUIC-LB](https://datatracker.ietf.org/doc/html/draft-ietf-quic-load-balancers) using this config option?

This PR doesn't make any assumption on the scheme used for the ConnectionIDs, so yes, from my understanding of that draft, one could provide a custom generator to the Config that follows the QUIC LB connection ID scheme.

joliveirinha:
I have updated the PR with more doc and the suggested improvements. 
The build for go1.18 failed but seems a flake. I am unable to re-run, tho

interface.go
```diff
@@ -211,6 +211,18 @@ type EarlyConnection interface {
 	NextConnection() Connection
 }
 
+// A ConnectionIDGenerator is a interface that allows clients to implement their own format
+// for the Connection IDs that servers/clients use as SrcConnectionID in QUIC packets.
+type ConnectionIDGenerator interface {
+
+	// GenerateConnectionID generates a new ConnectionID.
```
marten-seemann, line 218:
```suggestion
	// GenerateConnectionID generates a new Connection ID.
```

Can you also add some text to clarify that Connection IDs need to unique, and must not allow an observer to correlate them.

interface.go
```diff
@@ -211,6 +211,18 @@ type EarlyConnection interface {
...
+	GenerateConnectionID() ([]byte, error)
```
marten-seemann, line 219: Is there any use case where you'd want to know which connection you're generating the Connection ID for? With the current interface you only know that you're generating a CID for _any_ connection the server handling.

interface.go
```diff
@@ -211,6 +211,18 @@ type EarlyConnection interface {
...
+
+	// ConnectionIDLen tells what is the length of the ConnectionIDs generated by the implementation of
+	// this interface.
```
marten-seemann, line 222: Can you add some more information how this is used, especially around the use of 0-length CIDs?

interface.go
joliveirinha: Initially I taught about extending the API to allow that. But I don't have a use-case for that now, and doing so it would just open the door for creating an implementation that would potentially introduce the correlation of connection IDs that shouldn't exist.

that's why I kept it like that.

joliveirinha:
@marten-seemann is this good to go, or do you want to address anything extra?

marten-seemann:
Sorry for the long delay!

If I understand correctly, a `ConnectionIDGenerator` needs to generate CIDs with constant length, right? It's not possible to encode the length into the first few bits of the CID, and then parse it when receiving a packet? If that's a requirement, this should probably be documented on the interface.

> This is useful for scenarios where we want to perform some kind selection on the QUIC packets at the L4 level.

Just to help me understand, can you explain a bit more about your use case?

joliveirinha:
> If I understand correctly, a ConnectionIDGenerator needs to generate CIDs with constant length, right? 
Yes. In this case I did that because the whole code already had that assumption. I didn't changed that assumption, and just made it in the interface. 
To be honest, I don't know if RFC states it is possible for QUIC connections to re-issue CID with different lengths during their lifetime. And if it does, what would the benefit be at this point.

So, in my opinion, unless there is a valid use-case, making that change would require changing other parts of the code unrelated to this feature. At least as far as my understanding of the code goes.


> It's not possible to encode the length into the first few bits of the CID, and then parse it when receiving a packet?
That would be possible yes, but that is already an implementation specification of the exact format.
here, this approach allows implementations of quic-lb or others that may not encode that info. 

For instance, in our use-case, our format doesn't do that.
It's very similar to QUIC-LB, in the sense it encodes server information but doesn't follow exactly that format.


> If that's a requirement, this should probably be documented on the interface.
yes. can do that.


> Just to help me understand, can you explain a bit more about your use case?
sure.

Basically we have a datacenter with servers using the same IP (anycast). So connections from clients have as dstIP the same anycast IP.
In front of the DC we have routes that use ECMP. These route packets based on the 5-tuple hash of the packet and the pool of servers of the DC. 
If the pool of servers change, then packets will be diverted to a different server and connection broken. So we need something that is smarter and looks into the QUIC DCID and routes packets to the right server.
That is why we need this.

marten-seemann:
Thank you for your explanation, that makes a lot of sense to me.

> So, in my opinion, unless there is a valid use-case, making that change would require changing other parts of the code unrelated to this feature. At least as far as my understanding of the code goes.

Indeed, one would need to call the generator when parsing packet headers. I'm ok with not implementing this (we can still add it later, when somebody makes a compelling case), as long as we're really clear about the API contract of the generator. So yes, a clarification of the documentation there would be appreciated.

joliveirinha:
Thanks. Done, updated doc to be more clear.

marten-seemann:
@joliveirinha The linter is unhappy. Can you please fix?

marten-seemann:
I was trying to add an integration test to this PR, but I don't permissions to push to the branch. Maybe you could apply this commit: https://github.com/lucas-clemente/quic-go/commit/6522274f9e2d7737f8b685fe9fc651f4ad5eb735?

joliveirinha:
@marten-seemann done.

marten-seemann:
@joliveirinha The linter is still unhappy with the PR.

joliveirinha:
you are right. missed that! should be fine now.

marten-seemann (approved):

marten-seemann:
@joliveirinha The linter is still unhappy :( 
I know this is annoying. Thank you for your patience and for helping push this PR over the finish line!

joliveirinha:
No problem. I think now it is ok :)

maybe we should have a makefile that runs these tests locally.

marten-seemann (approved):

marten-seemann:
> maybe we should have a makefile that runs these tests locally.

Unfortunately, it's not that easy. The CI config specifies a certain version of golangci-lint, and test failure will depend on that. Would be great if there was a way to run a GitHub Actions workflow locally, but as far as I know GitHub doesn't provide an easy way to do that.

PR #3459, opened 2022-06-22
merge select blocks, close #3458

### LINKED ISSUE #3458: Are there reasons to use two select blocks in baseServer run method
There are two select blocks in the [baseServer run method](https://github.com/lucas-clemente/quic-go/blob/706a482340141f8edb26fefae993e96b1581b034/server.go#L231), could I ask that are there some reasons such as reserve a position to implement a new feature?

Is it possible to merge those two select blocks as this?
```go
func (s *baseServer) run() {
	defer close(s.running)
	for {
		select {
		case <-s.errorChan:
			return
		case p := <-s.receivedPackets:
			if bufferStillInUse := s.handlePacketImpl(p); !bufferStillInUse {
				p.buffer.Release()
			}
                default:
		}
	}
}
```

Thanks.

xieyuschen:


xieyuschen:
> 

If the `case p := <-s.receivedPackets:` is selected, the loop will return in the next turn if there is an error during packet handling. I'm not sure what's the `error to be selected even if there are received packets` means as we cannot capture the error when executing the handler for receiving packets. 

Thanks.

PR #3461, opened 2022-06-23
dedupe Alt-Svc header values

mattrobenolt:
Addresses https://github.com/lucas-clemente/quic-go/pull/3432#issuecomment-1164796769

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -325,10 +325,16 @@ func (s *Server) generateAltSvcHeader() {
 	if s.QuicConfig != nil && len(s.QuicConfig.Versions) > 0 {
 		supportedVersions = s.QuicConfig.Versions
 	}
+
+	// keep track of which have been seen so we don't yield duplicate values
+	seen := make(map[string]struct{}, len(supportedVersions))
 	var versionStrings []string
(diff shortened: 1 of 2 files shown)


http3/server.go
```diff
@@ -313,6 +313,15 @@ func extractPort(addr string) (int, error) {
 	return portInt, nil
 }
 
+func findString(haystack []string, needle string) int {
```
marten-seemann, line 316: You're not using the `int` return value. What about `contains([]string, string) bool` instead?

marten-seemann: Or just inline the entire function, then we don't need to think about naming ;)

marten-seemann (commented):
Very nice find! Thank you for the PR, just a minor stylistic comment.

marten-seemann:
Could you also add a test case for this please?

marten-seemann:
@mattrobenolt We're planning to cut a new quic-go release really soon. Would be great if we could get your patch in. Will you be able to incorporate the comments?

mattrobenolt:
@marten-seemann test added, should be good to go.

marten-seemann (approved):
Thanks!

PR #3467, opened 2022-07-01
optimize FirstOutstanding in the sent packet history

### LINKED ISSUE #3462: Optimizing (*sentPacketHistory).FirstOutstanding
Hi @marten-seemann,

quic-go currently spends a lot of CPU time on `(*sentPacketHistory).FirstOutstanding` when transmitting at high speed over a high packet loss connection. (Of course, usually this can't happen at all, as the built-in CUBIC congestion control slows the connection down to a crawl when packet loss rate is high. I have replaced it with a custom constant-speed CC.)

The following call graph was generated by a transmission test between 2 servers ($5 AWS Lightsail) @ 200 Mbps, simulated 30ms, 4% packet loss link

profile001

image

```
Showing nodes accounting for 3080ms, 28.33% of 10870ms total
----------------------------------------------------------+-------------
      flat  flat%   sum%        cum   cum%   calls calls% + context
----------------------------------------------------------+-------------
                                            1640ms 53.25% |   github.com/lucas-clemente/quic-go/internal/ackhandler.(*sentPacketHistory).HasOutstandingPackets (inline)
                                            1440ms 46.75% |   github.com/lucas-clemente/quic-go/internal/ackhandler.(*sentPacketHandler).PeekPacketNumber (inline)
    2490ms 22.91% 22.91%     3080ms 28.33%                | github.com/lucas-clemente/quic-go/internal/ackhandler.(*sentPacketHistory).FirstOutstanding
                                             590ms 19.16% |   github.com/lucas-clemente/quic-go/internal/ackhandler.(*PacketElement).Next (inline)
----------------------------------------------------------+-------------
                                             590ms   100% |   github.com/lucas-clemente/quic-go/internal/ackhandler.(*sentPacketHistory).FirstOutstanding (inline)
     590ms  5.43% 28.33%      590ms  5.43%                | github.com/lucas-clemente/quic-go/internal/ackhandler.(*PacketElement).Next
----------------------------------------------------------+-------------
```

`FirstOutstanding` was taking up nearly 30% of the runtime and was a major CPU bottleneck. The program occupied the entire CPU core, couldn't get faster than a merely 100 Mbps because of it. And it gets worse as I crank up the loss rate. Switching to servers with better CPU performance does improve the situation - I was able to hit ~700 Mbps between two Ryzen 7 machines over the same link.

Do you have any thoughts on what could be done to optimize this part of the code? Perhaps we can add some kind of cache, or pre-calculate the first outstanding packet every time the linked list is modified, instead of iterating it every call?

pprof sample attached:
[pprof.hy_linux_amd64.samples.cpu.002.pb.gz](https://github.com/lucas-clemente/quic-go/files/9006604/pprof.hy_linux_amd64.samples.cpu.002.pb.gz)

tobyxdd:
Closes #3462

### DIFF
--- a/internal/ackhandler/sent_packet_history.go
+++ b/internal/ackhandler/sent_packet_history.go
@@ -132,6 +132,8 @@ func (h *sentPacketHistory) DeclareLost(p *Packet) *Packet {
 	if !ok {
 		return nil
 	}
+	// try to remove it from both lists, as we don't know which one it currently belongs to.
+	// Remove is a no-op for elements that are not in the list.
 	h.outstandingPacketList.Remove(el)
 	h.etcPacketList.Remove(el)
 	p.declaredLost = true
--- a/internal/ackhandler/sent_packet_history.go
+++ b/internal/ackhandler/sent_packet_history.go
@@ -9,18 +9,20 @@ import (
 )
 
 type sentPacketHistory struct {
-	rttStats    *utils.RTTStats
-	packetList  *PacketList
-	packetMap   map[protocol.PacketNumber]*PacketElement
-	highestSent protocol.PacketNumber
+	rttStats              *utils.RTTStats
+	outstandingPacketList *PacketList
+	etcPacketList         *PacketList
+	packetMap             map[protocol.PacketNumber]*PacketElement
+	highestSent           protocol.PacketNumber
 }
 
 func newSentPacketHistory(rttStats *utils.RTTStats) *sentPacketHistory {
 	return &sentPacketHistory{
-		rttStats:    rttStats,
-		packetList:  NewPacketList(),
-		packetMap:   make(map[protocol.PacketNumber]*PacketElement),
-		highestSent: protocol.InvalidPacketNumber,
+		rttStats:              rttStats,
+		outstandingPacketList: NewPacketList(),
+		etcPacketList:         NewPacketList(),
+		packetMap:             make(map[protocol.PacketNumber]*PacketElement),
+		highestSent:           protocol.InvalidPacketNumber,
 	}
 }
 
@@ -30,7 +32,7 @@ func (h *sentPacketHistory) SentPacket(p *Packet, isAckEliciting bool) {
 	}
 	// Skipped packet numbers.
 	for pn := h.highestSent + 1; pn < p.PacketNumber; pn++ {
-		el := h.packetList.PushBack(Packet{
+		el := h.etcPacketList.PushBack(Packet{
 			PacketNumber:    pn,
 			EncryptionLevel: p.EncryptionLevel,
 			SendTime:        p.SendTime,
@@ -41,7 +43,12 @@ func (h *sentPacketHistory) SentPacket(p *Packet, isAckEliciting bool) {
 	h.highestSent = p.PacketNumber
 
 	if isAckEliciting {
-		el := h.packetList.PushBack(*p)
+		var el *PacketElement
+		if p.outstanding() {
+			el = h.outstandingPacketList.PushBack(*p)
+		} else {
+			el = h.etcPacketList.PushBack(*p)
+		}
 		h.packetMap[p.PacketNumber] = el
 	}
 }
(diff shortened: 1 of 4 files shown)


tobyxdd:
Working on the integration test failures

tobyxdd:
@marten-seemann I need some help 🤨 still haven't figured out the cause yet

tobyxdd:
nvm, found the problem.

internal/ackhandler/sent_packet_history.go
```diff
@@ -81,28 +101,49 @@ func (h *sentPacketHistory) Remove(p protocol.PacketNumber) error {
...
 func (h *sentPacketHistory) HasOutstandingPackets() bool {
-	return h.FirstOutstanding() != nil
+	return h.outstandingPacketList.Len() > 0
 }
 
 func (h *sentPacketHistory) DeleteOldPackets(now time.Time) {
 	maxAge := 3 * h.rttStats.PTO(false)
 	var nextEl *PacketElement
-	for el := h.packetList.Front(); el != nil; el = nextEl {
+	for el := h.etcPacketList.Front(); el != nil; el = nextEl {
```
marten-seemann, line 117: Can you add a comment here why we don't have to consider the `oustandingPacketList`?

internal/ackhandler/sent_packet_history.go
```diff
@@ -81,28 +101,49 @@ func (h *sentPacketHistory) Remove(p protocol.PacketNumber) error {
...
+	}
+}
+
+func (h *sentPacketHistory) DeclareLost(p *Packet) *Packet {
+	el, ok := h.packetMap[p.PacketNumber]
+	if !ok {
+		return nil
+	}
+	h.outstandingPacketList.Remove(el)
+	h.etcPacketList.Remove(el)
```
marten-seemann, line 134: Removing from both lists doesn't seem very clean. Should we save in the map which list we've stored it in?

internal/ackhandler/sent_packet_history.go
```diff
@@ -61,15 +82,14 @@ func (h *sentPacketHistory) Iterate(cb func(*Packet) (cont bool, err error)) err
...
 func (h *sentPacketHistory) FirstOutstanding() *Packet {
-	for el := h.packetList.Front(); el != nil; el = el.Next() {
-		p := &el.Value
-		if !p.declaredLost && !p.skippedPacket && !p.IsPathMTUProbePacket {
-			return p
-		}
+	el := h.outstandingPacketList.Front()
+	if el == nil {
+		return nil
+	} else {
```
marten-seemann, line 90: Same here.

marten-seemann (commented):
Thank you! This looks pretty good.

It would probably be good to have a test that `Iterate` iterates in the right order, when skipping between the two lists.

internal/ackhandler/sent_packet_history.go
tobyxdd: `Remove` already checks whether `e.list == l`. Do we really need to add an extra step?

tobyxdd:
> It would probably be good to have a test that `Iterate` iterates in the right order, when skipping between the two lists.

I think the tests involving `expectInPacketHistory` already cover it.

tobyxdd:
> It would probably be good to have a test that `Iterate` iterates in the right order, when skipping between the two lists.

done

tobyxdd:
@marten-seemann the latest commit https://github.com/lucas-clemente/quic-go/pull/3467/commits/7cedeed64a5bd52ac00e84249b33c8415b26a061 tries to solve https://github.com/lucas-clemente/quic-go/issues/300 using AVL tree for gaps.

I've already done most of the work but there's a nasty bug somewhere in handling the start/end gaps causing the tests to fail. I spent the entire night trying to debug without any luck... maybe you can take a look?

tobyxdd:
I've found the problem and fixed it. Can confirm that it works perfectly in real-world applications too.

With these 2 changes I can get a stable 400-500 Mbps on $5 AWS Lightsail - more than quadrupling the performance.

> Can you please separate this into a separate branch / PR though?

Maybe we can just put them together as one optimization PR?

marten-seemann:
Where’s that tree implementation coming from? I’m not convinced that quic-go should be in the business of implementing a tree data structure from scratch.

marten-seemann:
> Or we can just put them together as one big optimization PR?

These are two completely different issues. They need to be resolved in separate PRs.

tobyxdd:
> Where’s that tree implementation coming from? I’m not convinced that quic-go should be in the business of implementing a tree data structure from scratch.

It's from https://github.com/ross-oreto/go-tree/blob/master/btree.go but I've made various modifications to meet our use case

tobyxdd:
frameSorter changes removed

internal/ackhandler/sent_packet_history.go
marten-seemann: Ok, given that this introduces a bunch of complexity, we probably don't need it.

Can you add a comment that `Remove` checks if the element is contained in the list, and is a no-op if it isn't.

marten-seemann (approved):
This looks good, thanks for your work @tobyxdd!

One tiny request for a comment, then we're ready to merge.

marten-seemann (approved):
Thank you!

PR #3493, opened 2022-08-03
Fill http.Request.TLS with TLS ConnectionState

### LINKED ISSUE #2879: Cannot access TLS ConnectionState from a http.Request

superfashi:
Fix #2879

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -552,6 +552,8 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 		return newStreamError(errorGeneralProtocolError, err)
 	}
 
+	// If handshake is already completed, we can safely acquire TLS connection state
+	// without breaking 0-RTT.
 	if conn.(quic.EarlyConnection).HandshakeComplete().Err() != nil {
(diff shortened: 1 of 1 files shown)


http3/server.go
```diff
@@ -552,6 +552,10 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 		return newStreamError(errorGeneralProtocolError, err)
 	}
 
+	if conn.(quic.EarlyConnection).HandshakeComplete().Err() != nil {
```
marten-seemann, line 555: Can you please add a comment here?

marten-seemann (changes requested):

http3/server.go
superfashi: done

PR #3498, opened 2022-08-10
add a function to distinguish between long and short header packets

marten-seemann:

Logically, this function belongs into the `wire` package.

### DIFF
--- a/integrationtests/self/datagram_test.go
+++ b/integrationtests/self/datagram_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lucas-clemente/quic-go"
 	quicproxy "github.com/lucas-clemente/quic-go/integrationtests/tools/proxy"
 	"github.com/lucas-clemente/quic-go/internal/protocol"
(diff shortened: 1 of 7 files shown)


integrationtests/self/datagram_test.go
```diff
@@ -73,7 +74,7 @@ var _ = Describe("Datagram test", func() {
 							return false
 						}
 						// don't drop Long Header packets
-						if packet[0]&0x80 == 1 {
```
lucas-clemente, line 76: Fwiw, the old version here was incorrect.

lucas-clemente (approved):

integrationtests/self/datagram_test.go
marten-seemann: I don't think so, this is selecting long header packets?

marten-seemann: Now I see it. It would have to be `&0x80 == 0x80`. Nice one.

PR #3501, opened 2022-08-12
implement a new API to let servers control client address verification

### LINKED ISSUE #3494: redesign the Retry Token API
Our current API (the `AcceptToken` config option) is confusing and error-prone. In particular:
1. It requires users to understand (in detail!) how QUIC's address validation works.
    * This includes the distinction between tokens sent in Retry packets and tokens sent in NEW_TOKEN frames.
    * This callback is called twice, when receiving the first packet on a new connection, and to validate the token.
1. It is possible to send more than a single Retry (which is forbidden by the RFC) by returning `false` multiple times for a connection attempt.
1. It requires users to reimplement the `defaultAcceptToken` logic, if all they want to do is require address validation when their application is under load.
1. It makes it impossible to know if the callback actually validated the address (in which case we can remove the 3x amplification limit), or just rubber-stamped the connection attempt (in which case we still need to apply the amplification limit). See https://github.com/lucas-clemente/quic-go/pull/3326#pullrequestreview-888657767 for details.

I suggest replacing the `AcceptToken` callback with 3 config options:
1. `RequireAddressValidation(net.Addr) bool`: if set and returns `true`, the server will perform a Retry. It is called:
    * for new connection attempts (Initial packet with empty token)
    * for Initial packets where the token cannot be decoded. This can happen if the server lost state (reboot) and can't decrypt the token any more.
    * for Initial packets with a token issued on a previous connection, if validation of that token fails
1. `MaxTokenAge time.Duration` (default: 24h): allows controlling the maximum age of a token issued on a previous connection. Most ISPs assign new IP addresses once a day, so 24h seems to be a reasonable default.
1. `MaxRetryTokenAge time.Duration` (default: 5s): allows controlling the maximum age of a token issued in a Retry packet. Such a token has to be used on the same connection attempt, so 5s seems to be a reasonable default.

In case the token is younger than the maximum age for the respective token type, the server would perform the following validation:
* if the address is a `net.UDPAddr` (this is the common case): validate that the IP (but not necessarily the port) from the token matches the current remote address
* if the address is not a `net.UDPAddr` (can happen if running on top of a custom `net.PacketConn`): validate that `addr.String()` from the token matches the current address

Open question: Is there any use case where a perfect match between the IP address in the token and the current IP address is not required? Are there cases where one would accept an entire address range as valid? What about IPv6?

marten-seemann:
Fixes #3494.

This PR makes 2 changes:
1. It implements the new API suggested in #3494.
2. It disables address verification by default. This means that, unless configured otherwise, the QUIC handshake will take a single RTT. This is in line with providing safe defaults, as we implement the 3x amplification  protection.

@mholt, would be happy about your feedback on this PR!

### DIFF
--- a/config.go
+++ b/config.go
@@ -46,7 +46,7 @@ func populateServerConfig(config *Config) *Config {
 		config.MaxRetryTokenAge = protocol.RetryTokenValidity
 	}
 	if config.RequireAddressValidation == nil {
-		config.RequireAddressValidation = func(net.Addr) bool { return true }
+		config.RequireAddressValidation = func(net.Addr) bool { return false }
 	}
 	return config
 }
--- a/config.go
+++ b/config.go
@@ -2,11 +2,11 @@ package quic
 
 import (
 	"errors"
+	"net"
 	"time"
 
-	"github.com/lucas-clemente/quic-go/internal/utils"
-
 	"github.com/lucas-clemente/quic-go/internal/protocol"
+	"github.com/lucas-clemente/quic-go/internal/utils"
 )
 
 // Clone clones a Config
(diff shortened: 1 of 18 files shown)


lucas-clemente (approved):

mholt:
This looks better!

So basically, if I understand right, I would implement `RequireAddressValidation(net.Addr) bool` to return `true` if the server is under load, correct? And I don't need to call any sort of `defaultAcceptToken()` function in addition?

marten-seemann:
> So basically, if I understand right, I would implement RequireAddressValidation(net.Addr) bool to return true if the server is under load, correct? And I don't need to call any sort of defaultAcceptToken() function in addition?

Exactly! All the token validation logic is now handled under the hood (tunable using the `MaxTokenAge` and `MaxRetryTokenAge`). All you have to do is say if you want to perform address validation in the first place (at the cost of 1 additional RTT during the handshake).

mholt (approved):
Sounds like just what I was hoping for then. Thank you!

mholt:
@marten-seemann Excellent! I updated https://github.com/caddyserver/caddy/pull/4707 and it seems to be working (haven't tested the "highLoad" scenario but HTTP/3 works by default).

PR #3507, opened 2022-08-17
fix datagram support detection

TNQOYxNU:
- treat zero MaxDatagramFrameSize as unsupported
- avoid send corresponding transport parameters when support is disabled
- report error when calling Send/ReceiveMessage and support is disabled

close #3505

PR #3511, opened 2022-08-19
fix datagram support detection

KevinZonda:
- treat zero MaxDatagramFrameSize as unsupported.
- avoid send corresponding transport parameters when support is disabled
- report error when calling Send/ReceiveMessage and support is disabled

Fixes https://github.com/lucas-clemente/quic-go/issues/3505.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -1977,7 +1977,7 @@ func (s *connection) onStreamCompleted(id protocol.StreamID) {
 }
 
 func (s *connection) SendMessage(p []byte) error {
-	if s.datagramQueue == nil || !s.supportsDatagrams() {
+	if !s.supportsDatagrams() {
 		return errors.New("datagram support disabled")
 	}
 
@@ -1991,7 +1991,7 @@ func (s *connection) SendMessage(p []byte) error {
 }
 
 func (s *connection) ReceiveMessage() ([]byte, error) {
-	if s.datagramQueue == nil || !s.config.EnableDatagrams {
+	if !s.config.EnableDatagrams {
 		return nil, errors.New("datagram support disabled")
 	}
 	return s.datagramQueue.Receive()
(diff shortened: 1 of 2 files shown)


TNQOYxNU:
This replace #3507

marten-seemann (commented):
It seems like the `datagramQueue` should be initialized when we receive the peer's transport parameters, not on setup, right?

TNQOYxNU:
> It seems like the `datagramQueue` should be initialized when we receive the peer's transport parameters, not on setup, right?

I'm neither familiar with code base nor with standard, so this PR doesn't address this problem, only add a nil check to avoid go panic.
To support "unidirectional datagram", it need to be initiated somewhere, initiate it on demand after peer's transport parameter is the best option.
Another option is always initiate it on setup regardless configuration, straightforward but may have some overhead.
We'd better also disable `Send/ReceiveMessage` on demand, improve doc, even redesign ConnectionState API to avoid confusion--it's two independent unidirectional channel.

TNQOYxNU:
Initiate `datagramQueue` after transport parameter (`connection.applyTransportParameters()`) not works, seems `newPacketPacker` need `datagramQueue` before transport parameter available?
Always initiate it in `connection.preSetup()` passed all test cases.

marten-seemann:
> Always initiate it in connection.preSetup() passed all test cases.

Sounds good to me. Then we can send datagrams, even if we don't support receiving datagrams. Can you update the PR?

connection.go
```diff
@@ -1979,7 +1977,7 @@ func (s *connection) onStreamCompleted(id protocol.StreamID) {
 }
 
 func (s *connection) SendMessage(p []byte) error {
-	if s.datagramQueue == nil {
+	if s.datagramQueue == nil || !s.supportsDatagrams() {
```
marten-seemann, line 1980: I think you can remove the `s.datagramQueue == nil` check. You initialize the `datagramQueue` in all cases.

marten-seemann (changes requested):

marten-seemann (approved):

PR #3512, opened 2022-08-20
add a logging.NullTracer and logging.NullConnectionTracer

marten-seemann:
The `NullTracer` and the `NullConnectionTracer` can be embedded, to make it easier to access just a few of the calls.

### DIFF
--- a/logging/null_tracer.go
+++ b/logging/null_tracer.go
@@ -1,17 +1,29 @@
 package logging
 
 import (
+	"context"
 	"net"
 	"time"
 )
 
-// The NullConnectionTracer is a ConnectionTracer that does nothing.
-// It is useful for embedding. Don't modify this variable!
-var NullConnectionTracer ConnectionTracer = &nullConnectionTracer{}
+var (
+	// The NullTracer is a Tracer that does nothing.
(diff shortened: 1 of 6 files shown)


logging/null_tracer.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+)
+
+var (
+	// The NullTracer is a Tracer that does nothing.
+	// It is useful for embedding. Don't modify this variable!
+	NullTracer Tracer = &nullTracer{}
+	// The NullConnectionTracer is a ConnectionTracer that does nothing.
+	// It is useful for embedding. Don't modify this variable!
+	NullConnectionTracer ConnectionTracer = &nullConnectionTracer{}
```
marten-seemann, line 15: Design decision: Should `NullConnectionTracer` be a variable, or should it be the type?

Advantage variable: There's only a single `NullConnectionTracer` instance.
Advantage type: makes it really easy to embed, which is not possible when it's a variable.

marten-seemann: Switched it to a type. This makes embedding easier. I don't think having a single instance is actually a problem.

marten-seemann:
Merging, since this needs to go into the next release.

PR #3524, opened 2022-08-25
write packets in batches

### LINKED ISSUE #2607: use batch read / writes, if available

### LINKED ISSUE #2877: Boosting performance: UDP Offloading and GSO
Hello, I've been instrumenting numerous QUIC implementations for performance testing.

In particular, I see [`quiche`] from CloudFlare implemented UDP offloading and GSO earlier this year. [[link]](https://blog.cloudflare.com/accelerating-udp-packet-transmission-for-quic/)
`quic-go` has given me 470.3 Mbps with loopback setting, so it seems to have no support for both accelerations yet. (I briefly looked up the code)

Would anyone confirm this for `quic-go`?
Maybe we could keep open this issue for future implementation, too.

marten-seemann:
Fixes #2607. Fixes #2877.

This required some work since while `recvmmsg` is only called from a single Go routine, `sendmmsg` is called concurrently if there are multiple QUIC connections running on the same socket.

### DIFF
--- a/client.go
+++ b/client.go
@@ -14,7 +14,7 @@ import (
 )
 
 type client struct {
-	sconn sendConn
+	sconn rawConn
 	// If the client is created with DialAddr, we create a packet conn.
 	// If it is started with Dial, we take a packet conn as a parameter.
 	createdPacketConn bool
@@ -194,7 +194,7 @@ func dialContext(
 	if err != nil {
 		return nil, err
 	}
-	c, err := newClient(pconn, remoteAddr, config, tlsConf, host, use0RTT, createdPacketConn)
+	c, err := newClient(pconn, config, tlsConf, host, use0RTT, createdPacketConn)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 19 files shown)


bt90:
Does this PR directly fix #2877? It thought that this would require a different API as described in #3563?

lucas-clemente (approved):

imkos:
When will this pr merge into the master?

marten-seemann:
We now use GSO (#3808). This should be a lot faster than `sendmmsg`.

PR #3525, opened 2022-08-27
reduce allocations of ackhandler.Packet

marten-seemann:
During a 1 GB transfer, quic-go allocates 500 MB on the sending side, as determined using the `allocs` function of pprof. Of those 500 MB, 140 MB are consumed for the `ackhandler.Packet`. With this change, we can bring this down to 25 MB.

We can easily solve this by:
1. storing a the `Packet` by pointer instead of by value in the linked lists used in the ackhandler
2. use a `sync.Pool` to reuse `Packet` structs

Before:
image

After:
image

### DIFF
--- a/internal/ackhandler/packet.go
+++ b/internal/ackhandler/packet.go
@@ -1,0 +1,49 @@
+package ackhandler
+
+import (
+	"sync"
+	"time"
+
+	"github.com/lucas-clemente/quic-go/internal/protocol"
+)
+
+// A Packet is a packet
+type Packet struct {
+	PacketNumber    protocol.PacketNumber
+	Frames          []Frame
+	LargestAcked    protocol.PacketNumber // InvalidPacketNumber if the packet doesn't contain an ACK
+	Length          protocol.ByteCount
+	EncryptionLevel protocol.EncryptionLevel
+	SendTime        time.Time
+
+	IsPathMTUProbePacket bool // We don't report the loss of Path MTU probe packets to the congestion controller.
+
+	includedInBytesInFlight bool
+	declaredLost            bool
+	skippedPacket           bool
+}
+
+func (p *Packet) outstanding() bool {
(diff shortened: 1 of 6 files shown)


internal/ackhandler/sent_packet_handler.go
```diff
@@ -334,7 +334,10 @@ func (h *sentPacketHandler) ReceivedAck(ack *wire.AckFrame, encLevel protocol.En
 			acked1RTTPacket = true
 		}
 		h.removeFromBytesInFlight(p)
+		putPacket(p)
 	}
+	// After this point, we must not use ackedPackets any longer!
```
lucas-clemente, line 339: `ackedPackets = nil`

internal/ackhandler/packet.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (p *Packet) outstanding() bool {
+	return !p.declaredLost && !p.skippedPacket && !p.IsPathMTUProbePacket
+}
+
+var packetPool = sync.Pool{New: func() any { return &Packet{} }}
+
+func GetPacket() *Packet {
+	p := packetPool.Get().(*Packet)
```
lucas-clemente, line 33: Assuming that callers will sanitize data seems dangerous to me, and could leak to cross-connection data leaks. Is there a downside to zeroing out the other struct members here?

lucas-clemente (approved):

internal/ackhandler/packet.go
marten-seemann: Makes sense.
At some point it might be nice to reuse the slice in `Packet.Frames`, but we're not there yet.

internal/ackhandler/sent_packet_handler.go
marten-seemann: Done.

PR #3529, opened 2022-08-28
use a struct containing an array to represent Connection IDs

### LINKED ISSUE #3527: use a more efficient type for Connection IDs
The Connection ID is currently a `[]byte`: https://github.com/lucas-clemente/quic-go/blob/07412be8a02ef0e55580ebf8db9c38a759c0a0e5/internal/protocol/connection_id.go#L10-L11

This struct consumes 24 bytes for the slice (8 bytes for the pointer, the length and the capacity, respectively), plus the size of the underlying array.

Given that QUIC Connection IDs in all currently defined QUIC versions are limited to 20 bytes, we can use a simpler datastructure:
```go
type ConnectionID [21]byte
```
The first byte would be used to encode the length of the Connection ID (a uint8 is sufficient for that), and the next 20 bytes would contain the actual Connection ID.
This would have several advantages:
* cache locality: no need for a pointer, as we have in a byte slice
* can be allocated on the stack
* actually consumes less memory _in all cases_
* could be used as map keys

There's a slight complication, namely that the server needs to be able to handle packets with Connection IDs of up to 255 bytes, as future QUIC versions might allow for longer lengths, see [Section 5.3 of RFC 8999](https://datatracker.ietf.org/doc/html/rfc8999#section-5.3), to be able to send Version Negotiation packets. The hope is that we can keep this code path as separated as possible from the rest of the packet processing.

marten-seemann:
Fixes #3527. Depends on #3528.


internal/protocol/connection_id.go
```diff
@@ -27,58 +29,76 @@ func (c ArbitraryLenConnectionID) String() string {
...
 }
 
-// A ConnectionID in QUIC
-type ConnectionID []byte
-
 const maxConnectionIDLen = 20
 
+// A ConnectionID in QUIC
+type ConnectionID struct {
+	l uint8
```
lucas-clemente, line 36: I think Go will pad this up to a word, so you could save a few bytes by having the array first.

lucas-clemente (approved):

internal/protocol/connection_id.go
marten-seemann: `unsafe.Sizeof(ConnectionID{})` is 21 in both cases, so it doesn't seem to make a difference.

This is the same change that @Joroppo suggested in https://github.com/lucas-clemente/quic-go/issues/3527#issuecomment-1229536697.

Issue #3526, opened 2022-08-28
Tracking Issue: reduce allocations during bulk data transfers


marten-seemann:
A 1 GB transfer using v0.29.0 currently creates (as measured by pprof's alloc tool, which measures _total_ allocations over the lifetime of the process) about 500 - 600 MB of allocations, both on the sender and on the receiver side.

This is a problem for performance, because allocating memory consumes resources, and more importantly, all of this memory has to be garbage-collected, putting a lot of pressure on the GC.

We need to drastically reduce the amount of allocations. Target is a reduction of an order of magnitude.

The worst offenders at the moment seem to be:
- [x] allocations of `wire.{Extended}Header`. This is a struct representing a QUIC packet header
- [x] allocations of `ackhandler.Packet` structs (used to track sent packets until they're acknowledged / declared lost): #3525
- [x] allocations of `wire.AckFrame`: https://github.com/lucas-clemente/quic-go/pull/3547
- [x] parsing the control message via `unix.ParseSocketControlMessage`. This will require us to either rewrite that code ourselves, or upstream a fix to `golang.org/x/net` (both?)
     Go issue: https://github.com/golang/go/issues/54714
     PR: https://github.com/lucas-clemente/quic-go/pull/3609
- [x] use a more efficient type for connection IDs: #3527
- [x] stop using a `bytes.Buffer` when packing packets. Allocating new `bytes.Buffer`s is responsible 32 MB of allocations. We can just append to a `[]byte`
     #3530
- [x] don’t allocate a new `bytes.Reader` for every packet when parsing frames. This is responsible for 32 MB of allocations on the receiver side.
    #3536
- [x] list allocations, especially when sending packets: https://github.com/lucas-clemente/quic-go/pull/3655

### Server
![server](https://user-images.githubusercontent.com/1478487/187061914-e1842fb0-e981-442a-a536-8cfcd7bb07b1.png)

### Client
![profile001](https://user-images.githubusercontent.com/1478487/187061924-70fb2940-d6f3-4567-8524-42a7b47239dd.png)

zllovesuki:
On `v0.30.0` the allocations look way better:
```
specter-linux-amd64
Type: alloc_space
Time: Oct 30, 2022 at 4:30am (PDT)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top5
Showing nodes accounting for 356.03MB, 51.02% of 697.81MB total
Dropped 309 nodes (cum <= 3.49MB)
[...]

PR #3536, opened 2022-08-29
use a single bytes.Reader for frame parsing

marten-seemann:
Part of #3526. Depends on #3534.

### DIFF
--- a/internal/wire/frame_parser.go
+++ b/internal/wire/frame_parser.go
@@ -11,6 +11,8 @@ import (
 )
 
 type frameParser struct {
+	r bytes.Reader // cached bytes.Reader, so we don't have to repeatedly allocate them
+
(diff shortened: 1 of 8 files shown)


internal/wire/frame_parser.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 type frameParser struct {
+	r *bytes.Reader // cached bytes.Reader, so we don't have to repeatedly allocate them
```
lucas-clemente, line 14: Nit: You can probably also avoid the ptr indirection here.

lucas-clemente (approved):

internal/wire/frame_parser.go
marten-seemann: Done.

PR #3569, opened 2022-10-03
prevent the connection timer to be reset to the same deadline

marten-seemann:
If the timer was already set in the same mode to the same deadline, we shouldn't reset it (so that it doesn't fire again). This prevents busy-looping when we're not able to act upon firing of the timer for any reason. Possible cases where this might happen include:
* the send queue is backed up, preventing us from sending new packets
* the maximum number of outstanding packets was reached

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -210,7 +210,7 @@ type connection struct {
 
 	peerParams *wire.TransportParameters
 
-	timer *utils.Timer
+	timer timer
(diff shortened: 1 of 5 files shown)


marten-seemann:
This PR is incorrect. We can end up in a situation where no timer is set at all (not even an idle timeout). That would be really bad.

PR #3570, opened 2022-10-05
use a monotonous timer

### LINKED ISSUE #3179: Sending data CPU usage is too high
I have a file server and my program send data by stream.Write(buf)。
I found the following problem：
1. CPU usage is tooo high.
2. stream.Write() failed and the error is **Timeout: No recent network activity.**
3. This phenomenon is prone to occur when the number of connections reaches 30, but it does not necessarily occur.

### LINKED ISSUE #3567: run loop can busy-loop when the send queue blocks for a long period
Apparently the Accelerated DHT client makes the send queue block for extended periods of time, at least on OSX: https://github.com/ipfs/kubo/issues/9309.

This leads to busy-looping when the ACK-alarm expires, but we fail to send out the ACK packet.

marten-seemann:
Fixes #3567. Fixes #3179.

There's no point in having the timer fire multiple times for the same
timestamp. By using a monotonous timer we avoid busy-looping in cases where the
timer fires, but we can't actually send a packet.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -210,7 +210,7 @@ type connection struct {
 
 	peerParams *wire.TransportParameters
 
-	timer *utils.Timer
+	timer connectionTimer
 	// keepAlivePingSent stores whether a keep alive PING is in flight.
 	// It is reset as soon as we receive a packet from the peer.
 	keepAlivePingSent bool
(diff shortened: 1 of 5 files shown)


timer.go
```diff
@@ -0,0 +1,51 @@
+package quic
```
lucas-clemente, line 1: I wouldn't have two almost identical structs that subtly differ in their behavior in two separate packages with a similar name. Maybe call this `connectionTimer` or something like that?

Alternatively, you could generalize this helper by making it non-specific to the connection?

lucas-clemente (approved):
This assumes that we never rely on this behavior to not get stuck. I guess the test suite is good enough for that…? :/

timer.go
marten-seemann: Done.

marten-seemann:
> This assumes that we never rely on this behavior to not get stuck. I guess the test suite is good enough for that…? :/

The assumption is that when _any_ timer fires at `t > t'`, all the actions that should have been performed at `t'` will get executed. This has been our general design, since with a timer, you can never be sure when exactly it fires (it may fire at _any_ time after the time you set it for).

That said, I'd consider this PR an improvement, because even if we miss something here, worst case is that that one connection is eventually killed by the idle timeout. Previously, we'd effectively kill the node by busy-looping.

PR #3584, opened 2022-10-09
don't drop more than 10 consecutive packets in drop test

marten-seemann:


lucas-clemente (approved):
Maybe it's worth documenting why no more than 10?

PR #3588, opened 2022-10-11
HTTP/3 Datagrams are now RFC 9297

kokes:
I found a few references to HTTP/3 Datagrams that seem to be out of date. The [referred doc](https://www.ietf.org/archive/id/draft-ietf-masque-h3-datagram-08.html) has been superseeded by [RFC 9297](https://www.rfc-editor.org/rfc/rfc9297.html), so I updated the links and changed a few error/frame type values that have changed since the draft (see the draft where it says "note that this will switch to a lower value before publication").

I have not updated the README to say quic-go supports RFC 9297 as I don't know if it's the case, especially since I don't see any references to Quarter Stream IDs for instance.


marten-seemann:
I'm hesitant to merge this PR. First of all, we don't actually support this RFC. We support QUIC DATAGRAMs, but not HTTP/3 DATAGRAMs. Second, Chrome's WebTransport implementation does some datagram support detection (not sure if QUIC or H3), and refuses to establish a WebTransport connection if datagrams are disabled, so we'd have to be extra careful not to break things there.

kokes:
Capsule support has been merged in #3607, just adding it here for context.

marten-seemann:
@kokes Sorry for the long delay. Can you please rebase?

kokes:
Rebased.

dongcarl (approved):
LGTM. I have a longer branch with a fully working RFC 9297 and this is what was needed to interop with hyper/h3 for RFC 9297.

PR #3604, opened 2022-10-21
Fix logic problem in send_stream.go

DrakenLibra:
Consider such a situation: this stream already has nextFrame, now new dataForWriting needs to be written, but function canBufferStreamFrame() returns false, and function Write() stops at '<-s.writeChan". When popNewStreamFrame(), if nextFrame.DataLen() > maxDataLen, nextFrame will be split, it has probability that canBufferStreamFrame() can be True this time,  so we shouled run signalWrite(). If it runs signalWrite() and canBufferStreamFrame() returns True, it means that dataForWriting  can be stored in buffer and function Write() can be over. Therefore, the consumption of computing resources will be reduced.

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -286,9 +286,9 @@ func (s *sendStream) popNewStreamFrame(maxBytes, sendWindow protocol.ByteCount) 
 			s.nextFrame.DataLenPresent = true
 			copy(s.nextFrame.Data, nextFrame.Data[maxDataLen:])
 			nextFrame.Data = nextFrame.Data[:maxDataLen]
-		} else {
-			s.signalWrite()
 		}
+		s.signalWrite()
+
 		return nextFrame, s.nextFrame != nil || s.dataForWriting != nil
 	}
 


marten-seemann:
What issue does this solve?

DrakenLibra:
> What issue does this solve?

### Sorry for bothering you
Oh sorry，actually this PR doesn's associate with any issue. Does there have to be an issue to be helped, then PR can be proposed? It is my firt contribution, I don't konw whether this rule exists in ths project. If this rule exists, I'm sorry for bothering you.

### Motivation about this PR
When I review code in send_stream.go, I find its logic is "buffer as much data as possible". But before this PR, there may occur situation that dataForWriting can be buffered but not buffered as I said in comments above. Actually code in send_stream.go is all right and doesn't occur any error. But after I change code as this PR shows, function Write() can finish earlier in some situations (bucause it won't be blocked by writeChan and further buffer data). I think this change can  reduce some computing cost in some circumstances.

marten-seemann:
> Sorry for bothering you

No need to apologize, we're always happy about contributions.

I'm just having a hard time understanding this change. Admittedly, the code in the stream is not really easy to understand. Can you explain in which situation you think the computing cost is too high?

DrakenLibra:
### Which situation I face
Actually I'm trying to do a research experiment using quic-go. In my experiment, I need to create lots of streams to send data inside a connection and schedule them. So I need to fully understand how sending streams works. Additionally, I think code in stream is difficult to understand, too. But when I spent a lot of time trying to understand code in this part, I think code in quic-go to realize stream-multiplex feature is awesome. In my experiment, my CPU load is high bacause I create lots of streams. I think its absolutely normal. But when I try to understand code in send_stream.go, I think there may be some improvement space in code logic. That's why I propose this PR. I will give an example.

### Example
There is a `stream` in lots of streams inside a connection. This stream already has `nextFrame` and we assume its `datalen=1000bytes`. Now, this stream call `Write()` to write `dataForWriting `whose `datalen=800bytes`. Because `nextFrame+dataForWriting=1800 bytes` is bigger than `protocol.MaxPacketBufferSize=1452ytes`. So `dataForWriting` can **not** be buffered in nextFrame. Thus `Write()` will **stop** at `<-s.writeChan` and wait for signal. Assume that it is this stream turn to pop stream frame, func `popNewStreamFrame() ` will be called. But packet only have `600bytes space`. So judgement `if nextFrame.DataLen() > maxDataLen` will be True. Now let us see changes before and after this PR:
#### Before this PR
NextFrame will be split and return but `signalWrite()` **isn't called**. And `Write()` stiill stops at `<-s.writeChan`. Meanwhile, look at `AppendStreamFrames()` in `framer.go`, this stream will be put back into the **end** of the queue and wait other active streams being asked to transmit data. If there is some data needed to call `Write()` in this stream to be written, it will be awful (If deadlineTimer exists, it will be more awful)  and writing data has to wait.
#### After this PR
NextFrame will be split and return. Meanwhile `signalWrite()` is called. Thus in func `Write()`, we will enter next loop. This time, because `nextFrame+dataForWriting=1200 bytes` is smaller than `protocol.MaxPacketBufferSize=1452ytes`, `s.canBufferStreamFrame()` will return True. So `dataForWriting` can be added behind nextFrame, bufferSize will be fully used, and `Write()` will return this time. If now new data calls `Write()`, it will be buffered (if its datalen is small such as 200bytes) or it will run as what I said above.

### In a conclusion
Anyway, I think bufferSize can be used more efficiently after this PR, at least logically. But admittedly I'm a new one to learn golang and quic-go, maybe I'm thoughtless in some aspects in this part of code. Thanks for your patiently asking and great contribution to quic-go. I will continually focus on quic-go project.

Issue #3605, opened 2022-10-22
http3: client response does not include request


Davincible:
The stdlib client will include the original request in the response, as to provide some context. I found that the HTTP3 client does not include it. Not a terrible bug but would be nice to be compliant with this small thing as well.

Only thing that needs to change is adding `Request: req,` to this struct;

https://github.com/lucas-clemente/quic-go/blob/af30cef57c5c20ebcb599200a96d89f7c33c3d50/http3/client.go#L388-L393

If welcome I can create a PR

Davincible:
Small side question;
After adding this I noticed the `req.Proto == "HTTP/1.1"`, why is it saying that?
Response does say `HTTP/3`

marten-seemann:
> If welcome I can create a PR

Yes please :)

> After adding this I noticed the req.Proto == "HTTP/1.1", why is it saying that?

Not sure if we ever set it correctly on the request struct. If we return it, it should be set correctly though.

PR #3608, opened 2022-10-31
http3: add request to response

Davincible:
fix: #3605

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -11,12 +11,13 @@ import (
 	"sync"
 	"time"
 
+	"github.com/marten-seemann/qpack"
+
 	"github.com/lucas-clemente/quic-go"
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 	"github.com/lucas-clemente/quic-go/internal/qtls"
 	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/lucas-clemente/quic-go/quicvarint"
-	"github.com/marten-seemann/qpack"
 )
 
 // MethodGet0RTT allows a GET request to be sent using 0-RTT.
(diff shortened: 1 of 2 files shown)


http3/client_test.go
```diff
@@ -741,6 +741,9 @@ var _ = Describe("Client", func() {
 			Expect(rsp.Proto).To(Equal("HTTP/3.0"))
 			Expect(rsp.ProtoMajor).To(Equal(3))
 			Expect(rsp.StatusCode).To(Equal(418))
+			Expect(rsp.Request).ToNot(Equal(nil))
```
marten-seemann, line 744:
```suggestion
			Expect(rsp.Request).ToNot(BeNil())
```

http3/client.go
```diff
@@ -251,6 +251,10 @@ func (c *client) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Respon
 		return nil, c.handshakeErr
 	}
 
+	req.Proto = "HTTP/3.0"
+	req.ProtoMajor = 3
+	req.ProtoMinor = 0
```
marten-seemann, line 256: Why do we need to set these? The documentation for `http.Request` says:
> For client requests, these fields are ignored.

Davincible: > If we return it, it should be set correctly though

Otherwise, the request added to the response will say it's http1

marten-seemann: I was just wondering what happens if you use the same `http.Request` struct multiple times, potentially on HTTP/3 and HTTP/2. Are we creating a race condition here by modifying the struct?

http3/client.go
```diff
@@ -11,12 +11,13 @@ import (
...
 	"time"
 
+	"github.com/marten-seemann/qpack"
+
 	"github.com/lucas-clemente/quic-go"
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 	"github.com/lucas-clemente/quic-go/internal/qtls"
 	"github.com/lucas-clemente/quic-go/internal/utils"
 	"github.com/lucas-clemente/quic-go/quicvarint"
-	"github.com/marten-seemann/qpack"
```
marten-seemann, line 19: Can you revert this one?

Davincible: fixed

marten-seemann (approved):
Thank you!

PR #3613, opened 2022-11-05
fix client SNI handling

tobyxdd:
The current code assumes that a hostname containing a colon is in the `host:port` format, but it could actually be an IPv6 address. Calling SplitHostPort in this case returns an error, and right now the Client will fail to initialize.

It is possible and not extremely rare to have a TLS cert with IPv6 SAN(s). Real-life examples include Cloudflare's https://[2606:4700:4700::1111]/ and some users of my project using self-signed CAs to issue certs for their IPv6 addresses.

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"strings"
 
 	"github.com/lucas-clemente/quic-go/internal/protocol"
 	"github.com/lucas-clemente/quic-go/internal/utils"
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
What if the host doesn't contain any port? `net.SplitHostPort` will error in that case.

tobyxdd:
> What if the host doesn't contain any port? `net.SplitHostPort` will error in that case.

When `net.SplitHostPort` returns an error it treats the whole string as hostname instead of returning an error. That's the point of this PR.

marten-seemann:
> My change is to handle this situation exactly. An IPv6 address (without a port) contains colons but is not in the `host:port` format. Instead of checking whether the address has a port by the presence of a colon, we should feed it to `net.SplitHostPort` to see if it returns an error.

Got it, that makes sense. I missed that the `if err != nil` doesn't return. Can you add a comment why we're not returning (i.e. could be a hostname or an IP address without a port number)?

marten-seemann (approved):

PR #3616, opened 2022-11-08
use `go run` for mockgen, goimports and ginkgo

kixelated:
Works without go install and uses the (pinned) version in go.mod


marten-seemann (approved):

internal/mocks/mockgen.go
```diff
@@ -1,20 +1,20 @@
...
-//go:generate sh -c "mockgen -package mocks -destination congestion.go github.com/lucas-clemente/quic-go/internal/congestion SendAlgorithmWithDebugInfos"
-//go:generate sh -c "mockgen -package mocks -destination connection_flow_controller.go github.com/lucas-clemente/quic-go/internal/flowcontrol ConnectionFlowController"
-//go:generate sh -c "mockgen -package mockackhandler -destination ackhandler/sent_packet_handler.go github.com/lucas-clemente/quic-go/internal/ackhandler SentPacketHandler"
-//go:generate sh -c "mockgen -package mockackhandler -destination ackhandler/received_packet_handler.go github.com/lucas-clemente/quic-go/internal/ackhandler ReceivedPacketHandler"
+//go:generate sh -c "go run github.com/golang/mock/mockgen -package mockquic -destination quic/stream.go github.com/lucas-clemente/quic-go Stream"
+//go:generate sh -c "go run github.com/golang/mock/mockgen -package mockquic -destination quic/early_conn_tmp.go github.com/lucas-clemente/quic-go EarlyConnection && sed 's/qtls.ConnectionState/quic.ConnectionState/g' quic/early_conn_tmp.go > quic/early_conn.go && rm quic/early_conn_tmp.go && goimports -w quic/early_conn.go"
```
marten-seemann, line 4: Should we do the same for `goimports`?

marten-seemann (changes requested):

internal/mocks/mockgen.go
kixelated: Yeaaah I was just about to suggest that. And ginkgo?

marten-seemann: > And ginkgo?

You mean in CI setup? sgtm

kixelated: Yeah, ginkgo doesn't really matter since I think `go install` uses the version in go.mod? But might as well.

marten-seemann (approved):

PR #3619, opened 2022-11-11
Enable certificate verification in "echo" demo

jason-i-vv:
this will make example more complete ,if some one want to know how to enable client certificate verification


marten-seemann:
This is just using a `crypto/tls`, unrelated to quic-go. We don't need to show people how to use the standard library.

PR #3620, opened 2022-11-14
expose the QUIC version of a connection

MarcoPolo:
We already track the QUIC version in a [connection](https://github.com/lucas-clemente/quic-go/blob/master/connection.go#L2141). This simply exposes this in the public interface.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -331,7 +331,7 @@ type Config struct {
 type ConnectionState struct {
 	TLS               handshake.ConnectionState
 	SupportsDatagrams bool
-	QUICVersion       protocol.VersionNumber
+	Version       protocol.VersionNumber
 }
 
 // A Listener for incoming QUIC connections
--- a/interface.go
+++ b/interface.go
@@ -331,7 +331,7 @@ type Config struct {
 type ConnectionState struct {
 	TLS               handshake.ConnectionState
 	SupportsDatagrams bool
-	Version       protocol.VersionNumber
+	Version           protocol.VersionNumber
 }
 
 // A Listener for incoming QUIC connections
--- a/interface.go
+++ b/interface.go
@@ -331,7 +331,7 @@ type Config struct {
 type ConnectionState struct {
 	TLS               handshake.ConnectionState
 	SupportsDatagrams bool
-	Version           protocol.VersionNumber
+	Version           VersionNumber
 }
 
 // A Listener for incoming QUIC connections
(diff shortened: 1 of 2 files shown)


interface.go
```diff
@@ -184,6 +184,9 @@ type Connection interface {
 	SendMessage([]byte) error
 	// ReceiveMessage gets a message received in a datagram, as specified in RFC 9221.
 	ReceiveMessage() ([]byte, error)
+
+	// GetVersion returns the QUIC version number used for this connection
+	GetVersion() protocol.VersionNumber
```
marten-seemann, line 189: Why not expose this on the ConnectionState?

MarcoPolo: You've thought about this API more than I have, but here are two reasons:
1. The `ConnectionState` is a essentially a wrapper around `tls.connectionState` and the Version there refers to the TLS version use by the connection.
2. We already have a `GetVersion` method on the `connection` struct.

marten-seemann: > 1. The `ConnectionState` is a essentially a wrapper around `tls.connectionState` and the Version there refers to the TLS version use by the connection.

Not only, we also expose if the peer supports the datagram extension. I think it would make sense to add the QUIC version here, similar to how TLS exposes the version on the `tls.ConnectionState`.

> 2\. We already have a `GetVersion` method on the `connection` struct.

True. This is probably not very well-reasoned, but exposing another method feels like a bigger API change than adding a field to the `ConnectionState` struct.

Would you mind updating this PR?

MarcoPolo: Sounds good! will do

interface.go
```diff
@@ -331,6 +331,7 @@ type Config struct {
 type ConnectionState struct {
 	TLS               handshake.ConnectionState
 	SupportsDatagrams bool
+	QUICVersion       protocol.VersionNumber
```
marten-seemann, line 334:
```suggestion
	Version       protocol.VersionNumber
```

marten-seemann (commented):
LGTM, but let's call this `Version` instead of `QUICVersion`.

marten-seemann (commented):
LGTM, but CI is unhappy (not gofmt-ed).

MarcoPolo:
Should we make the version numbers non-internal? https://github.com/lucas-clemente/quic-go/blob/master/internal/protocol/version.go#L19

marten-seemann:
> Should we make the version numbers non-internal? https://github.com/lucas-clemente/quic-go/blob/master/internal/protocol/version.go#L19

It already is, there's an alias in the quic package. We should probably use that one.

MarcoPolo:
ah I can't believe I didn't see that at the top of the file.

Does this PR win for most back and forths relative to the code change?

PR #3623, opened 2022-11-17
http3: ContentLength must be -1 for unknown body length

mattrobenolt:
According to the http.Request docs, on a server request, a ContentLength of -1 indicates an unknown length.

So in the case of a request without an explicit Content-Length header, the ContentLength was being set to 0, which explicitly indicates there's no body to read. Whereas -1 would indicate to read until EOF.

From my understanding, requestFromHeaders() is only used within the server context, and if it were used for the client context, this would be wrong, since a 0 is correct as unknown length on a client Request.

Please correct me if I am misunderstanding, but this caused me a bug integrating an http3.Server into my project.

I worked around it on my end by doing something similar in my handler:

```go
if req.Body != nil && req.ContentLength == 0 && req.Header.Get("Content-Length") == "" {
    req.ContentLength = -1
}
```

This is less than ideal, but works in my case.

### DIFF
--- a/http3/request.go
+++ b/http3/request.go
@@ -83,10 +83,21 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
 	}
 
 	var contentLength int64
-	if len(contentLengthStr) > 0 {
-		contentLength, err = strconv.ParseInt(contentLengthStr, 10, 64)
-		if err != nil {
-			return nil, err
+	if mayHaveBody(method) {
+		if len(contentLengthStr) > 0 {
+			// parse the Content-Length header as a uint, without the sign bit, so it can be
+			// safely packed back into an int64. The header itself is an uint64, but a Content-Length header
+			// cannot be negative.
+			if cl, err := strconv.ParseUint(contentLengthStr, 10, 63); err != nil {
+				return nil, err
+			} else {
+				contentLength = int64(cl)
+			}
+		} else {
+			// if there is allowed to be a body based on the http method,
+			// and there's no content-length header, we set explicitly to -1, which
+			// indicates "unknown length"
+			contentLength = -1
 		}
 	}
 
@@ -110,4 +121,16 @@ func hostnameFromRequest(req *http.Request) string {
 		return req.URL.Host
 	}
 	return ""
+}
+
+func mayHaveBody(method string) bool {
+	switch method {
+	// HACK: methods which permit sending a body
+	// this isn't entirely accurate, but is a tradeoff we're willing to make
+	// to attempt to be more correct. It's technically possible to send a GET
+	// request with a body, and this is within spec, it's just extremely frowned upon.
+	case http.MethodPost, http.MethodPatch, http.MethodDelete, http.MethodPut:
+		return true
+	}
+	return false
 }
--- a/http3/request_test.go
+++ b/http3/request_test.go
@@ -15,6 +15,7 @@ var _ = Describe("Request", func() {
 			{Name: ":path", Value: "/foo"},
 			{Name: ":authority", Value: "quic.clemente.io"},
 			{Name: ":method", Value: "GET"},
+			// content-length should be ignored since it's a GET request
 			{Name: "content-length", Value: "42"},
 		}
 		req, err := requestFromHeaders(headers)
@@ -25,7 +26,7 @@ var _ = Describe("Request", func() {
 		Expect(req.Proto).To(Equal("HTTP/3.0"))
 		Expect(req.ProtoMajor).To(Equal(3))
 		Expect(req.ProtoMinor).To(BeZero())
-		Expect(req.ContentLength).To(Equal(int64(42)))
+		Expect(req.ContentLength).To(Equal(int64(0)))
 		Expect(req.Header).To(BeEmpty())
 		Expect(req.Body).To(BeNil())
 		Expect(req.Host).To(Equal("quic.clemente.io"))
@@ -47,6 +48,7 @@ var _ = Describe("Request", func() {
 		Expect(req.URL.Host).To(BeEmpty())
 		Expect(req.Host).To(Equal("quic.clemente.io"))
 		Expect(req.RequestURI).To(Equal("//foo"))
+		Expect(req.ContentLength).To(Equal(int64(0)))
 	})
 
 	It("concatenates the cookie headers", func() {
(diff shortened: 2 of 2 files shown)


marten-seemann:
> According to the http.Request docs, on a server request, a ContentLength of -1 indicates an unknown length.

I think I remember that this only applies to requests that actually have a body. Is that correct? What does the Go standard library?

mattrobenolt:
On a server request, Request.Body will never be nil, according to the docs, so even on a GET request, there is a non-nil body, it just immediately returns EOF when attempting to read.

I can put together a small demo if you'd like, but reading the docs on the Request struct make it pretty clear: https://pkg.go.dev/net/http#Request

> 	// For server requests, the Request Body is always non-nil
>	// but will return EOF immediately when no body is present.
>	// The Server will close the request body. The ServeHTTP
>	// Handler does not need to.

> 	// ContentLength records the length of the associated content.
>	// The value -1 indicates that the length is unknown.
>	// Values >= 0 indicate that the given number of bytes may
>	// be read from Body.

mattrobenolt:
Also based on the integration tests failing, I'm not sure of the more widespread ramification of this within the library, but I think to adhere to spec, this is the correct approach. Other stuff internally might need to account for -1?

marten-seemann:
> On a server request, Request.Body will never be nil, according to the docs, so even on a GET request, there is a non-nil body, it just immediately returns EOF when attempting to read.

That doesn't necessarily mean that the value needs to be -1 though. I could also see how 0 would make sense in that case (no body, no length).

mattrobenolt:
Sure it does, according to the ContentLength spec:

> The value -1 indicates that the length is unknown.

If I'm implementing a server, and I want to read a body, I would use req.ContentLength to say, allocate a fixed []byte to read my body into. If it's 0, I can't exactly differentiate between "unknown" and "0" without also double checking against the Content-Length header.

But either way, this behavior differs from the stdlib http server, and caused an issue on my end that I needed to work around. I'm fine keeping the workaround, but it feels incorrect on http3.Server's side, according to their docs.

mattrobenolt:
I guess what I'm saying is this isn't about the 0 length case, it's about an unknown length explicitly. If I'm a client, and I send a request without a Content-Length header, but I'm streaming in a long body, saying req.ContentLength of 0, to me as a server handler, indicates the body is a length of 0, rather than an unknown length, as -1 would indicate.

marten-seemann:
I agree with that case (unknown length of the body). We should definitely fix that.

I'm just saying that we should figure out how to handle the case where there's no body.

mattrobenolt:
> I'm just saying that we should figure out how to handle the case where there's no body.

Yeah, I just think that's entirely tangential.

Poking around at what http stdlib does, I think there's a bit of magic happening, since it seems to fill a correct ContentLength in normal cases, even when there's no Content-Length header, so my assumption is it's reading the body into a buffer and then it knows the actual length.

The only case req.ContentLength is coming through as -1 is when Transfer-Encoding: chunked is set on the request headers.

So to be fair, there could be a bug elsewhere, since I'm also using the http3.RoundTripper, and when using that, somehow a Content-Length isn't being set while using an http2.RoundTripper, it is.

I'm going to continue investigating, but I still believe defaulting here to -1 is more correct than 0.

mattrobenolt:
Oh, and I guess to clarify my current findings, I _am_ sending the request as Transfer-Encoding: chunked, which is why through normal http.Server, I get a ContentLength of -1, whereas through http3.Server, I get 0. So maybe we can also check the Transfer-Encoding header.

mattrobenolt:
In stdlib, I guess the logic here is much more in depth: https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/http/transfer.go;l=669-737;drc=8c17505da792755ea59711fc8349547a4f24b5c5

I'm still inclined to say, sending back -1 when unknown is better than 0. I think the difference is an efficiency gain for clients when they know the length, they can allocate buffers and whatnot to read it efficiently.

mattrobenolt:
Ok, so I really think this boils down to simply being Transfer-Encoding chunked. It seems to be correct in other cases.

marten-seemann:
Chunked encoding only applies to HTTP 1.1, there's no such thing in HTTP/2 and H3.

The relevant code snippet in HTTP/2 seems to be this one: https://cs.opensource.google/go/x/net/+/refs/tags/v0.2.0:http2/server.go;l=2135-2154

If I understand correctly, this:
1. doesn't set the content length if there's no body
2. parses the content length header, if it exists
3. sets -1 otherwise

Do you agree with my reading of that code? Is this the logic we should implement here as well?

mattrobenolt:
Ah, that makes more sense. I got the Transfer-Encoding because I was using the httputil.DumpRequest, which apparently isn't fully accurate. :) This has been a bit tricky to debug I'll admit.

But it seems that logic is correct to me. I'm not sure off hand how to know in http3 if there is a body or not. Happy to look into it later, or you can commandeer this PR since I'm sure you'll be faster with it.

marten-seemann:
Maybe we can go by the request method? There's a nice list here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET.

It looks like only POST, PATCH, DELETE and PUT (please double check) may have request bodies.

mattrobenolt:
Personally I wouldn't do that since the specs _allow_ bodies on other methods, even if they're not typically supported. That's just a choice your implementation can make, say, not supporting a body on a GET. I haven't checked this against what stdlib stuff does since it's not very applicable to me.

I'd say if anything, it's an easy iteration and should be sufficient for a lot of cases.

I wonder if the inverse would be problematic tho? Since you could POST/PUT whatever without bodies at all. So in theory if I POST with no body, I'd expect ContentLength to be 0, not -1. But I think we're splitting hairs here.

mattrobenolt:
@marten-seemann I have updated this with more tests, and to add a set of methods with expected bodies.

This gets us I think to where we wanted, where a `GET` request ends up with a `ContentLength=0`, and a `POST` without a `Content-Length` header, gets us to `ContentLength=-1`.

mattrobenolt:
@marten-seemann :wave: let me know if you have any other suggestions or opinions here.

mattrobenolt:
@marten-seemann reaching out here again. I'd like to remove my hacky middleware to work around this issue if I can. :)

marten-seemann:
I believe this is now resolved with the numerous changes regarding the Content-Length header in the recent releases. Please lmk if it's not.

PR #3624, opened 2022-11-19
implement Server.RequestContext

omz13:
Implement the equivalent of net/http's Server.ConnContext in http3

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -601,7 +601,7 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 	if cc := s.RequestContext; cc != nil {
 		ctx = cc(ctx, conn)
 		if ctx == nil {
-			panic("ConnContext returned nil")
+			panic("RequestContext returned nil")
 		}
 	}
 	req = req.WithContext(ctx)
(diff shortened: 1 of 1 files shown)


mattrobenolt:
Just to chime in since I've been looking for this to. This is entirely not the equivalent to stdlib and it'd be very disingenuous to commit to this feature and it behave this way.

In this implementation, this is happening per-request, not per-connection. The entire point of the ConnContext is to be one per connection.

mattrobenolt:
To be clear, to do this correctly, this needs to be hooked much much sooner in the flow and ensure it's passed all the way through. This should be hooked as close to the Accept() call as is realistic.

omz13:
> In this implementation, this is happening per-request, not per-connection. The entire point of the ConnContext is to be one per connection.

I made a mistake on the naming. So instead of mucking around with the context at the connection level, I could achieve what was necessary by mucking around at the request level. Naming and description updated so it should be clearer.

mattrobenolt:
At this point, why not just use a middleware pattern and wrap the server handler?

Still going to strongly suggest instead of any of this, we just implement ConnContext correctly.

omz13:
I've refactored how to solve my original issue... this PR is not the path I'm going to pursue anymore; it can be ignored

PR #3635, opened 2022-12-23
add Allow0RTT opt in the quic.Config to control 0-RTT on the server side

marten-seemann:
See https://github.com/lucas-clemente/quic-go/issues/3634 for the motivation and explanation of this API change.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -176,7 +176,6 @@ type Connection interface {
 	// Context returns a context that is cancelled when the connection is closed.
 	Context() context.Context
 	// ConnectionState returns basic details about the QUIC connection.
-	// It blocks until the handshake completes.
 	// Warning: This API should not be considered stable and might change soon.
 	ConnectionState() ConnectionState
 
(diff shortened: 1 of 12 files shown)


lucas-clemente (approved):

interface.go
```diff
@@ -325,6 +324,10 @@ type Config struct {
 	// This can be useful if version information is exchanged out-of-band.
 	// It has no effect for a client.
 	DisableVersionNegotiationPackets bool
+	// Allow0RTT allows the application to decide if a 0-RTT connection attempt should be accepted.
+	// When set, 0-RTT is enabled. When not set, 0-RTT is disabled.
+	// Only valid for the server.
+	Allow0RTT func(net.Addr) bool
```
MarcoPolo, line 330: This makes sense, but what's the use case for custom logic to switch depending on address?

MarcoPolo (approved):

interface.go
marten-seemann: The address is not particularly useful indeed, especially since it's not validated yet at this point.

On the other hand, I really want to introduce _some kind of_ callback here, because it might be interesting to have the application make the decision based on _something_. For example, we could allow the application to save some state into the session ticket at some point, and then allow / reject 0-RTT based on the restored stayed when the connection is resumed.

I'll add a comment saying that this API is unstable and we'll add more fields to the callback at some point.

PR #3638, opened 2022-12-23
update module path of qpack to quic-go/qpack

marten-seemann:


marten-seemann:
Closing in favor of #3676.

PR #3653, opened 2022-12-29
http3: re-dial connection if a previously used connection is closed

### LINKED ISSUE #765: handle closed clients in http3.RoundTripper

marten-seemann:
Fixes #765. Closes #3157.

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -17,6 +17,7 @@ import (
 
 type roundTripCloser interface {
 	RoundTripOpt(*http.Request, RoundTripOpt) (*http.Response, error)
+	IsClosed() bool
 	io.Closer
 }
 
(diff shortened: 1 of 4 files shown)


http3/roundtrip.go
```diff
@@ -175,9 +179,13 @@ func (r *RoundTripper) getClient(hostname string, onlyCached bool) (roundTripClo
 		if err != nil {
 			return nil, err
 		}
-		r.clients[hostname] = client
+		r.clients[hostname] = cl
+	}
+	if cl.IsClosed() {
```
lucas-clemente, line 184: Don't we have a race condition here, when the client is closed in between here and where it's returned?

marten-seemann: Yeah. Any suggestion how to fix this? There’s actually a quite long window here between the client being returned and the request being received by the server, in which the connection could die. Seems like the only way to properly fix this is by having some (bounded) retry logic?

zllovesuki: `when the client is closed in between here and where it's returned` might actually not be a problem, it should get recreated the next time `.getClient` is called. For bounded retry, you can add a parameter `retries` since it is calling itself recursively.

marten-seemann:
Good point! Especially interesting is the part where they look at context errors: https://cs.opensource.google/go/x/net/+/refs/tags/v0.5.0:http2/client_conn_pool.go;l=292-311;drc=c63010009c802314a29324ce49987897f9838e29
Of course we shouldn't retry when the user cancels the request. But I think this is covered by the `net.Error.Timeout()` check.

marten-seemann:
Additional complication: We need to handle the (probably very common) case where a client fires off a couple of requests to the same server. If we didn't have connection before, we don't want to retry the connection if we run into a handshake timeout.

I think this can be solved by only considering a connection as being "reused" once it has completed the handshake.

PR #3656, opened 2022-12-30
use a sync.Pool for ackhandler.Frames

marten-seemann:
Depends on #3648. Part of #3526.

### DIFF
--- a/internal/ackhandler/packet.go
+++ b/internal/ackhandler/packet.go
@@ -10,7 +10,7 @@ import (
 // A Packet is a packet
 type Packet struct {
 	PacketNumber    protocol.PacketNumber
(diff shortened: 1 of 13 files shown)


internal/ackhandler/packet.go
```diff
@@ -10,7 +10,7 @@ import (
 // A Packet is a packet
 type Packet struct {
 	PacketNumber    protocol.PacketNumber
-	Frames          []Frame
+	Frames          []*Frame
```
marten-seemann, line 13: It's a bit sad that we need to use a pointer here now, but that's the price we're paying for not allocating.

lucas-clemente (approved):

PR #3660, opened 2023-01-01
 optimizing  http3/server.go

aviscode:
Signed-off-by: Avi Rosenberg <avrumi96@gmail.com>

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -210,16 +210,14 @@ func (s *Server) ListenAndServe() error {
 //
 // If s.Addr is blank, ":https" is used.
 func (s *Server) ListenAndServeTLS(certFile, keyFile string) error {
-	var err error
-	certs := make([]tls.Certificate, 1)
-	certs[0], err = tls.LoadX509KeyPair(certFile, keyFile)
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
 		return err
 	}
 	// We currently only use the cert-related stuff from tls.Config,
 	// so we don't need to make a full copy.
 	config := &tls.Config{
-		Certificates: certs,
+		Certificates: []tls.Certificate{cert},
 	}
 	return s.serveConn(config, nil)
 }
@@ -280,8 +278,10 @@ func (s *Server) serveConn(tlsConf *tls.Config, conn net.PacketConn) error {
 		quicConf.EnableDatagrams = true
 	}
 
-	var ln quic.EarlyListener
-	var err error
+	var (
+		err error
+		ln  quic.EarlyListener
+	)
 	if conn == nil {
 		addr := s.Addr
 		if addr == "" {
@@ -308,11 +308,11 @@ func (s *Server) serveListener(ln quic.EarlyListener) error {
 		if err != nil {
 			return err
 		}
-		go func() {
+		go func(conn quic.EarlyConnection) {
 			if err := s.handleConn(conn); err != nil {
 				s.logger.Debugf(err.Error())
 			}
-		}()
+		}(conn)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
I don't understand the motivation behind most of these changes.

Using the status code constants is a worthwhile change though.

aviscode:
adding comments on the changes

http3/server.go
```diff
@@ -210,16 +210,14 @@ func (s *Server) ListenAndServe() error {
 //
 // If s.Addr is blank, ":https" is used.
 func (s *Server) ListenAndServeTLS(certFile, keyFile string) error {
-	var err error
-	certs := make([]tls.Certificate, 1)
-	certs[0], err = tls.LoadX509KeyPair(certFile, keyFile)
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
```
aviscode, line 213: changed to not allocate the slice with a make before we are sure we don't have any errors 
and if all ok no need for the make we can directly add to the tls.Config struct

http3/server.go
```diff
@@ -308,11 +308,11 @@ func (s *Server) serveListener(ln quic.EarlyListener) error {
 		if err != nil {
 			return err
 		}
-		go func() {
+		go func(conn quic.EarlyConnection) {
```
aviscode, line 311: because it's a in a for loop we need to pass the conn as a value to the go-routine to avoid race conditions

http3/server.go
```diff
@@ -321,12 +321,7 @@ func extractPort(addr string) (int, error) {
 	if err != nil {
 		return 0, err
 	}
-
-	portInt, err := net.LookupPort("tcp", portStr)
-	if err != nil {
-		return 0, err
-	}
-	return portInt, nil
+	return net.LookupPort("tcp", portStr)
```
aviscode, line 324: no need to check if err the return  0 and err net.LookupPort is doing it in case of error net.LookupPort returns 0,err

http3/server.go
```diff
@@ -446,7 +441,7 @@ func (s *Server) handleConn(conn quic.Connection) error {
 			}
 			return fmt.Errorf("accepting stream failed: %w", err)
 		}
-		go func() {
+		go func(str quic.Stream) {
```
aviscode, line 444: because it's a in a for loop we need to pass the conn as a value to the go-routine to avoid race conditions

http3/server.go
```diff
@@ -690,16 +685,14 @@ func ListenAndServeQUIC(addr, certFile, keyFile string, handler http.Handler) er
 // The correct Alt-Svc headers for QUIC are set.
 func ListenAndServe(addr, certFile, keyFile string, handler http.Handler) error {
 	// Load certs
-	var err error
-	certs := make([]tls.Certificate, 1)
-	certs[0], err = tls.LoadX509KeyPair(certFile, keyFile)
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
```
aviscode, line 688: changed to not allocate the slice with a make before we are sure we don't have any errors 
and if all ok no need for the make we can directly add to the tls.Config struct

http3/server.go
marten-seemann: I don't really care about allocs in the error case.

http3/server.go
marten-seemann: That's not correct. It only applies when you pass in the loop variable.

http3/server.go
marten-seemann: This is fine.

http3/server.go
marten-seemann: Same here.

aviscode:
@marten-seemann Any updates?

marten-seemann:
I reviewed this PR last week. I don't think these a worthwhile changes, except for using the constants for the HTTP status codes. Feel free to open a new PR with those changes.

PR #3673, opened 2023-01-08
http3: use HTTP/2 as the default protocol in ListenAndServe

mstmdev:
Running the [example](https://github.com/lucas-clemente/quic-go/blob/master/example/main.go) code with `-tcp` flag, and request the url `https://localhost:6121/demo/tiles`.

The current default protocol is `HTTP/1.1`.
http/1.1

Use `HTTP/2` as the default protocol in this PR.
http2

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -717,19 +717,6 @@ func ListenAndServe(addr, certFile, keyFile string, handler http.Handler) error 
 	}
 	defer udpConn.Close()
 
-	tcpAddr, err := net.ResolveTCPAddr("tcp", addr)
-	if err != nil {
-		return err
-	}
-	tcpConn, err := net.ListenTCP("tcp", tcpAddr)
-	if err != nil {
-		return err
-	}
-	defer tcpConn.Close()
-
-	tlsConn := tls.NewListener(tcpConn, config)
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
Why does `http.ListenAndServeTLS` use HTTP/2, but `http.Server.Serve` doesn't?

mstmdev:
> HTTP/2 support is only enabled if the Listener returns *tls.Conn
   connections and they were configured with "h2" in the TLS
   Config.NextProtos.

We can just set the "h2"  in the `Config.NextProtos` like the below code.

```go
config := &tls.Config{
	Certificates: certs,
}
config.NextProtos = append(config.NextProtos, "h2")
```

But why not simplify the code by using the `http.ListenAndServeTLS` function?

And here are some code links: [shouldConfigureHTTP2ForServe](https://github.com/golang/go/blob/master/src/net/http/server.go#L2995),[http2ConfigureServer](https://github.com/golang/go/blob/master/src/net/http/h2_bundle.go#L4035)

@marten-seemann

marten-seemann (approved):

marten-seemann:
Thank you @mstmdev, and sorry for the long time it took me to review your PR!

PR #3674, opened 2023-01-11
fix: return typed errors when after cancelling actions

Jorropo:
This is errors.Is'ed by libp2p and the fmt.Errorf messages didn't passed that test: https://github.com/libp2p/go-libp2p/blob/82315917f76ca1e9e1c3aca3ad957080f0ac870c/p2p/transport/quic/stream.go#L23

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -41,11 +41,20 @@ const (
 	NoViablePathError         = qerr.NoViablePathError
 )
 
+type StreamErrorAction uint8
+
+const (
+	StreamErrorActionUnknown = iota
+	StreamErrorActionRead
+	StreamErrorActionWrite
+)
+
 // A StreamError is used for Stream.CancelRead and Stream.CancelWrite.
 // It is also returned from Stream.Read and Stream.Write if the peer canceled reading or writing.
 type StreamError struct {
 	StreamID  StreamID
 	ErrorCode StreamErrorCode
+	Action    StreamErrorAction
 }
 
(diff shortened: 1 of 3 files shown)


MarcoPolo (approved):

marten-seemann (commented):
Good catch! We definitely need to return a `StreamError` here.

The QUIC stream state machine allows the following transition when it comes to abnormal stream termination.

The receive side of a stream:
* can be reset by the peer (it would send us a RESET_STREAM frame)
* can be reset by us by calling `CancelRead` (this would trigger sending of a STOP_SENDING frame)

The send side of a stream:
* can be reset by us by calling `CancelWrite` (this triggers sending of the RESET_STREAM frame)
* can be reset by the peer (it would send us a STOP_SENDING frame)

I think we can make the error more helpful by exposing which side reset the stream. wdyt?

Jorropo:
> I think we can make the error more helpful by exposing which side reset the stream. wdyt?

Would make sense, could you pickup this PR please tho ?

marten-seemann:
> > I think we can make the error more helpful by exposing which side reset the stream. wdyt?
> 
> Would make sense, could you pickup this PR please tho ?

Done. Should be ready to go once approved.

marten-seemann:
Closing in favor of https://github.com/quic-go/quic-go/pull/3681 since I'm not allowed to push to this branch.

PR #3683, opened 2023-01-26
 http3: change status codes to const

aviscode:
Signed-off-by: Avi Rosenberg <avrumi96@gmail.com>


marten-seemann (commented):
Thank you for this PR!

There's a few places where we check that a status code is a 2xx, for example. I don't think it makes the code easier to understand if we substitute the specific status codes there. Other than that, this looks good to me.

marten-seemann (approved):
LGTM. Thank you!

PR #3684, opened 2023-01-26
http3: correctly handle closed clients

marten-seemann:
Fixes https://github.com/quic-go/quic-go/issues/765. Closes https://github.com/quic-go/quic-go/pull/3157.

@zllovesuki, would you mind taking a look at this PR? It now implements the logic described in https://github.com/quic-go/quic-go/pull/3653#issuecomment-1402718763. Specifically,
* We immediately remove a client when a request errored. This is an important cleanup, otherwise the client map will fill up with failed clients over time, with no way to garbage collect it.
* When using an existing client, and we run into a timeout `net.Error`, we re-dial the connection before returning to the user.
* This re-dial logic only applies to clients that have already completed the handshake. This means that when we get two requests from the application at the same time, and the handshake fails, we won't retry.


zllovesuki:
Implementation looks fine to me. ginkgo test formats really hurt my eyes

marten-seemann:
> Implementation looks fine to me.

Thank you Rachel!

> ginkgo test formats really hurt my eyes

I know. I wish we hadn’t decided to use Ginkgo when we started the project. But now I can’t really justify spending 2 weeks full time to rewrite all the tests. If you have any idea, please let me know :)

zllovesuki:
Technical debts 🤔

for future features and implementations, let's try using `testing`, then move things off ginkgo if needed at that time

marten-seemann:
Yep, that’s what I’m doing. webtransport-go already uses testing.

PR #3711, opened 2023-02-15
update qtls to include the Go 1.20.1 / 1.19.6 changes to crypto/tls

marten-seemann:
quic-go is **not** vulnerable to https://github.com/golang/go/issues/58001 due to tight limits imposed on the amount of handshake data that can be sent: https://github.com/quic-go/quic-go/blob/ad7182ce8407d1f4684e5cd2aeb60058b04acc00/crypto_stream.go#L41-L46

Nevertheless we should still update our qtls implementation to include the fix.


marten-seemann:
Tested on the QUIC interop runner.

PR #3677, opened 2023-01-20
ci: update GitHub action workflow versions

GiddyGoatGaming:
Updates actions in the workflow folder.

PR #3712, opened 2023-02-15
ci: fix interop Docker builder

marten-seemann:
Bug introduced by https://github.com/quic-go/quic-go/pull/3677.


marten-seemann:
Confirmed that this successfully builds a new Docker image, which then successfully completes the QUIC interop runner tests.

PR #3713, opened 2023-02-15
interop: fix server configuration for the 0-RTT test case

marten-seemann:


marten-seemann:
Confirmed that this fixes the test failures I saw in the QUIC interop runner.

Issue #3573, opened 2022-10-07
http3: sniff HTTP Content Type from first Write, analogous to http.ResponseWriter


marten-seemann:
This should be pretty straightforward change which can mostly be copied from the standard library HTTP implementation. See https://github.com/caddyserver/caddy/issues/5096#issuecomment-1268784887 for more discussion and links to code.

Glonee:
Hello @marten-seemann, does this issue still need help? If so, would you please take a look at my PR?

PR #3715, opened 2023-02-16
http3: sniff HTTP Content Type

Glonee:
Fiexs #3573 
Seems like quic-go is testing code using ginkgo and I am not familar with it. But since my code is mostly copied from standard library, I don't think anything will be broken.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -82,6 +82,7 @@ func (w *responseWriter) WriteHeader(status int) {
 func (w *responseWriter) Write(p []byte) (int, error) {
 	bodyAllowed := bodyAllowedForStatus(w.status)
 	if !w.headerWritten {
+		//If body is not allowed, we don't need to(and we can't) sniff the body.
 		if bodyAllowed {
 			// If no content type, apply sniffing algorithm to body.
 			// We can't use `w.header.Get` here since if the Content-Type was set to nil, we shound't do sniffing.
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -82,18 +82,16 @@ func (w *responseWriter) WriteHeader(status int) {
 func (w *responseWriter) Write(p []byte) (int, error) {
 	bodyAllowed := bodyAllowedForStatus(w.status)
 	if !w.headerWritten {
-		// If body is not allowed, we don't need to(and we can't) sniff the body.
+		// If body is not allowed, we don't need to (and we can't) sniff the content type.
 		if bodyAllowed {
 			// If no content type, apply sniffing algorithm to body.
-			// We can't use `w.header.Get` here since if the Content-Type was set to nil,we shound't do sniffing.
+			// We can't use `w.header.Get` here since if the Content-Type was set to nil, we shoundn't do sniffing.
 			_, haveType := w.header["Content-Type"]
 
 			// If the Transfer-Encoding or Content-Encoding was set and is non-blank,
 			// we shouldn't sniff the body.
-			te := w.header.Get("Transfer-Encoding")
-			hasTE := te != ""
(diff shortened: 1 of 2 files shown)


Glonee:
I found a mistake that the sniff will never run. Let me fix it and I will pull request again.

Glonee:
I worte some test code for sniff. Ginkgo is so werid.

Glonee:
Test failed on go1.19 but passed on go1.20? Doesn't make any sense😂
The test was failed at integrationtests/self/handshake_drop_test.go:91. My changes shoud not effect that

http3/response_writer.go
```diff
@@ -80,10 +80,28 @@ func (w *responseWriter) WriteHeader(status int) {
...
+			// If no content type, apply sniffing algorithm to body.
+			// We can't use `w.header.Get` here since if the Content-Type was set to nil,we shound't do sniffing.
+			_, haveType := w.header["Content-Type"]
+
+			// If the Transfer-Encoding or Content-Encoding was set and is non-blank,
+			// we shouldn't sniff the body.
+			te := w.header.Get("Transfer-Encoding")
+			hasTE := te != ""
+			ce := w.header.Get("Content-Encoding")
+			hasCE := ce != ""
```
marten-seemann, line 96:
```suggestion
			hasTE := w.header.Get("Transfer-Encoding") != ""
			hasCE := w.header.Get("Content-Encoding") != ""
```

http3/response_writer.go
```diff
@@ -80,10 +80,28 @@ func (w *responseWriter) WriteHeader(status int) {
 }
 
 func (w *responseWriter) Write(p []byte) (int, error) {
+	bodyAllowed := bodyAllowedForStatus(w.status)
 	if !w.headerWritten {
+		// If body is not allowed, we don't need to(and we can't) sniff the body.
```
marten-seemann, line 85:
```suggestion
		// If body is not allowed, we don't need to (and we can't) sniff the content type.
```

http3/response_writer.go
```diff
@@ -80,10 +80,28 @@ func (w *responseWriter) WriteHeader(status int) {
...
+		if bodyAllowed {
+			// If no content type, apply sniffing algorithm to body.
+			// We can't use `w.header.Get` here since if the Content-Type was set to nil,we shound't do sniffing.
```
marten-seemann, line 88:
```suggestion
			// We can't use `w.header.Get` here since if the Content-Type was set to nil, we shouldn't do sniffing.
```

http3/response_writer_test.go
```diff
@@ -147,4 +147,13 @@ var _ = Describe("Response Writer", func() {
 		Expect(n).To(BeZero())
 		Expect(err).To(MatchError(http.ErrBodyNotAllowed))
 	})
+
+	It("first call to Write do sniff if Content-Type is not set", func() {
```
marten-seemann, line 151:
```suggestion
	It("first call to Write sniffs if Content-Type is not set", func() {
```

marten-seemann (commented):
Thank you @Glonee! This looks good to me, just a few comments.

For future reference, this is similar to the standard library code: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/net/http/server.go;l=1420-1431

marten-seemann:
> I wrote some test code for sniff. Ginkgo is so werid.

Yes, sorry for that!!

Glonee:
> Thank you @Glonee! This looks good to me, just a few comments.
> 
> For future reference, this is similar to the standard library code: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/net/http/server.go;l=1420-1431

Ok, sorry for the typo!

marten-seemann (approved):
Thank you!

PR #3716, opened 2023-02-20
http3: Close client at errors

khasanovbi:


marten-seemann:
What problem does this solve?

khasanovbi:
Under load, we encountered an error that the client cannot allocate a new UDP port: `listen udp 0.0.0.0:0: bind: address already in use`
```
# sudo lsof -p 512346 | grep UDP | wc -l
57142
```
Host always the same.
Not fully sure that this is the real place where the problem happen, but this maybe one of the reasons.

marten-seemann:
I don't think that's related. The HTTP/3 client doesn't keep the UDP connection open.

That being said, I think we can optimize things here. In principle, you only need _a single_ UDP socket for _all_ of your QUIC connections. That's because QUIC can demultiplex based on the QUIC Connection ID.
In quic-go, this is achieved by not using `quic.DialAddr` (which will create and manage a `net.UDPConn` for you), but using `quic.Dial`, which takes a `net.UDPConn` as an argument. In the HTTP/3 package, we could create a single `UDPConn` per `RoundTripper`, and use that for all outgoing requests.

PR #3720, opened 2023-02-22
http3: use a single UDPConn in RoundTripper

### LINKED ISSUE #3718: http3: use a single UDPConn in RoundTripper
In principle, you only need _a single_ UDP socket for _all_ of your QUIC connections. That's because QUIC can demultiplex based on the QUIC Connection ID.
In quic-go, this is achieved by not using `quic.DialAddr` (which will create and manage a `net.UDPConn` for you), but using `quic.Dial`, which takes a `net.UDPConn` as an argument. In the HTTP/3 package, we could create a single `UDPConn` per `RoundTripper`, and use that for all outgoing requests.

_Originally posted by @marten-seemann in https://github.com/quic-go/quic-go/issues/3716#issuecomment-1437832124_

Glonee:
fix #3718 
I think this will work

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -69,7 +69,7 @@ type RoundTripper struct {
 
 	// Dial specifies an optional dial function for creating QUIC
 	// connections for requests.
-	// If Dial is nil, quic.DialAddrEarlyContext will be used.
+	// If Dial is nil, makeDialer will be used to create a dialer.
 	Dial func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error)
 
 	// MaxResponseHeaderBytes specifies a limit on how many response bytes are
@@ -79,6 +79,7 @@ type RoundTripper struct {
 
 	newClient func(hostname string, tlsConf *tls.Config, opts *roundTripperOpts, conf *quic.Config, dialer dialFunc) (roundTripCloser, error) // so we can mock it in tests
 	clients   map[string]roundTripCloser
+	udpConn   *net.UDPConn
 }
 
 // RoundTripOpt are options for the Transport.RoundTripOpt method.
@@ -173,6 +174,16 @@ func (r *RoundTripper) getClient(hostname string, onlyCached bool) (rtc roundTri
 		if r.newClient != nil {
 			newCl = r.newClient
 		}
+		dial := r.Dial
+		if dial == nil {
+			if r.udpConn == nil {
+				r.udpConn, err = net.ListenUDP("udp", &net.UDPAddr{IP: net.IPv4zero, Port: 0})
+				if err != nil {
+					return nil, false, err
+				}
+			}
+			dial = makeDialer(r.udpConn)
+		}
 		client, err = newCl(
 			hostname,
 			r.TLSClientConfig,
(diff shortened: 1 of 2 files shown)


http3/roundtrip.go
```diff
@@ -69,7 +69,7 @@ type RoundTripper struct {
 
 	// Dial specifies an optional dial function for creating QUIC
 	// connections for requests.
-	// If Dial is nil, quic.DialAddrEarlyContext will be used.
+	// If Dial is nil, makeDialer will be used to create a dialer.
```
marten-seemann, line 72: That's not really a helpful comment for the user of this package. Can we describe instead what happens (a single packet conn is reused)?

http3/roundtrip.go
```diff
@@ -173,6 +174,16 @@ func (r *RoundTripper) getClient(hostname string, onlyCached bool) (rtc roundTri
 		if r.newClient != nil {
 			newCl = r.newClient
 		}
+		dial := r.Dial
+		if dial == nil {
+			if r.udpConn == nil {
+				r.udpConn, err = net.ListenUDP("udp", &net.UDPAddr{IP: net.IPv4zero, Port: 0})
```
marten-seemann, line 180: What about IPv6?

http3/roundtrip.go
```diff
@@ -245,3 +260,13 @@ func validMethod(method string) bool {
 func isNotToken(r rune) bool {
 	return !httpguts.IsTokenRune(r)
 }
+
+func makeDialer(udpConn *net.UDPConn) func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error) {
```
marten-seemann, line 264: Would it make sense to have this as a method on the roundtripper?

marten-seemann (commented):
Sorry for the long time it took me to review your PR @Glonee! I somehow missed the notification.

This mostly looks good to me, I'm just not sure how we should handle IPv6. What do you think?

http3/roundtrip.go
Glonee: This is just the same as what we did in `quic.dialAddrContext`:
https://github.com/quic-go/quic-go/blob/a92238b73cc14632f2244fbf13bad9e2a4e54f26/client.go#L109
Golang treat `0.0.0.0` or `[::]` as unspecified IP address, their behaivor is the same as no ip address is specified.
We can simply leave `IP` to `nil` to avoid misunderstanding.

http3/roundtrip.go
Glonee: How about this:
```
If Dial is nil, a single packet conn will be reused for dialer.
```

http3/roundtrip.go
Glonee: It does make sense. In this way we don't need to pass a `*net.UDPConn` to it, it can use the `udpConn` inside the `RoundTripper` directly.

Glonee:
@marten-seemann I committed a new patch, would you please take a look?

http3/roundtrip.go
```diff
@@ -69,7 +69,8 @@ type RoundTripper struct {
 
 	// Dial specifies an optional dial function for creating QUIC
 	// connections for requests.
-	// If Dial is nil, quic.DialAddrEarlyContext will be used.
+	// If Dial is nil, a UDPConn will be created at the first request
+	// and will be reused for creating QUIC connections.
```
marten-seemann, line 73:
```suggestion
	// and will be reused for subsequent connections to other servers.
```

marten-seemann (commented):
Code looks good to me. Would you mind adding a test case?

Glonee:
Sure. Added a test case for this.

marten-seemann (approved):
Thank you @Glonee!

PR #3725, opened 2023-03-16
Correct the description of MaxIncomingStreams and MaxIncomingUniStreams

fbzhong:
Based on the current implementation, it appears that the configuration parameters `MaxIncomingStreams` and `MaxIncomingUniStreams` define `the maximum number of streams that a peer is allowed to open`, rather than `the maximum number of concurrent streams that can be opened`.


marten-seemann:
When a stream is closed, it doesn’t count towards this limit any more, so I believe the current description is correct.

fbzhong:
Sorry that I make a mistake on my own test code, and you are correct.

PR #3726, opened 2023-03-18
qtls: don't set the tls.Config.CipherSuites for cipher suite tests

marten-seemann:
The standard library crypto/tls implementation doesn't allow setting of `tls.Config.CipherSuites`, and this won't change when QUIC support is added.

This PR introduces a workaround suggested by Filippo. It uses `go:linkname` to (temporarily) modify the slice of TLS 1.3 cipher suites in https://github.com/golang/go/blob/20e9b7f1b53d49fd66e0344b1d0d42d3cf5e47b6/src/crypto/tls/cipher_suites.go#L201-L205. This is exactly how we'll be able to test this once we switch to crypto/tls.

### DIFF
--- a/client.go
+++ b/client.go
@@ -47,7 +47,6 @@ var generateConnectionIDForInitial = protocol.GenerateConnectionIDForInitial
 // DialAddr establishes a new QUIC connection to a server.
 // It uses a new UDP connection and closes this connection when the QUIC connection is closed.
 // The hostname for SNI is taken from the given address.
-// The tls.Config.CipherSuites allows setting of TLS 1.3 cipher suites.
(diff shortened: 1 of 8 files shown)


marten-seemann:
We can now drop the respective [commit](https://github.com/quic-go/qtls-go1-20/commit/7344d2a2a7747581e46733c6e471cc51e12400bc) in qtls. It doesn't really hurt, but it's easy enough to remove (rebase without any merge conflicts), so we might as well reduce the diff to crypto/tls.
I'll cut new qtls releases and update them here.

lucas-clemente (approved):

PR #3728, opened 2023-03-22
release unused buffer when error

wwqgtxx:


marten-seemann:
What’s the motivation? These errors are exceedingly rare, happening at most once per connection. Recovering the buffer doesn’t really have any effect.

wwqgtxx:
For example, when testing whether the server can be connected with h3, there will be a high probability of errors. Returning the buffer in time can effectively reduce the GC overhead.

marten-seemann:
> For example, when testing whether the server can be connected with h3, there will be a high probability of errors

Why?

wwqgtxx:
> > For example, when testing whether the server can be connected with h3, there will be a high probability of errors
> 
> 

> > For example, when testing whether the server can be connected with h3, there will be a high probability of errors
> 
> Why?

First of all, there are still a large number of servers in the current network environment that have not yet supported h3. 
Secondly, quic-go can pass in a custom net.PacketConn, which may directly report an error when the data packet cannot be sent normally instead of simply discarding the data packet like the system udp stack.

marten-seemann:
> First of all, there are still a large number of servers in the current network environment that have not yet supported h3.

They would never get this error.

> Secondly, quic-go can pass in a custom net.PacketConn, which may directly report an error when the data packet cannot be sent successfully instead of simply discarding the data packet like the system udp stack.

Then the listener is closed.

I'll close this PR for now, as I still don't see the motivation. If you have a pprof (or some other kind of data) that shows that this _actually_ is a problem, I'm open to reconsider.

PR #3739, opened 2023-03-28
wire: use constants for frame types

marten-seemann:
Depends on #3736.

I haven't figured out how to do this for STREAM and DATAGRAM frames yet. Any suggestions? Anyway, I think this is a worthwhile change as is, even though it leaves out those two frame types.


lucas-clemente (approved):

lucas-clemente:
Does having a mask constant work?

PR #3743, opened 2023-03-28
make http3 error codes public and consistent with http2 package

jfgiorgi:


jfgiorgi:
sorry bad account for CLA. switching account did help... i'll refork and rePR from this account.

Issue #3737, opened 2023-03-27
http3: design a proper error API


kgersen:
We're using a `context.WithTimeout` for a HTTP/3 connection.
In case of a timeout, we'd like to catch the `quic.StreamError` with `errorRequestCanceled`  (`errorCode = 0x10c)`.
The `ErrorCode` field is exported but its possible values are not so we have to do a test with `0x10c` (or duplicate [http3/error_codes.go](https://github.com/quic-go/quic-go/blob/master/http3/error_codes.go) and cast)

We could eventually test against the string representation `H3_REQUEST_CANCELLED`  which is more user friendly than `0x10c` but `errorCode` is not exported.

May be I've missing something and there is a better way to test for an `errorRequestCanceled`  than using 0x10c ?

Our current error handling code :
                   
```
...
// err has the error
var qerr *quic.StreamError
if errors.As(err, &qerr) && qerr.ErrorCode == 0x10c {
  //this is an errorRequestCanceled error, handle it
}
...

**Comment by marten-seemann** [comment]
How would you do this in HTTP/2?

**Comment by kgersen** [comment]
> How would you do this in HTTP/2?

It depends:

**general case**: we currently use the exported values from https://github.com/golang/net/blob/master/http2/errors.go (for instance we use this to check for `http2.ErrCodeInternal` if the server `WriteTimeout` expires).

**our specific context.WithTimeout case**: the HTTP/2 client returns  a `context.DeadlineExceeded` error instead of a `http2.StreamError` or a `http2.ConnectionError`. It's the same behavior as net/http (HTTP/1.1 mode) so we don't even need to distinguish between HTTP/1.1 and HTTP/2. we just do:

```
....
// err has the error
if errors.Is(err, context.DeadlineExceeded) {
  // this request was timed out, handle it
}
...
```

if HTTP/3 could return context errors that would be also nice to have but that is probably a separate issue from this one. Not all errors are context errors so it would be nice to have public "user friendly" values for http3 & quic errors.

marten-seemann:
It's nice to have the http2 package as a precedent. I think it makes sense to expose a `http3.StreamError` that looks like a `http2.StreamError`. @kgersen, would you want to contribute a patch? Happy to review.

Returning a (wrapped?) context error in case of context expiration definitely makes sense.

marten-seemann:
> Our current error handling code :
> 
> ```
> ...
> // err has the error
> var qerr *quic.StreamError
[...]

PR #3744, opened 2023-03-28
http3: make error codes public and consistent with http2 package

jfgiorgi:
simple renaming and making public error code values. 
consistent with https://github.com/golang/net/blob/master/http2/errors.go
partial fix for #3737

### DIFF
--- a/http3/error_codes.go
+++ b/http3/error_codes.go
@@ -6,66 +6,66 @@ import (
 	"github.com/quic-go/quic-go"
 )
 
-type errorCode quic.ApplicationErrorCode
+type ErrCode quic.ApplicationErrorCode
 
 const (
-	errorNoError              errorCode = 0x100
-	errorGeneralProtocolError errorCode = 0x101
-	errorInternalError        errorCode = 0x102
-	errorStreamCreationError  errorCode = 0x103
-	errorClosedCriticalStream errorCode = 0x104
-	errorFrameUnexpected      errorCode = 0x105
-	errorFrameError           errorCode = 0x106
-	errorExcessiveLoad        errorCode = 0x107
-	errorIDError              errorCode = 0x108
-	errorSettingsError        errorCode = 0x109
-	errorMissingSettings      errorCode = 0x10a
-	errorRequestRejected      errorCode = 0x10b
-	errorRequestCanceled      errorCode = 0x10c
-	errorRequestIncomplete    errorCode = 0x10d
-	errorMessageError         errorCode = 0x10e
-	errorConnectError         errorCode = 0x10f
-	errorVersionFallback      errorCode = 0x110
-	errorDatagramError        errorCode = 0x4a1268
+	ErrrNoError             ErrCode = 0x100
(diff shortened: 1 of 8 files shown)


marten-seemann:
Does this allow you to get the context cancelation error?

jfgiorgi:
> Does this allow you to get the context cancelation error?

not it's just a renaming of exiting private values. It just  allows us to use `http3.ErrRequestCanceled` instead of 0x10c which is already a good thing.

Context cancellation is more complex to do (unless we treat all `ErrRequestCanceled` as a context cancellation which we probably should not).

http3/error_codes.go
```diff
@@ -6,66 +6,66 @@ import (
...
-	errorSettingsError        errorCode = 0x109
-	errorMissingSettings      errorCode = 0x10a
-	errorRequestRejected      errorCode = 0x10b
-	errorRequestCanceled      errorCode = 0x10c
-	errorRequestIncomplete    errorCode = 0x10d
-	errorMessageError         errorCode = 0x10e
-	errorConnectError         errorCode = 0x10f
-	errorVersionFallback      errorCode = 0x110
-	errorDatagramError        errorCode = 0x4a1268
+	ErrrNoError             ErrCode = 0x100
```
marten-seemann, line 12: Typo.

kgersen: typo fixed

http3/error_codes.go
```diff
@@ -6,66 +6,66 @@ import (
...
-	errorMissingSettings      errorCode = 0x10a
-	errorRequestRejected      errorCode = 0x10b
-	errorRequestCanceled      errorCode = 0x10c
-	errorRequestIncomplete    errorCode = 0x10d
-	errorMessageError         errorCode = 0x10e
-	errorConnectError         errorCode = 0x10f
-	errorVersionFallback      errorCode = 0x110
-	errorDatagramError        errorCode = 0x4a1268
+	ErrNoError              ErrCode = 0x100
+	ErrGeneralProtocolError ErrCode = 0x101
```
marten-seemann, line 13: The H2 package uses slightly different names, adding `Code` to the name (e.g. `ErrCodeGeneralProtocolError`). Should we do the same?

marten-seemann (commented):
Sorry for the late review. This looks good to me. I'd still like to actually resolve the issue, but this is a well-contained change, so let's just get this PR merged.

One question regarding naming consistency with the standard library, see below.

http3/error_codes.go
kgersen: I think (but not I've not verified) the Go team convention/idiomatic naming if that the named values of a pseudo "enum"  have to all start with the full name of the type or not. So ErrCode*. It's a good convention when using completion in an IDE for instance.

marten-seemann: Can you make the change then?

marten-seemann (approved):

marten-seemann:
Thank you @kgersen!

PR #3770, opened 2023-04-13
http3: prefer IPv6 over IPv4 when dialing

### LINKED ISSUE #3755: http3: client always prefers IPv4
Because of a [`bug of ResolveUDPAddr`](https://github.com/golang/go/issues/28666), when trying to connect to a dual stacked host (= host with both ipv6 and ipv4 addresses), `quic-go` client always uses IPv4 instead of prefering IPv6 if available.
`net/http` (both versions 1.x and 2.0) doesn't behave like that. It prefers IPv6 if available. 

As a workaround we're forced to explicitly resolve names before using quic-go http3 client.

demo of the issue: https://github.com/kgersen/h3dial
running on a dual stack machine:

```
QUIC_GO_LOG_LEVEL="INFO" go run main.go
QUIC-GO:
2023/04/03 15:29:41 client Starting new connection to cloudflare-quic.com ([::]:46976 -> 104.22.8.38:443), source connection ID (empty), destination connection ID d160c75cbc7ae26fb3, version v1
2023/04/03 15:29:41 client Queueing packet (1147 bytes) for later decryption
2023/04/03 15:29:41 client Queueing packet (1059 bytes) for later decryption
2023/04/03 15:29:41 client Queueing packet (1147 bytes) for later decryption
  got status:  200 OK
  proto:  HTTP/3.0
net/http:
  connected to  [2606:4700:10::6816:926]:443
  got status:  200 OK
  proto:  HTTP/2.0

```

marten-seemann:
Fixes #3755.

I'm not sure this is the correct fix. It seems like we should be able to handle multiple DNS responses, and fall back from IPv6 to IPv4 if IPv6 is not available (for example). That would be a bigger change, and would probably need to be done in the quic and not the http3 package.

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -269,10 +269,14 @@ func isNotToken(r rune) bool {
 // makeDialer makes a QUIC dialer using r.udpConn.
 func (r *RoundTripper) makeDialer() func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error) {
 	return func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error) {
-		udpAddr, err := net.ResolveUDPAddr("udp", addr)
+		// Don't use net.ResolveUDPAddr here.
+		// Due to a standard library bug (https://github.com/golang/go/issues/28666) that function prefers IPv4 over IPv6.
+		// net.Dial prefers IPv6 over IPv4.
+		// See https://github.com/quic-go/quic-go/issues/3755 for discussion.
+		udpAddr, err := net.Dial("udp", addr)
 		if err != nil {
 			return nil, err
 		}
-		return quicDialer(ctx, r.udpConn, udpAddr, addr, tlsCfg, cfg)
+		return quicDialer(ctx, r.udpConn, udpAddr.RemoteAddr(), addr, tlsCfg, cfg)
 	}
 }


marten-seemann:
If you're setting the `Dial` option yourself, you can implement a similar workaround as in this PR.

Issue for the fix of `DialAddr` here: https://github.com/quic-go/quic-go/issues/3772.

kgersen:
Does this means `quic.DialAddrEarlyContext` and `quic.DialAddrContext` will always prefer IPv4 ? 

I've updated [h3dial](https://github.com/kgersen/h3dial) with a use case of a custom `http3.RoundTripper.Dial` (which should be called `DialContext`  ?)

lucas-clemente (approved):
Feels unfortunate to have a workaround here rather than fix this in the stdlib, but I agree with the agenda of pushing ipv6 ;)

marten-seemann:
The problem now is that if you do a `Get("https://localhost:1234")`, it only does IPv6, which doesn't work if you're just listening on IPv4.

bt90:
Isn't the fallback mechanism defined by the IETF?

https://en.m.wikipedia.org/wiki/Happy_Eyeballs

bt90:
https://datatracker.ietf.org/doc/rfc8305/

bt90:
This is also what curl is doing as explained in @bagder blog post:

https://daniel.haxx.se/blog/2023/01/12/

kgersen:
Fallback (happy eyeball)  is another issue. see #3772

bt90:
Strictly speaking, the preference of IPv6 is part of happy eyeballs.

PR #3773, opened 2023-04-13
ci: allow changing runners through config vars

galargh:
###### Description

This PR:
- allows changing runners for jobs in integration and cross-compile workflows via configuration variables
- changes workflow triggers to: pull requests, pushes to the default branches, pushes to version tags
- configures concurrency groups for workflows: cancel in-progress runs in pull requests if the HEAD moves

###### Testing
- [x] ran integration workflow on `['self-hosted', 'linux', 'x64', 'xlarge']` https://github.com/pl-strflt/tf-aws-gh-runner/actions/runs/4687529039
- [x] ran cross-compile workflow on `['self-hosted', 'linux', 'x64', '2xlarge']` https://github.com/pl-strflt/tf-aws-gh-runner/actions/runs/4687351779

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -4,6 +4,10 @@ on:
     tags: [v*]
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}-${{ github.event_name == 'push' && github.sha || '' }}
+  cancel-in-progress: true
+
 jobs:
   unit:
     strategy:
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,6 +1,7 @@
 on:
   push:
     branches:  [master]
+    tags: [v*]
   pull_request:
 
 jobs:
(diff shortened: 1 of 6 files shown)


.github/workflows/unit.yml
```diff
@@ -1,5 +1,12 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:  [master]
+    tags: [v*]
```
marten-seemann, line 4: Not a big fan of not running CI on branches. It's useful to detect failures when working on a long feature branch.

marten-seemann (commented):
Thank you for the PR @galargh! What do I need to set for the environment variables to try this out?

Ideally, the cross compile workflow would use a machine that allows for very fast building (fast CPU, but not sure if more cores help).

.github/workflows/unit.yml
marten-seemann: Conclusion from a sync conversation with @galargh: This actually makes sense. When working on a long feature branch, one can just open a draft PR to trigger CI builds.

marten-seemann:
Closing in favor of #3783.

PR #3783, opened 2023-04-23
ci: allow changing runners through config vars

marten-seemann:
Minimal version of #3773, making **only** the changes required to run on custom machines.


marten-seemann:
@galargh Any idea why the runners are not picking up the jobs?

lucas-clemente (approved):

PR #3790, opened 2023-04-26
http3: add compatibility with net/http.ResponseController

dunglas:
[Go 1.20 added a new `"net/http".ResponseController` type](https://tip.golang.org/doc/go1.20#http_responsecontroller) that allows, among other things, to set per-request timeouts.

This patch adds support for this new type. It is necessary for caddyserver/caddy#5509 and dunglas/mercure#764.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/quic-go/quic-go"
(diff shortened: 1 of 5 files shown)


dunglas:
@marten-seemann better integration tests added (I also tested manually that this patch works well with Mecure and Caddy).

marten-seemann (approved):

marten-seemann:
Thank you @dunglas!

PR #3804, opened 2023-05-05
use SO_RCVBUFFORCE to force receive buffer increase on Linux

### LINKED ISSUE #3795: use SO_RCVBUFFORCE to increase receive buffer size
> SO_RCVBUFFORCE (since Linux 2.6.14): Using this socket option, a privileged (CAP_NET_ADMIN) process can perform the same task as SO_RCVBUF, but the rmem_max limit can be overridden.

MarcoPolo:
Closes #3795


sys_conn_helper_linux.go
```diff
@@ -17,3 +22,23 @@ const (
 )
 
 const batchSize = 8 // needs to smaller than MaxUint8 (otherwise the type of oobConn.readPos has to be changed)
+
+func forceSetReceiveBuffer(c interface{}, bytes int) error {
```
marten-seemann, line 26: There's no downside to always calling SO_RCVBUFFORCE, is there?

If that's the case, you could just do it as part of `setReceiveBuffer`, no?

MarcoPolo: A couple a reasons to keep it here:
1. `unix.SO_RCVBUFFORCE` is only defined for GOOS=linux. So this would fail if we put it in `setReceiveBuffer` as is, since that's shared.
2. Having it as a separate function lets us test only this thing. However, I'm not sure this will actually work since CI runner might not have CAP_NET_ADMIN (I don't when running this as a normal user).

marten-seemann (approved):
Thank you for working on this @MarcoPolo!

One minor change was still needed: We need to ignore the error from `SetReadBuffer`. Other than that, I modified the CI setup a bit, so that we can test this (it's even included in Codecov, which is pretty cool).

PR #3807, opened 2023-05-06
add Connection.SetIdleTimeout

mcluseau:
Hi,

For the context, I'm testing QUIC non-HTTP use cases and I use NextProtos to select the application I want so I can mix multiple applicatons on one port (ie: `dns` and `h3`).

Applications may have different idle timeouts requirements (ie "heartbeat of 1s" vs "keep my HTTP/3 connection for 2 minutes just in case I want to send another request").

Since I use a single port, I have a single listener and thus a single `quic.Config`, so I can't set the idle timeout here. `Config.GetConfigForClient` is also the wrong place as I require the TLS negociation done before setting the timeout. I can also imagine protocols that would like to change the timeout dynamically.

Thus, I propose in this PR a interface change to allow this dynamic change.

Since there's already a `maybeResetTimer`, it's very simple to implement. One caveat could be some race condition; if it's a concern, I can convert `idleTimeout` to an `atomic.Uint64`.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -2176,6 +2176,11 @@ func (s *connection) ReceiveMessage() ([]byte, error) {
 	return s.datagramQueue.Receive()
 }
 
+func (s *connection) SetIdleTimeout(timeout time.Duration) {
+	s.idleTimeout = timeout
+	s.maybeResetTimer()
+}
+
 func (s *connection) LocalAddr() net.Addr {
 	return s.conn.LocalAddr()
 }
--- a/interface.go
+++ b/interface.go
@@ -187,6 +187,9 @@ type Connection interface {
 	SendMessage([]byte) error
 	// ReceiveMessage gets a message received in a datagram, as specified in RFC 9221.
 	ReceiveMessage() ([]byte, error)
+
+	// SetIdleTimeout sets this connection's idle timeout.
+	SetIdleTimeout(time.Duration)
 }
 
 // An EarlyConnection is a connection that is handshaking.


marten-seemann:
This is not a clean API. It also doesn't work, since the idle timeout is negotiated during the handshake.

For the future, I suggest opening an issue if you're suggesting an API change, so we discuss it first.

> `Config.GetConfigForClient` is also the wrong place as I require the TLS negociation done before setting the timeout. I can also imagine protocols that would like to change the timeout dynamically.

Can you explain why `GetConfigForClient` doesn't work. I don't understand how TLS negotiation (what is that?) comes into play here.

mcluseau:
Hi Marten, thanks for your answer.

`GetConfigForClient` provides only the remote address, and I suspect it's called before the TLS negociation.
https://github.com/quic-go/quic-go/blob/22d6b0e626c56f311ff05705042d8caf506e5919/interface.go#L329

It's doesn't work for my prototype because I need the `NegotiatedProtocol` from `tls.ConnectionState`.

marten-seemann:
Yes, it's called before the the TLS `GetClientHelloInfo` call. At that point during the connection, we might not even have received a full ClientHello.

We could potentially call it after processing the ClientHello, but that would mean that the `HandshakeTimeout` option wouldn't make sense any more.

mcluseau:
I agree that only a subset of the Config makes sense after the handshake. Taking another look at the Config, this subset may even be everything but:
- `HandshakeTimeout`
- `Allow0RTT`
- `Tracer`

Another way to look at it is to examine what makes sense at the application level, opening the way to something like `GetApplicationConfig(negociatedProtocol string) *AppConfig`.

I seems useful for a developer to have access to both stages of the negociation though, especially is MaxIdleTimeout can't be change after the initial negociation.

I'm open to working on any solution that this project's owners see fit :-)

We can discuss things here but I can also open an issue as you requested.

PR #3812, opened 2023-05-08
increase the UDP send buffer size to 2 MB

marten-seemann:


sys_conn_buffers.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+	"fmt"
+	"net"
+	"syscall"
+
+	"github.com/quic-go/quic-go/internal/protocol"
+	"github.com/quic-go/quic-go/internal/utils"
+)
+
+//go:generate sh -c "echo '// Code generated by go generate. DO NOT EDIT.\n// Source: sys_conn_buffers.go\n' > sys_conn_buffers_write.go && sed -e 's/SetReadBuffer/SetWriteBuffer/g' -e 's/setReceiveBuffer/setSendBuffer/g' -e 's/inspectReadBuffer/inspectWriteBuffer/g' -e 's/protocol\\.DesiredReceiveBufferSize/protocol\\.DesiredSendBufferSize/g' -e 's/forceSetReceiveBuffer/forceSetSendBuffer/g' -e 's/receive buffer/send buffer/g' sys_conn_buffers.go | sed '/^\\/\\/go:generate/d' >> sys_conn_buffers_write.go"
```
MarcoPolo, line 13: This is kind of gross, but if this is the last time this gets touched it's probably okay.

internal/protocol/params.go
```diff
@@ -5,6 +5,9 @@ import "time"
 // DesiredReceiveBufferSize is the kernel UDP receive buffer size that we'd like to use.
 const DesiredReceiveBufferSize = (1 << 20) * 2 // 2 MB
 
+// DesiredSendBufferSize is the kernel UDP send buffer size that we'd like to use.
+const DesiredSendBufferSize = (1 << 20) * 2 // 2 MB
```
MarcoPolo, line 9: Maybe a note why 1 << 20 (1 MiB) is actually 2MB here? (the linux kernel doubles this for bookkeeping I believe)

MarcoPolo (approved):

internal/protocol/params.go
marten-seemann: There's a *2. It might be worth incorporating the doubling into the check in `setReceiveBuffer` though.

sys_conn_buffers.go
marten-seemann: It totally is. However, the alternative was to either have a function with 6 additional parameters, totally obfuscating what's actually going on, or copy-pasting that code manually.

internal/protocol/params.go
MarcoPolo: oh, I missed the `*2` not sure how. Nvm.

PR #3816, opened 2023-05-11
ci: run linter on all supported platforms

marten-seemann:
Depends on #3808.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,6 +7,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/lint.yml
```diff
@@ -7,6 +7,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
+          skip-pkg-cache: true
```
marten-seemann, line 10: Unfortunately, this is needed to avoid running into https://github.com/golangci/golangci-lint-action/issues/135.

lucas-clemente (approved):

Issue #3814, opened 2023-05-10
http: implement http3.RoundTripper.CloseIdleConnections


marten-seemann:
The `http.Client` has a `CloseIdleConnections` method that call this method on the round tripper, if it exists. This can be useful when dealing with HTTP/3 connections as well, and shouldn’t be hard to implement.

Glonee:
I am interested in implementing this.

marten-seemann:
Great! Go for it, happy to review :)

PR #3820, opened 2023-05-12
implement http3.RoundTripper.CloseIdleConnections

Glonee:
http: implement http3.RoundTripper.CloseIdleConnections #3814

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"golang.org/x/net/http/httpguts"
 
@@ -23,6 +24,11 @@ type roundTripCloser interface {
 	RoundTripOpt(*http.Request, RoundTripOpt) (*http.Response, error)
 	HandshakeComplete() bool
 	io.Closer
+}
+
+type roundTripCloserWithCount struct {
+	roundTripCloser
+	useCount atomic.Int64
 }
 
 // RoundTripper implements the http.RoundTripper interface
@@ -82,7 +88,7 @@ type RoundTripper struct {
 	MaxResponseHeaderBytes int64
 
 	newClient func(hostname string, tlsConf *tls.Config, opts *roundTripperOpts, conf *quic.Config, dialer dialFunc) (roundTripCloser, error) // so we can mock it in tests
-	clients   map[string]roundTripCloser
+	clients   map[string]*roundTripCloserWithCount
 	udpConn   *net.UDPConn
 }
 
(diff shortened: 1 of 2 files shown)


Glonee:
I am not sure my implemention is correct. In the standard library, the `http2ClientConn` has a method called `closeIfIdle`, see https://github.com/golang/go/blob/a25688d406f2a4296d39b22a2b10aea7178eddc6/src/net/http/h2_bundle.go#L8044-L8059.
I didn't find any similar method in quic.Connection nor method to find out if a connection is idle. So I count the use of every connection, and those have zero count are considered to be idle.
Let me know if you have any thought about this.

http3/roundtrip.go
```diff
@@ -204,10 +211,13 @@ func (r *RoundTripper) getClient(hostname string, onlyCached bool) (rtc roundTri
 		if err != nil {
 			return nil, false, err
 		}
+		client = new(roundTripCloserWithCount)
+		client.roundTripCloser = c
```
marten-seemann, line 215: Nit:
```suggestion
		client = &roundTripCloserWithCount{roundTripCloser: c}
```

http3/roundtrip.go
```diff
@@ -25,6 +26,11 @@ type roundTripCloser interface {
 	io.Closer
 }
 
+type roundTripCloserWithCount struct {
+	roundTripCloser
+	useCount atomic.Int64
```
marten-seemann, line 31: Both `getClient` and `CloseIdleConnections` run with the `Roundtripper.mutex` locked. Would it be sufficient to make this an `int`?

marten-seemann (commented):
Thank you @Glonee! Just a few comments, this mostly looks good.

Counting the number of active `RoundTrip` calls seems correct. The server can't initiate any streams in HTTP/3 (other than server push, which we don't support).

http3/roundtrip.go
```diff
@@ -143,6 +149,7 @@ func (r *RoundTripper) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.
 	if err != nil {
 		return nil, err
 	}
+	defer cl.useCount.Add(-1)
```
Glonee, line 152: We reduce the count here, out of the lock. Unless we add a lock here, or we should use atomic.

http3/roundtrip.go
Glonee: https://github.com/quic-go/quic-go/blob/b6c36841a5e5063864b628dd2beab780a8d043f2/http3/roundtrip.go#L152
We reduce the count here, out of the lock. Unless we add a lock here, or we should use atomic.

marten-seemann: Makes sense, thank you!

marten-seemann (approved):
LGTM, thank you @Glonee!

PR #3825, opened 2023-05-13
embed packetInfo in receivedPacket struct, use netip.Addr

marten-seemann:
This completely eliminates allocations when reading packets from the socket. Here's a profile of a 10 GB transfer (receive side):
profile001

Out of the 2.66 GB of allocations, 2.40 GB (that's > 90%) are caused in the UDP send / receive path in the standard library. This once again highlights how absolutely crucial the implementation of the new UDP API (https://github.com/golang/go/issues/45886) would be.

### DIFF
--- a/closed_conn.go
+++ b/closed_conn.go
@@ -30,7 +30,7 @@ func newClosedLocalConn(sendPacket func(net.Addr, *packetInfo), pers protocol.Pe
 	}
 }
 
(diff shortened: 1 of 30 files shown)


lucas-clemente (approved):

marten-seemann:
Stupid GitHub changed this PR, with no option to reopen. Great.

PR #3837, opened 2023-05-15
ci: fix coverage report

Glonee:
I noticed that the coverage report is not working in quic-go since last week, due to:
https://github.com/quic-go/quic-go/blob/cec79d338c645522923c92ffc88fee580c711315/.github/workflows/unit.yml#L47
Actually we should use a comma instead of a space to separate files here.


marten-seemann (approved):
Looks correct: https://github.com/codecov/codecov-action#arguments

Thank you @Glonee!

PR #3841, opened 2023-05-16
ackhandler: use a slice to keep track of sent packets

### LINKED ISSUE #3836: use an efficient data structure for tracking sent packets
Recent benchmarks (that include GSO (#3808)) show that tracking sent packets accounts for roughly 10% of the CPU on the sender side, equally split between registering sent packets and processing of the ACK frames. With the recent optimizations, that percentage is only going to increase. 
image

This is caused by the inefficient data structure used for tracking sent packets. Obviously, the combination of a linked list and a map won't be fast. It also requires another `sync.Pool` to amortize allocations of the list elements, which is not great.

---

## Proposal for a better data structure

We can make use of the fact that
1. Packet numbers are monotonically increasing (with the small exception of skipped packet numbers) and
1. Packets are usually received and acknowledged in order.

This suggests that a simple slice (or, once we have #3826, a ring buffer) is the most performant data structure, since:
1. Knowing the packet number of the first element, it's trivial to calculate where the requested packet number is located
1. When packets are missing from an ACK frame, they're declared lost after a relatively short timer. They can probably stay in the same slice. ACKed packets would be set to `nil` in the slice.
1. The size of the ring buffer would be limited to a small multiple (2?) of the congestion window (in packets). A peer trying to run a DoS on us would, worst case, be able to make us save 2x the congestion window until we'd stop sending new packets.

marten-seemann:
Fixes #3836.

### DIFF
--- a/internal/ackhandler/sent_packet_history.go
+++ b/internal/ackhandler/sent_packet_history.go
@@ -142,28 +142,6 @@ func (h *sentPacketHistory) HasOutstandingPackets() bool {
 	return h.numOutstanding > 0
 }
 
-func (h *sentPacketHistory) DeleteBefore(t time.Time) {
-	for i, p := range h.packets {
-		if p == nil {
-			continue
-		}
-		if !p.SendTime.Before(t) {
-			break
-		}
-		if !p.declaredLost {
-			continue
-		}
-		if p.outstanding() {
-			h.numOutstanding--
-			if h.numOutstanding < 0 {
-				panic("negative number of outstanding packets")
-			}
-		}
-		h.packets[i] = nil
-	}
-	h.cleanupStart()
-}
-
 // delete all nil entries at the beginning of the packets slice
 func (h *sentPacketHistory) cleanupStart() {
 	for i, p := range h.packets {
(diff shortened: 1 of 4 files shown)


internal/ackhandler/sent_packet_history.go
```diff
@@ -2,158 +2,197 @@ package ackhandler
...
+		h.packets[i] = nil
+	}
+	h.cleanupStart()
+}
+
+// delete all nil entries at the beginning of the packets slice
+func (h *sentPacketHistory) cleanupStart() {
+	for i, p := range h.packets {
+		if p != nil {
+			h.packets = h.packets[i:]
```
lucas-clemente, line 171: I'm a bit rust on Go slices, is there any chance for a leak here?

lucas-clemente (approved):

internal/ackhandler/sent_packet_history.go
marten-seemann: No, since all the packets up to `i` have been set to `nil`.
At some point we'll need to allocate a new `packets` slice (on `append`), and the old one will be GC'ed. That could be improved by using a ring buffer or similar data structure. We'll tackle this in a future PR.

Issue #3765, opened 2023-04-07
limit number of Go routines spawned when receiving unexpected packets


marten-seemann:
Specifically, we shouldn't spawn a new Go routine for sending stateless resets, INVALID_TOKEN and version negotiation packets.

Instead, we should have one or a few worker Go routines that receive from a channel, and then put packets on the wire as quickly as they can. This also provides a way backpressuring: If we're overwhelmed and can't send out these packets quickly enough, we can just drop the response, which will eventually lead to a retransmission from the peer.

marten-seemann:
Hint: This concerns most of the Go routines started in server.go and transport.go.

PR #3842, opened 2023-05-17
transport: send stateless reset packets from a single Go routine

sukunrt:
1/2 of https://github.com/quic-go/quic-go/issues/3765

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -66,7 +66,8 @@ type Transport struct {
 
 	conn rawConn
 
-	closeQueue chan closePacket
+	closeQueue          chan closePacket
+	statelessResetQueue chan sendStatelessResetArgs
 
 	listening   chan struct{} // is closed when listen returns
 	closed      bool
@@ -182,6 +183,7 @@ func (t *Transport) init(isServer bool) error {
 		t.listening = make(chan struct{})
 
 		t.closeQueue = make(chan closePacket, 4)
+		t.statelessResetQueue = make(chan sendStatelessResetArgs, 4)
 
 		if t.ConnectionIDGenerator != nil {
 			t.connIDGenerator = t.ConnectionIDGenerator
@@ -196,7 +198,7 @@ func (t *Transport) init(isServer bool) error {
 		}
 
 		go t.listen(conn)
-		go t.runCloseQueue()
+		go t.runSendQueue()
 	})
 	return t.initErr
 }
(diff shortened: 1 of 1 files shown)


transport.go
```diff
@@ -369,9 +383,7 @@ func (t *Transport) maybeSendStatelessReset(p *receivedPacket, connID protocol.C
 	rand.Read(data)
 	data[0] = (data[0] & 0x7f) | 0x40
 	data = append(data, token[:]...)
-	if _, err := t.conn.WritePacket(data, p.remoteAddr, p.info.OOB()); err != nil {
```
marten-seemann, line 372: `maybeSendStatelessReset` is still called in a separate Go routine. I'm wondering if the send queue Go routine should be the one that's composing the packet though, that would free up the Go routine that's reading from the connection.

sukunrt: fixed the bug. 
In case we do the computation in the sendQueue goroutine, do you think it makes sense to spawn multiple such go routines?

transport.go
```diff
@@ -210,13 +212,25 @@ func (t *Transport) enqueueClosePacket(p closePacket) {
 	}
 }
 
-func (t *Transport) runCloseQueue() {
+func (t *Transport) enqueueStatelessResetPacket(p packetDetails) {
+	select {
+	case t.statelessResetPacketQueue <- p:
+	default:
+		// It is fine to drop stateless reset packets
```
marten-seemann, line 219:
```suggestion
		// it's fine to not send a stateless reset when we're busy
```

transport.go
marten-seemann: No, doing it in a single Go routine is fine. We don't expect to be sending many stateless resets under normal circumstances.

transport.go
```diff
@@ -66,7 +66,8 @@ type Transport struct {
 
 	conn rawConn
 
-	closeQueue chan closePacket
+	closeQueue          chan closePacket
+	statelessResetQueue chan sendStatelessResetArgs
```
marten-seemann, line 70: Probably easier: just pass the `receivedPacket` around.

transport.go
```diff
@@ -363,13 +367,22 @@ func (t *Transport) maybeSendStatelessReset(p *receivedPacket, connID protocol.C
...
+	select {
+	case t.statelessResetQueue <- sendStatelessResetArgs{addr: p.remoteAddr, info: p.info, connID: connID}:
+	default:
+		// it's fine to not send a stateless reset when we're busy
+	}
+}
+
+func (t *Transport) sendStatelessReset(addr net.Addr, info *packetInfo, connID protocol.ConnectionID) {
 	token := t.handlerMap.GetStatelessResetToken(connID)
-	t.logger.Debugf("Sending stateless reset to %s (connection ID: %s). Token: %#x", p.remoteAddr, connID, token)
```
marten-seemann, line 367: You can re-parse the connection ID here. Or, for extra performance, even wrap this in a `if t.logger.Debug()`, then we only do the parsing and logging if we're actually logging.

marten-seemann (commented):
Logic is mostly fine, except for a race condition: You return the buffer to its pool, but keep a reference to the data slice.

transport.go
sukunrt: I would still need the connId for call to GetStatelessResetToken.

transport.go
sukunrt: that does sound better.

marten-seemann (approved):
Looks great! Thank you @sukunrt!

PR #3848, opened 2023-05-20
fix comment claiming ParseConnectionID reuses the data slice

sukunrt:
This is protocol.ParseConnectionID. It copies data from b.
```
func ParseConnectionID(b []byte) ConnectionID {
	if len(b) > maxConnectionIDLen {
		panic("invalid conn id length")
	}
	var c ConnectionID
	c.l = uint8(len(b))
	copy(c.b[:c.l], b)
	return c
}
```

Feel free to close if this is too small etc. Apologies if this is just plain mistaken.

### DIFF
--- a/internal/wire/header.go
+++ b/internal/wire/header.go
@@ -13,8 +13,6 @@ import (
 )
 
 // ParseConnectionID parses the destination connection ID of a packet.
-// It uses the data slice for the connection ID.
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):
Thank you! That's correct. Must have missed to update the comment in https://github.com/quic-go/quic-go/pull/3529.

PR #3854, opened 2023-05-23
server: send version negotiation and invalid token packets from a single Go routine

sukunrt:
closes: https://github.com/quic-go/quic-go/issues/3765


marten-seemann (approved):
Solid work! Thank you!

PR #3857, opened 2023-05-24
use a ring buffer in the framer

### LINKED ISSUE #3826: use a ring buffer for the frame queue in the framer
The framer uses a slice of stream IDs, containing an (ordered) list of the streams that have data to send:
https://github.com/quic-go/quic-go/blob/e9fea086134e61073d443ea1ac0ec256b88e4864/framer.go#L31

Using a slice here is problematic, as for every packet we send (assuming it contains one STREAM frame), we consume and append one element to the slice. It therefore leads to regular allocations in the data send path.

This allocation is responsible for more than 10% of the runtime of `AppendStreamFrames`:
image

Instead, we should use a ring buffer. It can start with a fairly small capacity (4 maybe?) and double that capacity once this is not enough. In this case, the maximum capacity is limited by the number of concurrent streams, so the ring buffer would not have to deal with this.

I imagine that it would make sense to implement a generic ring buffer and use that one in the framer.

Glonee:
Fixes #3826. Part of https://github.com/quic-go/quic-go/issues/3526.

### DIFF
--- a/framer.go
+++ b/framer.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/quic-go/quic-go/internal/ackhandler"
 	"github.com/quic-go/quic-go/internal/protocol"
+	"github.com/quic-go/quic-go/internal/utils/ringbuffer"
 	"github.com/quic-go/quic-go/internal/wire"
 	"github.com/quic-go/quic-go/quicvarint"
 )
@@ -28,7 +29,7 @@ type framerI struct {
 	streamGetter streamGetter
 
 	activeStreams map[protocol.StreamID]struct{}
-	streamQueue   []protocol.StreamID
+	streamQueue   ringbuffer.RingBuffer[protocol.StreamID]
 
 	controlFrameMutex sync.Mutex
 	controlFrames     []wire.Frame
@@ -45,7 +46,7 @@ func newFramer(streamGetter streamGetter) framer {
 
 func (f *framerI) HasData() bool {
 	f.mutex.Lock()
-	hasData := len(f.streamQueue) > 0
+	hasData := !f.streamQueue.Empty()
 	f.mutex.Unlock()
 	if hasData {
 		return true
@@ -84,7 +85,7 @@ func (f *framerI) AppendControlFrames(frames []*ackhandler.Frame, maxLen protoco
 func (f *framerI) AddActiveStream(id protocol.StreamID) {
 	f.mutex.Lock()
 	if _, ok := f.activeStreams[id]; !ok {
-		f.streamQueue = append(f.streamQueue, id)
+		f.streamQueue.PushBack(id)
 		f.activeStreams[id] = struct{}{}
 	}
 	f.mutex.Unlock()
@@ -95,13 +96,12 @@ func (f *framerI) AppendStreamFrames(frames []*ackhandler.Frame, maxLen protocol
 	var lastFrame *ackhandler.Frame
 	f.mutex.Lock()
 	// pop STREAM frames, until less than MinStreamFrameSize bytes are left in the packet
-	numActiveStreams := len(f.streamQueue)
+	numActiveStreams := f.streamQueue.Len()
 	for i := 0; i < numActiveStreams; i++ {
 		if protocol.MinStreamFrameSize+length > maxLen {
 			break
 		}
-		id := f.streamQueue[0]
-		f.streamQueue = f.streamQueue[1:]
+		id := f.streamQueue.PopFront()
 		// This should never return an error. Better check it anyway.
 		// The stream will only be in the streamQueue, if it enqueued itself there.
 		str, err := f.streamGetter.GetOrOpenSendStream(id)
@@ -117,7 +117,7 @@ func (f *framerI) AppendStreamFrames(frames []*ackhandler.Frame, maxLen protocol
 		remainingLen += quicvarint.Len(uint64(remainingLen))
 		frame, hasMoreData := str.popStreamFrame(remainingLen, v)
 		if hasMoreData { // put the stream back in the queue (at the end)
-			f.streamQueue = append(f.streamQueue, id)
+			f.streamQueue.PushBack(id)
 		} else { // no more data to send. Stream is not active any more
 			delete(f.activeStreams, id)
 		}
@@ -146,7 +146,7 @@ func (f *framerI) Handle0RTTRejection() error {
 	defer f.mutex.Unlock()
 
 	f.controlFrameMutex.Lock()
-	f.streamQueue = f.streamQueue[:0]
+	f.streamQueue.Clear()
 	for id := range f.activeStreams {
 		delete(f.activeStreams, id)
 	}
--- a/internal/utils/ringbuffer/ringbuffer.go
+++ b/internal/utils/ringbuffer/ringbuffer.go
@@ -1,15 +1,19 @@
 package ringbuffer
 
+// A RingBuffer is a ring buffer.
+// It acts as a heap that doesn't cause any allocations.
 type RingBuffer[T any] struct {
 	ring             []T
 	headPos, tailPos int
 	full             bool
 }
 
+// Init preallocs a buffer with a certain size.
 func (r *RingBuffer[T]) Init(size int) {
 	r.ring = make([]T, size)
 }
 
+// Len returns the number of elements in the ring buffer.
 func (r *RingBuffer[T]) Len() int {
 	if r.full {
 		return len(r.ring)
@@ -20,10 +24,13 @@ func (r *RingBuffer[T]) Len() int {
 	return r.tailPos - r.headPos + len(r.ring)
 }
 
+// Empty says if the ring buffer is empty.
 func (r *RingBuffer[T]) Empty() bool {
 	return !r.full && r.headPos == r.tailPos
 }
 
+// PushBack adds a new element.
+// If the ring buffer is full, its capacity is increased first.
 func (r *RingBuffer[T]) PushBack(t T) {
 	if r.full || len(r.ring) == 0 {
 		r.grow()
@@ -38,6 +45,9 @@ func (r *RingBuffer[T]) PushBack(t T) {
 	}
 }
 
+// PopFront returns the next element.
+// It must not be called when the buffer is empty, that means that
+// callers might need to check if there are elements in the buffer first.
 func (r *RingBuffer[T]) PopFront() T {
 	if r.Empty() {
 		panic("github.com/quic-go/quic-go/internal/utils/ringbuffer: pop from an empty queue")
@@ -66,6 +76,7 @@ func (r *RingBuffer[T]) grow() {
 	r.headPos, r.tailPos, r.full = 0, len(oldRing), false
 }
 
+// Clear removes all elements.
 func (r *RingBuffer[T]) Clear() {
 	var zeroValue T
 	for i := range r.ring {
--- a/internal/utils/ringbuffer/ringbuffer.go
+++ b/internal/utils/ringbuffer/ringbuffer.go
@@ -1,0 +1,75 @@
+package ringbuffer
+
+type RingBuffer[T any] struct {
+	ring             []T
+	headPos, tailPos int
+	full             bool
+}
+
+func (r *RingBuffer[T]) Init(size int) {
+	r.ring = make([]T, size)
+}
+
+func (r *RingBuffer[T]) Len() int {
+	if r.full {
+		return len(r.ring)
+	}
+	if r.tailPos >= r.headPos {
+		return r.tailPos - r.headPos
+	}
+	return r.tailPos - r.headPos + len(r.ring)
+}
+
+func (r *RingBuffer[T]) Empty() bool {
+	return !r.full && r.headPos == r.tailPos
+}
+
+func (r *RingBuffer[T]) PushBack(t T) {
+	if r.full || len(r.ring) == 0 {
+		r.grow()
+	}
+	r.ring[r.tailPos] = t
+	r.tailPos++
+	if r.tailPos == len(r.ring) {
+		r.tailPos = 0
+	}
+	if r.tailPos == r.headPos {
+		r.full = true
+	}
+}
+
+func (r *RingBuffer[T]) PopFront() T {
+	if r.Empty() {
+		panic("github.com/quic-go/quic-go/internal/utils/ringbuffer: pop from an empty queue")
+	}
+	r.full = false
+	t := r.ring[r.headPos]
+	r.ring[r.headPos] = *new(T)
+	r.headPos++
+	if r.headPos == len(r.ring) {
+		r.headPos = 0
+	}
+	return t
+}
+
+// Grow the maximum size of the queue.
+// This method assume the queue is full.
+func (r *RingBuffer[T]) grow() {
+	oldRing := r.ring
+	newSize := len(oldRing) * 2
+	if newSize == 0 {
+		newSize = 1
+	}
+	r.ring = make([]T, newSize)
+	headLen := copy(r.ring, oldRing[r.headPos:])
+	copy(r.ring[headLen:], oldRing[:r.headPos])
+	r.headPos, r.tailPos, r.full = 0, len(oldRing), false
+}
+
+func (r *RingBuffer[T]) Clear() {
+	var zeroValue T
+	for i := range r.ring {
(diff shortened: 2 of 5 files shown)


marten-seemann (commented):
Nice, thank you @Glonee! Could you add some test cases for the ring buffer?

Glonee:
Sure, should I write the test without Genkgo? Since we are going to get rid of it.

marten-seemann:
Yes, please still write with Ginkgo. I imagine these tests won't be very complicated, and we can hopefully use AI at some point to rewrite them. The more complicated tests (integration tests, connection tests, etc.) will probably require some manual work though.

marten-seemann:
However, when writing benchmarks, please use normal Go benchmarking. A benchmark test would be a great idea for this component, actually :)

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,60 @@
...
+	r.tailPos++
+	if r.tailPos == len(r.ring) {
+		r.tailPos = 0
+	}
+	if r.tailPos == r.headPos {
+		r.full = true
+	}
+}
+
+func (r *RingBuffer[T]) PopFront() T {
```
marten-seemann, line 37: What happens if the buffer is empty? Should this be a `Pop() (_ T, ok bool)`?

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,60 @@
...
+	r.full = false
+	t := r.ring[r.headPos]
+	r.headPos++
```
marten-seemann, line 40: To be on the safe side, should we overwrite the existing element with the zero value?

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,60 @@
...
+	oldRing := r.ring
+	newSize := len(oldRing) * 2
+	r.ring = make([]T, newSize)
+	headLen := copy(r.ring, oldRing[r.headPos:])
+	copy(r.ring[headLen:], oldRing[:r.headPos])
+	r.headPos, r.tailPos, r.full = 0, len(oldRing), false
+}
+
+func (r *RingBuffer[T]) Clear() {
+	r.headPos, r.tailPos, r.full = 0, 0, false
```
marten-seemann, line 59: Here too?

internal/utils/ringbuffer/ringbuffer.go
marten-seemann: > I was thinking effeciency is all we care about, since it is a internal package.

Kind of, but this is really easy to miss. This only adds a single if statement, which the branch predictor will take care of anyway, so the performance impact is likely negligible.

Do you see a big difference in the benchmark test?

Jorropo (changes requested):

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,72 @@
+package ringbuffer
+
+type RingBuffer[T any] struct {
+	zeroValue        T
```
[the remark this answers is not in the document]
Glonee: Thank you for pointing it out! My bad.

Glonee:
I did a test on a few queue implementions. Code can be found here: https://github.com/Glonee/QueueBench.
```
BenchmarkTwoStagesQueue-2                       1000000000	         2.724 ns/op	       0 B/op	       0 allocs/op
BenchmarkTwoStagesQueueWithGrow-2               1000000000	        10.25 ns/op	      32 B/op	       0 allocs/op
BenchmarkRingBuffer-2                   	1000000000	         4.057 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferWithGrow-2           	1000000000	        13.34 ns/op	      31 B/op	       0 allocs/op
BenchmarkRingBufferWithCheck-2          	1000000000	         5.832 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferWithCheckAndGrow-2   	1000000000	        16.27 ns/op	      31 B/op	       0 allocs/op
BenchmarkSlice-2                        	1000000000	         4.329 ns/op	      15 B/op	       0 allocs/op
BenchmarkSliceWithGrow-2                	1000000000	        10.81 ns/op	      44 B/op	       0 allocs/op
```
It seems like the two stages queue has the best performance, as it is used in the standard library(see https://github.com/golang/go/blob/e4e8f9b8ffff9d1bcbaaf4b98307d0b88c26678f/src/net/http/transport.go#L1264-L1303).
I think we might be able to use it as our queue implemention? Or use ringbuffer anyway.

marten-seemann:
Interesting benchmark results. Where do you think does the performance difference come from? Is it all cache locality?

Why does `BenchmarkTwoStagesQueueWithGrow` allocate though (haven't looked at the code). There shouldn't be any allocations, unless you keep continuously growing the buffer (which doesn't make too much sense to benchmark).

How does the ring buffer perform vs. the two-stage queue for the case where you only add and retrieve a single element to the list?

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,73 @@
+package ringbuffer
+
+type RingBuffer[T any] struct {
+	ring             []T
+	headPos, tailPos int
+	full             bool
```
Jorropo, line 6: This might be dodgy but you could do a short ring optimisation by using `full` as signaling. Due to how the GC work this would only work with pointer free types (like stream ids as you are using it now).
You could make it a `uint8`, have `0` be false `1` be true, and values above being len in an aliased array.
Then you can use `(*T)(unsafe.Add(unsafe.Pointer(r), unsafe.Sizeof(zerovalue) * (r.full - 2)))` (note this code only work if T's size match the pointer size, slightly different expression would always work and assuming I'm not off by one)
You would setup the array as this:
```go
// Setup memory layout like this:
// type RingBuffer struct{
//   arrayPointer ----+
//   len <------------+
//  cap
//  tailPos
//  headPos
//  full (signaling for short ring optimisation)
// }
*r = RingBuffer[T]{ring: unsafe.Slice((*T)(unsafe.Add(unsafe.Pointer(r), unsafe.Sizeof(uintptr(0)))), 0, 0), full: 2}
```
As you pop element from the front you would advance the pointer in the slice:
```go
// type RingBuffer struct{
//   arrayPointer ----+
//   len              |
//  cap <-------------+
//  tailPos
//  headPos
//  full (signaling for short ring optimisation)
// }
```
You wouldn't implement a ring, for such short runs if you ever need to grow you just shift all elements back to the first aliased pointer configuration I've shown.

Also the go compiler makes assumptions about the return values from the `len` and `cap` builtins which isn't compatible with using them as aliased storage so all short operations would need to go through unsafe.

This would be very cache efficient for short ring, assuming 64 bits arch with 64 bits T you could store 4 of them this way (which is the default usage you show in framer).

Jorropo: This is a usual optimisation to find in very well optimised C++ strings.

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,73 @@
...
+// This method assume the queue is full.
+func (r *RingBuffer[T]) grow() {
+	oldRing := r.ring
+	newSize := len(oldRing) * 2
+	if newSize == 0 {
+		newSize = 1
+	}
+	r.ring = make([]T, newSize)
+	headLen := copy(r.ring, oldRing[r.headPos:])
+	copy(r.ring[headLen:], oldRing[:r.headPos])
```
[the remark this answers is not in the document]
Glonee: Do you mean this?
```go
r.ring = make([]T, newSize)
head := append(r.ring[:0], oldRing[r.headPos:]...)
_ = append(head, oldRing[:r.headPos]...)
```

Glonee:
> Why does `BenchmarkTwoStagesQueueWithGrow` allocate though (haven't looked at the code). There shouldn't be any allocations, unless you keep continuously growing the buffer (which doesn't make too much sense to benchmark).

I kept growing the buffer😢
> How does the ring buffer perform vs. the two-stage queue for the case where you only add and retrieve a single element to the list?

Here is the result.
```
BenchmarkTwoStagesQueue-2        	1000000000	         4.777 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBuffer-2            	1000000000	         4.735 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferWithCheck-2   	1000000000	         6.753 ns/op	       0 B/op	       0 allocs/op
BenchmarkSlice-2                 	1000000000	        30.31 ns/op	       8 B/op	       1 allocs/op

```

internal/utils/ringbuffer/ringbuffer.go
marten-seemann: I'm not opposed to using `unsafe` in general, but there are tradeoffs. For starters, every time we find an unexplained crash, we'll come back to the places in our code where `unsafe` is used, so it does create a maintenance burden.

As a rule of thumb, I think we should only use `unsafe` where its critical for _overall_ performance, i.e. having a direct and significant (more than 1%? 2%?) impact on transfer speed.

framer.go
```diff
@@ -95,13 +98,12 @@ func (f *framerI) AppendStreamFrames(frames []*ackhandler.Frame, maxLen protocol
...
 	// pop STREAM frames, until less than MinStreamFrameSize bytes are left in the packet
-	numActiveStreams := len(f.streamQueue)
+	numActiveStreams := f.streamQueue.Len()
 	for i := 0; i < numActiveStreams; i++ {
 		if protocol.MinStreamFrameSize+length > maxLen {
 			break
 		}
-		id := f.streamQueue[0]
-		f.streamQueue = f.streamQueue[1:]
+		id, _ := f.streamQueue.PopFront()
```
marten-seemann, line 106: Handle the error? Panic should be fine, since this can only happen if there's a bug in the queue.

marten-seemann:
> It says when the queue is not too short(i.e length is 1), the two-stage queue has a better performance. This is because the two-stage queue has a simpler `Push` method since it is just a append operation. The ringbuffer's `Push` method have to determine if we are at the end of the ring or wether the queue is full each time.

That makes sense, thank you for the detailed benchmarks.

Having a queue of 1 is a very common case here: If there's only one stream that's sending data, the queue will only ever have one element.

Glonee:
But if the `Pop` method check wether the queue is empty every time, the ringbuffer is slower than two-stage queue in all ways. In the two-stage queue, pop from a empty queue is an out-of-range access to a slice, so it will also panic, and it's faster.

The other option to speed up is to use unsafe. This can aviod unnecessary out-of-range checks in the ringbuffer.

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (r *RingBuffer[T]) PopFront() (t T, ok bool) {
+	if r.Len() == 0 {
+		return
+	}
+	ok = true
+	r.full = false
+	t = r.ring[r.headPos]
+	var zeroValue T
+	r.ring[r.headPos] = zeroValue
```
marten-seemann, line 45:
```suggestion
	r.ring[r.headPos] = *new(T)
```

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,73 @@
...
+	if r.tailPos == len(r.ring) {
+		r.tailPos = 0
+	}
+	if r.tailPos == r.headPos {
+		r.full = true
+	}
+}
+
+func (r *RingBuffer[T]) PopFront() (t T, ok bool) {
+	if r.Len() == 0 {
```
marten-seemann, line 38: This should be correct, right?
Will probably speed up the benchmark.

```suggestion
	if r.tailPos == r.headPos {
```

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,73 @@
...
+		return
+	}
+	ok = true
```
marten-seemann, line 41:
```suggestion
```

internal/utils/ringbuffer/ringbuffer.go
```diff
@@ -0,0 +1,73 @@
...
+	r.full = false
+	t = r.ring[r.headPos]
+	var zeroValue T
+	r.ring[r.headPos] = zeroValue
+	r.headPos++
+	if r.headPos == len(r.ring) {
+		r.headPos = 0
+	}
+	return
```
marten-seemann, line 50:
```suggestion
	return t, true
```

internal/utils/ringbuffer/ringbuffer.go
Glonee: `r.tailPos == r.headPos` means the queue is full or empty. So it should be
```
if !r.full && r.tailPos == r.headPos
```

marten-seemann: Right. Probably still more performant than calling `Len`.

internal/utils/ringbuffer/ringbuffer.go
Glonee: We should panic here since pop from an empty queue must be a bug.

marten-seemann:
How are the benchmarks looking now?

Glonee:
```
BenchmarkTwoStagesQueue-8                       349521519                3.423 ns/op
BenchmarkTwoStagesQueueBuffered-8               524167392                2.216 ns/op
BenchmarkRingBufferWithCheck-8                  325348203                3.635 ns/op
BenchmarkRingBufferWithCheckBuffered-8          305492370                3.925 ns/op
BenchmarkSlice-8                                47949556                21.15 ns/op
BenchmarkSliceBuffered-8                        144393367                8.239 ns/op
```
Faster now but still slower than two-stage queue.

marten-seemann:
@Glonee Why did you close the PR? This was good code!

marten-seemann:
Regarding the benchmark, how do the numbers look if you add a similar check to the queue: https://github.com/Glonee/QueueBench/blob/main/queue.go#L16

Glonee:
Closed by mistake :(
I will reopen it later.

Glonee:
> Regarding the benchmark, how do the numbers look if you add a similar check to the queue

I have updated the benchmark in https://github.com/Glonee/QueueBench.
Here is the result on my laptop.
```
BenchmarkTwoStagesQueue
BenchmarkTwoStagesQueue-8               339773840                3.443 ns/op           0 B/op          0 allocs/op
BenchmarkTwoStagesQueueBuffered
BenchmarkTwoStagesQueueBuffered-8       477947502                2.486 ns/op           0 B/op          0 allocs/op
BenchmarkRingBuffer
BenchmarkRingBuffer-8                   357796366                3.428 ns/op           0 B/op          0 allocs/op
BenchmarkRingBufferBuffered
BenchmarkRingBufferBuffered-8           337476703                3.573 ns/op           0 B/op          0 allocs/op
```

marten-seemann:
Interesting results. I just ran your benchmarks:

M1 Macbook Pro:
```
goos: darwin
goarch: arm64
pkg: queue
BenchmarkTwoStagesQueue-10            	479846928	         2.370 ns/op	       0 B/op	       0 allocs/op
BenchmarkTwoStagesQueueBuffered-10    	563241345	         2.192 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBuffer-10                	480112345	         2.548 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferBuffered-10        	469203624	         2.543 ns/op	       0 B/op	       0 allocs/op
PASS
```

Ubuntu cloud server:
```
goos: linux
goarch: amd64
pkg: queue
cpu: AMD EPYC 7542 32-Core Processor
BenchmarkTwoStagesQueue-2           	162548380	         7.373 ns/op	       0 B/op	       0 allocs/op
BenchmarkTwoStagesQueueBuffered-2   	184529250	         6.631 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBuffer-2               	196241442	         6.180 ns/op	       0 B/op	       0 allocs/op
BenchmarkRingBufferBuffered-2       	194265243	         6.172 ns/op	       0 B/op	       0 allocs/op
```

marten-seemann (approved):
Added some documentation for the ringer buffer. Once applied, this should be good to go.

Thank you @Glonee for your efforts here, this was really helpful!

marten-seemann (approved):

Issue #3850, opened 2023-05-22
Respect minimum idle timeout of 3 PTOs


birneee:
Currently the idle timeout can be lower than 3 PTOs.

RFC 9000 Section 10.1 states:
> [...] To avoid excessively small idle timeout periods, endpoints **MUST** increase the idle timeout period to be at least three times the current Probe Timeout (PTO). [...]

This is a for example a problem if `Config.MaxIdleTimeout` is set to a value below the path RTT. 
The connection then times out immediately after the handshake.

Suggestion: add `connection.nextIdleTimeoutTime()` that respects the lower bound.
This should also affect the `connection.nextKeepAliveTime()`.
The `Config.HandshakeIdleTimeout` should remain unaffected, since no RTT measurements are available anyways.

marten-seemann:
Makes sense. Want to submit a patch?

marten-seemann:
@birneee I had to revert #3859, since it was causing CI failures on master (changed mock interface).

Could you reopen the PR, rebased on top of the current master. Your changes were good, and I'd like to get them into the next release.

PR #3859, opened 2023-05-24
Respect minimum idle timeout of 3 PTOs

birneee:
#3850

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -194,6 +194,7 @@ type connection struct {
 	versionNegotiated   bool
 	receivedFirstPacket bool
 
+	// the minimum of the max_idle_timeout values advertised by both endpoints
 	idleTimeout  time.Duration
 	creationTime time.Time
 	// The idle timeout is set based on the max of the time we received the last packet...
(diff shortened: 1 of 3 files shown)


marten-seemann:
The code lgtm, but I'm seeing an idle-timeout-related flaky test now:
```
Timeout tests
  does not time out if keepalive is set
  /home/runner/work/quic-go/quic-go/integrationtests/self/timeout_test.go:317
------------------------------
• [FAILED] [0.650 seconds]
Timeout tests
/home/runner/work/quic-go/quic-go/integrationtests/self/timeout_test.go:45
  [It] does not time out if keepalive is set
  /home/runner/work/quic-go/quic-go/integrationtests/self/timeout_test.go:317

  Expected an error, got nil
  In [It] at: /home/runner/work/quic-go/quic-go/integrationtests/self/timeout_test.go:371

  Full Stack Trace
    github.com/quic-go/quic-go/integrationtests/self_test.glob..func24.7()
    	/home/runner/work/quic-go/quic-go/integrationtests/self/timeout_test.go:371 +0x5ac
```

I don't think this test was flaky before this change.

birneee:
The problem is that the configured idle timeout of 100 ms can now be shorter than the minimum timeout calculated from the PTO.
Therefore, the timeout in the integration tests is no longer predictable.

On my machine, for those tests, the PTO is usually at 30 ms, but I think it likely fails under slightly different circumstances.

This makes the following test in `timeout_test.go` flaky:
- returns net.Error timeout errors when an idle timeout occurs
- times out after inactivity
- times out after sending a packet
- does not time out if keepalive is set

A simple fix would be to increase the used idleTimeout to 500 ms or 1 s, which should work probably on every machine.

marten-seemann (approved):

marten-seemann:
Reverted on master (via force push), since it was causing CI to fail.

PR #3866, opened 2023-05-31
config: handle overflows of stream and flow control limits

marten-seemann:
This will make implementing https://github.com/libp2p/go-libp2p/issues/2290 more convenient.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,13 +1,13 @@
 package quic
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"time"
 
 	"github.com/quic-go/quic-go/internal/protocol"
 	"github.com/quic-go/quic-go/internal/utils"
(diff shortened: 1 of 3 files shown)


MarcoPolo:
How does it make it more convenient?

MarcoPolo:
I'm generally wary of changing the config silently from the user. Although, this is being at the limits, I wouldn't expect much of a behavior change except when someone sets this to max int.

Would it be better to expose the max values that can be passed in instead?

marten-seemann:
> Would it be better to expose the max values that can be passed in instead?

That would be an option as well. The max is `quicvarint.Max`, and we could error if the value is larger than that. Although this wouldn't surface the error when using `GetConfigForClient`.

marten-seemann:
You have a good point that this is kind of magic behavior though. I'll add a sentence to the documentation to make it less magical.

PR #3867, opened 2023-05-31
http3: set tls.Config.ServerName for outgoing requests, if unset

### LINKED ISSUE #3865: v0.35.0 broke http3.RoundTripper and servername from url logic
Since v0.35.0 we have to manually set `tls.Config.ServerName` for any http3 connections to work.
This broke the backward compatibility with stdlib and can lead to numerous issues because we have to keep in sync `ServerName` in `tls.Config` and the url in the `http.Request` albeit these are in completely separated levels in the net/http API logic.

the cultrip is here: [client.go/newClient()](https://github.com/quic-go/quic-go/commit/d683b841c4aa8b8b1434ebd27cebc43be7c2ba16#diff-4b667feae66c9d46b21b9ecc19e8958cf4472d162ce0a47ac3e8386af8bbd8cfL217)

Glonee:
fix #3865
bug caused by: https://github.com/quic-go/quic-go/commit/d683b841c4aa8b8b1434ebd27cebc43be7c2ba16#diff-4b667feae66c9d46b21b9ecc19e8958cf4472d162ce0a47ac3e8386af8bbd8cfL231-L239
Copied this logic into http3 package.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -99,7 +99,6 @@ func newClient(hostname string, tlsConf *tls.Config, opts *roundTripperOpts, con
 			// It's ok if net.SplitHostPort returns an error - it could be a hostname/IP address without a port.
 			sni = hostname
 		}
-
 		tlsConf.ServerName = sni
 	}
 	// Replace existing ALPNs by H3
--- a/http3/client.go
+++ b/http3/client.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"strconv"
 	"sync"
@@ -91,6 +92,15 @@ func newClient(hostname string, tlsConf *tls.Config, opts *roundTripperOpts, con
 		tlsConf = &tls.Config{}
 	} else {
 		tlsConf = tlsConf.Clone()
+	}
+	if tlsConf.ServerName == "" {
+		sni, _, err := net.SplitHostPort(hostname)
+		if err != nil {
+			// It's ok if net.SplitHostPort returns an error - it could be a hostname/IP address without a port.
+			sni = hostname
+		}
+
+		tlsConf.ServerName = sni
 	}
 	// Replace existing ALPNs by H3
 	tlsConf.NextProtos = []string{versionToALPN(conf.Versions[0])}
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -90,8 +90,8 @@ var _ = Describe("Client", func() {
 		Expect(dialAddrCalled).To(BeTrue())
 	})
 
-	It("set ServerName in the tls.Config, if not set", func() {
-		host := "foo.bar"
+	It("sets the ServerName in the tls.Config, if not set", func() {
+		const host = "foo.bar"
 		dialCalled := false
 		dialFunc := func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error) {
 			Expect(tlsCfg.ServerName).To(Equal(host))
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -90,6 +90,22 @@ var _ = Describe("Client", func() {
 		Expect(dialAddrCalled).To(BeTrue())
 	})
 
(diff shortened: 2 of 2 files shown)


marten-seemann:
@kgersen Could you check if this PR fixes the regression you reported?

kgersen:
partially. it works for simple cases. but when the connection is reused, for instance when following a redirect, it panics:

    go run example/client/main.go -q https://www.google.com/
ok

but

    go run example/client/main.go -q https://google.com/

panic: connection already exists.

it's the default behavior of std lib http client to follow redirects so http3 client should too.

I haven't looked further why this panic occurs. if @Glonee can look at too ? otherwise I'll try to fix it.

v0.34.0 was fine and handled the redirect fine.

**side note**: the new 'Transport' (`quic.Transport`) introduced in v0.35.0 add a bit of confusion. In  std lib, `Transport` is usually understood as `http.Transport` (or `http2.Transport`), ie the implementation of HTTP/x protocol over TCP. Here it's called `http3.RoundTripper` and `quic.Transport` is something not specific to HTTP/3. 
It's fine, it is just names and different packages but it's not very consistent with the stdlib.

marten-seemann:
@kgersen Thanks for testing.

> panic: connection already exists.

The panic comes from here: https://github.com/quic-go/quic-go/blob/v0.35.0/multiplexer.go#L59. It shows that the http3 package is not using the `quic.Transport` correctly. That's a very annoying oversight, and shouldn't have made it into the final release. We definitely should have a test case for that (I'll create one right away).

> side note: the new 'Transport' (quic.Transport) introduced in v0.35.0 add a bit of confusion. In std lib, Transport is usually understood as http.Transport (or http2.Transport), ie the implementation of HTTP/x protocol over TCP. Here it's called http3.RoundTripper and quic.Transport is something not specific to HTTP/3.

Noted. Improving documentation (both GoDoc and README) is on my TODO list.

Glonee:
@kgersen Can you check if the problem still exists?

kgersen:
seems ok. thanks.
Both the provided example and our product (once we adapted / changed from using net.UDPConn to using quic.Transport) worked fine.
We didn't launch extensive tests, just our basic ones. I'll conduct more tests next week.

marten-seemann:
@Glonee Can you split the last 2 commits into a separate PR? It would also be great if we had a HTTP integration test that dials two QUIC connections. Any idea?

Glonee:
@marten-seemann  `localhost` and `127.0.0.1` are considered as different server names. Maybe we can dial to `localhost` and `127.0.0.1` to make two QUIC connections?

marten-seemann:
Sounds like a good idea. You could change the self-signed certificate to be valid for both the IP and localhost: https://github.com/quic-go/quic-go/blob/fce0261c3433532fecc88bd322f985b55c2381df/integrationtests/tools/crypto.go#L46

http3/client_test.go
```diff
@@ -90,6 +90,22 @@ var _ = Describe("Client", func() {
 		Expect(dialAddrCalled).To(BeTrue())
 	})
 
+	It("set ServerName in the tls.Config, if not set", func() {
```
marten-seemann, line 93:
```suggestion
	It("sets the ServerName in the tls.Config, if not set", func() {
```

http3/client_test.go
```diff
@@ -90,6 +90,22 @@ var _ = Describe("Client", func() {
...
+		host := "foo.bar"
```
marten-seemann, line 94:
```suggestion
		const host = "foo.bar"
```

http3/client.go
```diff
@@ -92,6 +93,15 @@ func newClient(hostname string, tlsConf *tls.Config, opts *roundTripperOpts, con
 	} else {
 		tlsConf = tlsConf.Clone()
 	}
+	if tlsConf.ServerName == "" {
+		sni, _, err := net.SplitHostPort(hostname)
+		if err != nil {
+			// It's ok if net.SplitHostPort returns an error - it could be a hostname/IP address without a port.
+			sni = hostname
+		}
+
```
marten-seemann, line 102:
```suggestion
```

marten-seemann (approved):
Thank you @Glonee!

PR #3869, opened 2023-05-31
http3: correctly use the quic.Transport

### LINKED ISSUE #3868: http3: incorrect use of the quic.Transport
@kgersen reports in https://github.com/quic-go/quic-go/pull/3867#issuecomment-1570038847 that the http3 package is not correctly reusing connections, leading to panics in the multiplexer: https://github.com/quic-go/quic-go/blob/v0.35.0/multiplexer.go#L59

Glonee:
Fixes #3868. Depends on #3867

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -17,8 +17,6 @@ import (
 	"github.com/quic-go/quic-go"
 )
 
-// declare this as a variable, such that we can it mock it in the tests
-
 type roundTripCloser interface {
 	RoundTripOpt(*http.Request, RoundTripOpt) (*http.Response, error)
 	HandshakeComplete() bool
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -18,7 +18,6 @@ import (
 )
 
 // declare this as a variable, such that we can it mock it in the tests
-var quicDialer = quic.DialEarly
 
 type roundTripCloser interface {
 	RoundTripOpt(*http.Request, RoundTripOpt) (*http.Response, error)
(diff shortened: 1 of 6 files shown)


Glonee:
The integration test won't pass until #3867 is merged.

marten-seemann:
I merged #3867. Could you rebase this PR?

http3/roundtrip.go
```diff
@@ -18,7 +18,6 @@ import (
 )
 
 // declare this as a variable, such that we can it mock it in the tests
```
marten-seemann, line 20:
```suggestion
```

integrationtests/self/http_test.go
```diff
@@ -122,6 +123,21 @@ var _ = Describe("HTTP tests", func() {
 		Expect(string(body)).To(Equal("Hello, World!\n"))
 	})
 
+	It("request to server with two different servrname", func() {
```
marten-seemann, line 126: I'd like to explicitly check that this was using the same UDP conn. What do you think of introducing an HTTP handler on the server side that outputs the remote address, and then checking that both requests came from the same address?

Glonee: I will do it right away.

Glonee:
Question: Does the `quic.Transport.Close()` close the underlying `UDPConn`? Otherwise we should close it manually in the `http3.RoundTripper.Close()`.

marten-seemann:
> Question: Does the `quic.Transport.Close()` close the underlying `UDPConn`? Otherwise we should close it manually in the `http3.RoundTripper.Close()`.

It doesn't. Want to send another PR before I cut v0.35.1?

Issue #3822, opened 2023-05-13
use a single ACK frame struct for sending and receiving (per connection)


marten-seemann:
It should be possible to use a single ACK frame struct for sending and receiving (each) per connection. This would allow us to get rid of the ACK frame `sync.Pool`.

**On the sender side**: When packing a packet with an ACK frame, fill the ACK frame struct, serialize it. There's no need to save the ACK frame with the sent packet. We only need to save some meta data, like the highest acked. We already do that.

**On the receiver side**: If a packet contains an ACK frame, use the ACK frame struct, parse the frame into that struct, and process it before touching the next packet. There's a little caveat here. The peer might send us a packet containing multiple ACK frames. There's absolutely no point in doing that, but it's allowed by RFC 9000, so we need to be able to handle that corner case.

This can be done by introducing a `AckFrame.Merge(AckFrame)` method. This can most likely be done in `O(N)`, where `N` is the number of ACK ranges.

It might even be possible to use a single ACK frame struct for the send and the receive path, combined, as we never process incoming packets and send out packets concurrently.

marten-seemann:
We don't even need the `Merge` function. Frames are processed one by one.

This also means that we can easily pull off the same trick for other frame types.

PR #3878, opened 2023-06-02
use a single ACK frame in the receive path

marten-seemann:
Part of #3822.


connection.go
```diff
@@ -1274,27 +1278,27 @@ func (s *connection) handleFrames(
...
-		// If we're logging, we need to make sure that the packet_received event is logged first.
-		if log == nil {
-			if err := s.handleFrame(frame, encLevel, destConnID); err != nil {
+		if log != nil {
+			frames = append(frames, logutils.ConvertFrame(frame))
+			// An error occurred handling a previous frame.
+			// Don't handle the current frame.
+			if handleErr != nil {
+				continue
+			}
```
sukunrt, line 1287: should this check be outside this block and before 
`if err := s.handleFrame(frame, encLevel, destConnID); err != nil`

sukunrt (approved):

connection.go
marten-seemann: Done.

PR #3895, opened 2023-06-12
Comment accuracy

arobase-che:


arobase-che:
Yes, but it's unclear that it listen on TCP port.
It's the only place in the code that it does and you have to look at the code to understand it.

PR #3900, opened 2023-06-14
http3: return http.ErrServerClosed instead of quic.ErrServerClosed

### LINKED ISSUE #3898: http3 - when closing, http3.Server.Serve returns quic.ErrServerClosed instead of http.ErrServerClosed
The returned error for `http3.Server.Serve`(...) is `quic.ErrServerClosed` in case of a normal shutdown/close. 
It should be `http.ErrServerClosed` instead. 
see method [serveConn](https://github.com/quic-go/quic-go/blob/master/http3/server.go#L276) (some return paths already use `http.ErrServerClosed`).

jfgiorgi:
fix #3898

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -274,6 +274,14 @@ func (s *Server) ServeListener(ln QUICEarlyListener) error {
 var errServerWithoutTLSConfig = errors.New("use of http3.Server without TLSConfig")
 
 func (s *Server) serveConn(tlsConf *tls.Config, conn net.PacketConn) error {
+	err := s.serveC(tlsConf, conn)
+	// replace quic package specific error with net/http equivalent ones
+	if err == quic.ErrServerClosed {
+		err = http.ErrServerClosed
+	}
+	return err
+}
+func (s *Server) serveC(tlsConf *tls.Config, conn net.PacketConn) error {
 	if tlsConf == nil {
 		return errServerWithoutTLSConfig
 	}
--- a/http3/server.go
+++ b/http3/server.go
@@ -262,27 +262,31 @@ func (s *Server) ServeQUICConn(conn quic.Connection) error {
 // Make sure you use http3.ConfigureTLSConfig to configure a tls.Config
 // and use it to construct a http3-friendly QUIC listener.
 // Closing the server does close the listener.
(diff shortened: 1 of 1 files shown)


http3/server.go
```diff
@@ -274,6 +274,14 @@ func (s *Server) ServeListener(ln QUICEarlyListener) error {
 var errServerWithoutTLSConfig = errors.New("use of http3.Server without TLSConfig")
 
 func (s *Server) serveConn(tlsConf *tls.Config, conn net.PacketConn) error {
+	err := s.serveC(tlsConf, conn)
+	// replace quic package specific error with net/http equivalent ones
+	if err == quic.ErrServerClosed {
+		err = http.ErrServerClosed
+	}
+	return err
+}
```
marten-seemann, line 283:
```suggestion
}

```

marten-seemann (approved):

http3/server.go
```diff
@@ -274,6 +274,15 @@ func (s *Server) ServeListener(ln QUICEarlyListener) error {
...
+
+func (s *Server) serveC(tlsConf *tls.Config, conn net.PacketConn) error {
```
marten-seemann, line 285: Do we need this new function? `quic.ErrServerClosed` can only be returned from `serveListener`, so it should be fine to just apply this logic for the last error returned, no?

jfgiorgi: Yes that was my first change initially.
The new function is more about "future proofing" allowing to catch all return paths and allowing to have one single place to eventually transform more quic errors into http errors if needed later.
up to you to choose which one you want.

marten-seemann: I think I'd prefer to just catch the error at the one place where it actually happens.

Also, can we have a test please? :)

jfgiorgi: I've reworked the code a bit. serveListener isn't needed anymore, it can be inlined in ServeListener  which can be used directly  in the original s.serveConn. I've updated the doc too.
I don't see what test can be add value here.
What is important is notify in the changelog that this is an API breaking change for people using the quic-go http3 package, since they  cannot test/check for `quic.ErrServerClosed` anymore.

marten-seemann (approved):

PR #3902, opened 2023-06-15
docs: add detailed description of the quic package

marten-seemann:
Left for follow-up work:
* describe how 0.5-RTT and 0-RTT work
* add a detailed description of the http3 package

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,15 +68,15 @@ defer cancel()
 conn, err := quic.Dial(ctx, conn, <server address>, <tls.Config>, <quic.Config>)
 ```
 
-When using this shortcut, it's not possible to reuse the same UDP socket for other outgoing connections, or to listen for incoming connections.
+Just as we saw before when used a similar shortcut to run a server, it's also not possible to reuse the same UDP socket for other outgoing connections, or to listen for incoming connections.
 
 ### Using a QUIC Connection
 
 #### Accepting Streams
 
-QUIC is a stream-multiplexed transport. A `quic.Connection` fundamentally differs from the `net.Conn` and the `net.PacketConn` interface defined in the standard library. Data is sent and received on (unidirectional and bidirectional) streams (and, if supported, in [datagrams](#quic-datagrams)), not on the connection itself. The stream state machine is described in details in [Section 3 of RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-3).
+QUIC is a stream-multiplexed transport. A `quic.Connection` fundamentally differs from the `net.Conn` and the `net.PacketConn` interface defined in the standard library. Data is sent and received on (unidirectional and bidirectional) streams (and, if supported, in [datagrams](#quic-datagrams)), not on the connection itself. The stream state machine is described in detail in [Section 3 of RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-3).
 
-A unidirectional stream is a stream that the initiator can only write to (`quic.SendStream`), and the receiver can only read from (`quic.ReceiveStream`). A bidirectional stream (`quic.Stream`) allows reading from and writing to for both sides.
+Note: A unidirectional stream is a stream that the initiator can only write to (`quic.SendStream`), and the receiver can only read from (`quic.ReceiveStream`). A bidirectional stream (`quic.Stream`) allows reading from and writing to for both sides.
 
 On the receiver side, streams are accepted using the `AcceptStream` (for bidirectional) and `AcceptUniStream` functions. For most user cases, it makes sense to call these functions in a loop:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,30 +7,183 @@
...
+This repository provides both a QUIC implementation, located in the `quic` package, as well as an HTTP/3 implementation, located in the `http3` package.
 
-*We currently support Go 1.19.x and Go 1.20.x*
+## Using QUIC
 
-Running tests:
+### Running a Server
 
-    go test ./...
+The central entry point is the `quic.Transport`. A transport manages QUIC connections running on a single UDP socket. Since QUIC uses Connection IDs, it can demultiplex a listener (accepting incoming connections) and an arbitrary number of outgoing connections on the same UDP socket.
```
p-shahi, line 23:
```suggestion
The central entry point is the `quic.Transport`. A transport manages QUIC connections running on a single UDP socket. Since QUIC uses Connection IDs, it can demultiplex a listener (accepting incoming connections) and an arbitrary number of outgoing connections on the same UDP socket.
```
"and an arbitrary number of outgoing connections on the same UDP socket"

missed a word after `and an` ?

README.md
```diff
@@ -7,30 +7,183 @@
...
+
+As a shortcut, `quic.Dial` and `quic.DialAddr` can be used without explictly initializing a `quic.Transport`:
+
+```go
+ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second) // 3s handshake timeout
+defer cancel()
+conn, err := quic.Dial(ctx, conn, <server address>, <tls.Config>, <quic.Config>)
+```
+
+When using this shortcut, it's not possible to reuse the same UDP socket for other outgoing connections, or to listen for incoming connections.
```
p-shahi, line 71:
```suggestion
Just as we saw before when used a similar shortcut to run a server, it's also not possible to reuse the same UDP socket for other outgoing connections, or to listen for incoming connections.
```

README.md
```diff
@@ -7,30 +7,183 @@
...
 
-Take a look at [this echo example](example/echo/echo.go).
+### Using a QUIC Connection
+
+#### Accepting Streams
+
+QUIC is a stream-multiplexed transport. A `quic.Connection` fundamentally differs from the `net.Conn` and the `net.PacketConn` interface defined in the standard library. Data is sent and received on (unidirectional and bidirectional) streams (and, if supported, in [datagrams](#quic-datagrams)), not on the connection itself. The stream state machine is described in details in [Section 3 of RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-3).
```
p-shahi, line 77:
```suggestion
QUIC is a stream-multiplexed transport. A `quic.Connection` fundamentally differs from the `net.Conn` and the `net.PacketConn` interface defined in the standard library. Data is sent and received on (unidirectional and bidirectional) streams (and, if supported, in [datagrams](#quic-datagrams)), not on the connection itself. The stream state machine is described in detail in [Section 3 of RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-3).
```

README.md
```diff
@@ -7,30 +7,183 @@
...
+
+A unidirectional stream is a stream that the initiator can only write to (`quic.SendStream`), and the receiver can only read from (`quic.ReceiveStream`). A bidirectional stream (`quic.Stream`) allows reading from and writing to for both sides.
```
p-shahi, line 79:
```suggestion
Note: A unidirectional stream is a stream that the initiator can only write to (`quic.SendStream`), and the receiver can only read from (`quic.ReceiveStream`). A bidirectional stream (`quic.Stream`) allows reading from and writing to for both sides.
```

README.md
```diff
@@ -7,30 +7,183 @@
...
+  // ... error handling
+  // handle the stream, usually in a new Go routine
+}
+```
+
+These functions return an error when the underlying QUIC connection is closed.
+
+#### Opening Streams
+
+There are two slightly different ways to open streams, one synchronous and one (potentially) asynchronous. The reason for this is that the receiver grants us a certain number of streams that we're allowed to open. It may grant us additional streams later on (typically when existing streams are closed), but it means that at the time we want to open a new stream, we might not be able to do so.
```
p-shahi, line 95:
```suggestion
There are two slightly different ways to open streams, one synchronous and one (potentially) asynchronous. The reason for this is that the receiver grants us a certain number of streams that we're allowed to open. It may grant us additional streams later on (typically when existing streams are closed), but it means that at the time we want to open a new stream, we might not be able to do so.
```
"The reason for this" -> are you referring to the async way? - it is a bit ambiguous

README.md
```diff
@@ -7,30 +7,183 @@
...
+conn.CloseWithError(0x42, "error 0x42 occurred")
+```
+
+Applications can transmit both an error code (an unsigned 62-bit number) as well as a UTF-8 encoded human-readable reason. The error code allows the receiver to learn why the connection was closed, and the reason can be useful for debugging purposes.
+
+On the receiver side, this is surfaced as a `quic.ApplicationError`.
+
+### QUIC Datagrams
+
+Unreliable datagrams is a QUIC extension ([RFC 9221](https://datatracker.ietf.org/doc/html/rfc9221)) that is negotiated during the handshake. Support can be enabled by setting the `quic.Config.EnableDatagram` flag. Note that this doesn't guarantee that the peer also supports datagrams. If the feature negotiation was successful can be learned from the `quic.ConnectionState.SupportsDatagrams` obtained from `quic.Connection.ConnectionState()`.
```
p-shahi, line 162:
```suggestion
Unreliable datagrams are a QUIC extension ([RFC 9221](https://datatracker.ietf.org/doc/html/rfc9221)) that is negotiated during the handshake. Support can be enabled by setting the `quic.Config.EnableDatagram` flag. Note that this doesn't guarantee that the peer also supports datagrams. Whether or not the feature negotiation succeeded can be learned from the `quic.ConnectionState.SupportsDatagrams` obtained from `quic.Connection.ConnectionState()`.
```

README.md
```diff
@@ -7,30 +7,183 @@
...
+
+QUIC DATAGRAMs are a new QUIC frame type sent in QUIC 1-RTT packets (i.e. after completion of the handshake). They're therefore end-to-end encrypted and congestion-controlled. However, if a DATAGRAM frame is deemed lost by QUIC's loss detection mechanism, they are not automatically retransmitted.
```
p-shahi, line 164:
```suggestion
QUIC DATAGRAMs are a new QUIC frame type sent in QUIC 1-RTT packets (i.e. after completion of the handshake). Therefore, they're end-to-end encrypted and congestion-controlled. However, if a DATAGRAM frame is deemed lost by QUIC's loss detection mechanism, they are not automatically retransmitted.
```

README.md
SahibYar: would be great if we can add a little more details of what "asynchronous streams" something like 
> An asynchronous stream is a sequence of asynchronous events i.e. instead of getting the next event when you ask for it, the stream tells you that there is an event when it is ready. The word asynchronous means that the data emitted can appear anywhere in time, after one milli-seconds or even after 10 seconds (if connection is not lost and stream is not closed)

marten-seemann: @SahibYar That's not how streams work.

README.md
```diff
@@ -7,30 +7,183 @@
...
+  // but it might be possible later, once the peer allowed us to do so.
+}
+```
 
-## Usage
+These functions return an error when the underlying QUIC connection is closed.
+
+#### Using Streams
+
+Using QUIC streams is pretty straightforward. The `quic.ReceiveStream` implements the `io.Reader` interface, and the `quic.SendStream` implements the `io.Writer` interface. A bidirectional stream (`quic.Stream`) implements both these interfaces. Conceptually, a bidirectional stream can be thought of as the composition of two bidirectional streams in opposite directions.
```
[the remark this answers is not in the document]
marten-seemann: Fixed.

README.md
marten-seemann: Not sure I understand the comment, even after re-reading the sentence for 10 sentences. I added "QUIC connections", hope that clarifies things.

README.md
SahibYar: is it right to add description of "async streams" in context of quic-go ?

marten-seemann: We don't have async streams.

lucas-clemente (approved):

Issue #3906, opened 2023-06-18
(*Transport.)close: nil pointer dereference of handlerMap


kelmenhorst:
Hi, 
I noticed that in some cases, `(*Transport.)close` causes a nil pointer dereference due to `(Transport.)handlerMap` being uninitialized. 

Minimal example to reproduce when using [DialEarly](https://github.com/quic-go/quic-go/blob/85764da3b65997a229a961d9704a03e6c02a682b/client.go#LL90C6-L90C15):
```
addr, _ := net.ResolveUDPAddr("udp", "google.com:443")
conn, _ := net.ListenUDP("udp", &net.UDPAddr{IP: net.IPv4zero, Port: 0, Zone: ""})
var invalidVersion quic.VersionNumber = 0x3      // <-- there is no v3, so validateConfig will fail
qconf := &quic.Config{
	Versions: []quic.VersionNumber{
		invalidVersion,
	},
}
quic.DialEarly(context.Background(), conn, addr, <*tls.Config>, qconf)
```

I get the following output:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x837680]

goroutine 18 [running]:
testing.tRunner.func1.2({0xa96480, 0xffe130})
	/usr/local/go/src/testing/testing.go:1526 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1529 +0x39f
panic({0xa96480, 0xffe130})
	/usr/local/go/src/runtime/panic.go:884 +0x213
github.com/quic-go/quic-go.(*Transport).close(0xc0003c41c0, {0xc6d140, 0xc0003943d0})
	/home/kelmenhorst/go/pkg/mod/github.com/quic-go/quic-go@v0.35.1/transport.go:270 +0x100
github.com/quic-go/quic-go.(*Transport).Close(0xc0003c41c0)
	/home/kelmenhorst/go/pkg/mod/github.com/quic-go/quic-go@v0.35.1/transport.go:232 +0x6b
github.com/quic-go/quic-go.DialEarly({0xc73a70, 0xc00003e0b8}, {0xc76620?, 0xc0003960a0?}, {0xc71358, 0xc000392a80}, 0xc0003c0300, 0x55cff4?)
[...]

PR #3908, opened 2023-06-19
check for uninitialized fields when closing the Transport

kelmenhorst:
The underlying observation that motivates this patch is that `(*Transport.)Close` and `(*Transport.)close` try to access possibly uninitialized fields of the `Transport` which can lead to nil pointer dereference failures (#3906) when the Transport's `init` function is not (fully) executed.

To fix this, I added respective checks in `(*Transport.)Close` and `(*Transport.)close`, so that the robustness of calling `(*Transport.)Close` does not depend on whether or how much of `(Transport.)init` was executed before. 
However, it might be that the underlying issue needs to be addressed at another place (I am not familiar enough with the code base to fully evaluate this).

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -244,7 +244,7 @@ func (t *Transport) runSendQueue() {
 func (t *Transport) Close() error {
 	t.close(errors.New("closing"))
 	if t.createdConn {
-		if err := t.conn.Close(); err != nil {
+		if err := t.Conn.Close(); err != nil {
 			return err
 		}
 	} else if t.conn != nil {
--- a/transport.go
+++ b/transport.go
@@ -247,11 +247,13 @@ func (t *Transport) Close() error {
 		if err := t.conn.Close(); err != nil {
 			return err
 		}
-	} else {
+	} else if t.conn != nil {
 		t.conn.SetReadDeadline(time.Now())
 		defer func() { t.conn.SetReadDeadline(time.Time{}) }()
 	}
-	<-t.listening // wait until listening returns
+	if t.listening != nil {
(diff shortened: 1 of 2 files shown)


marten-seemann (commented):
1. Can you please sign the CLA?
2. A test case would be appreciated.

kelmenhorst:
Sorry, I had forgot to refresh the CLA check after signing it.
I added a test case to transport_test.go.

Apart from the nil-checks, I changed `(*Transport.)Close` such that it closes the underlying `Conn` (net.PacketConn), rather than `conn` (rawConn), since `conn` might not be initialized.

Btw, what's the reason for allowing uninitialized `Transport`'s anyway?

marten-seemann:
> Btw, what's the reason for allowing uninitialized Transport's anyway?

What do you mean? It's an exported struct, so users can just use it.

marten-seemann:
> Btw, what's the reason for allowing uninitialized Transport's anyway?

What do you mean? It's an exported struct, so users can just use it.

marten-seemann (approved):
Thank you @kelmenhorst!

kelmenhorst:
> What do you mean? It's an exported struct, so users can just use it.

Yes, I guess I was asking whether it would make sense to not export the struct itself, and rather export a constructor that initializes the Transport.
However, I can imagine that there are use cases that I am not aware of where you need more flexibility with the Transport :)

marten-seemann:
I see the point, but it's nice to have some configuration fields on the `Transport`. And it matches the standard library pattern (e.g. `net.Dialer`).

Issue #3911, opened 2023-06-23
GSO fails on some platforms / interfaces


zllovesuki:
I'm seeing a lot of `INTERNAL_ERROR (local): write udp [::]:443->[REDACTED]:42830: sendmsg: invalid argument` on the server side, causing the connection to be disconnected. They only happen when the clients are `aarch64` (raspberry pi).

Could this be related to the GSO change?

marten-seemann:
GSO would have been my first guess as well. I suspect that for some reason we manage to enable GSO, but don't correctly recognize that it was successfully enabled. This then causes the failure you're seeing because once enabled, the kernel expects the segment length message for the `sendmsg` to be set: https://github.com/quic-go/quic-go/blob/21388c86bb2be0de9b5468d8b77f37281b64fcf0/sys_conn_df_linux.go#L70-L83

Can you try manually disabling GSO via the `QUIC_GO_DISABLE_GSO` environment variable?
https://github.com/quic-go/quic-go/blob/21388c86bb2be0de9b5468d8b77f37281b64fcf0/sys_conn_df_linux.go#L46-L50
This should make the failure go away.

If it's actually due to GSO, can you confirm if the syscall to enable it is successful in
https://github.com/quic-go/quic-go/blob/21388c86bb2be0de9b5468d8b77f37281b64fcf0/sys_conn_df_linux.go#L52-L57

---

If my suspicion is correct, this would mean that we can't rely on the return value of the syscall that enables GSO. Maybe we need to instead read the value back using `GetSockoptInt` in `maybeSetGSO`?

zllovesuki:
It is indeed GSO:

With QUIC_GO_DISABLE_GSO on v0.36.0:
```
$ sudo k3s kubectl -n kube-system logs deployment/specter-tunnel -f
{"level":"info","ts":1687509533.946454,"caller":"client/client.go:150","msg":"Connected to specter server","id":8319038888190,"addr":"..."}
{"level":"info","ts":1687509534.009357,"caller":"client/client.go:366","msg":"Reusing existing client certificate","id":8319038888190,"identity":{"id":8319038888190,"token":"..."}}
{"level":"info","ts":1687509534.8210588,"caller":"client/client.go:150","msg":"Connected to specter server","id":8319038888190,"addr":"..."}
{"level":"info","ts":1687509535.0624511,"caller":"client/client.go:150","msg":"Connected to specter server","id":8319038888190,"addr":"..."}
[...]

PR #3912, opened 2023-06-23
add DisableGSO to quic.Config

jfgiorgi:
The recent GSO optimization doesn't work well on some hardware/vm. 
We have the case with IPv6 inside a lxd container inside a crosvm vm (Linux on Chromebook): the connection fails if GSO is enabled.
There is an environment variable `QUIC_GO_DISABLE_GSO` to deactivate GSO but it's global (all connections of the program) and not in the hand of programmers (a sub package cannot explicitly disable GSO for instance)

This PR  adds the `DisableGSO` boolean field to `quic.Config`to solve this.

It is implemented where `connection.sendConn.capabilities().GSO` is used and not where it is defined so `connection.sendConn.capabilities().GSO` still remains true if supported.


marten-seemann:
I don’t want to do this. This should be automatic. We should figure out how to properly enable GSO and detect if it was successful. Until then, the environment variable seems to be a good pressure valve on platforms where this currently fails.

jfgiorgi:
it's still automatic. it's just an opt-out. it's not related to the issue #3911 
quiche has it.
gvisor too.
in some case it's handy to able to disable GSO for just some connections and not globally. For instance to do benchmarks and see where more optimization can be done. 
Without this we will have to fork `quic-go` for our app which is a lot of work for just a few extra lines in quic-go.

marten-seemann:
> in some case it's handy to able to disable GSO for just some connections and not globally. For instance to do benchmarks and see where more optimization can be done.

I assume that would be for some automated A/B testing done in production? I can see your point, but does that mean we'll have to introduce similar config knobs once we implement GRO, RSS, kernel pacing, pacing within GSO batches, etc.?

What are the other use cases you have in mind?

jfgiorgi:
yes config knobs for all optimization features would be nice to have. may be under a sub struct in quic.Config.

> What are the other use cases you have in mind?

Fairness between clients supporting or not gso. For instance one might want all clients to send to a server without gso so some clients are not faster than others.

And also: 

Futureproofing an issue with a kernel/hardware bug like #3911 on specific platforms is very important.

This allows devs to fix their app quicker than waiting for  quic-go to fix it. 
QUIC_GO_DISABLE_GSO isn't a fix for developers, it's a fix for integrators/devops/sysadmin, etc usually not the people with the source code and/or access to the build chain/tools.
Imagine if quic-go was part of Go std lib and #3911 was discovered 2 months after release ? not sure the Go team would push an emergency Go version to fix it. People impacted would have to wait 4 months for next version. 

Leaving control to devs to turn off GSO in their code is safer. Much like you can turn off HTTP/2 in std lib even if it's on by default (and without using the env var for this).
It allows to leave "breathing time" for quic-go devs to let people adopt and use a new feature and in case of issues there is less urgency to fix it upstream.

That been said, may be future features of quic-go should be released as opt-in first then switched to opt-out after a few months of feedback. A lot of software feature are released like this ("experimental features" in Chrome for instance).

v 0.35.0 : introduce GSO but off by default, use a  knob (in quic.Config) to turn it on (and with the env var too).
let some months go by
v 0.X.0 : switch GSO to on by default. A new knob in quic.Config can be used to turn it off. The old knob is deprecated and will be removed later on (usually years later).

To be honest, I'm not convinced a feature like GSO should be opt-out. A lot of software only use it opt-in, for instance nginx quic ( https://nginx.org/en/docs/quic.html ).

Now I also understand that quic-go is still version 0.x an not 1.x  so not really "production' ready and this way of introducing a new feature could look overkill.

marten-seemann:
> Fairness between clients supporting or not gso. For instance one might want all clients to send to a server without gso so some clients are not faster than others.

Fairness is guaranteed by the congestion controller. If you're able to dump packets on the wire very fast, you'd still be limited by the congestion window. The only situation where GSO gives you an advantage in this respect is if the non-GSO client is CPU-limited and not able to send packets out as quickly as the congestion controller allows. This is not specific to GSO, it applies to any performance optimization.





> Futureproofing an issue with a kernel/hardware bug like #3911 on specific platforms is very important.
> 
> This allows devs to fix their app quicker than waiting for quic-go to fix it. QUIC_GO_DISABLE_GSO isn't a fix for developers, it's a fix for integrators/devops/sysadmin, etc usually not the people with the source code and/or access to the build chain/tools. Imagine if quic-go was part of Go std lib and #3911 was discovered 2 months after release ? not sure the Go team would push an emergency Go version to fix it. People impacted would have to wait 4 months for next version.

I thought this PR was supposed to be independent of #3911. Yes, there's a bug somewhere, and we need to fix it. quic-go should correctly enable GSO, and maybe needs to check if GSO actually works. Let's discuss the fix on #3911. I'll only be satisfied with a fix once the end result is that the user doesn't need to worry about GSO at all.

I'm not sure if I agree with your assumption that - assuming quic-go was part of the standard library - this wouldn't constitute a bug that would warrant a patch release. However, this discussion is very hypothetical: If quic-go was part of the standard library, the release policy regarding "risky" features might well look different, especially if the bug wouldn't qualify for a patch release. But that's not the situation we're in, for better or for worse. We should do what's right for quic-go as an independent project.

You can use the environment variable as a developer as well, you could call `os.Setenv` in `init`. It's not nice, but as a pressure valve it should be more than sufficient.

---

To be clear, I'm not opposed to a config knob in principle, if there's a convincing use case. I'm a bit careful though expanding the API surface.

jfgiorgi:
> Fairness is guaranteed by the congestion controller

That's not this kind of fairness I was talking about. It's about user experience (and congestion controller is another issue we're dealing with everyday and where fairness isn't guaranteed too but we're out of scope here).

if I do an app for Linux, Android, iOS, MacOs and Windows and my clients on Android and Linux have a 'faster' experience than those on Windows, Mac, MacOS it can be problematic to justify/explain to the average user. I might want to do an app where no one has GSO or everyone.

My view here is that `quic-go` is library not a end-user product. As the dev of the end user product, the more knobs the libraires I use have the better for me. That's all.

I also made libraries so I do understand the cost of maintaining a larger API surface and whatnot.

And no the env var is really not something we want to use, we have decades of bad experiences with these.

marten-seemann:
> if I do an app for Linux, Android, iOS, MacOs and Windows and my clients on Android and Linux have a 'faster' experience than those on Windows, Mac, MacOS it can be problematic to justify/explain to the average user. I might want to do an app where no one has GSO or everyone.

I'm having trouble following this reasoning. There's hundreds of different factors that have an effect on transport performance, and that will vary widely between Linux, OSX and Windows. For example, timer resolution (in Go) on Windows are less precise than on Linux, which can have an impact on quic-go's timers (especially pacing). Do we now need to implement a (configurable) lower-resolution timer for Linux, just so things are "fair" between Linux and Windows? As another example, we've been using `recvmmsg` on Linux for years, which also provides a significant speedup.

I also still don't understand why this would be a per-connection option. It seems like you're arguing for a way to globally disable it, which is already possible via the environment variable, and might be solved a bit more cleanly via a package-level or `Transport`-level configuration knob.

kgersen:
I'm not at this level of details or precision. It's just that GSO gives a 3x or 4x boost in speed. A client sending a large file on Android/Linux then using the same app on Windows would notice a huge difference. 

We're wasting time on this for nothing, both of us. There is really no point to that. Let's wait & see, may other `quic-go` users will ask for this or not.

We do need it and we will fork for now and that's fine for me.

jfgiorgi:
closing since it's implemented in quic.Transport (see #3993)

marten-seemann:
It's not, #3993 was closed, not merged.

jfgiorgi:
my bad misread. 
Well then let's keep this one closed too.

PR #3931, opened 2023-06-29
Transport: Remove t.Conn from multiplexer when init fails

kelmenhorst:
This diff removes a net.PacketConn from the multiplexer when (*Transport).init fails.

Background: As far as I am aware, a PacketConn is only removed from the multiplexer, [when (*Transport).listen returns](https://github.com/quic-go/quic-go/blob/435444af7eb81b922a412ee80d4caddff26744de/transport.go#L300).
If (*Transport).init [returns early](https://github.com/quic-go/quic-go/blob/435444af7eb81b922a412ee80d4caddff26744de/transport.go#L190), the listen() goroutine is not started, while the PacketConn was already added to the multiplexer, which results in the inability of the Transport to remove the PacketConn entry from the multiplexer.

### DIFF
--- a/transport_test.go
+++ b/transport_test.go
@@ -6,6 +6,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"net"
+	"syscall"
 	"time"
 
 	mocklogging "github.com/quic-go/quic-go/internal/mocks/logging"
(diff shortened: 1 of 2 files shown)


marten-seemann (commented):
Have you considered calling `getMultiplexer().AddConn` later, after the code paths that can error (maybe even directly before `t.listen`)?

kelmenhorst:
Thanks! Yes, I think your suggestion makes sense, I will change it accordingly.

transport_test.go
```diff
@@ -307,4 +308,27 @@ var _ = Describe("Transport", func() {
 		pconn.EXPECT().Close()
 		Expect(tr.Close()).To(Succeed())
 	})
+
+	It("removes underlying PacketConn from multiplexer after (*Transport).init failed", func() {
```
marten-seemann, line 312:
```suggestion
	It("doesn't add the PacketConn to the multiplexer if (*Transport).init fails", func() {
```

marten-seemann (approved):

PR #3934, opened 2023-06-30
disable GSO unless QUIC_GO_ENABLE_GSO is set

marten-seemann:
Looks like the proper fix for #3911 is not easy. This PR disables GSO unless the user explicitly enables it by setting the QUIC_GO_ENABLE_GSO environment variable.

This could be included in a v0.36.1 patch release.


lucas-clemente (approved):

mholt:
Thanks for this!

FYI the release mentions `QUIC_GO_LOG_LEVEL=true` -- I assume that's a mistake?

marten-seemann:
> FYI the release mentions `QUIC_GO_LOG_LEVEL=true` -- I assume that's a mistake?

Yes. Thanks for proofreading! Just fixed the release notes.

Issue #3851, opened 2023-05-22
add a PING frame, when acknowledging packets during the handshake


marten-seemann:
When acknowledging packets during the handshake, we should add a PING frame to the packet if otherwise we'd just send an ACK frame. This commonly happens when the ClientHello is split across two packets, and one of them is lost.

This will allow the server to generate an RTT sample, speeding up the rest of the handshake.

marten-seemann:
@SahibYar You were asking for good first issue you could pick up. Want to take a look at this one?
Let me know if you need any pointers.

SahibYar:
Sure, thank you @marten-seemann, let me see this.

SahibYar:
Adding these lines inside [handleAckFrame()](https://github.com/quic-go/quic-go/blob/master/connection.go#L1498) in `connection.go` will be sufficient ?
```
if s.perspective == protocol.PerspectiveClient && !s.handshakeConfirmed {
   s.handleHandshakeConfirmed()
   
   // See https://github.com/quic-go/quic-go/issues/3851 for details
   // This is not strictly necessary, but it speeds up the handshake.
   s.queueControlFrame(&wire.PingFrame{})
}
```

marten-seemann:
`handleAckFrame` is called when we receive an ACK frame. This needs to be part of the packet packer, probably somewhere in the `PackCoalescedPacket` code path.

SahibYar:
Please review this PR https://github.com/quic-go/quic-go/pull/3935/files and guide me if it suffice the needs of this ticket.

SahibYar:
@marten-seemann you got the chance to review this PR ?

marten-seemann:
Sorry, pretty busy right now. I haven't forgotten about it.

PR #3935, opened 2023-07-01
#3851 ping frame added in handshake

SahibYar:
Ping Frame added when packet is nil and handShake is not confirmed yet. 

Commenting this line took 108 seconds to complete all tests
image

With this line took 82 seconds to complete all tests
image

### DIFF
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -331,9 +331,14 @@ func (p *packetPacker) PackCoalescedPacket(onlyAck bool, maxPacketSize protocol.
 		initialHdr, initialPayload = p.maybeGetCryptoPacket(maxPacketSize-protocol.ByteCount(initialSealer.Overhead()), protocol.EncryptionInitial, onlyAck, true, v)
 		if initialPayload.length > 0 {
 			size += p.longHeaderPacketLength(initialHdr, initialPayload, v) + protocol.ByteCount(initialSealer.Overhead())
+
+			if size >= maxPacketSize-protocol.MinCoalescedPacketSize {
+				ping := ackhandler.Frame{Frame: &wire.PingFrame{}}
+				initialPayload.frames = append(initialPayload.frames, ping)
+				initialPayload.length += ping.Frame.Length(v)
+			}
 		}
 	}
-
 	// Add a Handshake packet.
 	var handshakeSealer sealer
 	if (onlyAck && size == 0) || (!onlyAck && size < maxPacketSize-protocol.MinCoalescedPacketSize) {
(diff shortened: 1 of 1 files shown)


marten-seemann (changes requested):
I'd prefer to make this change in the packet packer instead of the connection.

Also please don't just modify random files that are unrelated to this PR.

SahibYar:
Updated as per PR comments, similar pattern of time consumption for all tests. 

When lines were commented, it took 116 seconds
image


When lines were not commented, it took 90 seconds on my machine
image

packet_packer.go
```diff
@@ -331,9 +331,14 @@ func (p *packetPacker) PackCoalescedPacket(onlyAck bool, maxPacketSize protocol.
 		initialHdr, initialPayload = p.maybeGetCryptoPacket(maxPacketSize-protocol.ByteCount(initialSealer.Overhead()), protocol.EncryptionInitial, onlyAck, true, v)
 		if initialPayload.length > 0 {
 			size += p.longHeaderPacketLength(initialHdr, initialPayload, v) + protocol.ByteCount(initialSealer.Overhead())
+
+			if size >= maxPacketSize-protocol.MinCoalescedPacketSize {
```
marten-seemann, line 335: This if doesn’t seem correct:
1. You’re not allowed to add a PING if `ackOnly` is set.
2. You should only add a PING if there’s no other frames in the packet.

SahibYar: > You should only add a PING if there’s no other frames in the packet.

is it a bug in our code ? because I confirmed this from [rfc9000#section-17.2.2-8](https://datatracker.ietf.org/doc/html/rfc9000#section-17.2.2-8) which says 

> The payload of an Initial packet includes a CRYPTO frame (or frames) containing a cryptographic handshake message, ACK frames, or both. PING, PADDING, and CONNECTION_CLOSE frames of type 0x1c are also permitted.

marten-seemann: Why do you think it’s a bug? All these frames are permitted, but we’re only concerned with the situation where we only happen to have an ACK frame to send.

SahibYar: so it means we only need to check these conditions `!onlyAck && len(initialPayload.frames) == 0` 
because `initialPayload.ack` can be there or cannot be there.

SahibYar: @marten-seemann you got the chance to review this PR again ?

SahibYar: @marten-seemann is this PR good to merge ?

SahibYar:
@marten-seemann is this PR good to merge ?

PR #3941, opened 2023-07-08
oss-fuzz: manually install Go, fix paths

marten-seemann:
Needs https://github.com/google/oss-fuzz/pull/10672.


marten-seemann:
Not sure why Codecov is complaining.

Issue #3793, opened 2023-04-27
set DF bit on OSX


marten-seemann:
I've [heard claims](https://quicdev.slack.com/archives/C68J3KSH3/p1682549680123499) that OSX finally supports setting the DF bit.

Comments in the Chromium source code support that claim: https://chromium.googlesource.com/chromium/src/net/+/master/socket/udp_socket_posix.cc?pli=1#537

sukunrt:
It works but will not work with dual stack listeners. I mean this will _not_ set the DF bit on IPv4 packets. 
```
net.ListenUDP("udp", conn, err := net.ListenUDP("udp", &net.UDPAddr{
	IP:   net.ParseIP("0.0.0.0"),
	Port: 0,
}))
rawConn, err := conn.SyscallConn()
errx := rawConn.Control(func(fd uintptr) {
	err2 = syscall.SetsockoptInt(int(fd), unix.IPPROTO_IPV6, unix.IPV6_DONTFRAG, 1)
	err1 = syscall.SetsockoptInt(int(fd), unix.IPPROTO_IP, unix.IP_DONTFRAG, 1)
})
```

This option `unix.IP_DONTFRAG` only works with "udp4".
The following will set DF bit correctly. 
```
net.ListenUDP("udp4", conn, err := net.ListenUDP("udp", &net.UDPAddr{
	IP:   net.ParseIP("0.0.0.0"),
	Port: 0,
}))
rawConn, err := conn.SyscallConn()
errx := rawConn.Control(func(fd uintptr) {
	err1 = syscall.SetsockoptInt(int(fd), unix.IPPROTO_IP, unix.IP_DONTFRAG, 1)
})
```
The comment in chromium source supports this:
https://chromium.googlesource.com/chromium/src/net/+/master/socket/udp_socket_posix.cc?pli=1#562

I think this makes this unusable for the majority of the use cases, but it's still better to just set it and ignore the error. If it were an IPv4 only conn it would work.

marten-seemann:
@sukun Thanks for digging into this. The code looks off, can you please take another look?

If I understand correctly, this is how things work out:
| listening on | Linux  | OSX  |
|---|---|---|
| upd4 | ✅  | ✅ |
| udp6  | ✅  | ✅ |
| udp  | ✅ (DF enabled for both udp4 and udp6) | ❌ (DF only enabled for upd6, but not udp4) |

Is my understanding correct?

[...]

PR #3946, opened 2023-07-10
enable the DF bit on macOS

sukunrt:
Closes: #3793

### DIFF
--- a/sys_conn_df_darwin.go
+++ b/sys_conn_df_darwin.go
@@ -14,13 +14,9 @@ import (
 )
 
 func setDF(rawConn syscall.RawConn) (bool, error) {
-	uname := &unix.Utsname{}
-	err := unix.Uname(uname)
-	if err != nil {
+	// Setting DF bit is only supported from macOS11
+	if supportsDF, err := isAtLeastMacOS11(); !supportsDF || err != nil {
 		return false, err
-	}
-	if !isAtLeastOS11(uname) {
-		return false, nil
 	}
 
 	// Enabling IP_DONTFRAG will force the kernel to return "sendto: message too long"
@@ -39,8 +35,10 @@ func setDF(rawConn syscall.RawConn) (bool, error) {
 		utils.DefaultLogger.Debugf("Setting DF for IPv4.")
 	case errDFIPv4 != nil && errDFIPv6 == nil:
 		utils.DefaultLogger.Debugf("Setting DF for IPv6.")
-		// On macOS DF bit for IPv4 cannot be set on dual-stack listeners.
-		// Better to be safe here.
+		// On macOS, the syscall for setting DF bit for IPv4 fails on dual-stack listeners.
+		// Treat the connection as not having DF enabled, even though the DF bit will be set
+		// when used for IPv6.
+		// See https://github.com/quic-go/quic-go/issues/3793 for details.
 		return false, nil
 	case errDFIPv4 != nil && errDFIPv6 != nil:
 		return false, errors.New("setting DF failed for both IPv4 and IPv6")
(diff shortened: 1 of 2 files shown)


sys_conn_df_darwin.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/quic-go/quic-go/internal/utils"
+)
+
+func setDF(rawConn syscall.RawConn) (bool, error) {
+	uname := &unix.Utsname{}
+	err := unix.Uname(uname)
+	if err != nil {
+		return false, err
+	}
+	if !isAtLeastOS11(uname) {
```
marten-seemann, line 22: Can you add a comment here why we're testing for this?

sys_conn_df_darwin.go
```diff
@@ -0,0 +1,67 @@
...
+		return false, errors.New("setting DF failed for both IPv4 and IPv6")
+	}
+	return true, nil
+}
+
+func isMsgSizeErr(err error) bool {
+	return errors.Is(err, unix.EMSGSIZE)
+}
+
+func isAtLeastOS11(uname *unix.Utsname) bool {
```
marten-seemann, line 55:
```suggestion
func isAtLeastMacOS11(uname *unix.Utsname) bool {
```

sys_conn_df_darwin.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/quic-go/quic-go/internal/utils"
+)
+
+func setDF(rawConn syscall.RawConn) (bool, error) {
+	uname := &unix.Utsname{}
+	err := unix.Uname(uname)
+	if err != nil {
+		return false, err
+	}
```
marten-seemann, line 21: Nit: Should these lines be moved into the `isAtLeast` function?

sys_conn_df_darwin.go
```diff
@@ -0,0 +1,67 @@
...
+	switch {
+	case errDFIPv4 == nil && errDFIPv6 == nil:
+		utils.DefaultLogger.Debugf("Setting DF for IPv4 and IPv6.")
+	case errDFIPv4 == nil && errDFIPv6 != nil:
+		utils.DefaultLogger.Debugf("Setting DF for IPv4.")
+	case errDFIPv4 != nil && errDFIPv6 == nil:
+		utils.DefaultLogger.Debugf("Setting DF for IPv6.")
+		// On macOS DF bit for IPv4 cannot be set on dual-stack listeners.
```
marten-seemann, line 42:
```suggestion
		// On macOS, the syscall for setting DF bit for IPv4 fails on dual-stack listeners.
		// Treat the connection as not having DF enabled, even though the DF bit will be set 
		// when used for IPv6.
		// See https://github.com/quic-go/quic-go/issues/3793 for details.
```

marten-seemann:
Looking at this logic and comparing it to what we do for Linux, I'm wondering why we have these two cases there:
https://github.com/quic-go/quic-go/blob/0fe21c7d6fc29a1f493a751ea524c3c9428f9848/sys_conn_df_linux.go#L31-L34

I assume that they occur for v4-only and v6-only sockets only, but that's just an assumption. It would be nice if we had a way to test if a socket is v4-only (or v6-only), and only do the correct syscall then.

Obviously, this is out of scope for this PR, but might be a nice follow-up.

sys_conn_df_darwin.go
sukunrt: I picked this up from chromium source code and also some random answer on stackoverflow / github. I couldn't find a more official source so I've just added a comment saying it's only supported from macOS 11.

marten-seemann: Could you link the Chromium source code and / or the Stackoverflow here?

sukunrt: done

marten-seemann (approved):
Thank you @sukunrt!

PR #3948, opened 2023-07-11
http3: validate Host header before sending

marten-seemann:
Not sure if / why this is necessary, but the standard library HTTP/2 package just added this check: https://go-review.googlesource.com/c/net/+/506995

This is apparently what triggered the Go 1.20.6 patch release.

### DIFF
--- a/http3/request_writer.go
+++ b/http3/request_writer.go
@@ -2,6 +2,7 @@ package http3
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


sukunrt (approved):

lifenjoiner:
https://github.com/golang/go/commit/312920c00aac9897b2a0693e752390b5b0711a5a explains it is the CVE-2023-29406 fix.

PR #3950, opened 2023-07-12
http3: don't write response headers if the handler panicked

WeidiDeng:
Fix [5125](https://github.com/caddyserver/caddy/issues/5125).


marten-seemann (approved):
This looks correct, assuming that it's the right thing to close to close the stream after a panic. In HTTP/2, the [stream is reset](https://cs.opensource.google/go/x/net/+/refs/tags/v0.12.0:http2/server.go;l=2299-2302), but that's probably not what we want here, as in QUIC, resetting a stream means that any packet loss for that stream is not repaired.

PR #3951, opened 2023-07-12
http3: implement FlushError for the response writer

WeidiDeng:
Move all of the logic to the new `FlushError` function and let `Flush` call it.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -114,10 +114,16 @@ func (w *responseWriter) Write(p []byte) (int, error) {
 	return w.bufferedStr.Write(p)
 }
 
-func (w *responseWriter) Flush() {
+func (w *responseWriter) FlushError() error {
 	if err := w.bufferedStr.Flush(); err != nil {
 		w.logger.Errorf("could not flush to stream: %s", err.Error())
+		return err
 	}
+	return nil
+}
+
+func (w *responseWriter) Flush() {
+	w.FlushError()
 }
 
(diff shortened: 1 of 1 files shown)


http3/response_writer.go
```diff
@@ -114,10 +114,16 @@ func (w *responseWriter) Write(p []byte) (int, error) {
 	return w.bufferedStr.Write(p)
 }
 
-func (w *responseWriter) Flush() {
+func (w *responseWriter) FlushError() error {
 	if err := w.bufferedStr.Flush(); err != nil {
 		w.logger.Errorf("could not flush to stream: %s", err.Error())
```
marten-seemann, line 119: No need to log if you're returning the error.

http3/response_writer.go
```diff
@@ -114,10 +114,16 @@ func (w *responseWriter) Write(p []byte) (int, error) {
...
+		return err
 	}
+	return nil
+}
+
+func (w *responseWriter) Flush() {
+	w.FlushError()
```
marten-seemann, line 126: Log here if you're ignoring the error.

marten-seemann (commented):
`Flush` is an implementation of `http.Flusher`. I don't see any corresponding `http.FlusherError` interface.

What is the motivation behind this change?

WeidiDeng:
@marten-seemann `FlushError` is used by [ResponseController](https://pkg.go.dev/net/http#NewResponseController), like setdeadline related methods.

marten-seemann (approved):
Thank you!

PR #3952, opened 2023-07-12
http3: add Date response header if not set

WeidiDeng:
All response will have a `Date` header by default.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -55,6 +55,10 @@ func (w *responseWriter) WriteHeader(status int) {
 
 	if status < 100 || status >= 200 {
 		w.headerWritten = true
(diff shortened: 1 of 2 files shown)


http3/response_writer.go
```diff
@@ -55,6 +55,10 @@ func (w *responseWriter) WriteHeader(status int) {
 
 	if status < 100 || status >= 200 {
 		w.headerWritten = true
+		// add Date header
```
marten-seemann, line 58:
```suggestion
		// Add Date header.
		// This is what the standard library does.
		// Can be disabled by setting the Date header to nil.
```

marten-seemann (approved):

PR #3953, opened 2023-07-12
http3: return http.ErrContentLength when writing too large response

WeidiDeng:
If handler sets a valid `Content-Length` header, it won't be able to write more than declared.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -23,6 +23,8 @@ type responseWriter struct {
 	header        http.Header
 	status        int // status code passed to WriteHeader
 	headerWritten bool
+	contentLen    int64 // if handler set valid Content-Length header
+	numWritten    int64 // bytes written
 
 	logger utils.Logger
 }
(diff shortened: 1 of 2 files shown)


marten-seemann (commented):
What about the opposite case, where a content length is set, but fewer bytes are written?

WeidiDeng:
@marten-seemann If content-length is set but handler writes fewer bytes, it's the error of the handler. Not something we can do about. stdlib doesn't either. The best we can do is to emit a warning after handler is done.

marten-seemann (approved):

marten-seemann:
Changes lgtm, but there’s a merge conflict. Can you rebase this PR?

marten-seemann:
Thank you!

marten-seemann:
@WeidiDeng Do we need something equivalent for the request writer?

WeidiDeng:
Yes, but request writer is controlled by `Request.ContentLength` not by a header value.

PR #3954, opened 2023-07-13
http3: add Content-Length header if the handler doesn't flush to the client

WeidiDeng:
stdlib will add `Content-Length` if handler doesn't flush to client usually due to no data written or the amount is small.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -23,6 +23,7 @@ type responseWriter struct {
 	header        http.Header
 	status        int // status code passed to WriteHeader
 	headerWritten bool
+	headerSent    bool
 	contentLen    int64 // if handler set valid Content-Length header
 	numWritten    int64 // bytes written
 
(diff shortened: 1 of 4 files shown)


http3/response_writer.go
```diff
@@ -50,6 +51,41 @@ func (w *responseWriter) Header() http.Header {
 	return w.header
 }
 
+// flushHeader send header frame in wire format when:
```
marten-seemann, line 54: I don't understand this comment.

http3/response_writer.go
```diff
@@ -50,6 +51,41 @@ func (w *responseWriter) Header() http.Header {
...
+// 1. 1XX status is written
+// 1. Flush is called
+// 2. Write can not longer buffer anymore data
+func (w *responseWriter) flushHeader() error {
+	var headers bytes.Buffer
+	// leave some room to encode frame header
+	headers.Write(w.buf[:cap(w.buf)])
```
marten-seemann, line 61: Is there a better way? This doesn't seem clean.

marten-seemann:
This is a lot of code, and makes things quite complicated. Is there really a lot of benefit to making this change?

http3/response_writer.go
WeidiDeng: This comment means when the header will be flushed to the client (i.e. encoded as valid quic datagrams and sent).

http3/response_writer.go
WeidiDeng: This part is very hacky, I'll admit. But I also used a similar technique when refactoring caddy's [fastcgi transport](https://github.com/caddyserver/caddy/pull/4978/files#diff-384ea25b96b47d84c242c7fe3341847cb468d867fafdea97d0920b605321a35dR120). Although fastcgi's length field is fixed size (2 bytes), and quic uses a variable length with a maximum size of 16 bytes.

The alternative is to use another buffer, but it's unnecessary. Since there may be data already buffered in `bufferedStr`, we can use that either.

I tend to not use too many hacks like this, this is one of the place I think it is sensible to use.

WeidiDeng:
The level of complication is within acceptable range I think. It's not that important as a feature but definitely nice to have.

marten-seemann:
I’m not sure I disagree. This PR adds a significant amount of complexity (and hacky code), and if the benefits are only minor, I’m inclined to not merge it.

WeidiDeng:
If you don't like this change, I can't force you to like it 🤷‍♂️ .

PR #3957, opened 2023-07-14
remove OptimizeConn, add a Transport.WriteTo method instead

### LINKED ISSUE #3947: Using OptimizeConn will cause future SetWriteBuffer calls to fail
The transport code docs say (https://github.com/MarcoPolo/quic-go/blob/master/transport.go#L33)

> 	// Calling WriteTo is only valid on the connection returned by OptimizeConn.

My understanding is that means if I want to use this packet conn for `WriteTo` later I need to manually call `OptimizeConn` on the conn before giving it to the `Transport`. However, if I do that the Transport fails to set the SendBuffer size and logs the error:

> connection doesn't allow setting of send buffer size. Not a *net.UDPConn?. See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for details.

This is probably because the returned connection from `OptimizeConn` doesn't implement the correct interface (e.g. `SyscallConn` or `SetWriteBuffer` methods). Maybe it would make sense to set the buffer sizes earlier as part of OptimizeConn?

While I'm here, it would be nice if `OptimizeConn` returned a concrete type that implements `PacketConn`. That way I can keep a reference to `OptimizedConn` and only use that for my `WriteTo` calls. Since otherwise I just have a `PacketConn` that is unclear whether or not it is safe to `WriteTo` to.

marten-seemann:
Fixes #3947.

### DIFF
--- a/integrationtests/self/mitm_test.go
+++ b/integrationtests/self/mitm_test.go
@@ -26,9 +26,10 @@ var _ = Describe("MITM test", func() {
 	const connIDLen = 6 // explicitly set the connection ID length, so the proxy can parse it
 
 	var (
-		serverUDPConn, clientUDPConn net.PacketConn
-		serverConn                   quic.Connection
-		serverConfig                 *quic.Config
+		clientUDPConn                    net.PacketConn
+		serverTransport, clientTransport *quic.Transport
+		serverConn                       quic.Connection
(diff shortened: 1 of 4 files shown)


MarcoPolo (approved):
Thanks. I think this a nicer api

PR #3960, opened 2023-07-16
http3: don't send more than http.Request.ContentLength bytes

marten-seemann:
The standard library HTTP/1.1 returns an error here (https://sourcegraph.com/github.com/golang/go/-/blob/src/net/http/transfer.go?L389-391), but we can't do this, since we're sending the request body async. Instead, the stream will be reset, which also means that there's no guarantee that the server will even receive the request (if packets for a reset stream are lost, they're not even retransmitted by the QUIC layer).

That's probably fine, since it's clearly an error to provide a `http.Request.Body` that reads more bytes than set on `http.Request.ContentLength`.

@WeidiDeng, could you take a look at this PR? Is this what you suggested in https://github.com/quic-go/quic-go/pull/3953#issuecomment-1636064584?

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -327,31 +327,43 @@ func (c *client) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Respon
 	return rsp, rerr.err
 }
 
-func (c *client) sendRequestBody(str Stream, body io.ReadCloser) error {
-	defer body.Close()
-	b := make([]byte, bodyCopyBufferSize)
-	for {
-		n, rerr := body.Read(b)
-		if n == 0 {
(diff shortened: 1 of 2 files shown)


WeidiDeng:
LGTM :shipit:

marten-seemann:
Thank you @WeidiDeng, your review is much appreciated!

Don't feel obligated to say yes, but would you mind if I tag you for review on some more PRs regarding HTTP/3. I have a few more in flight (mostly around header and content length handling) that I would like to get into the next release. 🙏

WeidiDeng:
I'll take a look when I have time and received the notification.

Issue #3958, opened 2023-07-15
how to access connection error


fholzer:
I am using quic directly, not http3. While working on my app I found that there's no way to easily access the error cause when a connection is closed. The only way to get the error is to call `conn.AcceptStream()`, though some use cases may not need one side of the connection to accept streams. That side would be unable to access the error reason (application error, transport error, etc) to learn why its connection might get closed. I might have missed something, maybe you can give me some pointers?

After some digging I thought it would be great to get these errors as cancel cause on the connection context, which works fine for me. Any thoughts? Only downside I came up with is that cancel causes were introduced in go1.20
My changes: https://github.com/fholzer/quic-go/compare/master...fholzer:quic-go:surface-conn-err

While I am writing this, I realize that it might be nice to do the same for stream errors on stream contexts.

marten-seemann:
> The only way to get the error is to call `conn.AcceptStream()`, though some use cases may not need one side of the connection to accept streams

That's correct, or `OpenStream`, or any read / write operation on the stream. This is equivalent to how a `net.TCPConn` works, you also only get the error when you do something with the connection.

That said, I like your solution :) We already have the context, and since Go 1.20 we can attach an error. I don't see any reason to not do it. Could you send us a PR? We might be able to include it in the v0.37 release, which will probably be released some time next week.

For the stream errors, could you send a separate PR?

PR #3961, opened 2023-07-16
connection: surface connection error as connection context cancellation cause

fholzer:
Currently connection errors surface on stream operations, as well as `Connection`'s `AcceptStream` and `OpenStream` calls. This change introduces an additional surface point in the form of the cancellation cause of a Connection's context.

Users can learn about connection errors via
```golang
// get error
err := context.Cause(conn.Context())
```

On server side a `quic.Connection`'s cancellation cause is:
* The error passed to `conn.CloseWithError(err)`
* `qerr.TransportError` when closing the transport's `Conn`, or the transport itself
* `context.Canceled` when the listener is closed, while connections aren't explicitly closed by the user

On client side a `quic.Connection`'s cancellation cause is:
* The error passed to `conn.CloseWithError(err)`
* `qerr.TransportError` when closing the transport's `Conn`, or the transport itself

There might be a race for setting the cancellation cause when users close the transport and also close connections concurrently, but I wasn't able to confirm this. Once a context's cancellation cause is set, it can't be changed by subsequent cancelFunc calls. (While i didn't test it, the same might be the case when closing a transport's `Conn`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ The `quic.Transport` contains a few configuration options that don't apply to an
 
 #### When the remote Peer closes the Connection
 
-In case the peer closes the QUIC connection, all calls to open streams, accept streams, as well as all methods on streams immediately return an error. Users can use errors assertions to find out what exactly went wrong:
+In case the peer closes the QUIC connection, all calls to open streams, accept streams, as well as all methods on streams immediately return an error. Additionally, it is set as cancellation cause of the connection context. Users can use errors assertions to find out what exactly went wrong:
 
 * `quic.VersionNegotiationError`: Happens during the handshake, if there is no overlap between our and the remote's supported QUIC versions.
 * `quic.HandshakeTimeoutError`: Happens if the QUIC handshake doesn't complete within the time specified in `quic.Config.HandshakeTimeout`.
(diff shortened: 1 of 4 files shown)


fholzer:
see #3958

marten-seemann:
> There might be a race for setting the cancellation cause when users close the transport and also close connections concurrently, but I wasn't able to confirm this.

This race necessarily exists, right? It's always possible that the user and the remote are closing the connection at the exact same time. The thing that matters is that we're consistent regarding which error we return. As far as I can tell, we'll return the same error for the `context.Cause` as we'd return for `Connection.AcceptStream` (and others), right?

marten-seemann (commented):
Can you also add a sentence to the documentation for `Connection.Context` (it's in interface.go)?

Other than that, this PR lgtm.

fholzer:
Done. Let me know if you want more details in the added documentation.

marten-seemann (approved):

marten-seemann:
LGTM. Thank you for the contribution!

Are you planning to submit a PR for the stream context as well?

fholzer:
Yes, I'm working on streams context now.

PR #3963, opened 2023-07-17
http3: set the Content-Length header in the http.Request.Header map

marten-seemann:


marten-seemann:
@WeidiDeng Would you mind taking a look at this PR? 🙏
It's my understanding that this is what the standard library HTTP implementation does.

WeidiDeng:
LGTM :shipit:

PR #3964, opened 2023-07-17
http3: reject header field that contain non-lowercase characters

marten-seemann:


marten-seemann:
cc @WeidiDeng 🙏

WeidiDeng:
:shipit:

PR #3965, opened 2023-07-17
http3: reject header field names with invalid characters

marten-seemann:
Depends on #3964.

cc @WeidiDeng 🙏


WeidiDeng:
:shipit:

PR #3966, opened 2023-07-17
http3: reject negative values for the Content-Length header

marten-seemann:
cc @WeidiDeng 🙏


WeidiDeng:
LGTM :shipit: . By the way, this should also be done for response writers, I remember stdlib uses `strconv.ParseUint` as well.

marten-seemann:
> By the way, this should also be done for response writers, I remember stdlib uses `strconv.ParseUint` as well.

That's correct. With https://github.com/quic-go/quic-go/pull/3969 we'll use the same code for header parsing for requests and responses.

WeidiDeng:
:shipit:

PR #3967, opened 2023-07-17
http3: reject header field values with invalid characters

marten-seemann:
Depends on #3965.

cc @WeidiDeng 🙏🙏🙏


WeidiDeng:
:shipit:

PR #3968, opened 2023-07-17
http3: enforce ordering requirement between pseudo and regular headers

marten-seemann:
Depends on #3967.

cc @WeidiDeng 🤷‍♂️

### DIFF
--- a/http3/request.go
+++ b/http3/request.go
@@ -17,6 +17,7 @@ func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
 	var path, authority, method, protocol, scheme, contentLengthStr string
 
 	httpHeaders := http.Header{}
+	var readFirstRegularHeader bool
 	for _, h := range headers {
 		// field names need to be lowercase, see section 4.2 of RFC 9114
 		if strings.ToLower(h.Name) != h.Name {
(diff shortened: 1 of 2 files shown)


sukunrt (commented):
1. The logic looks okay, but convoluted. Is this incorrect? 
```go
	for _, h := range headers {
		// field names need to be lowercase, see section 4.2 of RFC 9114
		if strings.ToLower(h.Name) != h.Name {
			return nil, fmt.Errorf("header field is not lower-case: %s", h.Name)
		}
		if !httpguts.ValidHeaderFieldValue(h.Value) {
			return nil, fmt.Errorf("invalid header field value for %s: %q", h.Name, h.Value)
		}
		if !h.IsPseudo() {
			if !httpguts.ValidHeaderFieldName(h.Name) {
				return nil, fmt.Errorf("invalid header field name: %q", h.Name)
			}
			readFirstRegularHeader = true
		} else {
			if readFirstRegularHeader {
				// all pseudo headers must appear before regular header fields, see section 4.3 of RFC 9114
				return nil, fmt.Errorf("received pseudo header %s after a regular header field", h.Name)
			}
		}
		switch h.Name {
		case ":path":
			path = h.Value
		case ":method":
			method = h.Value
		case ":authority":
			authority = h.Value
		case ":protocol":
			protocol = h.Value
		case ":scheme":
			scheme = h.Value
		case "content-length":
			contentLengthStr = h.Value
		default:
			httpHeaders.Add(h.Name, h.Value)
		}
	}
```

2. Any reason to not add `content-length` to httpHeaders?

sukunrt (approved):

marten-seemann:
Thank you for the code suggestion. That's indeed cleaner. I applied it (and flipped the `if` around).

PR #3969, opened 2023-07-17
http3: unify handling of request and response headers

marten-seemann:
Depends on #3968.

A lot of the validation logic introduced in the previous PRs applies to both request and response headers. This PR reuses the same code path for processing of both.

cc @WeidiDeng 🙏

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -419,27 +419,13 @@ func (c *client) doRequest(req *http.Request, conn quic.EarlyConnection, str qui
 		return nil, newConnError(ErrCodeGeneralProtocolError, err)
 	}
 
+	res, err := responseFromHeaders(hfs)
+	if err != nil {
+		return nil, newStreamError(ErrCodeMessageError, err)
+	}
 	connState := conn.ConnectionState().TLS
-	res := &http.Response{
-		Proto:      "HTTP/3.0",
-		ProtoMajor: 3,
-		Header:     http.Header{},
-		TLS:        &connState,
-		Request:    req,
-	}
-	for _, hf := range hfs {
-		switch hf.Name {
-		case ":status":
-			status, err := strconv.Atoi(hf.Value)
-			if err != nil {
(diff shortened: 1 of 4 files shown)


sukunrt (approved):

http3/client.go
```diff
@@ -419,27 +419,13 @@ func (c *client) doRequest(req *http.Request, conn quic.EarlyConnection, str qui
...
-				return nil, newStreamError(ErrCodeGeneralProtocolError, errors.New("malformed non-numeric status pseudo header"))
-			}
-			res.StatusCode = status
-			res.Status = hf.Value + " " + http.StatusText(status)
-		default:
-			res.Header.Add(hf.Name, hf.Value)
-		}
+	res, err := responseFromHeaders(hfs)
+	if err != nil {
+		return nil, newStreamError(ErrCodeMessageError, err)
```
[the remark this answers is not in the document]
marten-seemann: Yes, I fixed that. Malformed headers should cause a message error.

http3/headers.go
```diff
@@ -164,3 +164,23 @@ func hostnameFromRequest(req *http.Request) string {
...
+		return nil, err
+	}
+	rsp := &http.Response{
+		Proto:         "HTTP/3.0",
+		ProtoMajor:    3,
+		Header:        hdr.Headers,
+		ContentLength: hdr.ContentLength,
+	}
+	status, err := strconv.Atoi(hdr.Status)
+	if err != nil {
```
sukunrt, line 180: Should we check that the resulting integer status is a valid HTTP status so that we reject values like 12345?

sukunrt (approved):
This is fine except for the 1 comment here and https://github.com/quic-go/quic-go/pull/3971#discussion_r1267252159

http3/headers.go
marten-seemann: There's some considerations on this in https://datatracker.ietf.org/doc/html/rfc9110#name-status-codes. Looks like there are reasons to not reject all values outside of the valid range.

PR #3970, opened 2023-07-17
send stream: surface error as stream context cancellation cause

fholzer:
Currently stream errors surface on stream operations only. This change introduces an additional surface point in the form of the cancellation cause of a SendStream's context.

Users can learn about stream errors via

```
// wait for stream to finish
<-str.Context().Done()

// get stream error
err := context.Cause(str.Context())
```

### DIFF
--- a/send_stream_test.go
+++ b/send_stream_test.go
@@ -2,6 +2,7 @@ package quic
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"io"
 	mrand "math/rand"
@@ -318,6 +319,7 @@ var _ = Describe("Send Stream", func() {
 			Expect(str.Context().Done()).ToNot(BeClosed())
 			Expect(str.Close()).To(Succeed())
 			Expect(str.Context().Done()).To(BeClosed())
+			Expect(context.Cause(str.Context())).To(MatchError(context.Canceled))
 		})
 
 		Context("flow control blocking", func() {
@@ -668,6 +670,7 @@ var _ = Describe("Send Stream", func() {
 				Expect(str.Context().Done()).ToNot(BeClosed())
 				str.closeForShutdown(testErr)
 				Expect(str.Context().Done()).To(BeClosed())
+				Expect(context.Cause(str.Context())).To(MatchError(testErr))
 			})
 		})
 	})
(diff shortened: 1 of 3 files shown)


send_stream_test.go
```diff
@@ -318,6 +319,7 @@ var _ = Describe("Send Stream", func() {
 			Expect(str.Context().Done()).ToNot(BeClosed())
 			Expect(str.Close()).To(Succeed())
 			Expect(str.Context().Done()).To(BeClosed())
+			Expect(context.Cause(str.Context())).To(MatchError(context.Canceled))
```
marten-seemann, line 322: Is there any way to return `nil` here?

send_stream_test.go
```diff
@@ -846,6 +849,8 @@ var _ = Describe("Send Stream", func() {
 				Expect(str.Context().Done()).ToNot(BeClosed())
 				str.CancelWrite(1234)
 				Expect(str.Context().Done()).To(BeClosed())
+				Expect(context.Cause(str.Context())).To(BeAssignableToTypeOf(&StreamError{}))
+				Expect(context.Cause(str.Context())).To(HaveField("ErrorCode", StreamErrorCode(1234)))
```
marten-seemann, line 853: Maybe better:
```suggestion
				Expect(context.Cause(str.Context()).(*StreamError).ErrorCode).To(Equal(StreamErrorCode(1234)))
```

send_stream_test.go
fholzer: Calling a context's cancel func with `nil` will result in the context being cancelled with `context.Canceled`.

This from the [`WithCancelCause`](https://pkg.go.dev/context#WithCancelCause) docs:
> Calling cancel with a non-nil error (the "cause") records that error in ctx; it can then be retrieved using Cause(ctx). Calling cancel with nil sets the cause to Canceled.

Or maybe I misunderstood what you are asking.

fholzer: Btw, the same thing happens when Closing the a `quic.Connection` by closing the `Listener` (which, unlike a `Connection`, doesn't accept an error argument to `Close`). There, the context also resolves to `context.Canceled` which, I think, is the same error that you get when on `conn.AcceptStream()` when calling `ln.Close()`

fholzer: What I meant to say is that we do call the context cancel fun with `nil`, see https://github.com/quic-go/quic-go/pull/3970/files#diff-0683df4767cffeaa4b1127b3e6027900d06dac7973a80454782e810871af742eR369, but the context API treats that as `context.Canceled`

marten-seemann: Makes sense, thanks for the explanation @fholzer!

marten-seemann (approved):

PR #3971, opened 2023-07-18
http3: refactor header field processing into a separate function

marten-seemann:
Splitting this commit out of #3969, to allow git to preserve the blame history.

This also changes the structure of the `if`, as suggested by @sukunrt in https://github.com/quic-go/quic-go/pull/3968#pullrequestreview-1534409408.

### DIFF
--- a/http3/request.go
+++ b/http3/request.go
@@ -13,118 +13,147 @@ import (
 	"github.com/quic-go/qpack"
 )
 
-func requestFromHeaders(headers []qpack.HeaderField) (*http.Request, error) {
-	var path, authority, method, protocol, scheme, contentLengthStr string
+type header struct {
+	// Pseudo header fields defined in RFC 9114
+	Path      string
+	Method    string
+	Authority string
+	Scheme    string
+	Status    string
+	// for Extended connect
+	Protocol string
+	// parsed and deduplicated
+	ContentLength int64
+	// all non-pseudo headers
+	Headers http.Header
+}
 
-	httpHeaders := http.Header{}
+func parseHeaders(headers []qpack.HeaderField, isRequest bool) (header, error) {
+	hdr := header{Headers: make(http.Header, len(headers))}
 	var readFirstRegularHeader bool
+	var contentLengthStr string
 	for _, h := range headers {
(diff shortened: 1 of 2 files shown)


http3/request.go
```diff
@@ -13,118 +13,147 @@ import (
...
+				hdr.Scheme = h.Value
+			case ":status":
+				hdr.Status = h.Value
+				isResponsePseudoHeader = true
+			}
+			if isRequest && isResponsePseudoHeader {
+				return header{}, fmt.Errorf("invalid request pseudo header: %s", h.Name)
+			}
+			if !isRequest && !isResponsePseudoHeader {
+				return header{}, fmt.Errorf("invalid response pseudo header: %s", h.Name)
```
sukunrt, line 68: This doesn't fail if there is no `:status` in response. 

```
diff --git a/http3/request_test.go b/http3/request_test.go
index f6a1a424..319b66fb 100644
--- a/http3/request_test.go
+++ b/http3/request_test.go
@@ -100,6 +100,14 @@ var _ = Describe("Request", func() {
 		Expect(err).To(MatchError("invalid request pseudo header: :status"))
 	})
 
+	It("rejects response with no status field", func() {
+		headers := []qpack.HeaderField{
+			{Name: "content-length", Value: "42"},
+		}
+		_, err := parseHeaders(headers, false)
+		Expect(err).ToNot(BeNil())
+	})
+
 	It("parses path with leading double slashes", func() {
 		headers := []qpack.HeaderField{
 			{Name: ":path", Value: "//foo"},
```

sukunrt (changes requested):

http3/request.go
marten-seemann: That's something that should go into `responseFromHeaders` in #3969, isn't it? Validation logic for required request headers is in `requestFromHeaders`, not in `parseHeader`.

sukunrt: Okay. Makes sense. Absent required fields are checked in requestForHeaders. So this is inline with it. Sorry for blocking here.

sukunrt (approved):

PR #3974, opened 2023-07-18
validate newly received max_datagram_frame_size parameter

tanghaowillow:
According to RFC 9221, a client should validate the new value of the max_datagram_frame_size transport parameter is greater than or equal to the stored value.  https://datatracker.ietf.org/doc/html/rfc9221#name-motivation:~:text=If%20a%20client,with%20error%20PROTOCOL_VIOLATION.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -1667,6 +1667,17 @@ func (s *connection) handleTransportParameters(params *wire.TransportParameters)
 		})
 		return
 	}
+
+	// validate that the new value of the max_datagram_frame_size transport parameter sent by
+	// the server in the handshake is greater than or equal to the stored value
+	if s.perspective == protocol.PerspectiveClient && s.peerParams != nil && s.peerParams.MaxDatagramFrameSize > params.MaxDatagramFrameSize {
+		s.closeLocal(&qerr.TransportError{
+			ErrorCode:    qerr.ProtocolViolation,
+			ErrorMessage: "new peer max_datagram_frame_size is smaller than stored value",
+		})
+		return
+	}
+
 	s.peerParams = params
 	// On the client side we have to wait for handshake completion.
 	// During a 0-RTT connection, we are only allowed to use the new transport parameters for 1-RTT packets.
--- a/connection.go
+++ b/connection.go
@@ -1668,8 +1668,8 @@ func (s *connection) handleTransportParameters(params *wire.TransportParameters)
 		return
 	}
 
-	// validate that the new value of the max_datagram_frame_size transport parameter sent by
-	// the server in the handshake is greater than or equal to the stored value
+	// RFC 9221 requires the client to check that the server didn't decrease the
+	// max_datagram_frame_size transport parameter when doing a 0-RTT resumption.
 	if s.perspective == protocol.PerspectiveClient && s.peerParams != nil && s.peerParams.MaxDatagramFrameSize > params.MaxDatagramFrameSize {
 		s.closeLocal(&qerr.TransportError{
 			ErrorCode:    qerr.ProtocolViolation,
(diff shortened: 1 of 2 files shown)


marten-seemann:
Interesting, any idea why this requirement exists in RFC 9221? RFC 9000 doesn't have a similar requirement for the transport parameters remembered:
> When sending frames in 0-RTT packets, a client MUST only use remembered transport parameters; importantly, it MUST NOT use updated values that it learns from the server's updated transport parameters or from frames received in 1-RTT packets. Updated values of transport parameters from the handshake apply only to 1-RTT packets. For instance, flow control limits from remembered transport parameters apply to all 0-RTT packets even if those values are increased by the handshake or by frames sent in 1-RTT packets.

connection.go
```diff
@@ -1667,6 +1667,17 @@ func (s *connection) handleTransportParameters(params *wire.TransportParameters)
 		})
 		return
 	}
+
+	// validate that the new value of the max_datagram_frame_size transport parameter sent by
+	// the server in the handshake is greater than or equal to the stored value
```
marten-seemann, line 1672:
```suggestion
	// RFC 9221 requires the client to check that the server didn't decrease the
	// max_datagram_frame_size transport parameter when doing a 0-RTT resumption.
```

marten-seemann (approved):

tanghaowillow:
> Interesting, any idea why this requirement exists in RFC 9221? RFC 9000 doesn't have a similar requirement for the transport parameters remembered:
> 
> > When sending frames in 0-RTT packets, a client MUST only use remembered transport parameters; importantly, it MUST NOT use updated values that it learns from the server's updated transport parameters or from frames received in 1-RTT packets. Updated values of transport parameters from the handshake apply only to 1-RTT packets. For instance, flow control limits from remembered transport parameters apply to all 0-RTT packets even if those values are increased by the handshake or by frames sent in 1-RTT packets.

I think it is to provide the server with an approach of informing the client that datagram is not supported anymore.  Actually this is also mentioned RFC 9000:

> If 0-RTT data is accepted by the server, the server MUST NOT reduce any limits or alter any values that might be violated by the client with its 0-RTT data.

marten-seemann:
> There is a similar rule in RFC 9000.
> 
> > If 0-RTT data is accepted by the server, the server MUST NOT reduce any limits or alter any values that might be violated by the client with its 0-RTT data.

This rule only applies on the server side though. There's no need for the client to check that. In fact, the server is allowed to reduce limits, if it is willing to handle the fact that the client might have already sent more data in 0-RTT packets.

tanghaowillow:
Thank you for the discussion and explanation.
> I opened #3977 and #3978 to track these. Do you want to submit patches for these issues (please open separate PRs per issue)?
Yes, I would like to handle these two.

marten-seemann:
We currently don't save the `max_datagram_frame_size` parameter for session resumption (neither on the client nor on the server side), so the check suggested here will never be executed: https://github.com/quic-go/quic-go/blob/56cd866840be2c93e2f80073c656199665c49385/internal/wire/transport_parameters.go#L434-L459

Since we don't store the value, we're not in violation of RFC 9221:
>  If a client stores the value of the max_datagram_frame_size transport parameter with their 0-RTT state, they MUST validate that the new value of the max_datagram_frame_size transport parameter sent by the server in the handshake is greater than or equal to the stored value

We only need to do this once we start saving the value for 0-RTT (#3977). Therefore I'm going to close this PR.



> > I opened #3977 and #3978 to track these. Do you want to submit patches for these issues (please open separate PRs per issue)?
> 
> Yes, I would like to handle these two.

Great! #3978 is probably the easier place to start, and will make it easy to add the check required by RFC 9221 later.

PR #3980, opened 2023-07-19
http3: enforce that DATA frames don't exceed Content-Length

marten-seemann:
cc @Mephue @mholt 

I'm not sure if we have to deal with requests / responses that don't set the Content-Length field, but where we could derive from the request / response type that it doesn't have a body.

This PR:
1. Limits `http.Request.Body.Read` and `http.Response.Body.Read` to never read more bytes than set by the Content-Length header (if it's set).
2. Resets the stream with a H3_MESSAGE_ERROR if the peer violates that requirement.


mholt (approved):
On a quick scan, LGTM. Elegant fix!

marten-seemann:
Thank you for the review, @mholt!

PR #3982, opened 2023-07-20
check for WSAEMSGSIZE errors when receiving UDP packets on Windows

### LINKED ISSUE #3956: potential failure on Windows if UDP payload > 1472 bytes
I've found at least one place where `Temporary()` isn't doing the right thing now:

If you send a UDP packet with a payload of 1472 bytes to a quic-go listener (where `listener.Accept()` is being called) on Windows, the packetconn read call returns `windows.WSAEMSGSIZE`, which is apparently _not_ `.Temporary()`. The Accept call fails, and It would seem that a lot of quic-go-using applications panic at this point.

A payload larger than `MaxPacketBufferSize`=1452 bytes is only possible when using IPv4 (or probably with ethernet jumbo frames, too), because IPv4 headers are smaller than IPv6 headers, and the math involved in coming up with `MaxPacketBufferSize` apparently assumes all packets are IPv6? I would argue that `MaxPacketBufferSize` is in fact wrong for this reason, but this is beside the point.

On Linux and other unix-y OSes, `recvfrom` doesn't return an error when the incoming packet is too large, so there's no problem there.

One way to solve the problem on Windows would be by adding a few lines to `(*Transport).listen()` after checking `nerr.Temporary()`:

```go
    if isMsgSizeErr(err) {
        continue
    }
```

But I imagine you might not want to get the function overly cluttered up with special checks like this, so I don't know if that's the "right" solution.

_Originally posted by @thepaul in https://github.com/quic-go/quic-go/issues/1737#issuecomment-1631672219_

marten-seemann:
Fixes #3956. @thepaul, could you give this patch a try?

### DIFF
--- a/sys_conn_df_darwin.go
+++ b/sys_conn_df_darwin.go
@@ -47,9 +47,8 @@ func setDF(rawConn syscall.RawConn) (bool, error) {
 	return true, nil
 }
 
-func isMsgSizeErr(err error) bool {
-	return errors.Is(err, unix.EMSGSIZE)
-}
+func isSendMsgSizeErr(error) bool { return false }
+func isRecvMsgSizeErr(error) bool { return false }
 
 func isAtLeastMacOS11() (bool, error) {
 	uname := &unix.Utsname{}
--- a/sys_conn_df_darwin.go
+++ b/sys_conn_df_darwin.go
@@ -47,7 +47,10 @@ func setDF(rawConn syscall.RawConn) (bool, error) {
 	return true, nil
 }
 
-func isSendMsgSizeErr(error) bool { return false }
+func isSendMsgSizeErr(err error) bool {
+	return errors.Is(err, unix.EMSGSIZE)
+}
+
 func isRecvMsgSizeErr(error) bool { return false }
 
 func isAtLeastMacOS11() (bool, error) {
--- a/send_queue.go
+++ b/send_queue.go
@@ -82,7 +82,7 @@ func (h *sendQueue) Run() error {
 				// 1. Checking for "datagram too large" message from the kernel, as such,
 				// 2. Path MTU discovery,and
 				// 3. Eventual detection of loss PingFrame.
-				if !isMsgSizeErr(err) {
+				if !isSendMsgSizeErr(err) {
 					return err
 				}
 			}
--- a/sys_conn_df.go
+++ b/sys_conn_df.go
@@ -11,7 +11,12 @@ func setDF(syscall.RawConn) (bool, error) {
 	return false, nil
 }
 
-func isMsgSizeErr(err error) bool {
+func isSendMsgSizeErr(err error) bool {
+	// to be implemented for more specific platforms
+	return false
+}
+
+func isRecvMsgSizeErr(err error) bool {
 	// to be implemented for more specific platforms
 	return false
 }
--- a/sys_conn_df_linux.go
+++ b/sys_conn_df_linux.go
@@ -57,10 +57,12 @@ func maybeSetGSO(rawConn syscall.RawConn) bool {
 	return true
 }
 
-func isMsgSizeErr(err error) bool {
+func isSendMsgSizeErr(err error) bool {
 	// https://man7.org/linux/man-pages/man7/udp.7.html
 	return errors.Is(err, unix.EMSGSIZE)
 }
+
+func isRecvMsgSizeErr(err error) bool { return false }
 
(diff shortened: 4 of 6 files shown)


sukunrt (approved):
For windows this looks okay.
What happens on linux here? https://man7.org/linux/man-pages/man3/recvfrom.3p.html says
>If a message is too long to fit in the supplied
       buffer, and MSG_PEEK is not set in the flags argument, the excess
       bytes shall be discarded.
       
Is this fine? 

You might have to do handing for macOS since we merged #3946

thepaul:
> Fixes #3956. @thepaul, could you give this patch a try?

This does indeed fix the problem, according to my tests.

> You might have to do handling for macOS since we merged https://github.com/quic-go/quic-go/pull/3946

No changes should be necessary- the DF bit on outgoing datagrams shouldn't affect recvfrom(). It would affect things if we were using recvmsg() instead, and able to receive `MSG_TRUNC` control messages, but quic-go isn't there yet I think.

sukunrt:
@thepaul yes, there's no real change involved here. My comment was related to changes in `sys_conn_df.go` done in https://github.com/quic-go/quic-go/pull/3946/files 

That commit changed the build flags on `sys_conn_df.go` and introduced `sys_conn_df_darwin.go`. An older version of these files is modified in this PR.

thepaul:
Oh yes, I did read that change- sorry I wasn't clear! That change seems to amount to "set the DF bit on outgoing packets", and my claim is that that won't affect what is received with `recvfrom()`.

sukunrt:
This PR splits `isMsgSizeErr` into `isSendMsgSizeErr` and `isRecvMsgSizeErr`. That change needs to be made in this file too https://github.com/quic-go/quic-go/blob/master/sys_conn_df_darwin.go

marten-seemann:
> This PR splits `isMsgSizeErr` into `isSendMsgSizeErr` and `isRecvMsgSizeErr`. That change needs to be made in this file too https://github.com/quic-go/quic-go/blob/master/sys_conn_df_darwin.go

I see. This PR was based on an older commit, that's why CI succeeded on `push`, but failed on `pull_request`. I rebased and fixed.

sys_conn_df_darwin.go
```diff
@@ -47,9 +47,8 @@ func setDF(rawConn syscall.RawConn) (bool, error) {
 	return true, nil
 }
 
-func isMsgSizeErr(err error) bool {
-	return errors.Is(err, unix.EMSGSIZE)
-}
+func isSendMsgSizeErr(error) bool { return false }
```
sukunrt, line 50: should this be `unix.EMSGSIZE`?

marten-seemann: If I understand EMSGSIZE correctly, it only applies on the receive side: That's why I split `isMsgSizeErr` into a send and a receive part.

On Linux, when a large packet is received, the kernel will just return you whatever fits into the buffer. This is fine, since QUIC packets are authenticated, and too short packets will be discarded.

thepaul: Respectfully, I think you might have that backwards- `sendmsg()` will return `EMSGSIZE` on both Darwin and Linux when the message exceeds what it thinks the MTU is. `recvmsg()` can also return `EMSGSIZE`, but it only does so for `iov_msglen <= 0` or `iov_msglen > IOV_MAX`. `MSG_TRUNC` only applies on the receive side though.

marten-seemann: https://man7.org/linux/man-pages/man7/udp.7.html says:
>  By default, Linux UDP does path MTU (Maximum Transmission Unit) discovery.  This means the kernel will keep track of the MTU to a specific target IP address and return EMSGSIZE when a UDP packet write exceeds it.  When this happens, the application should decrease the packet size.

sukunrt: My understanding was `EMSGSIZE` is returned when we write a datagram larger than the MTU and the DF bit is set. So, `isSendMsgSizeErr` should return true when we receive `unix.EMSGSIZE` just like it does in sys_conn_df_linux.go

On the receive side, we never get any indication on macOS and Linux and the received datagram is just truncated.

marten-seemann: @sukunrt Sorry, I misunderstood what you were saying. I added the check for `isSendMsgSizeErr` on darwin.

sukunrt (approved):

PR #3984, opened 2023-07-20
http3: panic in ResponseWriter.WriteHeader for invalid status codes

WeidiDeng:
panic for status codes that aren't 3 digits. As there is no way to propagate this information to caller.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -3,6 +3,7 @@ package http3
 import (
 	"bufio"
 	"bytes"
+	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


marten-seemann (commented):
Is that what the standard library does? Could you add a test for this?

WeidiDeng:
For [http1](https://github.com/golang/go/blob/master/src/net/http/server.go#L1155). For [http2](https://github.com/golang/net/blob/master/http2/server.go#L2854). Both use the same function to check status code.

Issue #3385, opened 2022-04-17
Configurable maximum packet size


neilalexander:
So far there is no way to configure a maximum packet size above 1252 bytes for IPv4 and 1232 bytes for IPv6. This means that on networks with a higher MTU (i.e. ethernet links in private networks with 9000 byte jumbo frames, or in my case, overlay networks with larger packet sizes up to 65535 bytes) we cannot take advantage of the extra packet sizes in streams and we cannot send datagrams larger than 1220 bytes.

It would be amazing if this was configurable for these environments.

marten-seemann:
How would you deal with cases where the actual MTU is smaller than you expect?

Furthermore, would this be a per-connection or a per-server option?

neilalexander:
I would naively expect that the message would be padded up to the expected size in the `HELLO` phase to test the viability of the MTU and if a problem is identified there, it would either fail to handshake or fall back to a lower MTU.

In our particular use-case, we have a solid guarantee from the layer beneath us that the MTU is always 65535 bytes so we don't even really want to do that (in fact it is quite an annoying/wasteful property of QUIC to have to send so much data up front, particularly over slow links, when you already know for certain that the MTU requirement is met, but I expect that's specced behaviour).

I would imagine that this would need to be configured per-connection if it is a part of the client initiation to send that much data to a given destination — the server can't possibly know if the return path is going to be asymmetric at any rate — but it's also not inconceivable that the server might want to limit the maximum allowable size for other reasons.

dimalukas:
Anything new regarding configuring mtu for QUIC? 
I need to set max mtu to 1200 for my cases.

YRoelvink:
I am actually running into a similar issue as @dimalukas: I am using a satellite link that has known compatibility issues with regards to PMTUD, so the ability to manually set a max MTU would help me greatly!

marten-seemann:
@YRoelvink What's the issue with PMTUD, and what's the MTU you'd like to set?

YRoelvink:
I have had contact with my link provider (I am testing in a dev environment with a dedicated link) and they, for now, only informed me that there have been previous issues with PMTUD using similar setups. hey did, however, suggest that decreasing the MTU could improve the performance of the overall system, so that is what I would like to verify.

marten-seemann:
I'd be curious what those issues what be. 

[...]

Issue #3300, opened 2021-12-26
Allow max datagram frame size to be configurable


chungthuang:
https://datatracker.ietf.org/doc/html/draft-ietf-quic-datagram#section-3 doesn't place a limit on the max datagram frame size. One of the use case for datagram frame is to tunnel UDP over a QUIC, but the current limit of 1220 bytes is too small. Can it be configured by [Config](https://pkg.go.dev/github.com/lucas-clemente/quic-go#Config)? 
I tried to open a PR, but I was getting 403 forbidden.

marten-seemann:
It's not that easy. The MTU on the outer connection must be large enough to hold the datagrams (plus the additional packet header and crypto overhead).

chungthuang:
Does a session have access to packet header and crypto overhead size? I was thinking about updating https://github.com/lucas-clemente/quic-go/blob/e6572e3d0a1e754e91c4d0648cbd598d7e5ca7bc/session.go#L1969 to check datagram frame size + packet header + crypto overhead against MTU. The session can get current MTU from a new method of mtuDiscoverer.

marten-seemann:
The size of the header unfortunately is not constant, it depends on the value of the packet number and on the connection ID, for example. If you want a guarantee that things will fit, the only safe thing to do would be to calculate the maximum size of the header. Unfortunately, this size will be quite a bit bigger than the actual size in almost all cases.

chungthuang:
Can you point me to how the header size is computed? Perhaps we can also signal the max payload size to the caller during dequeue https://github.com/lucas-clemente/quic-go/blob/b8f07c728cf901da96ad445306a3190a51b2b399/packet_packer.go#L594-L599
by returning an error via the dequeue channel in https://github.com/lucas-clemente/quic-go/blob/b8f07c728cf901da96ad445306a3190a51b2b399/datagram_queue.go#L45-L46

marten-seemann:
@LPardue's suggestion definitely makes sense for the receive side.
The complication here is that the send side is not very sophisticated: https://github.com/lucas-clemente/quic-go/blob/496fbe9fda4bb094b1cbdb162a52130b6beda6b1/session.go#L1973-L1981

By using a datagram size that's 60 bytes smaller than the (minimum) packet size quic-go uses, we can make sure that the DATAGRAM frame is guaranteed to fit into the payload.
I haven't spent too much time thinking about the best API here, as I've not used datagrams in production yet.

marten-seemann:
[...]

Issue #3955, opened 2023-07-13
PMTU reduction causes packet loss


DevinCarr:
The quic-go library uses an MTU discovery process after the handshake is completed between the client and the server (https://github.com/quic-go/quic-go/blob/master/mtu_discoverer.go). This code does not handle the case that the PMTU is lowered below the "current" byte count.

If you have a long-lived QUIC connection between a server and a client and the PMTU is adjusted to a lower value than previous, QUIC packets that were sending with full payloads will begin to get dropped.
We noticed this in cloudflared that if the connection is sending ping frames but no other payloads, a PMTU reduction will not cause any issues as the ping frames are always smaller than the PMTU (usually under 100 bytes). But, when the client-side attempts to send payloads that would normally fill the QUIC packet size, we began to see an increase in the packet loss metric. We investigated further into the MTU discovery code in quic-go and noticed that once the PMTU is established, it will never be lowered: https://github.com/quic-go/quic-go/blob/master/mtu_discoverer.go#L109

Here is a repro of the issue: https://github.com/DevinCarr/pmtu-failure-quic-go-repro
_note_: it does require two machines to repro as you will need to adjust the MTU on the client side sending the payload.

Would there be some way that quic-go would be able to adjust this PMTU after the connection has been established and the PMTU has been calculated?

As a remediation, I thought that once the PMTU is established, quic-go could reduce the interval at which it evaluates the PMTU by attempting to send a ping packet of the current size to validate at least that the current PMTU has not been adjusted along the route.

Attempting something like this led me to the code which controls the congestion control: https://github.com/quic-go/quic-go/blob/master/internal/congestion/cubic_sender.go#L308
Reducing the PMTU via the `mtuDiscoverer` causes this panic. Is there a reason why the congestion control cannot reduce the datagram size? (I apologize in advance for my ignorance on the QUIC congestion control and how it's implemented here)

Finally, beyond leveraging [`DisablePathMTUDiscovery`](https://github.com/quic-go/quic-go/blob/master/interface.go#L320), is there any mechanism that we would be able to account for this issue?
Sadly, this configuration is currently our only lever beyond restarting the connection and establishing the PMTU again for the new connection.

marten-seemann:
Hi @DevinCarr, thank you for your detailed issue.

[...]

PR #3988, opened 2023-07-21
Add configurable max_udp_payload_size transport parameter

DevinCarr:
Helps partially address #3300 and #3385 by adding a configuration parameter for the `max_udp_payload_size`.

Even though the RFC defines the `max_udp_payload_size` to be defaulted to 65527[*](https://datatracker.ietf.org/doc/html/rfc9000#name-transport-parameter-definit) this PR continues the assumption of a default UDP payload size of 1452 (MaxPacketBufferSize). It does this because the MaxPacketBufferSize is used all throughout the code to allocate max capacity for buffers. By maintaining this maximum value we can implement these changes without needing to (currently) address larger packet sizes. It should work with lower `max_udp_payload_size` values between >1200 and <=1452.

Adjusting the MaxPacketBufferSize to be larger would also require that the `buffer_pool` can be adjusted ahead of opening a connection and would be unique to that connection. Or, another solution would be to increase the `buffer_pool` to a larger value of maybe 1600 per buffer as a maximum since it's unlikely (currently) many would want to have a larger UDP packet size than the jumbograms size. This would also require the transport parameter to also be clipped to this upper limit value. Unsure which is the best solution to approach here though, but defaulting to the 65527 limit seems a bit large and unlikely to be utilized by many except in perhaps unique networks.

Additionally, Path MTU Discovery can operate as normal with this parameter, as it just helps support the case that one side of a quic connection knows that it's PMTU can adjust to a certain lower threshold.

I have held off on adding tests for these changes as I feel you will want to iterate on the implementation and to address the above adjustments that need to be made before merging. I'm willing to continue work here, but I wanted to pause at this point for your feedback.


marten-seemann:
Can you help me understand the motivation behind this PR? Why do you need packet sizes beyond 1452 bytes?

I understand that there’s the jumbo-frames use case on local networks, but if I understand correctly, that’s not what you’re trying to solve here.

DevinCarr:
We and thereby, my use-case does not have a need for _payload_ sizes to go beyond 1452 bytes. I didn't want to limit the case that someone does want to go higher than that as that's what was described by other folks in #3385. If you feel that it would be fine to just provide the `max_udp_payload_size` as a configurable parameter but it can only go lower (but greater than 1200), I would be fine with that for this PR.

marten-seemann:
Why does it need to be configurable at all? PMTUD will find the appropriate values, right?

DevinCarr:
Currently the PMTUD will only discover the MTU values at the beginning of the connection, but if the PMTU were to decrease it causes packet loss (#3955).

To avoid this entirely, if a network operator were to know that the PMTU isn't always stable at 1500 and the PMTU could only ever drop to 1400 for example, setting the `max_udp_payload_size` is the tool that could be reached for to solve this. The long term goal would be to have on-going PMTU discovery, but that seems like more of an advanced use-case, that can pursued separately. The `max_udp_payload_size` is defined in the RFC already and can be made available to a user to configure and that's what this PR is attempting to accomplish.

marten-seemann:
I see. With the transport parameter, you’re making sure that *the peer* (and not yourself) is not sending packets larger than the configured size. In other words, you’re protecting against a peer that’s not properly implementing PMTUD. 

That’s not really what this TP was intended for. The TP exists to inform your peer or your maximum receive buffer size. 

I’d prefer to do “the right thing” and fix our PMTUD algorithm, since it doesn’t fully implement the respective RFC. I don’t think it’s _that much_ work to get those changes in, and it would probably be reasonable to target that for the v0.38 release. What do you think?

DevinCarr:
Sure, if you think that would be more valuable! I was trying this avenue since it seemed like a smaller scoped change instead of addressing the PMTUD algorithm to adjust for decreases (or increases).

We can bring this discussion back to the #3955 issue if you want to help provide me with some pointers for sections in the relevant RFC(s) and locations of where you think the changes would best be done. Not sure the timeline of getting this for the v0.38 release since I don't know what your timeline is for that, but I would be willing to at least get started on it.

marten-seemann:
No fixed timeline for v0.38 yet. We just released v0.37 today, so I’d expect v0.38 to ship towards the end of August / beginning of September, depending on the changes that make it in by then.

And yes, let’s move the discussion to #3955.

joliveirinha:
Can we still pivot this PR to address https://github.com/quic-go/quic-go/issues/3300 only?

It is independent of the PTMU issue, and we can solve that problem.

marten-seemann:
As a short term fix, I’d recommend simply turning it PMTUD.

marten-seemann:
Closing since we're not planning to add another config flag.

PR #3989, opened 2023-07-22
http3: automatically add content-length for small responses

WeidiDeng:
…all enough and doesn't call Flush

Another iteration of [3954](https://github.com/quic-go/quic-go/pull/3954).

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -15,9 +15,8 @@ import (
 	"github.com/quic-go/qpack"
 )
 
-// The maximum length of an encoded quic frame header is 10.
-// 1 (header type) + variable (frame length, the length of the largest encoded quic varint is 9).
-// 16 is more than enough.
+// The maximum length of an encoded HTTP/3 frame header is 16:
+// The frame has a type and length field, both QUIC varints (maximum 8 bytes in length)
 const frameHeaderLen = 16
 
 // headerWriter wraps the stream, so that the first Write call flushes the header to the stream
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -15,7 +15,12 @@ import (
 	"github.com/quic-go/qpack"
 )
 
-// headerWriter wrap the stream, so first Write will flush header to the stream
+// The maximum length of an encoded quic frame header is 10.
+// 1 (header type) + variable (frame length, the length of the largest encoded quic varint is 9).
+// 16 is more than enough.
+const frameHeaderLen = 16
+
+// headerWriter wraps the stream, so that the first Write call flushes the header to the stream
 type headerWriter struct {
 	str     quic.Stream
 	header  http.Header
@@ -37,15 +42,12 @@ func (hw *headerWriter) writeHeader() error {
 		}
 	}
 
-	buf := make([]byte, 0, 16+headers.Len())
+	buf := make([]byte, 0, frameHeaderLen+headers.Len())
 	buf = (&headersFrame{Length: uint64(headers.Len())}).Append(buf)
 	hw.logger.Infof("Responding with %d", hw.status)
 	buf = append(buf, headers.Bytes()...)
 
 	_, err := hw.str.Write(buf)
-	if err != nil {
-		hw.logger.Errorf("could not write headers frame: %s", err.Error())
-	}
 	return err
 }
 
@@ -85,7 +87,7 @@ func newResponseWriter(str quic.Stream, conn quic.Connection, logger utils.Logge
 	}
 	return &responseWriter{
 		headerWriter: hw,
-		buf:          make([]byte, 16),
+		buf:          make([]byte, frameHeaderLen),
 		conn:         conn,
 		bufferedStr:  bufio.NewWriter(hw),
 	}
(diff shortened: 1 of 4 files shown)


WeidiDeng:
@marten-seemann Do you have time to take a look at this pr?

http3/response_writer.go
```diff
@@ -15,19 +15,60 @@ import (
 	"github.com/quic-go/qpack"
 )
 
+// headerWriter wrap the stream, so first Write will flush header to the stream
```
marten-seemann, line 18:
```suggestion
// headerWriter wraps the stream, so that the first Write call flushes the header to the stream
```

http3/response_writer.go
```diff
@@ -15,19 +15,60 @@ import (
...
+	buf := make([]byte, 0, 16+headers.Len())
+	buf = (&headersFrame{Length: uint64(headers.Len())}).Append(buf)
+	hw.logger.Infof("Responding with %d", hw.status)
+	buf = append(buf, headers.Bytes()...)
+
+	_, err := hw.str.Write(buf)
+	if err != nil {
+		hw.logger.Errorf("could not write headers frame: %s", err.Error())
+	}
+	return err
```
marten-seemann, line 49: Either return the error (and log at the call site), or log the error here and don't return it.

http3/response_writer.go
```diff
@@ -15,19 +15,60 @@ import (
...
+	enc := qpack.NewEncoder(&headers)
+	enc.WriteField(qpack.HeaderField{Name: ":status", Value: strconv.Itoa(hw.status)})
+
+	for k, v := range hw.header {
+		for index := range v {
+			enc.WriteField(qpack.HeaderField{Name: strings.ToLower(k), Value: v[index]})
+		}
+	}
+
+	buf := make([]byte, 0, 16+headers.Len())
```
marten-seemann, line 40: Where's the 16 coming from? Can you add a comment explaining that magic number?

http3/response_writer.go
```diff
@@ -37,13 +78,16 @@ var (
...
+	}
 	return &responseWriter{
-		header:      http.Header{},
-		buf:         make([]byte, 16),
-		conn:        conn,
-		str:         str,
-		bufferedStr: bufio.NewWriter(str),
-		logger:      logger,
+		headerWriter: hw,
+		buf:          make([]byte, 16),
```
marten-seemann, line 88: Where's the 16 coming from? Can you add a comment explaining that magic number?

marten-seemann (commented):
Thank you for this PR!

Can you add an integration test (in http_test.go) that tests end-to-end that this actually works?

I'm also wondering if we should have any logic to send out large (> 1 MTU?) headers without waiting for the write call. What do you think?

WeidiDeng:
I don't known if we should flush headers immediately for large headers 😅 , as far as I know, I use this branch and [previous one](https://github.com/quic-go/quic-go/pull/3954) without encountering any problems at all.

WeidiDeng:
I don't know why these connection tests failed 🤷 .

http3/response_writer.go
```diff
@@ -15,7 +15,12 @@ import (
 	"github.com/quic-go/qpack"
 )
 
-// headerWriter wrap the stream, so first Write will flush header to the stream
+// The maximum length of an encoded quic frame header is 10.
+// 1 (header type) + variable (frame length, the length of the largest encoded quic varint is 9).
```
marten-seemann, line 19: It's 8.

marten-seemann: The type is not guaranteed to be 1 byte though, is it?

WeidiDeng: 16 comes from the original code for encoding frame header. The type is 1 byte for header and data frame according to the code.

It's 8 byte. My mistake.

http3/response_writer.go
```diff
@@ -15,19 +15,62 @@ import (
 	"github.com/quic-go/qpack"
 )
 
+// The maximum length of an encoded quic frame header is 10.
+// 1 (header type) + variable (frame length, the length of the largest encoded quic varint is 9).
+// 16 is more than enough.
```
marten-seemann, line 20:
```suggestion
// The maximum length of an encoded HTTP/3 frame header is 16:
// The frame has a type and length field, both QUIC varints (maximum 8 bytes in length)
```

marten-seemann (approved):

integrationtests/self/http_test.go
```diff
@@ -128,6 +128,18 @@ var _ = Describe("HTTP tests", func() {
 		Expect(string(body)).To(Equal("Hello, World!\n"))
 	})
 
+	It("sets content-length for small response", func() {
+		mux.HandleFunc("/small", func(w http.ResponseWriter, r *http.Request) {
+			defer GinkgoRecover()
+			w.Write([]byte("foobar"))
+		})
+
+		resp, err := client.Get("https://localhost:" + port + "/small")
```
marten-seemann, line 137:
```suggestion
		resp, err := client.Get(fmt.Sprintf("https://localhost:%d/small", port))
```

marten-seemann (approved):

PR #3992, opened 2023-07-23
add a method to retrieve non-QUIC packets from the Transport

### LINKED ISSUE #3929: Make quic-go more STUN friendly
The QUIC protocol is ideal for P2P applications which was a major factor in the adoption of quic-go for syncthing and libp2p.

As already discussed during the design phase of the new connection establishment API(see https://github.com/quic-go/quic-go/issues/3727#issuecomment-1477371806) syncthing and probably others need a better way to integrate STUN.

While it's easy to pass an existing PacketConn in order to send non-QUIC data, there's noch such mechanism yet to receive non-QUIC data.

With added GSO support in v0.36.0 and GRO on the horizon, solutions like https://github.com/AudriusButkevicius/pfilter/ are becoming less viable.

see also https://github.com/syncthing/syncthing/issues/8930

marten-seemann:
Fixes #3929.

Tests still missing.
Resolution to https://mailarchive.ietf.org/arch/msg/quic/oR4kxGKY6mjtPC1CZegY1ED4beg/ still missing.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -7,12 +7,12 @@ import (
 	"errors"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
-
-	"github.com/quic-go/quic-go/internal/wire"
 
 	"github.com/quic-go/quic-go/internal/protocol"
 	"github.com/quic-go/quic-go/internal/utils"
+	"github.com/quic-go/quic-go/internal/wire"
 	"github.com/quic-go/quic-go/logging"
 )
 
@@ -84,6 +84,9 @@ type Transport struct {
 	closed      bool
 	createdConn bool
 	isSingleUse bool // was created for a single server or client, i.e. by calling quic.Listen or quic.Dial
+
+	readingNonQUICPackets atomic.Bool
+	nonQUICPackets        chan receivedPacket
 
 	logger utils.Logger
 }
@@ -341,6 +344,13 @@ func (t *Transport) listen(conn rawConn) {
 }
 
 func (t *Transport) handlePacket(p receivedPacket) {
+	if len(p.data) == 0 {
+		return
+	}
+	if !wire.IsPotentialQUICPacket(p.data[0]) && !wire.IsLongHeaderPacket(p.data[0]) {
+		t.handleNonQUICPacket(p)
+		return
+	}
 	connID, err := wire.ParseConnectionID(p.data, t.connIDLen)
 	if err != nil {
 		t.logger.Debugf("error parsing connection ID on packet from %s: %s", p.remoteAddr, err)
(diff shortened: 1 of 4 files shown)


transport.go
```diff
@@ -85,6 +85,9 @@ type Transport struct {
 	createdConn bool
 	isSingleUse bool // was created for a single server or client, i.e. by calling quic.Listen or quic.Dial
 
+	readingNonQUICPackets atomic.Bool
```
AudriusButkevicius, line 88: Don't see the point of this bool. Why wouldn't we always have the buffer allocated (based on size param somewhere) and just drop on overflow (or when sized to zero).

transport.go
```diff
@@ -430,3 +439,34 @@ func (t *Transport) maybeHandleStatelessReset(data []byte) bool {
...
+
+func (t *Transport) handleNonQUICPacket(p receivedPacket) {
+	// Strictly speaking, this is racy,
+	// but we only care about receiving packets at some point after ReadNonQUICPacket has been called.
+	if !t.readingNonQUICPackets.Load() {
+		return
+	}
+	select {
+	case t.nonQUICPackets <- p:
+	default:
```
AudriusButkevicius, line 451: Is there some stats we could track how many packets are dropped?

transport.go
```diff
@@ -342,6 +345,12 @@ func (t *Transport) listen(conn rawConn) {
 }
 
 func (t *Transport) handlePacket(p receivedPacket) {
+	if len(p.data) == 0 {
+		return
+	}
+	if !wire.IsPotentialQUICPacket(p.data[0]) && !wire.IsLongHeaderPacket(p.data[0]) {
+		t.handleNonQUICPacket(p)
+	}
```
AudriusButkevicius, line 353: Return?

transport.go
marten-seemann: We can invoke the drop call on the tracer here, good point.

transport.go
marten-seemann: Filled with packet, this would add 32*1500 = ~48 kB of wasted memory. I'd like to avoid that.

transport.go
marten-seemann: Oops ;)

transport.go
AudriusButkevicius: Not if the count is configurable and set to 0

marten-seemann: I was trying to avoid introduce yet more configuration API.

AudriusButkevicius: It's only configurable if you need it. It's zero by default, so doesn't seem like you need to provide a value?

marten-seemann: There's still the race condition of calling this function before `Listen` or `Dial`. Then the channel hasn't been created yet.

AudriusButkevicius: Not if it's initializes at point the transport is constructed? The channel can be nil by default at which point we'd just drop the packets?

transport.go
marten-seemann: Done.

marten-seemann:
@bt90 I see you're blocked on this PR in syncthing. Have you tried out this branch yet? Does it work for your use case and allow you to retire the filter implementation?

bt90:
@AudriusButkevicius

AudriusButkevicius:
I don't think we are blocked. We just can't pick up the latest optimizations as we rely on this type of functionality which we are currently implementing with a separate package.

lucas-clemente (approved):

transport.go
AudriusButkevicius: I'd still argue that this is not needed.

calmh:
Maybe not blocked-blocked, but given that migrating to the new transport stuff is a refactor that doesn't fit cleanly into our existing way of doing it (since we can't write to the socket directly, nor read from the new transport), this would be welcome so that we don't need to rearchitect twice for the same change.

transport.go
AudriusButkevicius: I would still expose the responsibility of the channel creation/assignment to the builder of the transport type (before its used). Then you can use a nil check, and a user can control the buffer size (which is now quite small).

PR #3993, opened 2023-07-23
add a Transport.DisableGSO config value

### LINKED ISSUE #3911: GSO fails on some platforms / interfaces

marten-seemann:
Currently this flag only has an effect if `QUIC_GO_ENABLE_GSO` is set, but once we tackle #3911, this will be more useful.

Closes #3912.


marten-seemann:
With https://github.com/quic-go/quic-go/issues/3911#issuecomment-1657214086 and https://github.com/quic-go/quic-go/issues/3911#issuecomment-1657218892, it seems like this won't be needed any more. Note that this would make it feasible to disable GSO on a per-connection basis (#3912), but I'm not convinced we should increase the API surface.

Issue #2727, opened 2020-08-20
Avoid relying on internals of crypto/tls


FiloSottile:
I've noticed on #2614 that you're relying on unsafe casting to crypto/tls internal structures. I'm writing this to ask to please not do that.

Of course, that's not covered by any compatibility promise, but you know that. What you might not be considering is that we might and eventually will change internals for a security release, and then your users will be forced to choose between a security vulnerability and dropping a dependency in a hurry. Even more worrying, is that we might replace a field with a different one of the same type, and you might end up violating security-critical invariants without noticing.

Instead, please fork crypto/tls, and apply our changes on your own schedule, integrating them with the rest of the module. I hear `git subtree` is nice for that. You can even try to expose everything you need as clear, separate APIs, so you can more easily audit the contact points, and maybe eventually even get them upstreamed. It might require some more work upfront, but the only way to safely do what you are doing is _anyway_ to be aware of every single change in crypto/tls.

marten-seemann:
Hello @FiloSottile,

Thank you for opening this issue. I'd **really really** love to do that, the dependency on crypto/tls has been a huge pain point in the past, and I've spent a fair amount of time thinking about solutions for this problem.

Conceptually, forking crypto/tls indeed seems like the cleanest solution.
The problem with that approach is that this would have API implications for quic-go: Currently, we expose a `Dial(..., *tls.Config)` and `Listen(..., *tls.Config)` functions. Furthermore, in the http3 package we use the `tls.Config` of the `http.Server`.

This kind of API makes it trivial to use the same `tls.Config` to run services over TCP / QUIC (or HTTP2 / HTTP3) at the same time. Due to the occasional blackholing of UDP traffic, it is expected that most applications will have to do this for the foreseeable future.
Building a `tls.Config` can involve a fair amount of complexity. For example, Caddy uses the `GetConfigForClient` callback to generate a config per SNI, and runs the ACME protocol to obtain a certificate (@mholt, is that correct?). All of this code would have to be duplicated if we used a fork of crypto/tls, to use a `qtls.Certificate` and a `qtls.ClientHelloInfo` then.

I've tried to mitigate the security impact of the unsafe operations as far as possible by [building a function](https://github.com/lucas-clemente/quic-go/blob/master/internal/handshake/unsafe.go) that compares the structs: Adding an (exported or unexported) field to a struct will cause that check to fail, as will a reordering and a renaming. This code is run on `init()` (I didn't find a way to run it at compile time).
[...]

PR #3994, opened 2023-07-23
ci: update Go 1.21 to rc3

marten-seemann:
https://github.com/golang/go/issues/60939 should be resolved in this version.


otbutz:
With #2727 fixed, should we also built with gotip as part of the ci?

marten-seemann:
> With #2727 fixed, should we also build with gotip as part of the ci?

What would be the advantage of doing that?

PR #3997, opened 2023-07-26
http3: support HTTP CONNECT and CONNECT-UDP (masque) protocols

jfwangece:
This commit makes changes to the HTTP3 stack so that HTTP CONNECT and CONNECT-UDP requests will not fail due to internal checks. Also, it modifies the HTTP3 setting frame code to reflect the up-to-date setting for HTTP3 datagram over QUIC datagram. Finally, it adds two examples to run proxies TCP and UDP connections via HTTP3.

### DIFF
--- a/example/connect-udp/main.go
+++ b/example/connect-udp/main.go
@@ -1,0 +1,199 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"context"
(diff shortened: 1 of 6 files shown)


RithvikChuppala:
Do you have the server-side implementation for this as well?

PR #3998, opened 2023-07-27
http3: fix check for content length of the response

imroc:


marten-seemann:
What are you trying to achieve with this change?

imroc:
> What are you trying to achieve with this change?

Just like the logic written in comments: `Check that the server doesn't send more data in DATA frames than indicated by the Content-Length header (if set).`

imroc:
A simple example to reproduce the bug:

```go
package main

import (
	"io"
	"net/http"
	"strings"
)

func main() {
	client := &http.Client{
		Transport: &http3.RoundTripper{},
	}
	req, err := http.NewRequest(http.MethodPost, "https://www.cloudflare.com/", strings.NewReader("test"))
	if err != nil {
		panic(err)
	}
	req.Header.Set("Content-Length", "4")
	res, err := client.Do(req)
	if err != nil {
		panic(err)
	}
	_, err = io.ReadAll(res.Body)
	if err != nil {
		panic(err)
	}
}
```

Run the code:

```bash
$ go run .
panic: peer sent too much data

goroutine 1 [running]:
main.main()
        /Users/roc/dev/test/quic-go/main.go:27 +0x2d4
exit status 2
```

imroc:
Current logic: Check if response body content-length if greater than `Content-Length` header in request.

Should change to: Check if response body content-length if greater than `Content-Length` header in response.

marten-seemann:
Oh, I see! I missed that you're using `res` and not `req`. I don't think the `textproto` conversion is needed though, headers are already normalized after we retrieve them from qpack.

imroc:
> Oh, I see! I missed that you're using `res` and not `req`. I don't think the `textproto` conversion is needed though, headers are already normalized after we retrieve them from qpack.

OK, `textproto` is removed.

http3/client.go
```diff
@@ -429,8 +429,9 @@ func (c *client) doRequest(req *http.Request, conn quic.EarlyConnection, str qui
 	// Check that the server doesn't send more data in DATA frames than indicated by the Content-Length header (if set).
 	// See section 4.1.2 of RFC 9114.
 	var httpStr Stream
-	if _, ok := req.Header["Content-Length"]; ok && req.ContentLength >= 0 {
-		httpStr = newLengthLimitedStream(hstr, req.ContentLength)
+
```
marten-seemann, line 432:
```suggestion
```

marten-seemann (approved):

PR #4001, opened 2023-07-29
set a net.Conn with the correct addresses on the tls.ClientHelloInfo

### LINKED ISSUE #4000: regression in v0.37.0: ClientHelloInfo.Conn is nil
See https://github.com/golang/go/issues/61639 for details. Either Go resolves it (and we have to backport it to qtls for Go 1.20), or we'll have to wrap `GetClientHelloInfo`.

marten-seemann:
Fixes #4000.

@mholt, would you mind taking a look at this PR?

### DIFF
--- a/integrationtests/self/handshake_test.go
+++ b/integrationtests/self/handshake_test.go
@@ -140,6 +140,30 @@ var _ = Describe("Handshake tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("has the right local and remote address on the ClientHelloInfo.Conn", func() {
+			var local, remote net.Addr
+			done := make(chan struct{})
+			tlsConf := &tls.Config{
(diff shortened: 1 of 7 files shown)


integrationtests/self/handshake_test.go
```diff
@@ -140,6 +140,30 @@ var _ = Describe("Handshake tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("has the right local and remote address on the ClientHelloInfo.Conn", func() {
```
marten-seemann, line 143: Now including an end-to-end test to make sure this API never accidentally breaks again.

mholt (approved):
Sorry it took me over the weekend. At a glance this looks good! If we can get the remote/local addr at time of handshake, that's good enough for our use case for now, I think.

Thank you :)

I can probably get this out in Caddy 2.7. When are you planning the next release?

PR #4012, opened 2023-08-02
update qtls to restrict RSA keys in certificates to <= 8192 bits

marten-seemann:
This is a backport of the Go standard library fix that triggered the Go 1.20.7 release. It's only needed for the quic-go version built with Go 1.20, as Go 1.21 uses the standard library crypto/tls.

@mholt Just a heads-up, there will be a v0.37.2 patch release very soon.


mholt:
Thanks Marten. I really had to cut a Caddy release yesterday but we will likely cut another one soon-ish, so I'll be sure to include this then.

PR #4013, opened 2023-08-03
save the max_datagram_frame_size transport parameter in the session ticket

### LINKED ISSUE #3977: save the max_datagram_frame_size transport parameter for 0-RTT
Otherwise, it's easy to violate the size requirement described in section 3 of RFC 9221.

tanghaowillow:
Add the max_datagram_frame_size parameter in session ticket. This enables client to send 0-RTT datagrams.
Fixes #3977

### DIFF
--- a/integrationtests/self/zero_rtt_oldgo_test.go
+++ b/integrationtests/self/zero_rtt_oldgo_test.go
@@ -856,7 +856,7 @@ var _ = Describe("0-RTT", func() {
 		fmt.Fprintf(GinkgoWriter, "Sent %d 0-RTT packets.", num0RTT)
 		Expect(num0RTT).ToNot(BeZero())
 		zeroRTTPackets := get0RTTPackets(tracer.getRcvdLongHeaderPackets())
-		Expect(len(zeroRTTPackets)).To(BeNumerically("==", 1))
+		Expect(zeroRTTPackets).To(HaveLen(1))
 		Expect(conn.CloseWithError(0, "")).To(Succeed())
 	})
 
--- a/integrationtests/self/zero_rtt_oldgo_test.go
+++ b/integrationtests/self/zero_rtt_oldgo_test.go
@@ -801,4 +801,116 @@ var _ = Describe("0-RTT", func() {
 		Expect(len(zeroRTTPackets)).To(BeNumerically(">", 10))
 		Expect(zeroRTTPackets[0]).To(Equal(protocol.PacketNumber(0)))
 	})
+
+	It("send 0-RTT datagrams", func() {
+		tlsConf, clientTLSConf := dialAndReceiveSessionTicket(getQuicConfig(&quic.Config{
+			EnableDatagrams: true,
+		}))
+
+		tracer := newPacketTracer()
+		ln, err := quic.ListenAddrEarly(
+			"localhost:0",
+			tlsConf,
+			getQuicConfig(&quic.Config{
+				Allow0RTT:       true,
+				EnableDatagrams: true,
+				Tracer:          newTracer(tracer),
+			}),
+		)
+		Expect(err).ToNot(HaveOccurred())
+		defer ln.Close()
+
+		proxy, num0RTTPackets := runCountingProxy(ln.Addr().(*net.UDPAddr).Port)
+		defer proxy.Close()
+
+		// second connection
+		var sentMessage = GeneratePRData(100)
+		var receivedMessage []byte
+		received := make(chan struct{})
+		go func() {
+			defer GinkgoRecover()
+			defer close(received)
+			conn, err := ln.Accept(context.Background())
+			Expect(err).ToNot(HaveOccurred())
+			receivedMessage, err = conn.ReceiveMessage(context.Background())
+			Expect(err).ToNot(HaveOccurred())
+			Expect(conn.ConnectionState().Used0RTT).To(BeTrue())
+		}()
+		conn, err := quic.DialAddrEarly(
+			context.Background(),
+			fmt.Sprintf("localhost:%d", proxy.LocalPort()),
+			clientTLSConf,
+			getQuicConfig(&quic.Config{
+				EnableDatagrams: true,
+			}),
+		)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(conn.ConnectionState().SupportsDatagrams).To(BeTrue())
+		Expect(conn.SendMessage(sentMessage)).To(Succeed())
+		<-conn.HandshakeComplete()
+		<-received
+
+		Expect(conn.ConnectionState().Used0RTT).To(BeTrue())
+		Expect(receivedMessage).To(Equal(sentMessage))
+		num0RTT := atomic.LoadUint32(num0RTTPackets)
+		fmt.Fprintf(GinkgoWriter, "Sent %d 0-RTT packets.", num0RTT)
+		Expect(num0RTT).ToNot(BeZero())
+		zeroRTTPackets := get0RTTPackets(tracer.getRcvdLongHeaderPackets())
+		Expect(len(zeroRTTPackets)).To(BeNumerically("==", 1))
+		Expect(conn.CloseWithError(0, "")).To(Succeed())
+	})
+
+	It("reject 0-RTT datagrams when the server doesn't support datagrams anymore", func() {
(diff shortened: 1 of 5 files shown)


integrationtests/self/zero_rtt_test.go
```diff
@@ -938,4 +938,119 @@ var _ = Describe("0-RTT", func() {
...
+		<-conn.HandshakeComplete()
+		<-received
+
+		Expect(conn.ConnectionState().Used0RTT).To(BeTrue())
+		Expect(receivedMessage).To(Equal(sentMessage))
+		num0RTT := atomic.LoadUint32(num0RTTPackets)
+		fmt.Fprintf(GinkgoWriter, "Sent %d 0-RTT packets.", num0RTT)
+		Expect(num0RTT).ToNot(BeZero())
+		zeroRTTPackets := get0RTTPackets(tracer.getRcvdLongHeaderPackets())
+		Expect(len(zeroRTTPackets)).To(BeNumerically("==", 1))
```
marten-seemann, line 997:
```suggestion
		Expect(zeroRTTPackets).To(HaveLen(1))
```

integrationtests/self/zero_rtt_test.go
```diff
@@ -938,4 +938,119 @@ var _ = Describe("0-RTT", func() {
...
+		Expect(conn.CloseWithError(0, "")).To(Succeed())
+	})
+
+	It("reject 0-RTT datagrams when the server doesn't support datagrams anymore", func() {
```
marten-seemann, line 1001: Would it make sense to test that the server doesn't process the received DATAGRAM?

internal/handshake/session_ticket_test.go
```diff
@@ -17,6 +18,7 @@ var _ = Describe("Session Ticket", func() {
 				InitialMaxStreamDataBidiLocal:  1,
 				InitialMaxStreamDataBidiRemote: 2,
 				ActiveConnectionIDLimit:        10,
+				MaxDatagramFrameSize:           protocol.MaxDatagramFrameSize,
```
marten-seemann, line 21: Can we just use a constant value here? E.g. 20 or something like that.

internal/wire/transport_parameters.go
```diff
@@ -478,7 +482,8 @@ func (p *TransportParameters) ValidFor0RTT(saved *TransportParameters) bool {
 		p.InitialMaxData >= saved.InitialMaxData &&
 		p.MaxBidiStreamNum >= saved.MaxBidiStreamNum &&
 		p.MaxUniStreamNum >= saved.MaxUniStreamNum &&
-		p.ActiveConnectionIDLimit == saved.ActiveConnectionIDLimit
+		p.ActiveConnectionIDLimit == saved.ActiveConnectionIDLimit &&
+		p.MaxDatagramFrameSize >= saved.MaxDatagramFrameSize
```
marten-seemann, line 486: This is relying on the fact that `protocol.InvalidByteCount < 0`, but that's not a guarantee.

internal/wire/transport_parameters.go
```diff
@@ -490,7 +495,8 @@ func (p *TransportParameters) ValidForUpdate(saved *TransportParameters) bool {
 		p.InitialMaxStreamDataBidiRemote >= saved.InitialMaxStreamDataBidiRemote &&
 		p.InitialMaxStreamDataUni >= saved.InitialMaxStreamDataUni &&
 		p.MaxBidiStreamNum >= saved.MaxBidiStreamNum &&
-		p.MaxUniStreamNum >= saved.MaxUniStreamNum
+		p.MaxUniStreamNum >= saved.MaxUniStreamNum &&
+		p.MaxDatagramFrameSize >= saved.MaxDatagramFrameSize
```
marten-seemann, line 499: Same here.

integrationtests/self/zero_rtt_test.go
tanghaowillow: I thought it could test the check in ValidFor0RTT from a higher level, like test that the tracer will not recevie the 0-RTT packet. But yes, it feels meaningless for a integration test considering datagram can be unreliable. Should I remove it?

integrationtests/self/zero_rtt_test.go
```diff
@@ -938,4 +938,119 @@ var _ = Describe("0-RTT", func() {
...
+		Expect(receivedMessage).To(Equal(sentMessage))
+		num0RTT := atomic.LoadUint32(num0RTTPackets)
+		fmt.Fprintf(GinkgoWriter, "Sent %d 0-RTT packets.", num0RTT)
+		Expect(num0RTT).ToNot(BeZero())
+		zeroRTTPackets := get0RTTPackets(tracer.getRcvdLongHeaderPackets())
+		Expect(zeroRTTPackets).To(HaveLen(1))
+		Expect(conn.CloseWithError(0, "")).To(Succeed())
+	})
+
+	It("reject 0-RTT datagrams when the server doesn't support datagrams anymore", func() {
```
marten-seemann, line 1001:
```suggestion
	It("rejects 0-RTT datagrams when the server doesn't support datagrams anymore", func() {
```

integrationtests/self/zero_rtt_oldgo_test.go
```diff
@@ -801,4 +801,116 @@ var _ = Describe("0-RTT", func() {
 		Expect(len(zeroRTTPackets)).To(BeNumerically(">", 10))
 		Expect(zeroRTTPackets[0]).To(Equal(protocol.PacketNumber(0)))
 	})
+
+	It("send 0-RTT datagrams", func() {
```
marten-seemann, line 805:
```suggestion
	It("sends 0-RTT datagrams", func() {
```

integrationtests/self/zero_rtt_oldgo_test.go
```diff
@@ -801,4 +801,116 @@ var _ = Describe("0-RTT", func() {
...
+		Expect(receivedMessage).To(Equal(sentMessage))
+		num0RTT := atomic.LoadUint32(num0RTTPackets)
+		fmt.Fprintf(GinkgoWriter, "Sent %d 0-RTT packets.", num0RTT)
+		Expect(num0RTT).ToNot(BeZero())
+		zeroRTTPackets := get0RTTPackets(tracer.getRcvdLongHeaderPackets())
+		Expect(zeroRTTPackets).To(HaveLen(1))
+		Expect(conn.CloseWithError(0, "")).To(Succeed())
+	})
+
+	It("reject 0-RTT datagrams when the server doesn't support datagrams anymore", func() {
```
marten-seemann, line 863:
```suggestion
	It("rejects 0-RTT datagrams when the server doesn't support datagrams anymore", func() {
```

integrationtests/self/zero_rtt_test.go
```diff
@@ -938,4 +938,119 @@ var _ = Describe("0-RTT", func() {
 		)
 		Expect(restored).To(BeTrue())
 	})
+
+	It("send 0-RTT datagrams", func() {
```
marten-seemann, line 942:
```suggestion
	It("sends 0-RTT datagrams", func() {
```

integrationtests/self/zero_rtt_test.go
marten-seemann: I'm suggesting to test that the server actually drops the 0-RTT packet containing the datagram. It would be nice have a test that makes sure that we don't at some point reject 0-RTT, but still process DATAGRAM frames arriving in 0-RTT packets.

marten-seemann: Oh I see, we have a misunderstanding here :)

What I was suggesting was calling `ReceiveMessage` on the server side, and making sure that it doesn't return the message sent by the client. Does that make sense?

tanghaowillow: Thanks for being so patient :)

The `ReceiveMessage` returns the error in this testcase because the `EnableDatagrams` is not set, which is indeed unrelated to 0-RTT. 
So checking the tracer doesn't have the 0-RTT packet (means the server has dropped it) is enough here?  Sorry if I misunderstood your words again..

marten-seemann: Oh right, I missed that you're already calling `ReceiveMessage`. All good. Sorry!

marten-seemann (approved):

marten-seemann (approved):

marten-seemann:
Thank you @tanghaowillow!

PR #4014, opened 2023-08-03
set a net.Conn for tls.ClientHelloInfo.Conn used by GetCertificate

marten-seemann:
`tls.ClientHelloInfo` is not only used for `GetConfigForClient` (which we fixed in #4001), but also in `GetGertificate`. See https://github.com/caddyserver/caddy/issues/5680#issuecomment-1664407962 for details.

@mholt, would you mind taking a look?


mholt:
LGTM. I scanned our code base for `tls.ClientHelloInfo` and sure enough the only places that we use it in connection with stdlib stuff is GetCertificate and GetConfigForClient.

Thank you!!

PR #4015, opened 2023-08-04
expose crypto/tls errors on the TransportError

### LINKED ISSUE #3730: return a tls.CertificateVerificationError when verifying the certificate fails
This error was added in Go 1.20.

marten-seemann:
Fixes #3730. This was a lot less painful than I expected.

@aschmahmann Could you try this out and see if it resolves the problem you encountered in https://github.com/libp2p/go-libp2p/pull/2451. A quick review of this PR would be appreciated :)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.9.5
 	github.com/onsi/gomega v1.27.6
 	github.com/quic-go/qpack v0.4.0
-	github.com/quic-go/qtls-go1-20 v0.3.1
+	github.com/quic-go/qtls-go1-20 v0.3.2
 	golang.org/x/crypto v0.4.0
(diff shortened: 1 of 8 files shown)


aschmahmann:
@marten-seemann thanks for the quick 😜 PR and review. From what I can tell this looks reasonable, although I'm not familiar with this code base.

However, I'm having trouble testing it because this change comes after dropping quic draft 29 and go-libp2p doesn't have a PR that deletes that dependency yet so I get compile errors. (Note: I accidentally pointed you at the what seems to be the right place to make a change since the code got shuffled a bit in between the version of quic-go I was using and master).

WDYT, should we wait until after there's a go-libp2p PR that drops draft-29 support and revisit this afterwards?

marten-seemann:
Yes, that makes sense. We're planning to update to quic-go v0.37 in our next release, so it should only be a week or two until we can test this. I'll ping you again then :)

lucas-clemente (approved):

PR #4016, opened 2023-08-04
add tls.ClientHelloInfo.Conn for recursive GetConfigForClient calls

marten-seemann:
See https://github.com/caddyserver/caddy/issues/5680#issuecomment-1665042702.

### DIFF
--- a/integrationtests/self/handshake_test.go
+++ b/integrationtests/self/handshake_test.go
@@ -142,13 +142,20 @@ var _ = Describe("Handshake tests", func() {
 
 		It("has the right local and remote address on the tls.Config.GetConfigForClient ClientHelloInfo.Conn", func() {
 			var local, remote net.Addr
+			var local2, remote2 net.Addr
 			done := make(chan struct{})
 			tlsConf := &tls.Config{
 				GetConfigForClient: func(info *tls.ClientHelloInfo) (*tls.Config, error) {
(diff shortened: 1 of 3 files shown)


mholt (approved):
@francislavoie was able to confirm that this fixes the reproducer he discovered, so that's good. Still awaiting further confirmation if anyone else in the original issue wants to try it out.

Thank you for working on this. :pray:

Looks like there may be a race condition as noted by the CI. Not clear to me, but it seems it may be specific to the integration test itself (the write to `local2` on L154, but I'm not quite sure where the concurrent read is coming from, the stack trace just says ginkgo basically). So I dunno, maybe it's just an issue with the test case and not the actual production code.

Thank you again, sorry this one is so complicated.

marten-seemann:
I'm sorry for the trouble this is causing. But you're right, this is quite complex.

The race condition should be fixed now.

mholt (approved):
Oh, no need to apologize -- you're going above and beyond to help with something that I should have been able to catch sooner in a no-stress way. :)

I'm waiting to get more confirmation that the issue is fixed but you can merge this whenever IMO.

mholt:
We have another confirmation of the fix. :+1:

marten-seemann:
Thanks for the review @mholt and thank you @francislavoie and everyone else involved in testing!

PR #4018, opened 2023-08-05
fix handling of ACK frames serialized after CRYPTO frames

marten-seemann:


mholt (approved):
I don't feel like I understand the protocol thoroughly enough to *actually* review this, but the basic logic LGTM. Nice and simple fix.

marten-seemann:
Thanks for the review @mholt! I really hope this is the last critical bug in the v0.37 release line. Turns out that switching out the underlying TLS stack (that's in fact what we did, given that the crypto/tls API is completely different (better!) from what we had in qtls) is a minefield.

connection.go
```diff
@@ -1724,6 +1743,7 @@ func (s *connection) checkTransportParameters(params *wire.TransportParameters)
 
 func (s *connection) applyTransportParameters() {
 	params := s.peerParams
+	fmt.Println("apply transport parameters", s.config.MaxIdleTimeout, params.MaxIdleTimeout)
```
marten-seemann, line 1746: Leftover debug statement. Merged via this PR, then fix using force-push on master. Sorry for that.

mholt:
Thank you @marten-seemann  -- absolutely understandable and no worries :) We have released Caddy 2.7.3 with the patches.

PR #4023, opened 2023-08-10
integrationtests: fix proxy test on Windows

egonelbre:
Windows returns a different error than non-windows.


marten-seemann (approved):
Thanks!

PR #4025, opened 2023-08-12
add utls

ghost:
add utls
build with `-tags=utls` can switch to utls branch
you need modify your code to support utls


marten-seemann:
That's a massive amount of changes. I don't think we can take on the maintenance burden for that.

marten-seemann:
It might be helpful to use an alias import, so you don’t have to change the code everywhere.

marten-seemann:
> Maybe we should move utls to a new branch instead of the main branch, and then merge it into the main branch after the merge window period

Not sure what you're referring to. You already created a new branch, and I told you that the quic-go maintainers won't be able to take on the maintenance burden of maintaining this change.

marten-seemann:
> It might be helpful to use an alias import, so you don’t have to change the code everywhere.

This comment still stands.

PR #4031, opened 2023-08-16
handshake: use the correct hash function for TLS_AES_256_GCM_SHA384

marten-seemann:
TLS_AES_128_GCM_SHA256 and TLS_CHACHA20_POLY1305_SHA256 work fine, but if TLS happens to select TLS_AES_256_GCM_SHA384 during the handshake, we're deriving the wrong keys, causing a handshake failure.

Thanks to @MarcoPolo for discovering this when interop-ing with zig-libp2p.


MarcoPolo (approved):

marten-seemann:
@mholt Yet another (and this time, very stupid copy-paste) bug introduced when switching to crypto/tls. Sorry for creating so much churn. Will ship v0.37.5 later today.

Lesson learned: We need some interop tests with a different QUIC implementation in this repo, running in CI.

mholt:
No worries :blush: I don't know of any impacts our users experienced.

PR #4035, opened 2023-08-16
ci: build interop Docker image for pushes to master, and for releases

marten-seemann:
This PR:
1. Builds a new interop Docker image for every push to master, so that the interop runner can always test the most recent image
2. Tags an image for releases. Those won't be picked up by the interop runner, but can be useful for debugging.

### DIFF
--- a/.github/workflows/build-interop-docker.yml
+++ b/.github/workflows/build-interop-docker.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - master
-      - build-interop-docker-image-for-tags
     tags:
       - 'v*'
   
--- a/.github/workflows/build-interop-docker.yml
+++ b/.github/workflows/build-interop-docker.yml
@@ -1,11 +1,17 @@
 name: Build interop Docker image
 on: 
   push:
-    branches: [ interop ]
+    branches:
+      - master
+      - build-interop-docker-image-for-tags # TODO: remove
+    tags:
+      - 'v*'
   
 jobs:
   interop:
     runs-on: ubuntu-latest
+    env:
+      TAG: "latest"
     steps:
       - uses: actions/checkout@v3
       - name: Set up QEMU
@@ -19,9 +25,12 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: set tag name
+        if: github.event_name == 'tag'
+        run: echo "TAG=${{ github.event.release.tag_name }}"
       - uses: docker/build-push-action@v4
         with:
           context: "{{defaultContext}}:interop"
           platforms: linux/amd64,linux/arm64
           push: true
-          tags: martenseemann/quic-go-interop:latest
+          tags: martenseemann/quic-go-interop:${{ env.TAG }}
--- a/.github/workflows/build-interop-docker.yml
+++ b/.github/workflows/build-interop-docker.yml
@@ -10,8 +10,6 @@ on:
 jobs:
   interop:
     runs-on: ${{ fromJSON(vars['DOCKER_RUNNER_UBUNTU'] || '"ubuntu-latest"') }}
-    env:
-      TAG: "latest"
     steps:
       - uses: actions/checkout@v3
       - name: Set up QEMU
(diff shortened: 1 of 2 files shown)


marten-seemann:
@galargh Could you take a look at this PR, especially the code path for releases? I don't really know how to test this, and you have a lot more experience with GHA workflows. Thank you!

.github/workflows/build-interop-docker.yml
```diff
@@ -1,11 +1,17 @@
 name: Build interop Docker image
 on: 
   push:
-    branches: [ interop ]
+    branches:
+      - master
+      - build-interop-docker-image-for-tags # TODO: remove
+    tags:
+      - 'v*'
```
galargh, line 8: How are the tags for this repo created? I'm asking because if they're created automatically with the default GitHub Actions token then this trigger won't work unfortunately. There are some workarounds for that though.

.github/workflows/build-interop-docker.yml
```diff
@@ -19,9 +25,14 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: set tag name
+        if: github.event_name == 'tag'
```
galargh, line 29: The event_name would be push for either tag or branch push.

We could replace the check with `startsWith(github.ref, 'refs/tags/')` but I suggest removing it altogether and deciding on the tag using the step logic.

```suggestion
        id: tag
```

.github/workflows/build-interop-docker.yml
```diff
@@ -19,9 +25,14 @@ jobs:
...
+        # Tagged releases won't be picked up by the interop runner automatically,
+        # but they can be useful when debugging regressions.
+        run: echo "TAG=${{ github.event.release.tag_name }}"
```
galargh, line 32: I don't think release object is attached to tag push events. We can extract it from ref if relevant instead.

Also, we need to echo to GitHub context file to be able to use the prepared value. Either `GITHUB_ENV` or `GITHUB_OUTPUT`. Generally, I prefer using output because it makes it more explicit where the values are produced and where they are being used.

```suggestion
        run: |
          if [[ $GITHUB_REF == refs/tags/* ]]; then
            echo "tag=${GITHUB_REF#refs/tags/}" | tee -a $GITHUB_OUTPUT;
          else
            echo 'tag=latest' | tee -a $GITHUB_OUTPUT;
          fi
```

.github/workflows/build-interop-docker.yml
```diff
@@ -19,9 +25,14 @@ jobs:
...
       - uses: docker/build-push-action@v4
         with:
           context: "{{defaultContext}}:interop"
           platforms: linux/amd64,linux/arm64
           push: true
-          tags: martenseemann/quic-go-interop:latest
+          tags: martenseemann/quic-go-interop:${{ env.TAG }}
```
galargh, line 38: With my proposed changes, we should update where we get the tag from too.

```suggestion
          tags: martenseemann/quic-go-interop:${{ steps.tag.outputs.tag }}
```

.github/workflows/build-interop-docker.yml
```diff
@@ -1,11 +1,17 @@
...
+      - build-interop-docker-image-for-tags # TODO: remove
+    tags:
+      - 'v*'
   
 jobs:
   interop:
-    runs-on: ubuntu-latest
+    runs-on: ${{ fromJSON(vars['DOCKER_RUNNER_UBUNTU'] || '"ubuntu-latest"') }}
+    env:
+      TAG: "latest"
```
galargh, line 14: With my proposed changes, we don't need to set TAG in global env.

```suggestion
```

galargh (commented):
With workflows like this one, I usually try running it on real triggers in a fork. If you're cool with the changes I proposed, I can try doing that for this repo before we merge.

.github/workflows/build-interop-docker.yml
marten-seemann: Currently I'm pushing them manually, but I was considering adopting the uCI release workflow.
How would those workaround work?

marten-seemann:
I tried it out in https://github.com/quic-go/docker-test, and it [seems to work](https://hub.docker.com/repository/docker/martenseemann/docker-test/general): 
image

Thank you for your help @galargh!

.github/workflows/build-interop-docker.yml
```diff
@@ -1,11 +1,15 @@
 name: Build interop Docker image
 on: 
   push:
-    branches: [ interop ]
+    branches:
+      - master
+      - build-interop-docker-image-for-tags
```
marten-seemann, line 6:
```suggestion
```

.github/workflows/build-interop-docker.yml
galargh: With uCI release workflows, we'd need to add a `workflow_run` trigger (for successful release workflow runs) to this job and then use https://github.com/pl-strflt/uci/tree/main/.github/actions/inspect-releaser to retrieve information about the release that the release workflow possibly created. You can ping me when you want to set it up, happy to help :)

Issue #3892, opened 2023-06-09
http3  - ResponseController/SetWriteDeadline timeout not detected client side


kgersen:
v0.35.0 added ResponseController support.
We can set `SetWriteDeadline  `on a server but it times out the client just see normal termination.

poc:

        git clone https://github.com/kgersen/h3ctx
        cd h3ctx
        go run main.go -st 1s

This will do a test with HTTP/2 then HTTP/3. 
In both case, after 1 second the server timeout (`SetWriteDeadline`). 
With quic-go the client doesn't know (reading the request body end normally with err is nil). Server side we do get an error.
With `net/http` we get an error client side too.

If haven't looked at what the official 'semantic' and behavior should be. 

But it's handy to know client side if a deadline occurred on the server. For instance if you download a file you don't know if you got it all or not. You need to use other meta  information like size or checksum which aren't always available and requires extra steps.

kgersen:
tl;dr version:
 - `net/http` with a `SetWriteDeadline(timeout)` : client "http get" a big file -> timeout server side -> error client side
 - same case with `quic-go` : client "http/3 get" a big file-> timeout server side -> no error client side (we just get the beginning of the file up to the timeout)

Glonee:
See https://datatracker.ietf.org/doc/html/rfc9114#name-request-cancellation-and-re
Looks like we didn't use any error code at `SetWriteDeadline` timeout.

francislavoie:
FYI @dunglas

PR #4039, opened 2023-08-21
http3: introduce an HTTP/3 error type

### LINKED ISSUE #3737: http3: design a proper error API

marten-seemann:
Fixes #3737.

@kgersen Would this resolve your issue? Would you mind reviewing / trying it out?

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -63,7 +63,8 @@ func (r *body) wasStreamHijacked() bool {
 }
 
 func (r *body) Read(b []byte) (int, error) {
-	return r.str.Read(b)
+	n, err := r.str.Read(b)
+	return n, maybeReplaceError(err)
 }
 
 func (r *body) Close() error {
@@ -106,7 +107,7 @@ func (r *hijackableBody) Read(b []byte) (int, error) {
 	if err != nil {
 		r.requestDone()
 	}
-	return n, err
+	return n, maybeReplaceError(err)
 }
 
 func (r *hijackableBody) requestDone() {
--- a/http3/client.go
+++ b/http3/client.go
@@ -318,13 +318,13 @@ func (c *client) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Respon
 			}
 			conn.CloseWithError(quic.ApplicationErrorCode(rerr.connErr), reason)
 		}
-		return nil, rerr.err
+		return nil, maybeReplaceError(rerr.err)
 	}
 	if opt.DontCloseRequestStream {
 		close(reqDone)
 		<-done
 	}
-	return rsp, rerr.err
+	return rsp, maybeReplaceError(rerr.err)
 }
 
 // cancelingReader reads from the io.Reader.
--- a/http3/error.go
+++ b/http3/error.go
@@ -18,7 +18,7 @@ type Error struct {
 
 var _ error = &Error{}
 
-func (e Error) Error() string {
+func (e *Error) Error() string {
 	s := e.ErrorCode.string()
 	if s == "" {
 		s = fmt.Sprintf("H3 error (%#x)", uint64(e.ErrorCode))
@@ -54,5 +54,5 @@ func maybeReplaceError(err error) error {
 		e.ErrorCode = ErrCode(appErr.ErrorCode)
 		e.ErrorMessage = appErr.ErrorMessage
 	}
-	return e
+	return &e
 }
--- a/http3/error.go
+++ b/http3/error.go
@@ -1,0 +1,58 @@
+package http3
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/quic-go/quic-go"
+)
+
+// Error is returned from the round tripper (for HTTP clients)
+// and inside the HTTP handler (for HTTP servers) if an HTTP/3 error occurs.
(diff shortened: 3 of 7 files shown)


jfgiorgi:
(kgersen here).
I've updated our poc: https://github.com/kgersen/h3ctx

when using it with this PR we get:
```
QUIC_GO_DISABLE_GSO=true go run main.go -t 1s
...
receiving data with HTTP/1.1
s->c: client got error of type context.deadlineExceededError
...
receiving data with HTTP/2.0
s->c: client got error of type context.deadlineExceededError
...
receiving data with HTTP/3.0
s->c: client got error of type http3.Error
client error for https://localhost:2222/10000000000: H3_REQUEST_CANCELLED (local)
...

```
so it works and it's better : we do get now a proper `http3.Error` instead of previoulsy a `quic.StreamError`
but
It's not a `context.deadlineExceededError` like http1.1 and http/2.

marten-seemann:
Which error are you looking at? Is this the error the client gets when performing a requests whose context has been canceled? I agree that the client should get a context error in that case, and the server should receive a H3_REQUEST_CANCELLED.

jfgiorgi:
yes it's the error received when performing a request whose context has been canceled (timeout after 1 second in the posted example).
We don't see anything server side (but I'll have to double check this tomorrow). Might be the same issue as #3892 ?

marten-seemann:
@jfgiorgi Your issue is a bit different, in that it's dealing with the local error code, not with the error sent by the remote.

I suggest merging this PR first, and then dealing with the request cancelation and #3892. Would you want to contribute a fix?

jfgiorgi:
This PR is ok for me because it's cleaner since it follows the HTTP/3 error RFC.

But when  a context is cancelled or expire, we get a HTTP/3 error instead of the context error.
I think the issue is here: https://github.com/quic-go/quic-go/blob/824fd8a2f2eb8a08fe6cef7a693fee6be3819e01/http3/client.go#L296-L300

- it returns a HTTP/3 error (H3_REQUEST_CANCELLED) instead of a context error
- so we lose the information of the context error (which is available in `req.Context().Err()`)

With `net/http` we can distinguish between a context cancel or timeout for instance because the context error is returned by the transport.

marten-seemann:
@jfgiorgi I agree. It's not immediately obvious though how to get the error, since this code is running in a separate Go routine, so you'll probably need to check the context before returning from `RoundTripOpt`.

jfgiorgi:
> @jfgiorgi I agree. It's not immediately obvious though how to get the error, since this code is running in a separate Go routine, so you'll probably need to check the context before returning from `RoundTripOpt`.

I looked at this part of the code and compared to how it's done in `net/http` _http2 roundtrip_. 
It's the other way:  the whole roundtrip is done in a separate Go routine. It seems cleaner to deal with contexts that way.
see https://github.com/golang/net/blob/master/http2/transport.go#L1233 (spawn of the Go routine)
and also : https://github.com/golang/net/blob/master/http2/transport.go#L1464-L1480 (main event loop for the go routine)

But doing that in `quic-go `might involve (or not) a lot of refactoring,

lucas-clemente (approved):

marten-seemann:
> It's the other way: the whole roundtrip is done in a separate Go routine. It seems cleaner to deal with contexts that way. see https://github.com/golang/net/blob/master/http2/transport.go#L1233 (spawn of the Go routine) and also : https://github.com/golang/net/blob/master/http2/transport.go#L1464-L1480 (main event loop for the go routine)
> 
> But doing that in `quic-go `might involve (or not) a lot of refactoring,

@jfgiorgi Want to give it a shot? I'd be happy to review a PR :)

jfgiorgi:
> @jfgiorgi Want to give it a shot? I'd be happy to review a PR :)

tbh we're very uncertain about investing more time in quic-go. The  Go team own QUIC implementation is around corner and their http/3 is probably next.
Submitting small PRs is fine, bigger ones not so sure.
And if `quic.http3` is to be refactored  it should be done to be interoperable with the api changes happening here: https://github.com/golang/go/discussions/60746
I need more understanding about all these parts before committing to anything. So it's a no for now.

marten-seemann:
Good point about API compatibility. Regarding the Go team’s QUIC effort, I wouldn’t place to much confidence in the reliability and performance of a QUIC stack that hasn’t even completed a single QUIC handshake so far…

PR #4042, opened 2023-08-22
save the RTT in non-0-RTT session tickets

### LINKED ISSUE #3853: save the RTT with the session ticket, for non-0-RTT tickets
For 0-RTT session tickets, we save the RTT in the session ticket: https://github.com/quic-go/quic-go/blob/cb3453db2508eedf4ea3ce49f86bbe029de5c1c9/internal/handshake/crypto_setup.go#L483-L489

We could do that as well for non-0-RTT tickets, allowing for a faster session resumption. This makes sense to implement once we have the new crypto/tls API (#2727).

tanghaowillow:
Save the RTT in non-0RTT session tickets.
Fixes #3853 for go 1.21.

### DIFF
--- a/internal/handshake/crypto_setup_test.go
+++ b/internal/handshake/crypto_setup_test.go
@@ -8,6 +8,8 @@ import (
 	"crypto/x509/pkix"
 	"math/big"
 	"net"
+	"runtime"
+	"strings"
 	"time"
 
 	mocktls "github.com/quic-go/quic-go/internal/mocks/tls"
@@ -449,6 +451,9 @@ var _ = Describe("Crypto Setup TLS", func() {
 				Expect(server.ConnectionState().DidResume).To(BeTrue())
 				Expect(client.ConnectionState().DidResume).To(BeTrue())
 				Expect(clientRTTStats.SmoothedRTT()).To(Equal(clientRTT))
+				if !strings.Contains(runtime.Version(), "go1.20") {
+					Expect(serverRTTStats.SmoothedRTT()).To(Equal(serverRTT))
+				}
 			})
 
 			It("doesn't use session resumption if the server disabled it", func() {
--- a/internal/handshake/crypto_setup_test.go
+++ b/internal/handshake/crypto_setup_test.go
@@ -449,7 +449,6 @@ var _ = Describe("Crypto Setup TLS", func() {
 				Expect(server.ConnectionState().DidResume).To(BeTrue())
 				Expect(client.ConnectionState().DidResume).To(BeTrue())
 				Expect(clientRTTStats.SmoothedRTT()).To(Equal(clientRTT))
-				Expect(serverRTTStats.SmoothedRTT()).To(Equal(serverRTT))
 			})
 
 			It("doesn't use session resumption if the server disabled it", func() {
--- a/internal/handshake/crypto_setup_test.go
+++ b/internal/handshake/crypto_setup_test.go
@@ -417,11 +417,13 @@ var _ = Describe("Crypto Setup TLS", func() {
 					close(receivedSessionTicket)
 				})
 				clientConf.ClientSessionCache = csc
+				const serverRTT = 25 * time.Millisecond // RTT as measured by the server. Should be restored.
 				const clientRTT = 30 * time.Millisecond // RTT as measured by the client. Should be restored.
+				serverOrigRTTStats := newRTTStatsWithRTT(serverRTT)
 				clientOrigRTTStats := newRTTStatsWithRTT(clientRTT)
 				client, _, clientErr, server, _, serverErr := handshakeWithTLSConf(
 					clientConf, serverConf,
-					clientOrigRTTStats, &utils.RTTStats{},
+					clientOrigRTTStats, serverOrigRTTStats,
 					&wire.TransportParameters{ActiveConnectionIDLimit: 2}, &wire.TransportParameters{ActiveConnectionIDLimit: 2},
 					false,
 				)
@@ -434,9 +436,10 @@ var _ = Describe("Crypto Setup TLS", func() {
 				csc.EXPECT().Get(gomock.Any()).Return(state, true)
 				csc.EXPECT().Put(gomock.Any(), gomock.Any()).MaxTimes(1)
 				clientRTTStats := &utils.RTTStats{}
+				serverRTTStats := &utils.RTTStats{}
 				client, _, clientErr, server, _, serverErr = handshakeWithTLSConf(
 					clientConf, serverConf,
-					clientRTTStats, &utils.RTTStats{},
+					clientRTTStats, serverRTTStats,
 					&wire.TransportParameters{ActiveConnectionIDLimit: 2}, &wire.TransportParameters{ActiveConnectionIDLimit: 2},
 					false,
 				)
@@ -446,6 +449,7 @@ var _ = Describe("Crypto Setup TLS", func() {
 				Expect(server.ConnectionState().DidResume).To(BeTrue())
 				Expect(client.ConnectionState().DidResume).To(BeTrue())
 				Expect(clientRTTStats.SmoothedRTT()).To(Equal(clientRTT))
+				Expect(serverRTTStats.SmoothedRTT()).To(Equal(serverRTT))
 			})
 
 			It("doesn't use session resumption if the server disabled it", func() {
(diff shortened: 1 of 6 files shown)


internal/qtls/go121.go
```diff
@@ -55,11 +55,9 @@ func SetupConfigForServer(qconf *QUICConfig, _ bool, getData func() []byte, acce
 	// add callbacks to save transport parameters into the session ticket
 	origWrapSession := conf.WrapSession
 	conf.WrapSession = func(cs tls.ConnectionState, state *tls.SessionState) ([]byte, error) {
-		// Add QUIC transport parameters if this is a 0-RTT packet.
-		// TODO(#3853): also save the RTT for non-0-RTT tickets
-		if state.EarlyData {
-			state.Extra = append(state.Extra, addExtraPrefix(getData()))
-		}
+		// Add QUIC transport parameters
+		state.Extra = append(state.Extra, addExtraPrefix(getData()))
```
marten-seemann, line 59: Is this also saving the transport parameters for non-0-RTT tickets?

tanghaowillow: Yes, currently it is. I was indeed hesitant to whether only include the RTT in the non-0RTT session ticket or keep it the same with 0-RTT session ticket.

marten-seemann: It's better to have a small ticket. There's also less that can go wrong when we receive it.

We need to be careful when parsing the ticket that we don't attempt to parse the transport parameters when they're not there.

tanghaowillow: Transport parameters are removed for non-0RTT tickets.

internal/handshake/session_ticket.go
```diff
@@ -21,7 +21,11 @@ func (t *sessionTicket) Marshal() []byte {
 	b := make([]byte, 0, 256)
 	b = quicvarint.Append(b, sessionTicketRevision)
 	b = quicvarint.Append(b, uint64(t.RTT.Microseconds()))
-	return t.Parameters.MarshalForSessionTicket(b)
+	if t.Parameters == nil {
+		return b
+	} else {
+		return t.Parameters.MarshalForSessionTicket(b)
+	}
```
marten-seemann, line 28:
```suggestion
	}
	return t.Parameters.MarshalForSessionTicket(b)
```

internal/handshake/session_ticket.go
```diff
@@ -37,11 +41,13 @@ func (t *sessionTicket) Unmarshal(b []byte) error {
 	if err != nil {
 		return errors.New("failed to read RTT")
 	}
-	var tp wire.TransportParameters
-	if err := tp.UnmarshalFromSessionTicket(r); err != nil {
-		return fmt.Errorf("unmarshaling transport parameters from session ticket failed: %s", err.Error())
+	if r.Len() > 0 {
```
marten-seemann, line 44: What do you think of passing a 0-RTT bool to this function. This would make the parsing logic stricter.

tanghaowillow: That sounds reasonable because the two kinds of ticket should be parsed with different logic and unmarshaling a 0-RTT ticket without transport parameters should fail. I have added it.

internal/handshake/session_ticket.go
```diff
@@ -37,11 +40,13 @@ func (t *sessionTicket) Unmarshal(b []byte) error {
...
-	if err := tp.UnmarshalFromSessionTicket(r); err != nil {
-		return fmt.Errorf("unmarshaling transport parameters from session ticket failed: %s", err.Error())
+	if using0RTT {
+		var tp wire.TransportParameters
+		if err := tp.UnmarshalFromSessionTicket(r); err != nil {
+			return fmt.Errorf("unmarshaling transport parameters from session ticket failed: %s", err.Error())
+		}
+		t.Parameters = &tp
 	}
-	t.Parameters = &tp
```
marten-seemann, line 44: Should we add a check here that the entire ticket has been consumed (`r.Len() == 0`)?

internal/handshake/session_ticket.go
```diff
@@ -21,10 +21,13 @@ func (t *sessionTicket) Marshal() []byte {
 	b := make([]byte, 0, 256)
 	b = quicvarint.Append(b, sessionTicketRevision)
 	b = quicvarint.Append(b, uint64(t.RTT.Microseconds()))
+	if t.Parameters == nil {
```
marten-seemann, line 24: Should we bump the `sessionTicketRevision`? Not sure if it's actually needed, but we'd be on the safe side.

internal/handshake/crypto_setup_test.go
```diff
@@ -418,11 +418,13 @@ var _ = Describe("Crypto Setup TLS", func() {
 					close(receivedSessionTicket)
 				})
 				clientConf.ClientSessionCache = csc
+				const serverRTT = 25 * time.Millisecond // RTT as measured by the server. Should be restored.
```
marten-seemann, line 421: There's no check that this is actually restored, is there?

tanghaowillow: There was a check of that, but I removed it because the RTT is only restored for go 1.21. Any advice for this test?

marten-seemann:
```go
!strings.Contains(runtime.Version(), "go1.20")
```

internal/handshake/session_ticket.go
tanghaowillow: I think maybe it is only needed for unmarshaling non-0-RTT session ticket.  A 0-RTT session ticket is end with transport parameters and unmarshaling transport parameters from session ticket always consumes all bytes of the session ticket(extra bytes would either be ignored as unknwon parameter IDs or cause an EOF error).

internal/handshake/session_ticket.go
tanghaowillow: It has been bumped to 4.

internal/handshake/session_ticket.go
```diff
@@ -21,10 +21,13 @@ func (t *sessionTicket) Marshal() []byte {
...
+		return b
+	}
 	return t.Parameters.MarshalForSessionTicket(b)
 }
 
-func (t *sessionTicket) Unmarshal(b []byte) error {
+func (t *sessionTicket) Unmarshal(using0RTT bool, b []byte) error {
```
marten-seemann, line 30: Nit: swap the parameters (`[]byte` first, then the `bool`).

internal/handshake/session_ticket_test.go
```diff
@@ -22,33 +22,61 @@ var _ = Describe("Session Ticket", func() {
 			RTT: 1337 * time.Microsecond,
 		}
 		var t sessionTicket
-		Expect(t.Unmarshal(ticket.Marshal())).To(Succeed())
+		Expect(t.Unmarshal(true, ticket.Marshal())).To(Succeed())
 		Expect(t.Parameters.InitialMaxStreamDataBidiLocal).To(BeEquivalentTo(1))
 		Expect(t.Parameters.InitialMaxStreamDataBidiRemote).To(BeEquivalentTo(2))
 		Expect(t.Parameters.ActiveConnectionIDLimit).To(BeEquivalentTo(10))
 		Expect(t.Parameters.MaxDatagramFrameSize).To(BeEquivalentTo(20))
 		Expect(t.RTT).To(Equal(1337 * time.Microsecond))
```
marten-seemann, line 30: Add a check that unmarshaling this ticket as a non-0-RTT packet fails?

internal/handshake/session_ticket_test.go
```diff
@@ -22,33 +22,61 @@ var _ = Describe("Session Ticket", func() {
...
 	})
 
+	It("marshals and unmarshals a non-0-RTT session ticket", func() {
+		ticket := &sessionTicket{
+			RTT: 1337 * time.Microsecond,
+		}
+		var t sessionTicket
+		Expect(t.Unmarshal(false, ticket.Marshal())).To(Succeed())
+		Expect(t.Parameters).To(BeNil())
+		Expect(t.RTT).To(Equal(1337 * time.Microsecond))
```
marten-seemann, line 40: Add a check that unmarshaling this ticket as a 0-RTT packet fails?

marten-seemann (commented):
Looks good, just one small nit, and two suggestions how to improve the tests.

internal/handshake/session_ticket.go
tanghaowillow: Fixed.

marten-seemann (approved):

marten-seemann:
Thanks for all your work @tanghaowillow!

Issue #3999, opened 2023-07-28
add ECN support


marten-seemann:
After talking with someone who really knows their congestion control / ECN stuff at IETF 117, here's what I learned:
* The validation algorithm described in https://datatracker.ietf.org/doc/html/rfc9000#name-sample-ecn-validation-algor is mostly fine. It doesn't really happen that ECN-marked packets are dropped completely.
* The two relevant failure modes are the following:
     1. Bleaching: ECN marks being removed completely.
     2. Mangling: ECN consistently being set to CE.

Based on that, the following validation logic makes sense:
1. Send 10 packets with ECN. If the ACK for them doesn't come back with any ECN marks, we're either dealing with bleaching or an endpoint that doesn't support ECN. In either case, turn off ECN for that path.
2. If they're all coming being acknowledged with CE, there's most likely some mangling going on. Turn ECN off.

---

- [ ] implement ECN support
- [ ] add README entry

marten-seemann:
Minimal example of how to set per-packet ECN bits: https://gist.github.com/marten-seemann/31f20a7da8e6ad69c950af93f6d9a55b.
Observation: Setting the ECN bits on macOS doesn't work, apparently due to https://github.com/apple-oss-distributions/xnu/blob/xnu-8796.101.5/bsd/netinet/udp_usrreq.c#L2970.

Slack thread: https://quicdev.slack.com/archives/C68J3KSH3/p1690758298785569

marten-seemann:
> Observation: Setting the ECN bits on macOS doesn't work, apparently due to https://github.com/apple-oss-distributions/xnu/blob/xnu-8796.101.5/bsd/netinet/udp_usrreq.c#L2970.

This is not correct. It works when using an int in the control message (and not a uint8):
```c
cmsg->cmsg_len = CMSG_LEN(sizeof(int));
int tos_value = 2;
*((int *) CMSG_DATA(cmsg)) = tos_value;
```

This will require the use of build flags so we can target macOS (and Freebsd?), which is annoying, but nothing we wouldn't be able to work around.

marten-seemann:
Design considerations:
* We're not running ECN validation during the handshake. All handshake packets will be sent unmarked.
* ECN validation starts as soon as we send (un-coalesced) 1-RTT packets. The first 1-RTT packet will be ECN-marked. This prevents on-path observers from deriving any information about the internal state from the presence of ECN marks (they can see packet types anyway).

[...]

PR #4056, opened 2023-08-28
move GSO control message handling to the oobConn

marten-seemann:
For #3999. For ECN, we'll need to append even more control messages, and the current construction would get too complicated.

### DIFF
--- a/sys_conn_oob.go
+++ b/sys_conn_oob.go
@@ -228,7 +228,6 @@ func (c *oobConn) ReadPacket() (receivedPacket, error) {
 }
 
 // WritePacket writes a new packet.
-// If the connection supports GSO, it's the caller's responsibility to append the right control mesage.
 func (c *oobConn) WritePacket(b []byte, addr net.Addr, packetInfoOOB []byte, gsoSize uint16) (int, error) {
 	oob := packetInfoOOB
 	if gsoSize > 0 {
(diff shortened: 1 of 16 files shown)


sys_conn_oob.go
```diff
@@ -229,7 +229,14 @@ func (c *oobConn) ReadPacket() (receivedPacket, error) {
 
 // WritePacket writes a new packet.
 // If the connection supports GSO, it's the caller's responsibility to append the right control mesage.
-func (c *oobConn) WritePacket(b []byte, addr net.Addr, oob []byte) (int, error) {
+func (c *oobConn) WritePacket(b []byte, addr net.Addr, packetInfoOOB []byte, gsoSize uint16) (int, error) {
+	oob := packetInfoOOB
```
sukunrt, line 233: Do you need this declaration? Is is possible to just use packetInfoOOB since it is local anyway?

sukunrt (approved):
LGTM. Just this one comment: https://github.com/quic-go/quic-go/pull/4056#discussion_r1311180689

sys_conn_oob.go
marten-seemann: I don't, but maybe it's cleaner? I want to make it clear that the caller should only pass in the packet info control message.

sys_conn_oob.go
```diff
@@ -229,7 +229,14 @@ func (c *oobConn) ReadPacket() (receivedPacket, error) {
 
 // WritePacket writes a new packet.
 // If the connection supports GSO, it's the caller's responsibility to append the right control mesage.
```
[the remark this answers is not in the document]
marten-seemann: Done.

sys_conn_oob.go
sukunrt: sure. That's fine by me.

PR #4057, opened 2023-08-29
use new gomock feature to generate type-safe methods in mocks

marten-seemann:


marten-seemann:
Tests are failing due to https://github.com/uber-go/mock/issues/80 and https://github.com/uber-go/mock/issues/81.

sukunrt (approved):

PR #4059, opened 2023-08-31
add ECN support

### LINKED ISSUE #3999: add ECN support

marten-seemann:
Fixes #3999.


internal/ackhandler/ecn.go
```diff
@@ -0,0 +1,257 @@
...
+	// If an ACK frame newly acknowledges a packet that the endpoint sent with either the ECT(0) or ECT(1)
+	// codepoint set, ECN validation fails if the corresponding ECN counts are not present in the ACK frame.
+	if (ackedECT0 > 0 || ackedECT1 > 0) && ect0 == 0 && ect1 == 0 && ecnce == 0 {
+		e.logger.Debugf("Disabling ECN. ECN-marked packet acknowledged, but no ECN counts on ACK frame.")
+		if e.tracer != nil {
+			e.tracer.ECNStateUpdated(logging.ECNStateFailed, logging.ECNFailedNoECNCounts)
+		}
+		e.state = ecnStateFailed
+		return false
+	}
```
sukunrt, line 172: Is this redundant with the two checks at the bottom?

`if newECT0+newECNCE < ackedECT0 {` and the one for ECT1

internal/ackhandler/ecn.go
```diff
@@ -0,0 +1,257 @@
...
+// It is designed only work for 1-RTT packets, it doesn't handle multiple packet number spaces.
+// In order to avoid revealing any internal state to on-path observers,
+// callers should make sure to start using ECN (i.e. calling Mode) for the very first 1-RTT packet sent.
+type ecnTracker struct {
+	state                          ecnState
+	numSentTesting, numLostTesting uint8
+
+	firstTestingPacket protocol.PacketNumber
+	lastTestingPacket  protocol.PacketNumber
+	firstCapablePacket protocol.PacketNumber
```
sukunrt, line 42: looks like this field is never set.

internal/ackhandler/ecn.go
```diff
@@ -0,0 +1,257 @@
...
+	case ecnStateInitial:
+		if e.tracer != nil {
+			e.tracer.ECNStateUpdated(logging.ECNStateTesting, logging.ECNTriggerNoTrigger)
+		}
+		e.state = ecnStateTesting
+		return e.Mode()
+	case ecnStateTesting, ecnStateCapable:
+		return protocol.ECT0
+	case ecnStateUnknown, ecnStateFailed:
+		return protocol.ECNNon
```
sukunrt, line 109: I'm not sure if this is a bug but there is an inconsistency between how state  `ecnStateUnknown` is handled. 

From `ecnStateTesting` we move to `ecnStateUnknown` after sending numTesting packets. 

In the method `Mode()` for `ecnStateUnknown` will use `ECNNon`

But in the `HandleNewlyAcked` we use `ecnMarking` which assumes everything other than `ecnStateFailed` uses ECT0 and then the method does count checking to tally that we've received the correct number of ECN counts relying on the output of `ecnMarking`.

sukunrt (commented):
Have closely reviewed the `ecn.go` file. Will finish the review after a couple of hours.

packet_packer.go
```diff
@@ -67,6 +67,11 @@ type coalescedPacket struct {
 	shortHdrPacket *shortHeaderPacket
 }
 
+// IsShortHeaderPacket says if this packet only contains a short header packet (and no long header packets).
+func (p *coalescedPacket) IsShortHeaderPacket() bool {
+	return len(p.longHdrPackets) == 0 && p.shortHdrPacket != nil
+}
```
sukunrt, line 73: NIT: Better name `IsOnlyShortHeaderPacket`

send_conn.go
```diff
@@ -3,12 +3,13 @@ package quic
...
 	"net"
 
+	"github.com/quic-go/quic-go/internal/protocol"
 	"github.com/quic-go/quic-go/internal/utils"
 )
 
 // A sendConn allows sending using a simple Write() on a non-connected packet conn.
 type sendConn interface {
-	Write(b []byte, gsoSize uint16) error
+	Write(b []byte, size uint16, ecn protocol.ECN) error
```
sukunrt, line 12: Is this rename gsoSize => size required? I see the implementation uses the same name gsoSize.

send_conn.go
```diff
@@ -42,9 +43,9 @@ func newSendConn(c rawConn, remote net.Addr, info packetInfo, logger utils.Logge
 	}
 
 	oob := info.OOB()
-	// add 32 bytes, so we can add the UDP_SEGMENT msg
+	// increase oob slice capacity, so we can add the UDP_SEGMENT and ECN control messages without allocating
 	l := len(oob)
-	oob = append(oob, make([]byte, 32)...)
+	oob = append(oob, make([]byte, 64)...)
 	oob = oob[:l]
```
sukunrt, line 49: NIT: will 
```
oob = append(oob, make([]byte, 64)...)[:len(oob)] 
```
have the same effect? 
Feel free to ignore this.

packet_packer.go
marten-seemann: I like it!

send_conn.go
marten-seemann: Keeping this `gsoSize` is definitely better.

send_conn.go
marten-seemann: It does, I tried it out: https://go.dev/play/p/C83Gg3GKzju.
But it's not obvious too me that this should work. I prefer the slightly longer but clearer version. I'll move the `[:l]` one line up though.

internal/ackhandler/ecn.go
marten-seemann: I think it is, but I'd prefer to keep is this way, because this is a special case that:
1. detects (total) bleaching.
2. detects a peer that doesn't implement ECN.

The validation logic implemented here follows the validation logic described in the RFC sentence by sentence, and it's complicated enough that I don't want to be smart here.

internal/ackhandler/ecn.go
```diff
@@ -0,0 +1,257 @@
...
+		if e.tracer != nil {
+			e.tracer.ECNStateUpdated(logging.ECNStateFailed, logging.ECNFailedTooFewECNCounts)
+		}
+		e.state = ecnStateFailed
+		return false
+	}
+	// Similarly, ECN validation fails if the sum of the increases to ECT(1) and ECN-CE counts is less than
+	// the number of newly acknowledged packets sent with an ECT(1) marking.
+	if newECT1+newECNCE < ackedECT1 {
+		e.logger.Debugf("Disabling ECN. Received less ECT(1) + ECN-CE than packets sent with ECT(1).")
```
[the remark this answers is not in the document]
marten-seemann: I struggled with the naming here as well. The problem is that we don't expect accurate counts (packets / ACKs can get lost), which is why we have all these inequalities here.

internal/ackhandler/ecn.go
marten-seemann: `ecnMarking` assumes that packets sent in Unknown are sent with Not-ECT (after I fixed the accounting for the first capable packet number), which I believe should fix this.

internal/ackhandler/ecn.go
```diff
@@ -0,0 +1,257 @@
...
+	Mode() protocol.ECN
+	HandleNewlyAcked(packets []*packet, ect0, ect1, ecnce int64) (congested bool)
+	LostPacket(protocol.PacketNumber)
+}
+
+// The ecnTracker performs ECN validation of a path.
+// Once failed, it doesn't do any re-validation of the path.
+// It is designed only work for 1-RTT packets, it doesn't handle multiple packet number spaces.
+// In order to avoid revealing any internal state to on-path observers,
+// callers should make sure to start using ECN (i.e. calling Mode) for the very first 1-RTT packet sent.
```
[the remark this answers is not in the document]
marten-seemann: I added a reference to the RFC.

internal/ackhandler/sent_packet_handler.go
```diff
@@ -333,6 +345,19 @@ func (h *sentPacketHandler) ReceivedAck(ack *wire.AckFrame, encLevel protocol.En
...
+	// Only inform the ECN tracker about new 1-RTT ACKs if the ACK increases the largest acked.
+	if encLevel == protocol.Encryption1RTT && h.ecnTracker != nil && largestAcked > pnSpace.largestAcked {
+		ecnCongestionDetected = h.ecnTracker.HandleNewlyAcked(ackedPackets, int64(ack.ECT0), int64(ack.ECT1), int64(ack.ECNCE))
+	}
+
+	pnSpace.largestAcked = utils.Max(pnSpace.largestAcked, largestAcked)
+
+	if ecnCongestionDetected {
+		h.congestion.OnCongestionEvent(largestAcked, 0, priorInFlight)
+	}
```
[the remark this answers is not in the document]

sukunrt (approved):

sys_conn_helper_freebsd.go
```diff
@@ -14,6 +14,8 @@ const (
 	ipv4PKTINFO  = 0x7
 )
 
+const ecnIPv4DataLen = 4
```
peterjeremy, line 17: This is wrong: It should be 1 for FreeBSD.  See https://cgit.freebsd.org/src/tree/sys/netinet/udp_usrreq.c#n1144
4 is the correct length for use with `setsockopt(2)` but not as a cmsg.

marten-seemann:
@peterjeremy You're commenting on a PR that was merged a couple of weeks ago. Please don't do this!

If you think you found an issue, please just open a new issue and we can continue discussing there.

PR #4060, opened 2023-09-01
integration tests: fix connection timeout in 0-RTT test

### LINKED ISSUE #4058: saving the datagram size in the session tickets breaks 0-RTT tests
After merging #4013, the 0-RTT datagram integration test takes 30s. The test still passes, but clearly there's something going wrong there.
I haven't debugged this, but this looks very much like we're running into a connection timeout somewhere.

@tanghaowillow, could you help debug and fix this?

tanghaowillow:
In 0-RTT datagram test the connection/tracer is not closed before fetching the traced packets, which makes it cannot return until the connection becomes timeout.
FIxes #4058


marten-seemann:
@tanghaowillow Thanks for the fix! Do you have any idea why this worked before #4013?

tanghaowillow:
Because these tests are introduced in #4013.. I should be have been more careful (

marten-seemann (approved):
Thanks for debugging and fixing this!

Issue #3549, opened 2022-09-08
How to best implement address validation callback


mholt:
Hey Marten,

Caddy is getting some attention on HN today due to HTTP/3 being enabled by default: https://news.ycombinator.com/item?id=32768454

[One of the comments asks](https://news.ycombinator.com/item?id=32770091 ) for an implementation of RequireAddressValidation that uses better metrics. Right now we just count `ServeHTTP()` methods (i.e. active HTTP requests), but the commenter says:

> In that case, I don't think your metric works at all, since legitimate users will increase it, but attackers won't. During an attack it will either remain the same, or even go down as fewer legitimate users manage to connect.
>
> This attack is the QUIC equivalent of a SYN flood, which results in half-open connections, because the attacker is unable to complete the connection by responding to message the server sends. RequireAddressValidation corresponds enabling syn cookies.

Do you have any ideas for what might be a better/safer way to determine whether address validation should occur?

Thanks!

CodesInChaos:
The attack `RequireAddressValidation` is designed to prevent is an attacker sending a large number SYN-packets\* where the sender IP-address is forged (similar to [SYN floods](https://en.wikipedia.org/wiki/SYN_flood) in TCP). Since the attacker does not receive responses sent to that forged IP-address, they're unable to complete the connection or respond to the address validation request, and those connections remain in some kind of [half-open/embryonic state](https://en.wikipedia.org/wiki/TCP_half-open#Embryonic_connection) (unlike TCP it can already send *early* data during this phase).

Caddy currently enables `RequireAddressValidation` if there is a large number active-connections. However it does not include half open connections in that metric. This metric will not increase during an attack, if anything it'll decrease as fewer legitimate clients manage to connect.  So it is unsuitable for the purpose of enabling address validation.

You should look into alternative metrics which are sensitive to the attack and insensitive to legitimate users instead. Some ideas:

* Number of half open connections. A connection is half open until the client proves that it can receive messages to its claimed IP-address, either via explicit address validation, or by responding to a regular packet from the server
* Number of recent connection attempts that did not eventually result in a full connection

---
\* I'm not too familiar with QUIC terminology, so I'm using TCP terminology here

marten-seemann:
Hi @mholt, thanks for opening this issue. This is a really good question. Sorry for the late response, I was on vacation. I'll reply when I'm caught up.

marten-seemann:
Relevant discussions on the QUIC Slack:
[...]

PR #4064, opened 2023-09-05
remove Config.MaxRetryTokenAge, set it to the handshake timeout

marten-seemann:
Depends on #4063. For #3549.

There is no good reason to manually set the validity period for Retry tokens. Retry tokens are only valid on a single connection during the handshake, so it makes sense to limit their validity to the configured handshake timeout.

### DIFF
--- a/config.go
+++ b/config.go
@@ -19,6 +19,10 @@ func (c *Config) handshakeTimeout() time.Duration {
 	return 2 * c.HandshakeIdleTimeout
 }
 
+func (c *Config) maxRetryTokenAge() time.Duration {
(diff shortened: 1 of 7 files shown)


sukunrt (approved):
can remove this too, I think. https://github.com/quic-go/quic-go/blob/master/config_test.go#L83

marten-seemann:
> can remove this too, I think. https://github.com/quic-go/quic-go/blob/master/config_test.go#L83

Done! Thanks for catching this!

sukunrt (approved):

PR #4066, opened 2023-09-05
add a Transport config option for the key used to encrypt tokens

### LINKED ISSUE #4065: make the key used for encrypting NEW_TOKEN tokens configurable
NEW_TOKEN tokes are supposed to be valid at all servers serving an origin (section 8.1.3 of RFC 9000):
> A token received in a NEW_TOKEN frame is applicable to any server that the connection is considered authoritative for (e.g., server names included in the certificate).

It would also be nice if there was an option to have NEW_TOKEN tokens survive reboots. This can easily be solved by not requesting a new key for every server: https://github.com/quic-go/quic-go/blob/96b1943cf58d228891e5298c429de85bdefd1ef4/internal/handshake/token_protector.go#L33-L37

Instead, we can define a `quic.TokenKey` (a `[32]byte`) on the `Transport`, similar to the `StatelessResetKey`.

marten-seemann:
Fixes #4065.


marten-seemann:
@sukunrt We can use this in libp2p. That way, resumption tokens will survive reboots of the node. (We're currently not saving tokens on the client side, but once we do, the server side is already deployed and it will just work).

lucas-clemente (approved):

PR #4069, opened 2023-09-06
randomize the serialization order of control frames

marten-seemann:
We recently discovered a bug where the receiver relied on frames being sent in a particular order. This PR randomizes the order of what we call control frames (all frames other than ACK or STREAM frames) when packing the packet.

I'd like to randomize the order of the ACK as well (currently it's always serialized first). We could do this, but we wouldn't have a way to (q)log this without changing the tracer call, so this would be a bigger change: https://github.com/quic-go/quic-go/blob/591d864e5ee1953b181644f15abc5fecda2b9fdc/logging/interface.go#L117

### DIFF
--- a/packet_packer.go
+++ b/packet_packer.go
@@ -1,8 +1,12 @@
 package quic
 
 import (
+	crand "crypto/rand"
+	"encoding/binary"
 	"errors"
 	"fmt"
+
+	"golang.org/x/exp/rand"
 
 	"github.com/quic-go/quic-go/internal/ackhandler"
 	"github.com/quic-go/quic-go/internal/handshake"
@@ -122,6 +126,7 @@ type packetPacker struct {
 	acks                ackFrameSource
 	datagramQueue       *datagramQueue
 	retransmissionQueue *retransmissionQueue
+	rand                rand.Rand
 
 	numNonAckElicitingAcks int
 }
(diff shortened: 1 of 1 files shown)


packet_packer.go
```diff
@@ -844,6 +853,11 @@ func (p *packetPacker) appendPacketPayload(raw []byte, pl payload, paddingLen pr
 	if paddingLen > 0 {
 		raw = append(raw, make([]byte, paddingLen)...)
 	}
+	// Randomize the order of the control frames.
+	// This makes sure that the receiver doesn't rely on the order in which frames are packed.
+	if len(pl.frames) > 1 {
+		p.rand.Shuffle(len(pl.frames), func(i, j int) { pl.frames[i], pl.frames[j] = pl.frames[j], pl.frames[i] })
```
marten-seemann, line 859: Using exp/rand here, since the `Shuffle` function doesn't allocate.

packet_packer.go
```diff
@@ -151,6 +159,7 @@ func newPacketPacker(
 		perspective:         perspective,
 		framer:              framer,
 		acks:                acks,
+		rand:                *rand.New(rand.NewSource(binary.BigEndian.Uint64(b[:]))),
```
marten-seemann, line 162: Using exp/rand here, since the `rand.Source` is small (16 bytes).

sukunrt (approved):
Change looks good. Just want to mention that we are now modifying payload.Frames. I can't see any case where this is an issue since it's not used anywhere after `appendPacketPayload`. So it looks fine to me.

marten-seemann:
> Change looks good. Just want to mention that we are now modifying payload.Frames. I can't see any case where this is an issue since it's not used anywhere after `appendPacketPayload`. So it looks fine to me.

I agree that it won't be a problem, but to be explicit about it, I added a comment to that function.

PR #4072, opened 2023-09-06
don't close established connections on Listener.Close, when using a Transport

### LINKED ISSUE #3962: clarify shutdown procedure
While writing up a summary for #3961 I realized that quic-go's `Listener.Close()` behaves differently than Go's `TCPListener.Close()` with the former closing all active connection, and the latter not doing that. 

Conceptually I like the idea of being able to first stop listening for new connections, then take care of gracefully shutting down any active connections. Another issue is that `listener.Close()` doesn't accept a user error like `connection.CloseWithError(error)` does. Even if it would accept an error, users may not want to close all active connections with the same error.

I also noticed that I can cancel calls to `listener.Accept(ctx)` by cancelling the passed context, though I am not sure if it would have an adverse effect when not calling Accept for some time and clients want to open new connections.

As it is now I have to decide between:
1. stopping to listen by calling `listener.Close()` and not being able to convey the desired error to the remote side (neither to have it surfaced via connections context cancellation cause, see #3961)
2. stopping to listen by calling `listener.Close()`, also calling `conn.CloseWithError(err)`, and hope that `conn.CloseWithError(err)` can cleanly close the connection before `listener.Close()` does what it does.
3. Canceling any pending `listener.Accept(ctx)` calls, closing all open connections via `conn.CloseWithError(err)`, once all connection contexts are done, close the listener, then the transport, then the transport.Conn.

I have a feeling that option 3 is probably the recommended way, but unsure about cancelling the context on `listener.Accept(ctx)`. Though upon further testing I found that even when cancelling the accept context, the library still continues to accept connections. At least on client side the call to `transport.Dial` completes with `err = nil`. Later, after `transport.Close` is called on the server, the client receives `Application error 0x0 (remote)`.

marten-seemann:
Part of #3962.

When `Listener.Close` is called, now:
* `Accept` will return `ErrServerClosed` as soon as all connections in the accept queue have been accepted, and
* QUIC handshakes that are still in flight will be rejected with a CONNECTION_REFUSED error.

Closing the listener doesn't have any effect on already established connections anmore.

### DIFF
--- a/server.go
+++ b/server.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -100,9 +101,11 @@ type baseServer struct {
 		protocol.VersionNumber,
 	) quicConn
 
-	errorChan               chan struct{} // is closed when the server is closed
-	closeErr                error
-	running                 chan struct{} // closed as soon as run() returns
+	closeOnce sync.Once
+	errorChan chan struct{} // is closed when the server is closed
+	closeErr  error
+	running   chan struct{} // closed as soon as run() returns
+
 	versionNegotiationQueue chan receivedPacket
 	invalidTokenQueue       chan rejectedPacket
 	connectionRefusedQueue  chan rejectedPacket
@@ -329,18 +332,15 @@ func (s *baseServer) Close() error {
 }
 
 func (s *baseServer) close(e error, notifyOnClose bool) {
-	select {
-	case <-s.errorChan: // already closed
-		return
-	default:
-	}
-	s.closeErr = e
-	close(s.errorChan)
+	s.closeOnce.Do(func() {
+		s.closeErr = e
+		close(s.errorChan)
 
-	<-s.running
-	if notifyOnClose {
-		s.onClose()
-	}
+		<-s.running
+		if notifyOnClose {
+			s.onClose()
+		}
+	})
 }
 
 // Addr returns the server's network address
--- a/server.go
+++ b/server.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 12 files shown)


server.go
```diff
@@ -321,38 +319,28 @@ func (s *baseServer) accept(ctx context.Context) (quicConn, error) {
...
-		s.serverError = ErrServerClosed
-	}
-	s.closed = true
-	close(s.errorChan)
-	s.mutex.Unlock()
-
-	<-s.running
-	s.onClose()
+	s.close(ErrServerClosed, true)
 	return nil
```
sukunrt, line 328: You still need the mutex, no? If two goroutines call Close simultaneously, they might close s.errorChan twice?

sukunrt (changes requested):

transport.go
```diff
@@ -294,7 +295,6 @@ func (t *Transport) Close() error {
 }
 
 func (t *Transport) closeServer() {
-	t.handlerMap.CloseServer()
 	t.mutex.Lock()
 	t.server = nil
 	if t.isSingleUse {
```
sukunrt, line 300: On line 304, we will end up Closing the underlying udp connection if t.createdConn was true, this will end up closing the established quic connections. Do we want this?

marten-seemann: I don't see any other way, unless we want to keep the Go routine reading from the socket running.

server.go
marten-seemann: Fixed by introducing a `sync.Once`.

transport.go
marten-seemann: Otherwise, we'd have to keep a reference counter to eventually stop listening once all connections are closed. That would actually be quite nice.

transport.go
```diff
@@ -322,7 +322,7 @@ func (t *Transport) close(e error) {
 		t.handlerMap.Close(e)
 	}
 	if t.server != nil {
-		t.server.setCloseError(e)
+		t.server.close(e, false)
```
sukunrt, line 325: Why do we need this? isn't this method called from `server.close`?

sukunrt (approved):

transport.go
sukunrt: Ignore this. I didn't realise this is being called from transport.close and not from transport.closeServer

Issue #3704, opened 2023-02-12
use a struct, not an interface, for tracers


marten-seemann:
Our tracing interface is quite large: https://github.com/quic-go/quic-go/blob/5b72f4c900f209b5705bb0959399d59e495a2c6e/logging/interface.go#L116-L146

It also quite expensive, especially since it causes a large amount of allocs when processing frames: https://github.com/quic-go/quic-go/blob/5b72f4c900f209b5705bb0959399d59e495a2c6e/connection.go#L1280-L1292

One option would be to convert the interface to a struct. Users could then set the specific tracing calls that they're interested in.

PR #4076, opened 2023-09-11
remove duplicate mocks for the Tracer and the ConnectionTracer

marten-seemann:
They were introduced to avoid an import loop in the tests in the logging package, but the same can be achieved by having a dedicated package for the tests (`logging_test`).

This will make work on #3704 easier.


marten-seemann:
Merging to avoid further merge conflicts.

PR #4079, opened 2023-09-11
Remove LocalAddr check

nekohasekai:
Due to a legacy workaround, quic-go requires PacketConn to have a unique LocalAddr.String, which is not mentioned in the documentation and can cause confusion.

This PR removes the relevant code.


marten-seemann:
We're not ready to remove this check yet. The `Transport` API refactor has only been released in [v0.35](https://github.com/quic-go/quic-go/releases/tag/v0.35.0) in May this year. We should probably keep it around for one year in total.

marten-seemann:
Maybe not in the general case, since `net.PacketConn` doesn't prescribe what `LocalAddr` should return.

Nevertheless we shouldn't remove this check yet. The old quic-go API asked users to pass the same `net.PacketConn` to multiple `Listen` and `Dial` calls, which the new API explicitly forbids. This was a dangerous change to make, and this check will catch it immediately.

PR #4084, opened 2023-09-14
move MaxTokenAge configuration option to the Transport

marten-seemann:
Even more API cleanup. Trying to do all the token-related API changes in one release, to reduce the amount of pain caused for users.

~~Depends on #4066.~~

### DIFF
--- a/internal/handshake/token_protector.go
+++ b/internal/handshake/token_protector.go
@@ -61,7 +61,7 @@ func (s *tokenProtectorImpl) DecodeToken(p []byte) ([]byte, error) {
 }
 
 func (s *tokenProtectorImpl) createAEAD(nonce []byte) (cipher.AEAD, []byte, error) {
-	h := hkdf.New(sha256.New, s.key[:], nonce[:], []byte("quic-go token source"))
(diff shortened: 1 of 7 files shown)


internal/handshake/token_protector.go
```diff
@@ -68,7 +61,7 @@ func (s *tokenProtectorImpl) DecodeToken(p []byte) ([]byte, error) {
 }
 
 func (s *tokenProtectorImpl) createAEAD(nonce []byte) (cipher.AEAD, []byte, error) {
-	h := hkdf.New(sha256.New, s.secret, nonce, []byte("quic-go token source"))
+	h := hkdf.New(sha256.New, s.key[:], nonce[:], []byte("quic-go token source"))
```
sukunrt, line 64: NIT: `nonce` instead of `nonce[:]`

sukunrt (approved):

internal/handshake/token_protector.go
marten-seemann, line 64:
```suggestion
	h := hkdf.New(sha256.New, s.key[:], nonce, []byte("quic-go token source"))
```

PR #4089, opened 2023-09-17
ackhandler: fix ECN mangling detection when packets are lost

marten-seemann:
Some of the 10 testing packets are might be lost, while others might be CE-marked. We need to detect mangling if all testing packets are either lost are CE-marked.

~~Depends on #4088.~~

### DIFF
--- a/internal/ackhandler/ecn.go
+++ b/internal/ackhandler/ecn.go
@@ -135,7 +135,10 @@ func (e *ecnTracker) LostPacket(pn protocol.PacketNumber) {
 			e.tracer.ECNStateUpdated(logging.ECNStateFailed, logging.ECNFailedLostAllTestingPackets)
 		}
 		e.state = ecnStateFailed
+		return
 	}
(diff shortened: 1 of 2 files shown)


internal/ackhandler/ecn.go
```diff
@@ -259,6 +261,18 @@ func (e *ecnTracker) HandleNewlyAcked(packets []*packet, ect0, ect1, ecnce int64
 	return e.state == ecnStateCapable && newECNCE > 0
 }
 
+// detectMangling fails ECN validation if all testing packets are lost or CE-marked.
+func (e *ecnTracker) detectMangling() {
```
sukunrt, line 265: NIT: `detectMangling` reads like it should return a bool. But I also don't have a good suggestion. The best I can think of is `failIfMangled`

sukunrt (approved):

internal/ackhandler/ecn.go
marten-seemann: Makes sense, I'll take it.

Issue #4077, opened 2023-09-11
provide Prometheus metrics


marten-seemann:
We should provide a way to collect Prometheus metrics for node operators.

The metrics tracer should probably be an implementation of the existing `Tracer` / `ConnectionTracer`. To make this performant, we'll need to resolve #3704 first. The code should live in a separate package, so that we're not forcing a Prometheus dependency onto anyone (there won't be a way to prevent it from being included in the `go.mod` though, unless we move this to a separate Go package).

We should also provide a sample Grafana dashboard.

PR #4092, opened 2023-09-17
don't spawn a new Go routine to send a Retry packet

marten-seemann:
Depends on #4091.

### DIFF
--- a/server.go
+++ b/server.go
@@ -111,6 +111,7 @@ type baseServer struct {
 	versionNegotiationQueue chan receivedPacket
 	invalidTokenQueue       chan rejectedPacket
 	connectionRefusedQueue  chan rejectedPacket
+	retryQueue              chan rejectedPacket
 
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -253,6 +254,7 @@ func newServer(
 		versionNegotiationQueue:   make(chan receivedPacket, 4),
 		invalidTokenQueue:         make(chan rejectedPacket, 4),
 		connectionRefusedQueue:    make(chan rejectedPacket, 4),
+		retryQueue:                make(chan rejectedPacket, 8),
```
sukunrt, line 257: Any reason for the queue lengths?

sukunrt (approved):

server.go
marten-seemann: This isn't backed by any data, but I'd expect Retries to be sent more frequently than the other types of packets. #4077 will give us more insights at some point.

PR #4101, opened 2023-09-24
move race detection to the testutils package

marten-seemann:


marten-seemann:
I can't remember why I proposed this change.

PR #4103, opened 2023-09-25
limit concurrency when closing all connections handled by a Transport

marten-seemann:
I'm not 100% sure we not a Go routine here in the first place, but limiting the concurrency definitely seems reasonable.

### DIFF
--- a/packet_handler_map.go
+++ b/packet_handler_map.go
@@ -247,12 +247,15 @@ func (h *packetHandlerMap) Close(e error) {
 
 	close(h.closeChan)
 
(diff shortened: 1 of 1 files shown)


sukunrt (commented):
Change seems okay but I'm not sure what is the deadlock leading to the test failures. The only difference I can think of is previously we released packatHandlerMap mutex and let the goroutines run to completion and now we don't release the mutex.

marten-seemann:
The problem is that the connections remove their connection IDs from the map, which requires them to take the mutex which we're now still holding.

PR #4111, opened 2023-10-13
catch EPERM sendmsg errors for the very first packet on Linux

marten-seemann:
This is a workaround for https://github.com/golang/go/issues/63322. On Linux, we now ignore the EPERM error for the very first sendmsg call on a new connection, and retry exactly once.

This is expected to dramatically reduce the flakiness of our test suite.

### DIFF
--- a/send_conn.go
+++ b/send_conn.go
@@ -28,6 +28,9 @@ type sconn struct {
 	packetInfoOOB []byte
 	// If GSO enabled, and we receive a GSO error for this remote address, GSO is disabled.
 	gotGSOError bool
+	// Used to catch the error sometimes returned by the first sendmsg call on Linux,
+	// see https://github.com/golang/go/issues/63322.
+	wroteFirstPacket bool
 }
 
 var _ sendConn = &sconn{}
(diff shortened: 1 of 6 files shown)


send_conn.go
```diff
@@ -28,6 +28,9 @@ type sconn struct {
 	packetInfoOOB []byte
 	// If GSO enabled, and we receive a GSO error for this remote address, GSO is disabled.
 	gotGSOError bool
+	// Used to catch the error sometimes returned by the first sendmsg call on Linux,
+	// see https://github.com/golang/go/issues/63322.
+	wroteFirstPacket bool
```
sukunrt, line 33: Is it confirmed that this happens only on the first packet? I didn't check this when I found that issue.

sukunrt (approved):

send_conn.go
marten-seemann: I tested it, and it never occurred after the first write. I realize this is not a very satisfying answer, but to find this out we'd have to dive very deep into the Go / Linux networking stack.

Issue #3522, opened 2022-08-24
http3: implement HTTP Datagrams and Capsules (RFC 9297)


marten-seemann:
https://www.rfc-editor.org/rfc/rfc9297.html

marten-seemann:
Writing and parsing of CAPSULEs was implement by #3607.

marten-seemann:
The API will be a bit tricky. Here's a proposal.

## On the client side

`RoundTripper.EnableDatagrams` enables both QUIC and HTTP datagrams.

HTTP Datagrams are always associated with an HTTP Request (section 2 of the RFC):
> HTTP Datagrams MUST only be sent with an association to an HTTP request that explicitly supports them. For example, existing HTTP methods GET and POST do not define semantics for associated HTTP Datagrams; therefore, HTTP Datagrams associated with GET or POST request streams cannot be sent.

It seems like the `http.RoundTripper` interface isn't suitable for that use case, since we need to be able to send datagrams right after sending the HTTP request (without having to wait for the response). However, `RoundTrip` blocks until the HTTP response has been received from the server. This means that we need to introduce a new interface:
```go
type RoundTripperWithDatagrams interface {
     RoundTripWithDatagrams(*http.Request) (Datagrammer, <-chan *http.Response, error)
}
```
Calling `RoundTripWithDatagrams` immediately returns to allow sending datagrams. The HTTP response is sent on the channel once it is received. `Datagrammer` is an interface that allows sending of HTTP datagrams:
```go
type Datagrammer interface {
     // SendMessage sends an HTTP Datagram associated with an HTTP request.
     // It must only be called while the send side of the stream is still open, i.e.
     // * on the client side: before calling Close on the request body
     // * on the server side: before calling Close on the response body
     SendMessage([]byte) error
     // SendMessage receives an HTTP Datagram associated with an HTTP request:
     // * on the server side: datagrams can be received while the request handler hasn't returned, AND 
     //      the client hasn't close the request stream yet
     // * on the client side: datagrams can be received with the server hasn't close the response stream
     ReceiveMessage(context.Context) ([]byte, error)
}
```
Note that while this interface looks like the functions exposed the `quic.Connection`, it does prepend the Quarter Stream ID on `SendMessage` as defined in section 2.1 of the RFC, and demultiplexes (and removes) it on `ReceiveMessage`.

## On the server side

`Server.EnableDatagrams` enables both QUIC and HTTP datagrams.

[...]

PR #4112, opened 2023-10-15
add datagrams support

curquhart:
- add send and receieve message to StreamCreator so that datagrams can be sent and received
required for https://github.com/quic-go/webtransport-go/pull/99


marten-seemann:
This is incorrect, or at the very least maximally confusing. What you’re exposing here is QUIC datagrams, not HTTP/3 datagrams.

marten-seemann:
API proposal for an HTTP datagram API: https://github.com/quic-go/quic-go/issues/3522#issuecomment-1773671123

PR #4115, opened 2023-10-18
http3: discard body from responses to HEAD requests

### LINKED ISSUE #4097: http3: HEAD method should not have a body
image

fix like http lib @ go/src/net/http/server.go line 377
image

Glonee:
fix #4097

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -70,6 +70,7 @@ type responseWriter struct {
 	headerWritten bool
 	contentLen    int64 // if handler set valid Content-Length header
 	numWritten    int64 // bytes written
+	isHead        bool
 }
 
 var (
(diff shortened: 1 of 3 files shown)


http3/server.go
```diff
@@ -599,6 +599,9 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 	ctx = context.WithValue(ctx, http.LocalAddrContextKey, conn.LocalAddr())
 	req = req.WithContext(ctx)
 	r := newResponseWriter(str, conn, s.logger)
+	if req.Method == "HEAD" {
```
marten-seemann, line 602:
```suggestion
	if req.Method == http.MethodHead {
```

http3/response_writer.go
```diff
@@ -70,6 +70,7 @@ type responseWriter struct {
 	headerWritten bool
 	contentLen    int64 // if handler set valid Content-Length header
 	numWritten    int64 // bytes written
+	isHead        bool
```
marten-seemann, line 73: Nit: if you move `headerWritten` down here, this struct will be 8 bytes smaller (I think).

Glonee: Good point out. I have made this change.

marten-seemann (approved):

PR #4120, opened 2023-10-23
use typed atomics in integration tests

marten-seemann:


integrationtests/self/hotswap_test.go
```diff
@@ -49,7 +49,7 @@ func (ln *fakeClosingListener) Accept(ctx context.Context) (quic.EarlyConnection
 func (ln *fakeClosingListener) Close() error {
 	if atomic.CompareAndSwapInt32(&ln.closed, 0, 1) {
```
sukunrt, line 50: You can also change `ln.closed` to `atomic.Int32`

sukunrt (approved):

integrationtests/self/hotswap_test.go
marten-seemann: Good catch! This could be an `atomic.Bool` actually.

PR #4121, opened 2023-10-23
ci: create separate artifact archives per workflow run

marten-seemann:
Also reduces the retention time to 7 days. This is only used for debugging, so one week should be fine.


marten-seemann:
Tested in https://github.com/quic-go/quic-go/actions/runs/6609176460?pr=4119.

PR #4122, opened 2023-10-23
ci: use bash on all platforms

marten-seemann:


marten-seemann:
This doesn't work. Windows is just terrible. A pain in every regard.

marten-seemann:
Tested in https://github.com/quic-go/quic-go/actions/runs/6609475017?pr=4119:
image

PR #4143, opened 2023-10-30
send large max_datagram_frame size, introduce a DatagramTooLargeError error

chungthuang:
The size can be overwritten to a lower value for testing. This implements the suggestion from https://github.com/quic-go/quic-go/issues/3300#issuecomment-1642475753.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -61,3 +61,15 @@ func (e *StreamError) Error() string {
 	}
 	return fmt.Sprintf("stream %d canceled by %s with error code %d", e.StreamID, pers, e.ErrorCode)
 }
+
+// DatagramTooLargeError is returned from Connection.SendDatagram if the payload is too large to be sent.
+type DatagramTooLargeError struct {
+	PeerMaxDatagramFrameSize int64
+}
+
+func (e *DatagramTooLargeError) Is(target error) bool {
+	_, ok := target.(*DatagramTooLargeError)
+	return ok
+}
+
+func (e *DatagramTooLargeError) Error() string { return "DATAGRAM frame too large" }
--- a/integrationtests/self/datagram_test.go
+++ b/integrationtests/self/datagram_test.go
@@ -19,7 +19,8 @@ import (
 )
 
 var _ = Describe("Datagram test", func() {
-	const num = 100
+	const concurrentSends = 100
+	const maxDatagramSize = 250
 
 	var (
 		serverConn, clientConn *net.UDPConn
@@ -47,11 +48,11 @@ var _ = Describe("Datagram test", func() {
 
 			if expectDatagramSupport {
 				Expect(conn.ConnectionState().SupportsDatagrams).To(BeTrue())
-
 				if enableDatagram {
+					f := &wire.DatagramFrame{DataLenPresent: true}
 					var wg sync.WaitGroup
-					wg.Add(num)
-					for i := 0; i < num; i++ {
+					wg.Add(concurrentSends)
+					for i := 0; i < concurrentSends; i++ {
 						go func(i int) {
 							defer GinkgoRecover()
 							defer wg.Done()
@@ -60,6 +61,11 @@ var _ = Describe("Datagram test", func() {
 							Expect(conn.SendDatagram(b)).To(Succeed())
 						}(i)
 					}
+					maxDatagramMessageSize := f.MaxDataLen(maxDatagramSize, conn.ConnectionState().Version)
+					b := make([]byte, maxDatagramMessageSize+1)
+					Expect(conn.SendDatagram(b)).To(MatchError(&quic.DatagramTooLargeError{
+						PeerMaxDatagramFrameSize: int64(maxDatagramMessageSize),
+					}))
 					wg.Wait()
 				}
 			} else {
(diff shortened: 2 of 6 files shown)


internal/wire/datagram_frame.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/quic-go/quic-go/quicvarint"
 )
 
+// MaxDatagramSize is the maximum size of a DATAGRAM frame (RFC 9221).
+// By setting it to a large value, we allow all datagrams that fit into a QUIC packet.
+// The value is chosen such that it can still be encoded as a 2 byte varint.
+var MaxDatagramSize uint64 = 16383
```
marten-seemann, line 14: Have you considered making this a `protocol.ByteCount`?

marten-seemann (commented):
You'll need to sign the CLA as well.

internal/wire/datagram_frame.go
chungthuang: Sure, I will change that.

chungthuang:
Hi @marten-seemann, thanks for reviewing. Do you have other feedback for this PR?

marten-seemann:
> The size can be overwritten to a lower value for testing.

Do you want to add a test case?

chungthuang:
> > The size can be overwritten to a lower value for testing.
> 
> Do you want to add a test case?
https://github.com/quic-go/quic-go/pull/4143/commits/257b644b4237bcd2be1957139065b0cce9170de6 adds a test case to make sure peer's max DATAGRAM size is respected.
I think the other case of making sure only DATAGRAM frame that can fit into a packet is sent is already covered in https://github.com/quic-go/quic-go/blob/master/packet_packer_test.go#L594.

internal/wire/datagram_frame.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/quic-go/quic-go/quicvarint"
 )
 
+// MaxDatagramSize is the maximum size of a DATAGRAM frame (RFC 9221).
+// By setting it to a large value, we allow all datagrams that fit into a QUIC packet.
+// The value is chosen such that it can still be encoded as a 2 byte varint.
```
marten-seemann, line 13:
```suggestion
// MaxDatagramSize is the maximum size of a DATAGRAM frame (RFC 9221).
// By setting it to a large value, we allow all datagrams that fit into a QUIC packet.
// The value is chosen such that it can still be encoded as a 2 byte varint.
// This is a var and not a const so it can be set in tests.
```

integrationtests/self/datagram_test.go
```diff
@@ -47,19 +46,23 @@ var _ = Describe("Datagram test", func() {
 
 			if expectDatagramSupport {
 				Expect(conn.ConnectionState().SupportsDatagrams).To(BeTrue())
-
 				if enableDatagram {
+					f := &wire.DatagramFrame{DataLenPresent: true}
+					maxDatagramMessageSize := f.MaxDataLen(num, conn.ConnectionState().Version)
```
marten-seemann, line 51: `num` is the concurrency parameter here. Can we use a different value? Maybe just set it to 250 bytes or something like that.

internal/qerr/errors.go
```diff
@@ -137,3 +137,15 @@ func getRole(remote bool) string {
 	}
 	return "local"
 }
+
+type DatagramTooLargeError struct {
```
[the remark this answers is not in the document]
chungthuang: Sure I will move it. For my understanding, what kind of errors should be defined in qerr package?

internal/qerr/errors.go
```diff
@@ -137,3 +137,15 @@ func getRole(remote bool) string {
...
+	PeerMaxDatagramFrameSize int64
+}
+
+func (e *DatagramTooLargeError) Timeout() bool   { return false }
+func (e *DatagramTooLargeError) Temporary() bool { return true }
```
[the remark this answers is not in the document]
chungthuang: No, I just did it because other errors in this package implement it. But it's not needed as we move to quic package.

internal/qerr/errors.go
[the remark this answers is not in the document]
marten-seemann: Good question!

The qerr package mostly exists to avoid import loops. Other sub-packages in this repo need to be able to emit specific errors.

integrationtests/self/datagram_test.go
chungthuang: Done.

chungthuang:
Hi @marten-seemann, I can squash the commits if you don't have other feedback.

chungthuang:
Hi @marten-seemann, just want to check if this is good to merge soon?

errors.go
```diff
@@ -61,3 +61,14 @@ func (e *StreamError) Error() string {
 	}
 	return fmt.Sprintf("stream %d canceled by %s with error code %d", e.StreamID, pers, e.ErrorCode)
 }
+
+type DatagramTooLargeError struct {
```
marten-seemann, line 65:
```suggestion
// DatagramTooLargeError is returned from Connection.SendDatagram if the payload is too large to be sent.
type DatagramTooLargeError struct {
```

marten-seemann (approved):

marten-seemann:
Sorry for the delay @chungthuang! Tests are failing, can you please take a look?

chungthuang:
> Sorry for the delay @chungthuang! Tests are failing, can you please take a look?

No worries. The tests were failing because `internal/wire/transport_parameter_test.go` was still using `protocol.MaxDatagramFrameSize`. This is fixed now.

marten-seemann (approved):

chungthuang:
Looks like all tests passed, but I don't have permission to merge.

gr0l0rg:
FINALLY! Thank you @chungthuang for the great work.

Wasem9926:
abohosam_3632@hotmail.com

Issue #4144, opened 2023-10-30
Prevent large datagrams from blocking the send queue


chungthuang:
The [packet packer logic](https://github.com/quic-go/quic-go/blob/v0.39.3/packet_packer.go#L607-L614) can get into a loop of peeking the same DATAGRAM frame without sending if it cannot fit in a packet. This blocks subsequent calls to `SendDatagram`. One possible workaround is passing a context to `SendDatagram` then `datagramQueue.AddAndWait`, so the datagram can be removed from queue when the context is cancelled.

chungthuang:
https://github.com/quic-go/quic-go/pull/4145 is a proposal for how the `SendDatagram` will look like with context.

marten-seemann:
How do other datagram APIs deal with this? For example, WebTransport has an API to send datagrams. They must be running into the same issue.

One option would be to return a `ErrDatagramTooLarge` error. However, one reason for the bad datagram performance we're currently suffering from (#3766) is that every call to `SendDatagram` waits until the datagram has been packaged. The fix for that seems easy: just have a queue of datagrams, allowing more datagrams to be queued, so we can efficiently send them out. However, this also means that we won't be able to return a separate error per call to `SendDatagram`.

chungthuang:
Looking at https://www.w3.org/TR/webtransport/#example-fixed-rate, WebTransport API provides an option for the caller to wait till the transport is ready. Perhaps `SendDatagram` can return an error channel of size 1 if we move to a queue of datagrams. That way the caller can decide if they want to wait for the feedback.
However, I think improving performance can be tackle separately. In the meantime to prevent large datagrams from blocking the queue, we can:
- Reject datagrams larger than current MTU. MTU is unlikely to increase in the near future
- For datagrams size less than MTU but larger than the remaining size of the QUIC packet, drop it if it cannot be sent after a few retries

Issue #3766, opened 2023-04-08
Poor datagram extension send performance


birneee:
I noticed that the sending rate of datagram frames is much lower than that of stream frames.

Example measurement with streams:

```bash
$ qperf client --addr [::1]:18080 --receive-stream
{"qlog_format":"NDJSON","qlog_version":"draft-02","title":"quic-go qlog","code_version":"v0.33.0","trace":{"vantage_point":{"type":"client"},"common_fields":{"ODCID":"a7a4f894c0eb3547b4","group_id":"a7a4f894c0eb3547b4","reference_time":1680959374739.0117,"time_format":"relative"}}}
{"time":0.009531,"name":"transport:connection_started","data":{"ip_version":"ipv6","src_ip":"::","src_port":36664,"dst_ip":"::1","dst_port":18080,"src_cid":"(empty)","dst_cid":"a7a4f894c0eb3547b4"}}
{"time":2.909691,"name":"qperf:handshake_completed","data":{}}
{"time":3.156259,"name":"qperf:handshake_confirmed","data":{}}
{"time":3.194904,"name":"qperf:first_app_data_received","data":{}}
{"time":1003.043641,"name":"qperf:report","data":{"stream_mbps_received":1795.8972,"stream_bytes_received":225188568,"max_rtt":2.245394,"packets_lost":0,"period":1003.1245}}
{"time":2003.067291,"name":"qperf:report","data":{"stream_mbps_received":1830.302,"stream_bytes_received":228793158,"max_rtt":0.318308,"packets_lost":0,"period":1000.0236}}
{"time":3003.103221,"name":"qperf:report","data":{"stream_mbps_received":1851.251,"stream_bytes_received":231414678,"max_rtt":0.323288,"packets_lost":0,"period":1000.0359}}
{"time":4003.111423,"name":"qperf:report","data":{"stream_mbps_received":1771.6084,"stream_bytes_received":221452902,"max_rtt":0.360415,"packets_lost":0,"period":1000.0083}}
{"time":5003.120854,"name":"qperf:report","data":{"stream_mbps_received":1726.5168,"stream_bytes_received":215816634,"max_rtt":0.314119,"packets_lost":0,"period":1000.0094}}
{"time":6003.125988,"name":"qperf:report","data":{"stream_mbps_received":1813.5583,"stream_bytes_received":226695942,"max_rtt":0.449681,"packets_lost":0,"period":1000.005}}
[...]

PR #4145, opened 2023-10-30
Support passing context to SendDatagram API (#4144)

chungthuang:
This can prevent a large datagram from blocking the send queue by dropping it when the context is cancelled.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -2343,7 +2343,7 @@ func (s *connection) onStreamCompleted(id protocol.StreamID) {
 	}
 }
 
-func (s *connection) SendDatagram(p []byte) error {
+func (s *connection) SendDatagram(ctx context.Context, p []byte) error {
 	if !s.supportsDatagrams() {
 		return errors.New("datagram support disabled")
 	}
@@ -2356,7 +2356,7 @@ func (s *connection) SendDatagram(p []byte) error {
 	}
 	f.Data = make([]byte, len(p))
 	copy(f.Data, p)
-	return s.datagramQueue.AddAndWait(f)
+	return s.datagramQueue.AddAndWait(ctx, f)
 }
 
 func (s *connection) ReceiveDatagram(ctx context.Context) ([]byte, error) {
--- a/connection.go
+++ b/connection.go
@@ -2343,7 +2343,7 @@ func (s *connection) onStreamCompleted(id protocol.StreamID) {
 	}
 }
 
-func (s *connection) SendDatagram(ctx context.Context, p []byte) error {
+func (s *connection) SendDatagram(p []byte) error {
 	if !s.supportsDatagrams() {
 		return errors.New("datagram support disabled")
 	}
@@ -2359,7 +2359,7 @@ func (s *connection) SendDatagram(ctx context.Context, p []byte) error {
 	}
 	f.Data = make([]byte, len(p))
 	copy(f.Data, p)
-	return s.datagramQueue.AddAndWait(ctx, f)
+	return s.datagramQueue.AddAndWait(f)
 }
 
 func (s *connection) ReceiveDatagram(ctx context.Context) ([]byte, error) {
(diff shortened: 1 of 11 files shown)


marten-seemann:
I don't think using a context is a clean API. This would require very intimate coordination with the application.

chungthuang:
I am trying a new approach:
1. Don't change `SendDatagram` API, but reject datagrams larger than current MTU. MTU is unlikely to increase in the near future
2. For datagrams size less than MTU but larger than the remaining size of the QUIC packet, drop it if it cannot be sent after a few retries. Each `Peek` counts as a retry

chungthuang:
Hi @marten-seemann, I tried with a simpler approach without changing the API. Can you share feedback or ideas for other approaches?

chungthuang:
Hi @marten-seemann, thank you for reviewing my previous PR. Can you also take a look at this one and let me know if you think this is the right direction?

tanghaowillow:
How about introducing an expiration duration config and set expiration for each datagram to be sent?  When the packet packer tries to find a datagram to send, the expired datagrams can be discarded at first. This is how chromium does for WebTransport datagrams: https://github.com/chromium/chromium/blob/main/services/network/web_transport.cc#L514 
This should also work even if we enlarge the datagram send queue size in the future to solve the performance issue.

joliveirinha:
@tanghaowillow I believe that would cause packets to hang until expiration is met. Here we just drop them in case they are not "valid size".

@marten-seemann is the PR in a state that we can merge? Ideally we would like to merge it to unblock an issue we saw, and this is the only thing that is preventing us from moving away from our own fork and just use upstream quic-go.

joliveirinha (approved):

joliveirinha:
@marten-seemann do you have a different opinion going forward? if you want we could setup a call to discuss it.
le me know what you think

marten-seemann:
Sorry for the late response here. I was very busy, and needed some time to come up with a possible way forward. I hope my ideas here help us move datagram support forward. If there's interest, we could make that a goal for the next release.
@joliveirinha Happy to jump on a call some time, please send me an email.

Here are some thoughts:
1. We know that datagram performance is pretty bad (#3766). We don't need to resolve this issue in this PR, but we need to keep the fix for that issue in mind when designing an API. Specifically resolving that issue means introducing a queue, so that multiple datagrams can be passed to quic-go before they're sent out. Otherwise we'll never be able to send datagrams at line rate.
2. We shouldn't try to send out a datagram 5 times. Instead the datagram queue should have a Peek / Pop API. The packing logic is then:
     * If the packet contains an ACK, call `Peek` and see if the DATAGRAM fits. If it does, `Pop` it. If it doesn't, fill the packet with other frames (which leaves the DATAGRAM frame in the queue).
     * If the packet doesn't contain an ACK, `Pop` the DATAGRAM as the first frame (then fill the packet with more frames). In case it doesn't fit, discard it immediately. There's no point in retrying, since we don't expect the available space to increase.
3. We should have a sanity check when `SendDatagram` is called that returns an error if the DATAGRAM exceeds the _current_ expected maximum size.
3. This means that queued DATAGRAM frames might get dropped if the available packet size is reduced. That seems ok, as this is a corner case and there's no expectation of reliable delivery. The API documentation should point this out.
4. It might be interesting to expose a notification to the application when a DATAGRAM frame is lost or acknowledged, similar to the [`FrameHandler` interface](https://github.com/quic-go/quic-go/blob/31a677cacd97b4e277f854f069fa9f2947ca7208/internal/ackhandler/frame.go#L7-L11). For example, we could change the method to `SendDatagram([]byte) (onAcked, onLost func(), err error)`. This would give the application some hint what happened to a particular datagram. It would also be really simple to implement, since we just need to pipe through the `FrameHandler` callbacks to the user. We could then manually call the `OnLost` callback when we discard a DATAGRAM frame in the corner case described in (3). Note however, that this is just a hint, since:
      * A packet might be declared lost even though it was actually received. This could happen if the ACK frame(s) for that packet are lost.
      * Even if a packet was acknowledged, the DATAGRAM might not have been delivered to the application. All that the acknowledgment tells us is that the peer's QUIC stack received the packet, but depending on the implementation, delivery to the application might not be reliable.

PR #4147, opened 2023-10-31
ackhandler: immediately acknowledge ECN-CE marked packets

### LINKED ISSUE #4146: send an immediate ACK for CE-marked packets
See section 13.2.1 of RFC 9000. Timely acknowledgement of CE-marked packets will speed up congestion control on the sending node.

This is likely very important for proper L4S support (#4002).

marten-seemann:
Fixes #4146.

Sending an ACK immediately when receiving an ECN-CE marked packet allows the sender to react to the marking faster. This is expected to be crucial for L4S to work properly, especially when used in conjunction with the ACK frequency extension.

### DIFF
--- a/internal/ackhandler/received_packet_tracker.go
+++ b/internal/ackhandler/received_packet_tracker.go
@@ -143,17 +143,13 @@ func (h *receivedPacketTracker) maybeQueueACK(pn protocol.PacketNumber, rcvTime 
 
 	// queue an ACK if there are new missing packets to report
 	if h.hasNewMissingPackets() {
-		if h.logger.Debug() {
-			h.logger.Debugf("\tQueuing ACK because there's a new missing packet to report.")
-		}
+		h.logger.Debugf("\tQueuing ACK because there's a new missing packet to report.")
 		h.ackQueued = true
 	}
 
 	// queue an ACK if the packet was ECN-CE marked
 	if ecn == protocol.ECNCE {
-		if h.logger.Debug() {
-			h.logger.Debugf("\tQueuing ACK because the packet was ECN-CE marked.")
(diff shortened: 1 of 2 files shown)


internal/ackhandler/received_packet_tracker.go
```diff
@@ -143,9 +141,19 @@ func (h *receivedPacketTracker) maybeQueueACK(pn protocol.PacketNumber, rcvTime
...
-		h.logger.Debugf("\tQueuing ACK because there's a new missing packet to report.")
+		if h.logger.Debug() {
+			h.logger.Debugf("\tQueuing ACK because there's a new missing packet to report.")
+		}
+		h.ackQueued = true
+	}
+
+	// queue an ACK if the packet was ECN-CE marked
+	if ecn == protocol.ECNCE {
+		if h.logger.Debug() {
```
marten-seemann, line 154: This is likely important for performance. When using L4S, we expect 2 CE-markings per RTT, so we want to avoid calls to `Debugf` when we're not actually outputting logs.

sukunrt: This will just be a function call with a few params, no? 

```
// Debugf logs something
func (l *defaultLogger) Debugf(format string, args ...interface{}) {
	if l.logLevel == LogLevelDebug {
		l.logMessage(format, args...)
	}
}
```

sukunrt (approved):

internal/ackhandler/received_packet_tracker.go
marten-seemann: Right. It's only important when passing arguments to the function, because that would allocate a slice.

Issue #3652, opened 2022-12-28
migrate away from Ginkgo


marten-seemann:
Ginkgo is getting in the way of properly testing our code more and more. There's big advantages to using the Go native testing (with an assertion helper like [require](https://pkg.go.dev/github.com/stretchr/testify/require)), especially:
* a native way to do benchmarking, and tooling to compare benchmarks like [benchstat](golang.org/x/perf/cmd/benchstat)
* a native way to do fuzzing (#3651)

Migrating away from Ginkgo will be a massive undertaking, and we won't be able to do it all at once. This has to be a step-by-step migration. Fortunately, it's possible to run Ginkgo and `go test` in parallel, in fact, the `ginkgo` command invokes `go test`.

New test cases aded would be Go tests (unless trivial), and when doing major refactors of a struct / package, this would the time to move away from Ginkgo.

The risk here is that we'll end up in an eternally inconsistent state. Migrating the connection tests (for example) won't be fun.
Alternatively, we could embrace Go's native tools only for benchmarks and for fuzzing, while sticking with Ginkgo for tests.

francislavoie:
Mildly off-topic, but it seems like having upgraded to ginkgo v2 in `quic-go` is causing ripple effects, causing `xcaddy` builds to fail, because of a bug in `go get`. I outlined the details here: https://github.com/golang/go/issues/56494#issuecomment-1425022660

marten-seemann:
That’s annoying. Is there anything we can do to fix your build short of migrating away from Ginkgo?

francislavoie:
I really don't know :grimacing: I wish Go would separate test/tool dependencies from runtime ones. I'm hoping @mholt will have some ideas soon.

mholt:
I don't think it's a problem with quic-go tbh. It's unfortunate, but it seems like a Go tooling problem.

We have a workaround for our users, short of having all Caddy plugins update their go.mod: https://github.com/caddyserver/caddy/issues/5301#issuecomment-1426005951

But we really hope Go will fix the bug soon.

marten-seemann:
It might be possible to use GPT-4 to rewrite our tests: 
![image](https://user-images.githubusercontent.com/1478487/227820142-62f671a8-3752-4f11-84c2-74987748aad9.png)

Problems:
* The ChatGPT typing animation is _sooo_ annoying.
* The current file size limit is too small to be actually useful, we'd need to be able to dump files with 1k LOC into it.

[...]

PR #4152, opened 2023-11-01
tests: qerr: remove ginkgo

hdonnay:
I saw #3652 and have done similar in other projects. The only way to go about it (IMO) is just like eating a buffalo: one bite at a time.

If this sort of PR is okay, this is the plan:

- Port a package file-by-file
- PR package-by-package

I'm open to doing this other ways, if that's preferred.


marten-seemann:
Thanks for the PR, but if we decide to do this, this will need to happen in a concerted effort.

PR #4168, opened 2023-11-15
http3: add settings frame handler for client and server

### LINKED ISSUE #4160: make the HTTP settings available to the client
This is needed to correctly implement WebTransport. The client has to check that the server sent the WebTransport setting in SETTINGS, for example.

tanghaowillow:
Add a peerSettingsHandler on both http3 client and server side.  This allows the client (the hijackableBody) and the server (the responseWriter) to check if the settings frame from peer is received and get peer's settings. I think this is needed in implementing http/3 datagrams as well as WebTransport where the settings frame must be checked before starting the communication.

fixes #4160

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -35,6 +35,7 @@ var _ StreamCreator = quic.Connection(nil)
 // It is used by WebTransport to create WebTransport streams after a session has been established.
 type Hijacker interface {
 	StreamCreator() StreamCreator
+	SettingsGetter() PeerSettingsGetter
 }
 
 // The body of a http.Request or http.Response.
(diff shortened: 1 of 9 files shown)


marten-seemann:
Thanks for the PR, but I don't think this works for WebTransport. We need to be able to access the SETTINGS _before_ issuing the CONNECT request. In general, please always discuss larger changes, especially ones that make API changes, on the issue before investing a lot of time into writing code.

tanghaowillow:
@marten-seemann Sorry for no discussion in advance.  I see, so it seems WebTransport needs the http3.roundTripCloser to provide the access to the settings, instead of the Hijacker(which is too late). Thanks for the feedback anyway!

PR #4181, opened 2023-11-23
http3: reset stream when a handler panics

WeidiDeng:
When debugging an [issue](https://github.com/caddyserver/caddy/issues/5951) for caddy, I found http3 server handler will not correctly propagate stream error. Code example [here](https://github.com/caddyserver/caddy/issues/5951#issuecomment-1823731441).

In this case `Content-Length` is set, http3 may use that info to infer the stream doesn't terminate properly. But in a streaming environment if there is an error upstream, client should be able to know there is an error at a stream level just like http2 regardless of the presence of `Content-Length`.

### DIFF
--- a/http3/frames.go
+++ b/http3/frames.go
@@ -19,6 +19,8 @@ type frame interface{}
 
 var errHijacked = errors.New("hijacked")
 
+var errPanicked = errors.New("panicked")
+
 func parseNextFrame(r io.Reader, unknownFrameHandler unknownFrameHandlerFunc) (frame, error) {
 	qr := quicvarint.NewReader(r)
 	for {
--- a/http3/frames.go
+++ b/http3/frames.go
@@ -19,8 +19,6 @@ type frame interface{}
 
 var errHijacked = errors.New("hijacked")
 
-var errPanicked = errors.New("panicked")
-
 func parseNextFrame(r io.Reader, unknownFrameHandler unknownFrameHandlerFunc) (frame, error) {
 	qr := quicvarint.NewReader(r)
 	for {
--- a/http3/server_test.go
+++ b/http3/server_test.go
@@ -272,7 +272,7 @@ var _ = Describe("Server", func() {
 			str.EXPECT().CancelRead(gomock.Any())
 
 			serr := s.handleRequest(conn, str, qpackDecoder, nil)
-			Expect(serr.err).ToNot(HaveOccurred())
+			Expect(serr.err).To(HaveOccurred())
 			Expect(responseBuf.Bytes()).To(HaveLen(0))
 		})
 
@@ -288,7 +288,7 @@ var _ = Describe("Server", func() {
 			str.EXPECT().CancelRead(gomock.Any())
 
 			serr := s.handleRequest(conn, str, qpackDecoder, nil)
-			Expect(serr.err).ToNot(HaveOccurred())
+			Expect(serr.err).To(HaveOccurred())
 			Expect(responseBuf.Bytes()).To(HaveLen(0))
 		})
 
--- a/http3/server_test.go
+++ b/http3/server_test.go
@@ -272,7 +272,7 @@ var _ = Describe("Server", func() {
 			str.EXPECT().CancelRead(gomock.Any())
 
 			serr := s.handleRequest(conn, str, qpackDecoder, nil)
-			Expect(serr.err).To(Equal(errPanicked))
+			Expect(serr.err).To(MatchError(errPanicked))
 			Expect(responseBuf.Bytes()).To(HaveLen(0))
 		})
 
@@ -288,7 +288,7 @@ var _ = Describe("Server", func() {
 			str.EXPECT().CancelRead(gomock.Any())
 
 			serr := s.handleRequest(conn, str, qpackDecoder, nil)
-			Expect(serr.err).To(HaveOccurred())
+			Expect(serr.err).To(MatchError(errPanicked))
 			Expect(responseBuf.Bytes()).To(HaveLen(0))
 		})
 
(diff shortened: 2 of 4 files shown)


marten-seemann (commented):
If the panic happens before we've written the HTTP header, should we just write the 500 and close the stream?

WeidiDeng:
I don't think status 500 is necessary in this case. Related [pr](https://github.com/quic-go/quic-go/pull/3950) and [issue](https://github.com/caddyserver/caddy/issues/5125).

marten-seemann:
What does HTTP/2 do in both these cases?

WeidiDeng:
HTTP/2 doesn't differentiate between these two cases. It will [reset the stream](https://github.com/golang/net/blob/v0.18.0/http2/server.go#L2353) if there is a panic.

WeidiDeng:
Per `http.Handler` [documentation](https://pkg.go.dev/net/http#Handler):

```
If ServeHTTP panics, the server (the caller of ServeHTTP) assumes that the effect of the panic was isolated to the active request. It recovers the panic, logs a stack trace to the server error log, and either closes the network connection or sends an HTTP/2 RST_STREAM, depending on the HTTP protocol. To abort a handler so the client sees an interrupted response but the server doesn't log an error, panic with the value ErrAbortHandler.
```

There is no mention of interpreting the panic and handling it the other way. And it's actually doing that.

integrationtests/self/http_test.go
```diff
@@ -140,6 +140,24 @@ var _ = Describe("HTTP tests", func() {
...
+			time.Sleep(5 * time.Second)
+			panic(http.ErrAbortHandler)
+		})
+
+		resp, err := client.Get(fmt.Sprintf("https://localhost:%d/writing_and_panicking", port))
+		Expect(err).ToNot(HaveOccurred())
+		body, err := io.ReadAll(resp.Body)
+		Expect(err).To(HaveOccurred())
+		// in reality the parts of the body is undetermined, but it's a test
+		Expect(body).To(Equal([]byte("foobar")))
```
marten-seemann, line 158: Why do you test for something that's undetermined?

http3/frames.go
```diff
@@ -19,6 +19,8 @@ type frame interface{}
 
 var errHijacked = errors.New("hijacked")
 
+var errPanicked = errors.New("panicked")
```
marten-seemann, line 22: Why not define this in server.go? This has nothing to do with frames.

http3/server_test.go
```diff
@@ -272,7 +272,7 @@ var _ = Describe("Server", func() {
 			str.EXPECT().CancelRead(gomock.Any())
 
 			serr := s.handleRequest(conn, str, qpackDecoder, nil)
-			Expect(serr.err).ToNot(HaveOccurred())
+			Expect(serr.err).To(HaveOccurred())
```
marten-seemann, line 275: Assert for `errPanicked`?

integrationtests/self/http_test.go
WeidiDeng: Should I change it so it just tests there are data that are just prefix of the data we just wrote?

marten-seemann: All that you can assert is that the error code is correct, right? The stream data might have been lost in transit.

WeidiDeng: Some data might have been lost, but those that have been received are sure be the prefix of what we had written (`nil` is prefix of the data). Without the `time.Sleep` the client won't even receive the response.

http3/server_test.go
```diff
@@ -272,7 +272,7 @@ var _ = Describe("Server", func() {
 			str.EXPECT().CancelRead(gomock.Any())
 
 			serr := s.handleRequest(conn, str, qpackDecoder, nil)
-			Expect(serr.err).ToNot(HaveOccurred())
+			Expect(serr.err).To(Equal(errPanicked))
```
marten-seemann, line 275:
```suggestion
			Expect(serr.err).To(MatchError(errPanicked))
```

http3/server_test.go
```diff
@@ -288,7 +288,7 @@ var _ = Describe("Server", func() {
 			str.EXPECT().CancelRead(gomock.Any())
 
 			serr := s.handleRequest(conn, str, qpackDecoder, nil)
-			Expect(serr.err).ToNot(HaveOccurred())
+			Expect(serr.err).To(HaveOccurred())
```
marten-seemann, line 291: Here too?

integrationtests/self/http_test.go
```diff
@@ -140,6 +140,24 @@ var _ = Describe("HTTP tests", func() {
...
+			w.(http.Flusher).Flush()
+			// simulate a large response body, some data were written
+			time.Sleep(5 * time.Second)
+			panic(http.ErrAbortHandler)
+		})
+
+		resp, err := client.Get(fmt.Sprintf("https://localhost:%d/writing_and_panicking", port))
+		Expect(err).ToNot(HaveOccurred())
+		body, err := io.ReadAll(resp.Body)
+		Expect(err).To(HaveOccurred())
```
marten-seemann, line 156: What are you asserting here? Does this test now take 5 seconds? That would be way too long.

WeidiDeng: To test the client has encountered stream errors when reading the body. Ideally this should not use sleep. I tried to write a large body, but as soon as I called panic, the response won't even arrive at the client (`client.Get` will return an error).

marten-seemann: You can block on a channel before panicking, and close that channel once `client.Get` has returned.

marten-seemann (approved):
Thank you @WeidiDeng, and sorry for the delay in reviewing this PR! This will be part of the next quic-go release.

PR #4183, opened 2023-11-26
qtls: only attempt 0-RTT resumption for 0-RTT enabled session tickets

marten-seemann:
In specific, these lines would overwrite the `EarlyData` property from `false` to `true`, when doing session resumption using `DialEarly` (code from master):
https://github.com/quic-go/quic-go/blob/2d7ea376729d317e2fbce7ff5ae44ad180ffefdb/internal/qtls/client_session_cache.go#L49-L54

In addition to fixing that, we now also avoid saving the transport parameters into the session ticket for non-0-RTT tickets (code from this PR): https://github.com/quic-go/quic-go/blob/3e984c14b18ee61a830cef348832f56bda45eab0/internal/handshake/crypto_setup.go#L320-L324

### DIFF
--- a/internal/handshake/crypto_setup.go
+++ b/internal/handshake/crypto_setup.go
@@ -25,7 +25,7 @@ type quicVersionContextKey struct{}
 
 var QUICVersionContextKey = &quicVersionContextKey{}
 
-const clientSessionStateRevision = 3
+const clientSessionStateRevision = 4
 
 type cryptoSetup struct {
 	tlsConf *tls.Config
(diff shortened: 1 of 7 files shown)


internal/handshake/crypto_setup.go
```diff
@@ -336,25 +341,28 @@ func (h *cryptoSetup) handleDataFromSessionState(data []byte) (allowEarlyData bo
...
+	}
+	rtt := time.Duration(rttEncoded) * time.Microsecond
+	if !earlyData {
+		return rtt, nil, nil
 	}
-	h.rttStats.SetInitialRTT(time.Duration(rtt) * time.Microsecond)
 	var tp wire.TransportParameters
 	if err := tp.UnmarshalFromSessionTicket(r); err != nil {
-		return nil, err
+		return rtt, nil, err
```
sukunrt, line 363: In the caller `rtt` will never be used if err is non nil.

internal/handshake/crypto_setup.go
```diff
@@ -313,19 +313,24 @@ func (h *cryptoSetup) handleTransportParameters(data []byte) error {
...
+		return h.peerParams.MarshalForSessionTicket(b)
+	}
+	return b
 }
 
-func (h *cryptoSetup) handleDataFromSessionState(data []byte) (allowEarlyData bool) {
-	tp, err := h.handleDataFromSessionStateImpl(data)
+func (h *cryptoSetup) handleDataFromSessionState(data []byte, earlyData bool) (allowEarlyData bool) {
+	rtt, tp, err := decodeDataFromSessionState(data, earlyData)
 	if err != nil {
```
sukunrt, line 329: rtt might still be valid if err is non nil. This will happen whenever unmarshalling of transport parameters fails.

sukunrt (approved):
one minor comment.

internal/handshake/crypto_setup.go
marten-seemann: This should never happen, unless something is going seriously wrong. We can just return 0 instead of `rtt` here.

internal/handshake/crypto_setup.go
marten-seemann: Unmarshaling the transport parameters shouldn't fail under normal circumstances. I'll change the function to always return 0, and then it's ok to ignore the RTT.

PR #4199, opened 2023-12-13
limit the number of queued PATH_RESPONSE frames to 256

marten-seemann:


marten-seemann:
Approved by @sukunrt oob.

Issue #4189, opened 2023-12-06
Enable and set destination for qlog output via QLOGDIR environment variable


birneee:
I ask myself if quic-go itself should act automatically on the [QLOGDIR](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-main-schema-07.html#name-set-file-output-destination.
) environment variable, and enable qlog output, or if this has to be manually enabled/implemented by the application, which is using `quic-go`.

I think this behavior should be similar to `QUIC_GO_LOG_LEVEL`, so it would unify debugging for all applications using quic-go.

If an application developer wants to disable it, an option `IgnoreQlogDir` could be added to `quic.Config`. 
Something similar could then also be added for `QUIC_GO_LOG_LEVEL`.

marten-seemann:
Would that be in addition to the tracer configuration? I'm a bit hesitant to make this make this a default, but on the other hand, it's mentioned in the draft.

I don't think we need any `quic.Config` option though. If you don't want qlogs, don't set the env.

birneee:
I think it is fine to have two ways to enable qlogs.
The QLOGDIR variable is for enabling the default behavior.
But if the application needs more control over the qlog, the tracer configuration should be the way to go.

marten-seemann:
Writing up a little bit more clearly what we already discussed on https://github.com/quic-go/quic-go/pull/4201.

We could introduce a `qlog.DefaultTracer` (suggestions for a better name highly appreciated) that is a `func(context.Context, logging.Perspective, ConnectionID) *logging.ConnectionTracer` (i.e. the exact type needed for `quic.Config.Tracer`).

It would do exactly what you suggested: read the directory from the QLOGDIR env, and if set create the directory and dump qlogs into that directory. Enabling qlogs would then be super simple, just use:
```go
quic.Config{
    Tracer: qlog.DefaultTracer,
}
```

I believe this would cover a lot of simple use cases, while still keeping the flexibility to configure your own tracing callback if you need something more sophisticated.

@birneee What do you think? Want to raise a PR?

PR #4201, opened 2023-12-13
Extend qlog support

birneee:
- support QLOGDIR environment variable
- support chosen_alpn qlog event
- support stream_data_moved event

Covers #4189

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,8 +4,9 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
-	"github.com/quic-go/quic-go/qlog"
 	"net"
+
+	"github.com/quic-go/quic-go/qlog"
 
 	"github.com/quic-go/quic-go/internal/protocol"
 	"github.com/quic-go/quic-go/internal/utils"
--- a/client.go
+++ b/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"github.com/quic-go/quic-go/qlog"
 	"net"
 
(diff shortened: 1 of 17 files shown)


marten-seemann (commented):
Can you split this up into 3 separate PRs?

1. The ALPN event totally makes sense to me.
1. What's the motivation for the stream_data_moved event? How would you use it?
1. The QLOGDIR should be optional, i.e. not be implemented in the `quic` package. Having a separate constructor in the `qlog` package makes sense to me.

birneee:
> Can you split this up into 3 separate PRs?

I can :) 

> What's the motivation for the stream_data_moved event? How would you use it?

I would like to test/benchmark the "end-to-end"-performance that I can expect on the application layer in some scenarios. From the current qlog it is not obvious how far a stream has progressed, because of loss and reordering. For now I have a Python script that tracks the stream gaps from the qlog stream 😅

> The QLOGDIR should be optional

Not sure if I understand you correctly, you mean just a separate NewQlogDirConnectionTracer?

marten-seemann:
> > The QLOGDIR should be optional
> 
> Not sure if I understand you correctly, you mean just a separate NewQlogDirConnectionTracer?

Yes, pretty much. Then we can tell users "If you want to record qlogs every time the QLOGDIR is set, set the `Config.Tracer` to `qlog.NewQLOGDirTracer`.

marten-seemann:
Closing, since we've had support for the ALPN event and the QLOGDIR for a while now. Still not sure what to do about the data moved event.

PR #4203, opened 2023-12-13
http3: return the context cancellation error from RoundTrip

### LINKED ISSUE #4196: context deadline can cause surprising errors
When a context deadline is added to the context of an `*http.Request` and the deadline elapses, the actual errors returned from this package can be a bit surprising. In particular, most context-deadline-checking code really would like to test errors using something like `errors.Is(err, context.DeadlineExceed)` to ascertain if the error is due to the context deadline.

However, that formulation returns false for both `quic.IdleTimeoutError` and `quic.HandshakeTimeoutError`. Ideally, both of these error types would return true for this formulation via their implementation of an `Is(error)` method. That is how the core runtime's `net.timeoutError` works: https://github.com/golang/go/blob/master/src/net/net.go#L600-L60. Both these types already have an `Is(error)` method, but they only handle formulations like `errors.Is(err, net.ErrClosed)` ([link](https://github.com/quic-go/quic-go/blob/master/internal/qerr/errors.go#L91)).

This is very similar to https://github.com/golang/go/issues/64449

marten-seemann:
@jhump Can you try out this PR? Does it fix your problem?

### DIFF
--- a/integrationtests/self/http_test.go
+++ b/integrationtests/self/http_test.go
@@ -299,6 +299,21 @@ var _ = Describe("HTTP tests", func() {
 		Expect(string(body)).To(Equal("Hello, World!\n"))
 	})
 
+	It("handles context cancellations", func() {
+		mux.HandleFunc("/cancel", func(w http.ResponseWriter, r *http.Request) {
+			<-r.Context().Done()
+		})
+
+		ctx, cancel := context.WithCancel(context.Background())
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("https://localhost:%d/cancel", port), nil)
+		Expect(err).ToNot(HaveOccurred())
+		time.AfterFunc(50*time.Millisecond, cancel)
+
+		_, err = client.Do(req)
+		Expect(err).To(HaveOccurred())
+		Expect(err).To(MatchError(context.Canceled))
+	})
+
 	It("cancels requests", func() {
 		handlerCalled := make(chan struct{})
 		mux.HandleFunc("/cancel", func(w http.ResponseWriter, r *http.Request) {
--- a/http3/client.go
+++ b/http3/client.go
@@ -254,6 +254,15 @@ func (c *client) maxHeaderBytes() uint64 {
 
 // RoundTripOpt executes a request and returns a response
 func (c *client) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Response, error) {
+	rsp, err := c.roundTripOpt(req, opt)
+	if err != nil && req.Context().Err() != nil {
+		// if the context was canceled, return the context cancellation error
+		err = req.Context().Err()
+	}
+	return rsp, err
+}
+
+func (c *client) roundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Response, error) {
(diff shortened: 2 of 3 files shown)


jhump:
@marten-seemann, on the face of it, this looks right, but I'm seeing something weird in my client program that I need to track down. One thing that is absent here is doing something similar for errors returned from `Body.Read`, since consuming the response data can also encounter context timeout/cancellation.

integrationtests/self/http_test.go
```diff
@@ -299,6 +299,21 @@ var _ = Describe("HTTP tests", func() {
...
+		ctx, cancel := context.WithCancel(context.Background())
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("https://localhost:%d/cancel", port), nil)
+		Expect(err).ToNot(HaveOccurred())
+		time.AfterFunc(50*time.Millisecond, cancel)
+
+		_, err = client.Do(req)
+		Expect(err).To(HaveOccurred())
+		Expect(err).To(MatchError(context.Canceled))
+	})
+
```
jhump, line 316: Another useful test would be to verify receipt of deadline exceeded, which looks very similar to this case.
```go
	It("handles context cancellations", func() {
		mux.HandleFunc("/cancel", func(w http.ResponseWriter, r *http.Request) {
			<-r.Context().Done()
		})

		ctx, cancel := context.WithTimeout(context.Background(), 50*time.Millisecond)
		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("https://localhost:%d/cancel", port), nil)

		_, err = client.Do(req)
		Expect(err).To(HaveOccurred())
		Expect(err).To(MatchError(context.DeadlineExceeded))
	})
```

jhump: A test case for the body-read error wouldn't be much different:
```go
	It("handles context cancellations when reading body", func() {
		mux.HandleFunc("/cancel", func(w http.ResponseWriter, r *http.Request) {
			w.WriteHeader(http.StatusOK)
			<-r.Context().Done()
		})

		ctx, cancel := context.WithCancel(context.Background())
		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("https://localhost:%d/cancel", port), nil)
		time.AfterFunc(50*time.Millisecond, cancel)

		resp, err = client.Do(req)
		_, err = resp.Body.Read(make([]byte, 1024))
		Expect(err).To(HaveOccurred())
		Expect(err).To(MatchError(context.Canceled))
	})
```

marten-seemann: > A test case for the body-read error wouldn't be much different:

That's a much bigger change. The Go documentation for the request context says:
> For an outgoing client request, the context controls the entire lifetime of a request and its response: obtaining a connection, sending the request, and reading the response headers and body.

So you're right, the body needs to return the context error as well. Currently, we don't even pass the context to the body, and it's not clear to me how to correctly respect the context without spawning a new Go routine, since the `Stream.Read` call is blocking call.

marten-seemann: I suggest doing this in a follow-up PR. I'll open a new issue.

marten-seemann:
Merging this PR, since it at least partially fixes #4196. The remainder of the fix is tracked in #4205.

PR #4208, opened 2023-12-14
http3: add remote address to request context

### LINKED ISSUE #4198: proposal: attach remote address to context in HTTP3 request
Hi there,

We are trying to move our control plane from gRPC to connect-go/HTTP3 (see scionproto/scion#4434).
We rely on a feature of gRPC that lets us access the peer address when handling requests.

This is currently not possible with HTTP3, because the remote address is only passed as a regular string on the HTTP request of the stdlib type.

By attaching the remote address to the HTTP request context, we would solve our use case. But I also believe that it is more generally useful for implementing server code.

Enabling this can be done with a very [minimal patch](https://github.com/scionproto/scion/blob/connectrpc/patches/com_github_quic_go_quic_go/http3_remote_addr.patch). If you would be interested, I will create a PR to upstrem it:

```diff
diff --git a/http3/server.go b/http3/server.go
index ac2e32a6..a4c4e327 100644
--- a/http3/server.go
+++ b/http3/server.go
@@ -115,6 +115,8 @@ func (k *contextKey) String() string { return "quic-go/http3 context value " + k
 // type *http3.Server.
 var ServerContextKey = &contextKey{"http3-server"}
 
+var RemoteAddrContextKey = &contextKey{"remote-addr"}
+
 type requestError struct {
 	err       error
 	streamErr ErrCode
@@ -597,6 +599,7 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 	ctx := str.Context()
 	ctx = context.WithValue(ctx, ServerContextKey, s)
 	ctx = context.WithValue(ctx, http.LocalAddrContextKey, conn.LocalAddr())
+	ctx = context.WithValue(ctx, RemoteAddrContextKey, conn.RemoteAddr())
 	req = req.WithContext(ctx)
 	r := newResponseWriter(str, conn, s.logger)
 	if req.Method == http.MethodHead {
```

oncilla:
Add the remote address of the underlying packet connection to the HTTP request context. This is useful for applications that need access to the actual remote address (wrapped in a net.Addr) rather than just its string representation.

Fixes #4198

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -115,6 +115,16 @@ func (k *contextKey) String() string { return "quic-go/http3 context value " + k
 // type *http3.Server.
 var ServerContextKey = &contextKey{"http3-server"}
 
+// RemoteAddrContextKey is a context key. It can be used in
+// HTTP handlers with Context.Value to access the remote
+// address of the connection. The associated value will be of
+// type net.Addr.
+//
+// Use this value instead of [http.Request.RemoteAddr] if you
+// require access to the remote address of the connection rather
+// than its string representation.
+var RemoteAddrContextKey = &contextKey{"remote-addr"}
+
 type requestError struct {
 	err       error
 	streamErr ErrCode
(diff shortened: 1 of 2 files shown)


marten-seemann (commented):
Code lgtm. Could you add a test please in integrationtests/self/http_test.go?

oncilla:
Done.

I was not sure how deep we want to go to assure the right value is set.
For now, it asserts that a net.Addr is present in the context.

Due to the dynamic nature of the requests, it is a bit harder to know
exactly how the remote address will look like. IPv4 vs IPv6, random high
port. I think it is fine to only assert that the value is present.

marten-seemann (approved):

marten-seemann:
> I was not sure how deep we want to go to assure the right value is set. For now, it asserts that a net.Addr is present in the context.
> 
> Due to the dynamic nature of the requests, it is a bit harder to know exactly how the remote address will look like. IPv4 vs IPv6, random high port. I think it is fine to only assert that the value is present.

Makes sense, thank you!

PR #3521, opened 2022-08-24
implement greasing of the QUIC bit (RFC 9287)

marten-seemann:
Fixes #3133.

We don't (yet) save the value of this transport parameter. This means that we'll never grease the QUIC bit on the first Initial packet and on 0-RTT packets. Implementing this is left as an exercise for a later PR.

Issue #3133, opened 2021-03-31
implement the QUIC Grease Bit (RFC 9287)


marten-seemann:
https://www.rfc-editor.org/rfc/rfc9287.html

As the QUIC bit can be used for demultiplexing QUIC with other protocols, we should probably add a `quic.Config` option to turn this off: `DisableQUICBitGreasing`.

marten-seemann:
The [grease-quic-bit](https://github.com/lucas-clemente/quic-go/tree/grease-quic-bit) contains an initial implementation, but we'll have to do another careful pass before we can advance this to a PR, paying special attention to greasing on 0-RTT connections.

marten-seemann:
I'm not convinced anymore we should implement this extension. It adds a non-trivial amount of complexity, for very little benefit: Greasing the QUIC bit is only useful for implementations that are _very_ common on the public internet. Only those implementations will be able to persuade middleboxes to not rely on the QUIC bit.

quic-go is not in this position. Unless and until that changes, we should prioritize (relative) simplicity over the benefit for the QUIC ecosystem as a whole.

PR #4213, opened 2023-12-19
Greasing QUIC bit (RFC 9287)

naveenrajm7:
Hello!

I wanted the greasing QUIC bit feature in quic-go. Hence I copied the implementation from #3521  on the current master which has highly refactored header handling (writing & parsing).
While the greasing and handling seems to work for long header packets. It is not working for short headers when the corresponding peer sends a short header with QUIC bit 0. Due to this the integration test for grease_quic_bit is also failing.

While I have been working on debugging further. I am sharing the draft PR to get some help
* Get feedback on what I might be missing, Since this was already done and folks here are well aware of the refactoring changes. 
* To keep up the #3521 effort from being lost due to refactoring

Any help is appreciated. 

Thanks.


marten-seemann:
Hi @naveenrajm7! I changed my mind on implementing this extension since working on #3521, see https://github.com/quic-go/quic-go/issues/3133#issuecomment-1862118677. That's the reason there hasn't been any progress on that PR. I should've documented my reasoning on the issue earlier, sorry for that!

naveenrajm7:
Interesting reason. I guess I'm on my own to fix this one.
Thanks for your clarification  :)

Issue #4212, opened 2023-12-18
http3: bad Errorf 


nanokatze:
https://github.com/quic-go/quic-go/blob/e0bf13be01aaa715329d32da90d526671b5245ce/http3/server.go#L424
will substitute `err` into the `%+v` instead of the `%s`, which is clearly not what was intended, but I'm not sure what the original intent was. Substituting `l` into `%+v`, which will print the address of that object, doesn't seem super helpful.

marten-seemann:
Would you mind submitting another PR? 🙏

PR #4214, opened 2023-12-19
http3: improve debug message when determining the listener port fails

nanokatze:
Fixes: #4212


marten-seemann (commented):
Should we instead print the `Addr().String()`?

nanokatze:
That seems like the more useful thing to report, indeed.

marten-seemann (approved):
Thank you @nanokatze!

PR #4219, opened 2023-12-28
http3: fix channel size in ListenAndServe

narqo:
This PR addresses two small issues in the `ListenAndServe` function:

1. I believe there is a typo in the functin's docs.
2. If an internal listener from one of the goroutines returns on an error, its peer's gorotine will leak on sending to a channel. This PR addresses this by making the channels buffered.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -723,7 +723,7 @@ func ListenAndServeQUIC(addr, certFile, keyFile string, handler http.Handler) er
 	return server.ListenAndServeTLS(certFile, keyFile)
 }
 
-// ListenAndServe listens on the given network address for both TCP and QUIC
+// ListenAndServe listens on the given network address for both TLS/TCP and QUIC
 // connections in parallel. It returns if one of the two returns an error.
 // http.DefaultServeMux is used when handler is nil.
 // The correct Alt-Svc headers for QUIC are set.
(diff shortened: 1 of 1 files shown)


http3/server.go
```diff
@@ -723,7 +723,7 @@ func ListenAndServeQUIC(addr, certFile, keyFile string, handler http.Handler) er
 	return server.ListenAndServeTLS(certFile, keyFile)
 }
 
-// ListenAndServe listens on the given network address for both, TLS and QUIC
```
marten-seemann, line 726: TLS is not a protocol you can listen on.

marten-seemann (commented):
The code fix is correct, but please revert the incorrect documentation change.

http3/server.go
```diff
@@ -723,7 +723,7 @@ func ListenAndServeQUIC(addr, certFile, keyFile string, handler http.Handler) er
...
+// ListenAndServe listens on the given network address for both TCP and QUIC
```
marten-seemann, line 726:
```suggestion
// ListenAndServe listens on the given network address for both TLS/TCP and QUIC
```

marten-seemann: Maybe this is better?

marten-seemann (approved):

PR #4220, opened 2023-12-29
qtls: remove unneeded type alias for the tls.QUICEncryptionLevel

marten-seemann:
Some more cleanup after #4195.


Jorropo (approved):
:tada:

PR #4221, opened 2023-12-29
discard DATAGRAM frames that don't fit into packets without an ACK

### LINKED ISSUE #4144: Prevent large datagrams from blocking the send queue

marten-seemann:
Fixes #4144. Closes #4145.

This PR implements part of my suggestion in https://github.com/quic-go/quic-go/pull/4145#issuecomment-1869316371.

If the packet doesn't contain an ACK (i.e. the payload is empty), there's no point retrying to pack it later. It's extremely unlikely that the available packet size will suddenly increase.

Since we're relaxing the API guarantees (`SendDatagram` doesn't *guarantee* that the DATAGRAM frame is actually going to be sent out), this PR should allow us to move away from the blocking API (`datagramQueue.AddAndWait`) that's the main cause for the bad datagram performance (#3766).

@joliveirinha @chungthuang @birneee Could you take a look at this PR please? Does this resolve your problem?

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -187,8 +187,12 @@ type Connection interface {
 	// Warning: This API should not be considered stable and might change soon.
 	ConnectionState() ConnectionState
 
-	// SendDatagram sends a message as a datagram, as specified in RFC 9221.
-	SendDatagram([]byte) error
+	// SendDatagram sends a message using a QUIC datagram, as specified in RFC 9221.
+	// There is no delivery guarantee for DATAGRAM frames, they are not retransmitted if lost.
+	// The payload of the datagram needs to fit into a single QUIC packet.
+	// In addition, a datagram may be dropped before being sent out if the available packet size suddenly decreases.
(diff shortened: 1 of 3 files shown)


chungthuang (approved):

chungthuang:
Thank you @marten-seemann. This fixes the blocking problem we are seeing.

chungthuang (approved):

marten-seemann:
Thanks for reviewing @chungthuang!

joliveirinha:
Thank you @marten-seemann for the fix.

PR #4223, opened 2023-12-30
use a ring buffer for the datagram queue

marten-seemann:


chungthuang (approved):

sukunrt:
Sorry for the drive by comment. Just to point it out in case it was actually missed. Please ignore if not relevant. 

Is a Peek followed by a Pop threadsafe considering that we only synchronise the calls individually and not the whole Peek + Pop operation? 
I can't see any issues currently because it seems to be called only from the connection run loop go routine.

marten-seemann:
> Sorry for the drive by comment.

Thanks for reviewing!



> Is a Peek followed by a Pop threadsafe considering that we only synchronise the calls individually and not the whole Peek + Pop operation?

The entire `RingBuffer` is not thread-safe (it doesn't try to: it doesn't even contain a mutex). It's the caller's responsibility to synchronize access.

I believe the `datagramQueue` does that: `Peek` will return a non-nil frame iff there's actually a frame queued. While it's possible that new frames are added between the `Peek` and `Pop` call, the only way that frame could disappear via `Pop` call, there's no other way to dequeue / delete a frame from the queue. Do you see any problem with that logic?

sukunrt:
This makes sense. 
Only `Add` happens from a different goroutine. `Peek` and `Pop` happen from the same goroutine so we will always `Pop` the element we received from `Peek` since `Add` puts it at the back of the queue.

sukunrt (approved):

PR #4225, opened 2024-01-01
ackhandler: refactor ACK queueing logic

marten-seemann:
Once an ACK has been queued, there's no need to check futher conditions that would lead to queueing of an ACK.


connection_test.go
```diff
@@ -1269,7 +1269,6 @@ var _ = Describe("Connection", func() {
 			sph.EXPECT().ECNMode(true).AnyTimes()
 			runConn()
 			packer.EXPECT().AppendPacket(gomock.Any(), gomock.Any(), conn.version).Return(shortHeaderPacket{}, errNothingToPack).AnyTimes()
-			conn.receivedPacketHandler.ReceivedPacket(0x035e, protocol.ECNNon, protocol.Encryption1RTT, time.Now(), true)
```
marten-seemann, line 1272: I don't understand what the purpose of this call was. And it now causes a race condition when accessing the `ackAlarm` field.

Issue #3603, opened 2022-10-20
http3: Is there an equivalent of http.Server.ConnContext


omz13:
I am trying to add http/3 support into an existing application.

For my existing application I hook into `http.Server.ConnContext` so I can insert my own `Context`. I was hoping that I could easily do something similar with `http3.Server`, but I can't see any equivalent mechanisms. Am I not seeing something obvious, or is this not implemented?

marten-seemann:
That makes sense. Thank you for the explanation!

> I see in func (s *Server) handleRequest that you are already wrapping the req with some package-specific context; what I am missing is the ability to add user-specific context somewhere into that context.

I'm wondering how that would work. In `handleRequest` we already use the context returned from `quic.Stream.Context`. In Go, it's not possible to "combine" two contexts, so we'd have to decide which context we use as a base context. Or am I missing something here.

omz13:
It would make sense to go via `quic.Stream.Context`: a` http3.Server.StreamContext(ctx context.Context, c quic.Connection) context.Context` to be the equivalent of `http.Server.ConnContext(ctx context.Context, c net.Conn) context.Context `?

carljkempe:
I too am looking to take advantage of something similar. In my case, I'm trying to hook-in and include `ClientHello` message info. 👋

marten-seemann:
> It would make sense to go via quic.Stream.Context: a http3.Server.StreamContext(ctx context.Context, c quic.Connection) context.Context to be the equivalent of http.Server.ConnContext(ctx context.Context, c net.Conn) context.Context ?

The stream context is not set by the user, it's set by the stream: https://github.com/lucas-clemente/quic-go/blob/af30cef57c5c20ebcb599200a96d89f7c33c3d50/send_stream.go#L66-L80

omz13:
> The stream context is not set by the user, it's set by the stream

Yes, it is set by the stream, and what I mean is that is probably the place where a subsequent call to a user-supplied function (if supplied) would be where the user could add to that context.

marten-seemann:
That’s the QUIC layer though, not http3. It would be very hard to set a context that deep in the stack.

mattrobenolt:
I am also going to put a +1 in here. Our use case is using a per-connection context to bind a logger, connection-id, and store other metadata about the actual connection.

Without this, there's no way to understand or track multiple requests happening over the same connection, etc.

marten-seemann:
[...]

PR #4230, opened 2024-01-02
http3: add ConnContext to the server

rthellend:
ConnContext can be used to modify the context used by a new http Request. This is equivalent to `ConnContext` in `http.Server`, but with a `quic.Connection` instead of a `net.Conn`.

```
// ConnContext optionally specifies a function that modifies
// the context used for a new connection c. The provided ctx
// has a ServerContextKey value.
ConnContext func(ctx context.Context, c quic.Connection) context.Context
```

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -618,7 +618,7 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 	if s.ConnContext != nil {
 		ctx = s.ConnContext(ctx, conn)
 		if ctx == nil {
-			panic("ConnContext returned nil")
+			panic("http3: ConnContext returned nil")
 		}
 	}
 	req = req.WithContext(ctx)
--- a/http3/server.go
+++ b/http3/server.go
@@ -211,6 +211,11 @@ type Server struct {
 	// In that case, the stream type will not be set.
 	UniStreamHijacker func(StreamType, quic.Connection, quic.ReceiveStream, error) (hijacked bool)
 
+	// ConnContext optionally specifies a function that modifies
+	// the context used for a new connection c. The provided ctx
+	// has a ServerContextKey value.
+	ConnContext func(ctx context.Context, c quic.Connection) context.Context
+
 	mutex     sync.RWMutex
 	listeners map[*QUICEarlyListener]listenerInfo
 
@@ -610,6 +615,9 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 	ctx = context.WithValue(ctx, ServerContextKey, s)
 	ctx = context.WithValue(ctx, http.LocalAddrContextKey, conn.LocalAddr())
 	ctx = context.WithValue(ctx, RemoteAddrContextKey, conn.RemoteAddr())
+	if s.ConnContext != nil {
+		ctx = s.ConnContext(ctx, conn)
+	}
 	req = req.WithContext(ctx)
 	r := newResponseWriter(str, conn, s.logger)
 	if req.Method == http.MethodHead {
--- a/http3/server.go
+++ b/http3/server.go
@@ -617,6 +617,9 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 	ctx = context.WithValue(ctx, RemoteAddrContextKey, conn.RemoteAddr())
 	if s.ConnContext != nil {
 		ctx = s.ConnContext(ctx, conn)
+		if ctx == nil {
+			panic("ConnContext returned nil")
+		}
 	}
 	req = req.WithContext(ctx)
 	r := newResponseWriter(str, conn, s.logger)
--- a/integrationtests/self/http_test.go
+++ b/integrationtests/self/http_test.go
@@ -528,4 +528,30 @@ var _ = Describe("HTTP tests", func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(resp.StatusCode).To(Equal(200))
 	})
+
+	It("sets conn context", func() {
+		type ctxKey int
+		server.ConnContext = func(ctx context.Context, c quic.Connection) context.Context {
+			ctx = context.WithValue(ctx, ctxKey(0), "Hello")
+			ctx = context.WithValue(ctx, ctxKey(1), c)
+			return ctx
+		}
+		mux.HandleFunc("/conn-context", func(w http.ResponseWriter, r *http.Request) {
+			defer GinkgoRecover()
+			v, ok := r.Context().Value(ctxKey(0)).(string)
+			Expect(ok).To(BeTrue())
+			Expect(v).To(Equal("Hello"))
+
+			_, ok = r.Context().Value(ctxKey(1)).(quic.Connection)
+			Expect(ok).ToNot(Equal(nil))
+
+			serv, ok := r.Context().Value(http3.ServerContextKey).(*http3.Server)
+			Expect(ok).To(BeTrue())
+			Expect(serv).To(Equal(server))
+		})
+
+		resp, err := client.Get(fmt.Sprintf("https://localhost:%d/conn-context", port))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(resp.StatusCode).To(Equal(200))
+	})
 })
--- a/integrationtests/self/http_test.go
+++ b/integrationtests/self/http_test.go
@@ -546,8 +546,9 @@ var _ = Describe("HTTP tests", func() {
 			Expect(ok).To(BeTrue())
 			Expect(v).To(Equal("Hello"))
 
-			_, ok = r.Context().Value(ctxKey(1)).(quic.Connection)
-			Expect(ok).ToNot(Equal(nil))
+			c, ok := r.Context().Value(ctxKey(1)).(quic.Connection)
+			Expect(ok).To(BeTrue())
+			Expect(c).ToNot(BeNil())
 
 			serv, ok := r.Context().Value(http3.ServerContextKey).(*http3.Server)
 			Expect(ok).To(BeTrue())
--- a/integrationtests/self/http_test.go
+++ b/integrationtests/self/http_test.go
@@ -532,6 +532,10 @@ var _ = Describe("HTTP tests", func() {
 	It("sets conn context", func() {
 		type ctxKey int
 		server.ConnContext = func(ctx context.Context, c quic.Connection) context.Context {
+			serv, ok := ctx.Value(http3.ServerContextKey).(*http3.Server)
+			Expect(ok).To(BeTrue())
+			Expect(serv).To(Equal(server))
+
 			ctx = context.WithValue(ctx, ctxKey(0), "Hello")
 			ctx = context.WithValue(ctx, ctxKey(1), c)
 			return ctx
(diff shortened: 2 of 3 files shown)


rthellend:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/quic-go/quic-go/pull/4230/checks?check_run_id=20107714809) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

This check is ✔️ now

marten-seemann:
Have you seen the discussion in #3603? Is this PR attempting to resolve that issue? If so, can you explain how your PR resolves the problems raised there?

rthellend:
> Have you seen the discussion in #3603? Is this PR attempting to resolve that issue? If so, can you explain how your PR resolves the problems raised there?

I had not actually seen that discussion. Thanks for pointing it out. I would generally agree with the motivation behind it.

The way I use `ConnContext` is pretty straight forward. I wrap the underlying connection (`tls.Conn` or `quic.Connection` here) in my own struct so that I can keep track of metrics and some annotations for each connection. The http handlers need access to some of that data to do what they need to do. (Happy to provide more details if you're interested)

Here's a concrete example the wrapper for `net.Conn`: https://github.com/c2FmZQ/tlsproxy/blob/main/proxy/internal/netw/netw.go#L82

I have something similar for `quic.Connection`.

With `http.Server`, I have:
```
    ConnContext: func(ctx context.Context, c net.Conn) context.Context {
        return context.WithValue(ctx, connCtxKey, c)
     },
```

With `http3.Server`, I would have:
```
    ConnContext: func(ctx context.Context, c quic.Connection) context.Context {
        return context.WithValue(ctx, connCtxKey, c)
    },
```

Then, when the handler needs to access the wrapper's data, it does something like:
```
if wrappedConn, ok := req.Context().Value(connCtxKey).(*netw.Conn); ok {
  // use wrappedConn
}
```
The alternative I've been using so far is to create a new `http3.Server` for each connection and then call `ServeQUICConn(conn)`:
```
    conn := ...

    serv := &http3.Server{
        Handler: http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
            req = req.WithContext(context.WithValue(req.Context(), connCtxKey, conn))
            realHandler.ServeHTTP(w, req)
         }),
    }
    serv.ServeQUICConn(conn);
```
It works, but it is not very elegant.

http3/server.go
```diff
@@ -610,6 +615,9 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
 	ctx = context.WithValue(ctx, ServerContextKey, s)
 	ctx = context.WithValue(ctx, http.LocalAddrContextKey, conn.LocalAddr())
 	ctx = context.WithValue(ctx, RemoteAddrContextKey, conn.RemoteAddr())
+	if s.ConnContext != nil {
+		ctx = s.ConnContext(ctx, conn)
```
marten-seemann, line 619: Should we have a `nil` check here, equivalent to what the stdlib does? https://github.com/golang/go/blob/6db1102605f227093ea95538f0fe9e46022ad7ea/src/net/http/server.go#L3275-L3281

marten-seemann (commented):
Could you also add an integration test for this in `integrationtests/self/http_test.go`? The test should also assert that this statement remains true:
```The provided ctx has a ServerContextKey value.```

File: http3/server.go (hunk 1 above)
Comment (rthellend) [reply http3/server.go:619]: done

**Comment by rthellend** [comment]
> Could you also add an integration test for this in `integrationtests/self/http_test.go`? The test should also assert that this statement remains true: `The provided ctx has a ServerContextKey value.`

done

File: integrationtests/self/http_test.go [hunk 2]
```diff
@@ -528,4 +528,30 @@ var _ = Describe("HTTP tests", func() {
...
+		mux.HandleFunc("/conn-context", func(w http.ResponseWriter, r *http.Request) {
+			defer GinkgoRecover()
+			v, ok := r.Context().Value(ctxKey(0)).(string)
+			Expect(ok).To(BeTrue())
+			Expect(v).To(Equal("Hello"))
+
+			_, ok = r.Context().Value(ctxKey(1)).(quic.Connection)
+			Expect(ok).ToNot(Equal(nil))
+
+			serv, ok := r.Context().Value(http3.ServerContextKey).(*http3.Server)
```
marten-seemann, line 548: This check should be in the `ConnContext` function, right?

integrationtests/self/http_test.go
```diff
@@ -528,4 +528,30 @@ var _ = Describe("HTTP tests", func() {
...
+			return ctx
+		}
+		mux.HandleFunc("/conn-context", func(w http.ResponseWriter, r *http.Request) {
+			defer GinkgoRecover()
+			v, ok := r.Context().Value(ctxKey(0)).(string)
+			Expect(ok).To(BeTrue())
+			Expect(v).To(Equal("Hello"))
+
+			_, ok = r.Context().Value(ctxKey(1)).(quic.Connection)
+			Expect(ok).ToNot(Equal(nil))
```
marten-seemann, line 546:
```suggestion
			c, ok := r.Context().Value(ctxKey(1)).(quic.Connection)
			Expect(ok).To(BeTrue())
			Expect(c).ToNot(BeNil())
```

http3/server.go
```diff
@@ -610,6 +615,12 @@ func (s *Server) handleRequest(conn quic.Connection, str quic.Stream, decoder *q
...
+		if ctx == nil {
+			panic("ConnContext returned nil")
```
marten-seemann, line 621:
```suggestion
			panic("http3: ConnContext returned nil")
```

integrationtests/self/http_test.go
rthellend: Sure. How about we put it in both the `ConnContext` function and the handler?

marten-seemann: Sure, that works as well.

marten-seemann (approved):
Thank you @rthellend! This looks good to me now.

I'll keep the PR open for a few more days to give @omz13, @carljkempe and @mattrobenolt a chance to chime in.

This will definitely make it into the next release, I just added it to the milestone.

omz13:
@marten-seemann What happened to waiting a few days?

A quick look and it looks ok. 👍 

Now its in master I guess I have something to play with 🎉

mattrobenolt:
I might be misunderstanding the implementation here, but I just started poking around, and one thing different, is this ConnContext gets called per-http3 request, not per-QUIC connection, like I'd anticipate.

In TCP based http handler, we get one ConnContext called the first time the TCP stream is connected, and that is called once for the duration of the TCP connection.

From my testing now with quic-go, it seems every single HTTP3 request triggers the ConnContext, which is not what I'd anticipate.

This also might be more my misunderstandings of how QUIC works or some implementation detail, just wanted to note the behavior differs than anticipated.

marten-seemann:
@nanokatze Any thoughts on this?

I would've expected `ConnContext` to work as it does on HTTP/2, since it also uses stream multiplexing. @mattrobenolt I haven't dug into this yet. Maybe you could point us to the code in the standard library where HTTP/2 handles this?

mattrobenolt:
@marten-seemann it doesn't happen in the http2 code path, it's purely in net/http right after Accepting a TCP connection.

https://cs.opensource.google/go/go/+/master:src/net/http/server.go;l=3257-3290?q=ConnContext&ss=go%2Fgo

So it is called as the first thing once a TCP connection is accepted, and this context is used for the lifetime until it disconnects.

For HTTP3/QUIC, I'd anticipate a similar behavior, once a QUIC connection is established, we'd share the context for the lifetime.

mattrobenolt:
No worries. The implementation as-is replaces the hack I had in place which was also per-request and not per-connection. So swapping to using this should yield a free win once that's fixed to behave more correct. :) It's just some minor optimization for performance to do some computations and cache them for the lifetime of a connection when possible.

rthellend:
The difference is that `net/http.Server` sets the context when a new TCP connection is received, whereas `http3.Server` sets it when a new quic stream is opened. I wouldn't expect the difference to be very significant.

mattrobenolt:
This is where my knowledge of QUIC and details may differ.

So let me explain why I use this functionality and what my goal is.

A simple example that we leverage is binding a logger to the context with the RemoteAddr. Then everywhere down the chain spawning from that connection should have the same IP address for multiple requests sharing that same connection.

In structured logging sense, effectively doing a `logger.With("remote_addr", conn.RemoteAddr().String())`

I would expect one QUIC connection to have the same RemoteAddr which multiple streams multiplex over.

Or are my fundamental understandings here wrong?

marten-seemann:
> I would expect one QUIC connection to have the same RemoteAddr which multiple streams multiplex over.

Not necessarily. In principle, QUIC allows connection migration, which means that the client could migrate, for example from a WiFi interface to a cellular interface. This of course will change its IP address.

mattrobenolt:
Right, so does that mean there is no start/end for a connection analagous to TCP? Is per-request the most general we can expect to get with QUIC?

marten-seemann:
I wouldn't put it that way. You still have an underlying QUIC connection, it's just that the remote address is not necessarily stable. But to give a counter example, if you authenticate on that connection somehow (TLS client certs?), this would apply to all requests sent on this connection.

mattrobenolt:
That's a good example, we don't do that, but I'd say similar vibes. I understand and would expect the remote addr to not be stable. But I would anticipate some signal for connection/start and end. There are other similar things we can do per-connection that are similar to auth, but we utilize some other caches and memory pooling on a per-connection basis when possible.

marten-seemann:
I’m not sure how the ConnContext would help with that. If you want to be able to perform some cleanup at the end of the connection, you can use the ServeQUICConn API.

mattrobenolt:
I'm not sure what you mean. Unless the reason is that QUIC behaves in a way that's different here than TCP. 

With TCP I can guarantee a single connection is at minimum the same client.

By cleanup, in Go it's just GCing away. But binding stuff in this case to the Context using WithValue or whatever, and maintaining that same Context across multiple requests.

Are you suggesting that the QUIC APIs just fundamentally can't support this with ConnContext? If not, this frankly doesn't differ from the context already on a Request and not sure what additional value a ConnContext brings other than having a hook to get the quic.Connection.

I don't mind, to be clear, if this is a limitation. I'm just trying to understand the disconnect. As-is currently this doesn't mirror the behavior of the typical net/http ConnContext, and if the behavior is explicitly different it might just be worth documenting that difference.

marten-seemann:
How does ConnContext allow you to find out when the underlying connection is closed?

mattrobenolt:
It doesn't. Maybe I confused things by mentioning that. My intent is that we use this hook to do effectively connection-local caching of some expensive computation rather than per request.

In our case, we often have TCP connections lasting many hours to days with thousands to millions of requests per second over 1 connection.

So being able to reuse anything connection wide is beneficial.

Currently our workloads and traffic over QUIC and HTTP/3 is relatively low, so this optimization hasn't been too important yet. 

And we rely on just normal Go GC to cleanup the connection-local caching we do when the connection is closed and the Context falls out of scope.

So we don't rely or expect this hook to handle connections closing. But as per the TCP hook, it's useful if we can establish one Context that's shared across every request over that same quic.Connection rather than a net.Conn.

I hope that helps explain.

PR #4233, opened 2024-01-03
Add qlog default tracer which writes to QLOGDIR if set

birneee:
Issue #4189

Split from #4201

### DIFF
--- a/qlog/qlog_dir.go
+++ b/qlog/qlog_dir.go
@@ -1,0 +1,57 @@
+package qlog
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"github.com/quic-go/quic-go/internal/utils"
+	"github.com/quic-go/quic-go/logging"
+	"log"
+	"os"
+	"strings"
+)
+
+// QlogDir contains the value of the QLOGDIR environment variable.
+// If it is the empty string ("") no qlog output is written.
+var QlogDir string
+
+func init() {
+	QlogDir = os.Getenv("QLOGDIR")
+	if QlogDir != "" {
+		if _, err := os.Stat(QlogDir); os.IsNotExist(err) {
+			if err := os.MkdirAll(QlogDir, 0o755); err != nil {
+				log.Fatalf("failed to create qlog dir %s: %v", QlogDir, err)
+			}
+		}
+	}
+}
+
+// DefaultTracer creates a qlog file in the qlog directory specified by the QLOGDIR environment variable.
+// File names are <odcid>_<perspective>.qlog.
+// Returns nil if QLOGDIR is not set.
+func DefaultTracer(ctx context.Context, p logging.Perspective, connID logging.ConnectionID) *logging.ConnectionTracer {
+	var label string
+	switch p {
+	case logging.PerspectiveClient:
+		label = "client"
+	case logging.PerspectiveServer:
+		label = "server"
+	}
+	return qlogDirTracer(p, connID, label)
+}
+
+// qlogDirTracer creates a qlog file in the qlog directory specified by the QLOGDIR environment variable.
+// File names are <odcid>_<label>.qlog.
+// Returns nil if QLOGDIR is not set.
+func qlogDirTracer(p logging.Perspective, connID logging.ConnectionID, label string) *logging.ConnectionTracer {
+	if QlogDir == "" {
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


qlog/qlog_dir.go
```diff
@@ -0,0 +1,58 @@
...
+	"os"
+	"strings"
+
+	"github.com/quic-go/quic-go/internal/utils"
+	"github.com/quic-go/quic-go/logging"
+)
+
+// QlogDir contains the value of the QLOGDIR environment variable.
+// If it is the empty string ("") no qlog output is written.
+var QlogDir string
```
marten-seemann, line 17: Why is this exported?

qlog/qlog_dir.go
```diff
@@ -0,0 +1,58 @@
...
+
+func init() {
+	QlogDir = os.Getenv("QLOGDIR")
```
marten-seemann, line 20: Should we read this outside of `init`? The last couple of times when I introduced environment variables, sooner or later people wanted to set them programmatically (`os.Setenv`), and that doesn't work if they're processed on initialization.

qlog/qlog_dir_test.go
```diff
@@ -0,0 +1,51 @@
...
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		QlogDir = ""
+		err := os.RemoveAll(tempTestDirPath)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("qlog dir ist set", func() {
```
marten-seemann, line 33:
```suggestion
	It("qlog dir is set", func() {
```

marten-seemann (commented):
Thank you for working on this @birneee!

It would be nice to have an integration test as well. In that test, you could assert that
1. The directory is created
2. It contains qlog files for both the client and server
3. These qlog files both contain the same connection ID

marten-seemann:
We can also simplify the qlog instructions in the README: https://github.com/quic-go/quic-go?tab=readme-ov-file#quic-event-logging-using-qlog

qlog/qlog_dir.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// qlogDirTracer creates a qlog file in the qlog directory specified by the QLOGDIR environment variable.
+// File names are <odcid>_<label>.qlog.
+// Returns nil if QLOGDIR is not set.
+func qlogDirTracer(p logging.Perspective, connID logging.ConnectionID, label string) *logging.ConnectionTracer {
+	if QlogDir == "" {
+		return nil
+	}
+	path := fmt.Sprintf("%s/%s_%s.qlog", strings.TrimRight(QlogDir, "/"), connID, label)
```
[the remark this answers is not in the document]
birneee: For me, this does not seem to be a big advantage, as one could also sort by file creation time.

I would rather stick with ODCID plus vantage point as recommended by the qlog draft.

qlog/qlog_dir.go
birneee: It is exposed to programmatically change it :)

qlog/qlog_dir.go
birneee: This would make the code and the interface simpler. The performance impact should also be negligible.

qlog/qlog_dir.go
```diff
@@ -0,0 +1,49 @@
...
+
+// qlogDirTracer creates a qlog file in the qlog directory specified by the QLOGDIR environment variable.
+// File names are <odcid>_<label>.qlog.
+// Returns nil if QLOGDIR is not set.
+func qlogDirTracer(p logging.Perspective, connID logging.ConnectionID, label string) *logging.ConnectionTracer {
+	qlogDir := os.Getenv("QLOGDIR")
+	if qlogDir == "" {
+		return nil
+	}
+	if _, err := os.Stat(qlogDir); os.IsNotExist(err) {
```
marten-seemann, line 37: 1. Should we handle non-nil errors that are not `os.IsNotExist` errors?
2. The documentation for `os.IsNotExist` says we should use `errors.Is(err, fs.ErrNotExist)`.

marten-seemann: I'll do this in a follow-up PR. I might also change the file name a bit.

marten-seemann (approved):

qlog/qlog_dir.go
birneee: Sounds good to me!

PR #4247, opened 2024-01-17
fix incorrect statement about connection ID lengths in the Transport

marten-seemann:


marten-seemann:
Tested by running some integration tests with 2-byte CIDs. Also tested 1-byte CIDs, which failed, but that's not unexpected, collisions are not unlikely.

PR #4248, opened 2024-01-18
add Transport config options to limit the number of handshakes

### LINKED ISSUE #3549: How to best implement address validation callback

marten-seemann:
Fixes #3549.

### DIFF
--- a/server.go
+++ b/server.go
@@ -699,7 +699,25 @@ func (s *baseServer) handleInitialImpl(p receivedPacket, hdr *wire.Header) error
 		s.numHandshakesUnvalidated.Add(1)
 	}
 	go conn.run()
-	go s.handleNewConn(conn, clientAddrValidated)
+	go func() {
+		completed := s.handleNewConn(conn)
+		if clientAddrValidated {
+			if s.numHandshakesValidated.Add(-1) < 0 {
+				panic("server BUG: number of validated handshakes negative")
+			}
+		} else if s.numHandshakesUnvalidated.Add(-1) < 0 {
+			panic("server BUG: number of unvalidated handshakes negative")
+		}
+		if !completed {
+			return
+		}
+
+		select {
+		case s.connQueue <- conn:
+		default:
+			conn.closeWithTransportError(ConnectionRefused)
+		}
+	}()
 	if conn == nil {
 		p.buffer.Release()
 		return nil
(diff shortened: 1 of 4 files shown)


transport.go
```diff
@@ -18,6 +18,11 @@ import (
 
 var errListenerAlreadySet = errors.New("listener already set")
 
+const (
+	defaultMaxNumUnvalidatedHandshakes = 32
+	defaultMaxNumHandshakes            = 64
```
marten-seemann, line 23: These values seem unreasonably low.

marten-seemann: Maybe we should set `defaultMaxNumHandshakes` to unlimited? It's really hard to pick a limit that fits for all.

server.go
```diff
@@ -710,10 +731,20 @@ func (s *baseServer) handleNewConn(conn quicConn) {
...
+	if clientAddrValidated {
+		if s.numHandshakesValidated.Add(-1) < 0 {
+			panic("server BUG: number of validated handshakes negative")
+		}
+	} else {
+		if s.numHandshakesUnvalidated.Add(-1) < 0 {
+			panic("server BUG: number of unvalidated handshakes negative")
+		}
+	}
+
```
sukunrt, line 743: I think this block needs to be deferred. There are a couple of returns in the block above and we need to *always* decrement this post a handshake completion.

sukunrt (changes requested):

transport.go
sukunrt: 64 still feels too low for `defaultMaxNumHandshakes`. This number is way too low to DoS for even very small machines. I think a high value > 1000 is better. I think unlimited is also better than this being < 1000.

server.go
marten-seemann: Good catch!

Deferring doesn't quite work, since we want to decrement the counter _before_ adding the connection to the server's accept queue. I refactored the function a bit in 3b6f7aa.

transport.go
marten-seemann: Changed to unlimited in 77ede93. Not entirely happy with that, but we can always change it later once we come up with a better value.

sukunrt (approved):

PR #4253, opened 2024-01-19
remove the RequireAddressValidation callback from the Config

marten-seemann:
Depends on #4248.

### DIFF
--- a/config.go
+++ b/config.go
@@ -2,7 +2,6 @@ package quic
 
 import (
 	"fmt"
-	"net"
 	"time"
 
 	"github.com/quic-go/quic-go/internal/protocol"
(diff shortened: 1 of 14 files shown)


sukunrt (approved):
This specific change looks good. But I think you need to defer this block that decrements the handshake count: https://github.com/quic-go/quic-go/blob/remove-require-address-validation-callback/server.go#L734-L745

There are a couple of returns just above that block.

marten-seemann:
> This specific change looks good. But I think you need to defer this block that decrements the handshake count: https://github.com/quic-go/quic-go/blob/remove-require-address-validation-callback/server.go#L734-L745
> 
> There are a couple of returns just above that block.

Done in #4248.

PR #4261, opened 2024-01-23
http3: only use :protocol pseudo-header for Extended CONNECT

taoso:
The default value should be "HTTP/3.0".

### DIFF
--- a/http3/headers.go
+++ b/http3/headers.go
@@ -140,15 +140,13 @@ func requestFromHeaders(headerFields []qpack.HeaderField) (*http.Request, error)
 			if err != nil {
 				return nil, err
 			}
+			protocol = hdr.Protocol
 		} else {
 			u.Path = hdr.Path
 		}
 		u.Scheme = hdr.Scheme
 		u.Host = hdr.Authority
 		requestURI = hdr.Authority
-		if hdr.Protocol != "" {
-			protocol = hdr.Protocol
-		}
 	} else {
 		u, err = url.ParseRequestURI(hdr.Path)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


marten-seemann:
Can you explain the motivation behind this PR?

taoso:
> Can you explain the motivation behind this PR?

hi @marten-seemann If the client fire a CONNECT request without :protocol, this req.Proto will be empty without this patch.

marten-seemann (commented):
I think this PR is correct, but I'd to tighten the validation logic here a bit. The `:protocol` pseudo header is only defined for Extended Connect requests (RFC 9220). We should reject any non-Extended Connected request that sets this header field.

@taoso Do you want to update the PR? We'd also need a test case that tests that we correctly reject requests with `:protocol` header.

http3/headers.go
```diff
@@ -124,11 +124,14 @@ func requestFromHeaders(headerFields []qpack.HeaderField) (*http.Request, error)
 		}
 	} else if len(hdr.Path) == 0 || len(hdr.Authority) == 0 || len(hdr.Method) == 0 {
 		return nil, errors.New(":path, :authority and :method must not be empty")
+	} else if len(hdr.Protocol) > 0 {
+		return nil, errors.New(":protocol must be empty")
```
marten-seemann, line 128: This branch needs to run `if !isExtendedConnected`. I think this is given by the current logic, but it's a bit subtle. I'd prefer to have this is in a separate if statement.

taoso: Fixed

marten-seemann (approved):
LGTM, thanks for pushing this through @taoso!

taoso:
Hi @marten-seemann  Could this PR be merged?

PR #4294, opened 2024-01-31
Introduce an API that lets the user get the current datagram size

nanokatze:
cc @marten-seemann @chungthuang @joliveirinha @tanghaowillow

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -638,6 +638,8 @@ runLoop:
 		} else {
 			sendQueueAvailable = nil
 		}
+
+		s.updateMaxDatagramDataSize()
 	}
 
 	s.cryptoStreamHandler.Close()
@@ -678,6 +680,22 @@ func (s *connection) ConnectionState() ConnectionState {
 	s.connState.Used0RTT = cs.Used0RTT
 	s.connState.GSO = s.conn.capabilities().GSO
 	return s.connState
+}
+
+func (s *connection) updateMaxDatagramDataSize() {
+	if s.peerParams == nil {
+		return
+	}
+	if !s.supportsDatagrams() {
+		return
+	}
+
+	maxDatagramFrameSize := min(s.peerParams.MaxDatagramFrameSize, s.packer.MaxPayloadSize(s.mtuDiscoverer.CurrentSize()))
+	maxDatagramDataSize := (&wire.DatagramFrame{DataLenPresent: true}).MaxDataLen(maxDatagramFrameSize, s.version)
+
+	s.connStateMutex.Lock()
+	s.connState.MaxDatagramSize = int(maxDatagramDataSize)
+	s.connStateMutex.Unlock()
 }
 
 // Time when the connection should time out
--- a/interface.go
+++ b/interface.go
@@ -343,6 +343,15 @@ type ConnectionState struct {
 	// If datagram support was negotiated, datagrams can be sent and received using the
 	// SendDatagram and ReceiveDatagram methods on the Connection.
 	SupportsDatagrams bool
+	// MaxDatagramSize specifies how big a datagram can be sent using
+	// SendDatagram. Datagrams bigger than MaxDatagramSize are silently dropped.
+	//
+	// Note: MaxDatagramSize can grow or shrink at any time. Users should call
+	// ConnectionState every once in a while to get up to date MaxDatagramSize.
+	//
+	// MaxDatagramSize is zero if datagrams are not supported on this
+	// connection.
+	MaxDatagramSize int
 	// Used0RTT says if 0-RTT resumption was used.
 	Used0RTT bool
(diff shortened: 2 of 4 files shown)


nanokatze:
I'm having trouble generating mocks using the latest https://github.com/uber-go/mock: running
`mockgen -typed -build_flags=-tags=gomock -package quic -self_package github.com/quic-go/quic-go -destination mock_packer_test.go github.com/quic-go/quic-go Packer` as it says in the comment on top of `packet_packer_test.go`, if I look at the diff, new mock types have a `Mock` prefix, for some reason, e.g.:
```diff
-// PackerAppendPacketCall wrap *gomock.Call
-type PackerAppendPacketCall struct {
+// MockPackerAppendPacketCall wrap *gomock.Call
+type MockPackerAppendPacketCall struct {
 	*gomock.Call
 }
```
which breaks lots of other code. Is this expected, and the other code using current names should be adjusted to use new names with `Mock` prefix or am I invoking it incorrectly?

marten-seemann:
You should only need to run `go generate ./...`.

connection.go
```diff
@@ -679,6 +681,22 @@ func (s *connection) ConnectionState() ConnectionState {
...
+	}
+	if !s.supportsDatagrams() {
+		return
+	}
+
+	maxDatagramFrameSize := min(s.peerParams.MaxDatagramFrameSize, s.packer.MaxPayloadSize(s.mtuDiscoverer.CurrentSize()))
+	maxDatagramDataSize := (&wire.DatagramFrame{DataLenPresent: true}).MaxDataLen(maxDatagramFrameSize, s.version)
+
+	s.connStateMutex.Lock()
+	s.connState.MaxDatagramSize = int(maxDatagramDataSize)
```
chungthuang, line 696: Have we consider using atomic uint32?

connection.go
```diff
@@ -527,6 +527,8 @@ runLoop:
 		default:
 		}
 
+		s.updateMaxDatagramDataSize()
```
marten-seemann, line 530: Can we be smarter than running this once per `runLoop` iteration? This is the hot path.

interface.go
```diff
@@ -344,6 +344,15 @@ type ConnectionState struct {
...
 	SupportsDatagrams bool
+	// MaxDatagramSize specifies how big a datagram can be sent using
+	// SendDatagram. Datagrams bigger than MaxDatagramSize are silently dropped.
+	//
+	// Note: MaxDatagramSize can grow or shrink at any time. Users should call
+	// ConnectionState every once in a while to get up to date MaxDatagramSize.
+	//
+	// MaxDatagramSize is zero if datagrams are not supported on this
+	// connection.
+	MaxDatagramSize int
```
marten-seemann, line 355: uint16?

connection.go
nanokatze: Do you mean this should be moved to the end of `runLoop` body? I wasn't sure under which conditions end of the `runLoop` body is or isn't reachable.

connection.go
nanokatze: I'm not convinced this would be a worthwhile change, right now there are lower hanging fruit to address first, and if connStateMutex contention turns out to be a problem, I think we might want to revisit how all of ConnectionState is accessed, not just this field.

interface.go
nanokatze: I don't mind making it `uint16`, but I don't see what improvement this would have. At worst I believe this would, in some cases, cause the users to perform additional conversion in their code, as dealing with sizes in Go conventionally is done in ints.

connection.go
marten-seemann: Not really. `runLoop` iterates once per (batch of) packets received and sent. Which is a lot on an active QUIC connection.

The available space in the packet however only changes under 2 conditions (I think):
1. When we update the connection ID.
2. When we discover a new MTU.

nanokatze: I think there's another reason: encoded packet number length increases, which IIUC grows with RTT and how fast we send packets (as in, packets per unit of time.)

I also just realized I misread and misunderstood your initial comment, oops.

The simplest way I can think of would be just to rate limit `updateMaxDatagramSize`. A more "precise" solution would be to run it in response to the 3 events: connection ID changed, PMTU changed, packet number length changed. Packet number length can change after every packet we send and on every ACK we receive. This reasoning is in fact why I initially went with running `updateMaxDatagramSize` every `runLoop`, as it runs in response to all of these events. In any case, do you think just rate limiting so that `updateMaxDatagramSize` runs once in a while, e.g. every 10 ms, or maybe it should be every N iterations of runLoop, would be good enough solution, or should we instead try to run `updateMaxDatagramSize` in response to 3 events listed above?

nanokatze:
After some thinking I've switched away from datagrams in favor of streams, in my application. I therefore won't pursue this change.

PR #4295, opened 2024-02-01
protocol: rename VersionNumber to Version

marten-seemann:


interface.go
```diff
@@ -16,8 +16,12 @@ import (
 // The StreamID is the ID of a QUIC stream.
 type StreamID = protocol.StreamID
 
+// A Version is a QUIC version number.
+type Version = protocol.Version
+
 // A VersionNumber is a QUIC version number.
-type VersionNumber = protocol.VersionNumber
+// Deprecated: VersionNumber was renamed to Version.
+type VersionNumber = Version
```
marten-seemann, line 24: Introducing a type alias here to make the transition less painful for users.

sukunrt (approved):

Issue #4124, opened 2023-10-23
http3: add qlog support


marten-seemann:
We currently only implement qlog support for QUIC, but not for HTTP/3.

It's not immediately clear how the package separation would work. It would be nice if we had an extensible system, otherwise, we'd have to pollute the qlog package with HTTP/3 events.

marten-seemann:
Here's an idea: We could expose the `qlog.Writer` and the `qlog.Event` (or move them to a separate package), and somehow expose it from the `quic.Connection`. This could be done using a new method on the struct, or via a context value.

The HTTP/3 package (and any other package as well) can then retrieve the `qlog.Writer`, and record its own events. And since `qlog.Event` is an interface, the events can be defined in that package, providing a clean separation.

However, this will require us to change the function signature of `Config.Tracer`: https://github.com/quic-go/quic-go/blob/3f3e12bd7c5b6a8759bd13e57e53846818898b8a/interface.go#L342
If this function returned both the `logging.ConnectionTracer` and the `qlog.Writer`, we could then pass the `qlog.Writer` to the `connection` constructor. Unfortunately, this is an API-breaking change.

@bt90 @sukunrt Any thoughts on this?

bt90:
Isn't the number of event types an application might may to use rather limited? e.g. [ApplicationError](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-quic-events-08.html#name-applicationerror)

marten-seemann:
This is about much more than just error codes. This is about an entire new set of events, defined by the application.

For example, the HTTP/3 events are described here: https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html

marten-seemann:
> Here's an idea: We could expose the `qlog.Writer` and the `qlog.Event` (or move them to a separate package), and somehow expose it from the `quic.Connection`. This could be done using a new method on the struct, or via a context value.

It would be preferable if we didn't have to attach any magic values to the connection context. One way to do this is to have the application create the `qlog.Writer` and attach it to the connection context themselves (under a context key chosen by the application).

marten-seemann:
Note that this feature would also be useful for non-HTTP/3 users of quic-go, since it would allow them to log any type of event in the same qlog file.

marten-seemann:
[...]

PR #4300, opened 2024-02-02
qlog: disentangle the ConnectionTracer from the qlog writer

marten-seemann:
The qlog writer simply records events, puts them into a channel, and consumes these events in a separate Go routine (by serializing them).

The `ConnectionTracer` is the one generating those events.

This will make it possible to export qlog using a `Tracer` (i.e. for events that don't belong to any connections), and it paves the way for serializing qlog events outside of the quic package (see #4124).

### DIFF
--- a/qlog/connection_tracer_test.go
+++ b/qlog/connection_tracer_test.go
@@ -3,12 +3,9 @@ package qlog
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"io"
-	"log"
 	"net"
 	"net/netip"
-	"os"
 	"time"
 
 	"github.com/quic-go/quic-go"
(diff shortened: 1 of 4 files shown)


qlog/writer.go
```diff
@@ -0,0 +1,84 @@
...
+		if w.encodeErr != nil { // if encoding failed, just continue draining the event channel
+			continue
+		}
+		if err := enc.Encode(ev); err != nil {
+			w.encodeErr = err
+			continue
+		}
+		if _, err := w.w.Write([]byte{'\n'}); err != nil {
+			w.encodeErr = err
+		}
```
sukunrt, line 67: Should we log this error? Right now it'll just stop processing events and not output this error till Close, right?

qlog/connection_tracer_test.go
sukunrt, line 1: NIT: Do you want to add a commit with just the rename?

sukunrt (approved):

qlog/writer.go
marten-seemann: It is logged when the qlog is closed. We need to continue draining the channel if we don't want to block the connection.

qlog/connection_tracer_test.go
marten-seemann: Does it make a difference, given that we're going to squash anyway? But really annoying that Git doesn't recognize it...

sukunrt: Would git blame work if we do two commits and then squash?

marten-seemann: I'll move it to a separate PR.

marten-seemann: Done in #4301.

PR #4307, opened 2024-02-03
ci: run macOS workflows on the new MacOS M1 runners

marten-seemann:


marten-seemann:
This doesn't really work. Asked in https://github.com/ipdxco/unified-github-workflows/issues/62.

Issue #4304, opened 2024-02-02
Handshake timeout when bound to 0.0.0.0 rather than specific interface(s)


thijsvandien:
Recently, I noticed that HTTP/3 on my Caddy server doesn't work anymore over IPv4, and reported that: https://github.com/caddyserver/caddy/issues/6073.  

Since this was identified as likely an upstream issue, I just tested with the example server:

`go run main.go -bind 0.0.0.0:443`

From commit 3b5950a1 connecting (`curl` or `quiche-client`) only works if I bind to the actual address of the interface used. 

I am running FreeBSD 14, which means https://github.com/quic-go/quic-go/issues/4105 might be somewhat related. Another user reported seeing the same on Ubuntu 22.04.3 LTS, but I haven't verified that.

marten-seemann:
I wasn't able to reproduce #4105. Can you help me debug this issue?

What are the packets that are received by quic-go? What do you think where things go wrong? https://github.com/quic-go/quic-go/commit/3b5950a1cea563ef974b8af784a8f8325b843c97 looks completely harmless to me, as it's only changing the way the IP is encoded internally, but maybe I'm missing something?

thijsvandien:
This is the first time I'm touching Go, so I asked ChatGPT to come up with a basic echo server, and went from there. Apologies if it's bad in any way. My goal was to investigate any differences in how `netip` treats IPv4 addresses on a shared v4/v6 socket vs. how `net` does that. 

Consider the following code:

```
func main() {
    pc, err := net.ListenPacket("udp", ":8080")
    if err != nil {
        log.Fatalf("Failed to listen on UDP port 8080: %v", err)
    }
    defer pc.Close()

    fmt.Println("UDP server listening on port 8080")

    for {
        buf := make([]byte, 1024)

        n, addr, err := pc.ReadFrom(buf)
        if err != nil {
            log.Printf("Error reading from UDP: %v", err)
            continue
        }

        log.Printf("Source before conversion: %s", addr)

        addrIP, ok := netip.AddrFromSlice(addr.(*net.UDPAddr).IP)
        if !ok {
            log.Printf("Error converting net.Addr to netip.Addr")
            continue
        }

        log.Printf("Source after conversion: %s", addrIP)
    }
}
```

On FreeBSD, it outputs:

```
Source before conversion: 192.168.0.1:51207
Source after conversion: ::ffff:192.168.0.1
```

[...]

Issue #4105, opened 2023-10-01
IPv4 connections to quic-go server fail on FreeBSD


peterjeremy:
quic-go uses dual-stack sockets so incoming IPv4 connections appear as IPv4-mapped IPv6 connections (see RFC2553 §3.7).  #3881 changed the behaviour of the OOB data associated with the connection (`quic.sconn`) in this case (see the change to `OOB()` in [sys_conn_oob.go](https://github.com/quic-go/quic-go/pull/3881/commits/3b5950a1cea563ef974b8af784a8f8325b843c97#diff-18d15af7517a1da90325d39231f5dba3c53c9be1a3770e256cddddb4d90a670f)):  Previously, an IPv4-mapped address was treated as IPv4 and would result in a `struct in_pktinfo` being attached (this doesn't exist on FreeBSD so the OOB data is empty).  The current behaviour is that an IPv4-mapped address is treated as IPv6 and results in a `struct in6_pktinfo` being attached.  FreeBSD doesn't support `struct in6_pktinfo` for IPv4 addresses and the `sendmsg` call that responds to the incoming connection will return EINVAL.

(Note that FreeBSD recommends that separate IPv4 and IPv6 sockets should be used rather than dual-stack sockets for security reasons - see the *Interaction between IPv4/v6 sockets* section of [inet6(4)](https://man.freebsd.org/cgi/man.cgi?query=inet6)).

peterjeremy:
The attached [sys_conn_oob.txt](https://github.com/quic-go/quic-go/files/12787012/sys_conn_oob.txt) reverts to the previous behaviour, where an IPv4-mapped IPv6 address is treated as an IPv4 address.  This corrects the behaviour on FreeBSD and still works on Linux (at least Ubuntu 22.04.3 LTS).  I can't check MacOS.

Note that a complete fix to make quic-go v0.39.0 server work on FreeBSD also requires the fix to #4106.

marten-seemann:
@peterjeremy Can you submit a PR with the fix please?

marten-seemann:
I can't reproduce this. Connecting to google.com with an IPv4-mapped IPv6 address (`go run example/client/main.go https://[::ffff:142.250.103.102]`) on a FreeBSD GCP VM was successful.

marten-seemann:
Closing, since this is not reproducible. Please comment here / reopen the issue with more details on how to reproduce it if you think this is incorrect.

MirkoGamerman:
[...]

PR #4308, opened 2024-02-04
Prevent using IPv4-mapped IPv6 address as source in OOB

thijsvandien:
Fixes https://github.com/quic-go/quic-go/issues/4304 and https://github.com/quic-go/quic-go/issues/4105. Thanks to @peterjeremy for identifying the issue.

### DIFF
--- a/sys_conn_oob_test.go
+++ b/sys_conn_oob_test.go
@@ -5,6 +5,7 @@ package quic
 import (
 	"fmt"
 	"net"
+	"net/netip"
 	"time"
 
 	"golang.org/x/net/ipv4"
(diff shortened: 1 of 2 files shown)


sys_conn_oob_test.go
```diff
@@ -310,6 +311,16 @@ var _ = Describe("OOB Conn Test", func() {
 		})
 	})
 
+	Context("serializing packet info", func() {
+		It("treats IPv4-mapped IPv6 addresses as IPv4", func() {
```
marten-seemann, line 315: Does this test fail on FreeBSD without your fix?

thijsvandien: In fact it should on any platform, because the packet will look different. Some are however insensitive to that and will happily send it.

thijsvandien:
The failing test was a flake? I cannot reproduce nor explain it. Also, my alternative implementation passed, while it essentially does the same.

sys_conn_oob_test.go
thijsvandien: Interestingly, on Windows it _doesn't_ fail. On macOS, FreeBSD and Ubuntu it does. This is very curious, so I will try to investigate further.

thijsvandien: Ah, these tests are skipped on Windows...

PR #4310, opened 2024-02-04
qlog: defer creation of the qlog file in QLOGDIR until the first write

marten-seemann:
If the syscall to create the qlog file takes a non-negligible amount of time, this can block the connection's run loop, or, even worse, the server's run loop.

By creating the qlog file on the first write, it will be created by a the qlogger's Go routine, and not be able to block the run loop. The qlogger uses a buffered channel (capacity: 50). The qlogger will therefore only block if more than 50 qlog events are queued before the syscall returns.

@birneee Would you mind taking a look at this PR?

### DIFF
--- a/qlog/qlog_dir.go
+++ b/qlog/qlog_dir.go
@@ -4,13 +4,54 @@ import (
 	"bufio"
 	"context"
 	"fmt"
-	"log"
+	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


birneee (commented):
generally looks good to me

PR #4313, opened 2024-02-07
avoid lock contention when accepting new connections

marten-seemann:
The server used to hold the packet handler map's lock while creating the connection struct for a newly accepted connection. This was intended to make sure that no two connections with the same Destination Connection ID could be created.

This is a corner case: it can only happen if two Initial packets with the same Destination Connection ID are received at the same time. If the second one is received after the first one has already been processed, it would be routed to the first connection. We don't need to optimized for this corner case. It's ok to create a new connection in that case, and immediately close it if this collision is detected.

### DIFF
--- a/server.go
+++ b/server.go
@@ -681,19 +681,21 @@ func (s *baseServer) handleInitialImpl(p receivedPacket, hdr *wire.Header) error
 		hdr.Version,
 	)
 	conn.handlePacket(p)
-	if q, ok := s.zeroRTTQueues[hdr.DestConnectionID]; ok {
-		for _, p := range q.packets {
-			conn.handlePacket(p)
-		}
-		delete(s.zeroRTTQueues, hdr.DestConnectionID)
-	}
 	// Adding the connection will fail if the client's chosen Destination Connection ID is already in use.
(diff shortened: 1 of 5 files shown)


server.go
```diff
@@ -636,63 +636,66 @@ func (s *baseServer) handleInitialImpl(p receivedPacket, hdr *wire.Header) error
...
+		s.logger,
+		hdr.Version,
+	)
+	conn.handlePacket(p)
+	if q, ok := s.zeroRTTQueues[hdr.DestConnectionID]; ok {
+		for _, p := range q.packets {
+			conn.handlePacket(p)
 		}
+		delete(s.zeroRTTQueues, hdr.DestConnectionID)
+	}
```
sukunrt, line 689: Can this block be after?

```
	if added := s.connHandler.AddWithConnID(hdr.DestConnectionID, connID, conn); !added {

```

sukunrt: I do not think it'd matter though as a collision here is rare and not exploitable since it needs to be two concurrent initial packets.

sukunrt (approved):

server.go
marten-seemann: Good catch! I agree that it's not exploitable, but this is definitely more correct.

Issue #4318, opened 2024-02-08
Possible data race in qtls.clientSessionCache


ainar-g:
We have detected this race once on one of our Windows builders, and unfortunately I'm unable to reproduce it again, see the backtrace below.  None of the goroutines seem to be from our code directly, so I've decided to report anyway, just in case.

Looking at the code, `qtls.clientSessionCache.Get` calls `cs.ResumptionState`, the documentation of which says:

```none
// It can be called by [ClientSessionCache.Put] to serialize (with
// [SessionState.Bytes]) and store the session.
```

I'm not entirely sure, but the wording *might* suggest that it is not supposed to be called from `tls.ClientSessionCache.Get`?  Additionally, `tls.NewResumptionState` docs state that:

```none
// state needs to be returned by [ParseSessionState], and the ticket and session
// state must have been returned by [ClientSessionState.ResumptionState].
```

Where, again, the wording is not clear.

Alternatively, it could be a `crypto/tls` race?

The full backtrace:

```none
==================
WARNING: DATA RACE
Write at 0x00c000386478 by goroutine 1677:
  github.com/quic-go/quic-go/internal/qtls.clientSessionCache.Get()
      C:/Users/runneradmin/go/pkg/mod/github.com/quic-go/quic-go@v0.41.0/internal/qtls/client_session_cache.go:53 +0x331
  github.com/quic-go/quic-go/internal/qtls.(*clientSessionCache).Get()
      <autogenerated>:1 +0x74
  crypto/tls.(*Conn).loadSession()
      C:/hostedtoolcache/windows/go/1.21.7/x64/src/crypto/tls/handshake_client.go:307 +0x2ac
  crypto/tls.(*Conn).clientHandshake()
      C:/hostedtoolcache/windows/go/1.21.7/x64/src/crypto/tls/handshake_client.go:188 +0x1bd
  crypto/tls.(*Conn).clientHandshake-fm()
      <autogenerated>:1 +0x47
  crypto/tls.(*Conn).handshakeContext()
      C:/hostedtoolcache/windows/go/1.21.7/x64/src/crypto/tls/conn.go:1552 +0x615
  crypto/tls.(*Conn).HandshakeContext()
      C:/hostedtoolcache/windows/go/1.21.7/x64/src/crypto/tls/conn.go:1492 +0x3a
  crypto/tls.(*QUICConn).Start.func1()
[...]

PR #4319, opened 2024-02-09
qtls: protect the tls.ClientSessionCache implementation with a mutex

marten-seemann:
This prevents a race condition when the underlying ClientSessionCache provided by the application returns the same session ticket for multiple connections. Reusing session tickets is explicitly recommended against by both RFC 8446 and RFC 9001, but it's not forbidden. This fix only benefits applications that compromise their users' privacy by reusing session tickets.

@ainar-g Could you give this PR a try? As a permanent fix, I'd recommend not reusing session tickets though.

### DIFF
--- a/internal/qtls/client_session_cache.go
+++ b/internal/qtls/client_session_cache.go
@@ -2,9 +2,11 @@ package qtls
 
 import (
 	"crypto/tls"
+	"sync"
 )
 
 type clientSessionCache struct {
+	mx      sync.Mutex
 	getData func(earlyData bool) []byte
 	setData func(data []byte, earlyData bool) (allowEarlyData bool)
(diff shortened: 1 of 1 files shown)


ainar-g:
As mentioned in #4318, we weren't able to reproduce the race in our code reliably yet.  As far as I can see, the original code is not explicitly reusing any session tickets.  It essentially is trying to establish both an HTTP/2 and an HTTP/3 connection to the same endpoint at the same time to see which protocol is faster.

The code in the PR looks good to me, but I wonder if I should file an issue about the unclear documentation in the Go repo?

marten-seemann:
Not sure what documentation you want to improve. Warn against the reuse of session tickets? That sounds like a worthwhile change.

Issue #4266, opened 2024-01-25
deadlock when closing transport and/or server


paralin:
I'm noticing this semaphore deadlock which occurs occasionally, but not always:

```
github.com/quic-go/quic-go.(*Transport).closeServer(0x14000398400)
        /quic-go/transport.go:306 +0x8c
github.com/quic-go/quic-go.(*baseServer).close.func1()
        /quic-go/server.go:341 +0x84
sync.(*Once).doSlow(0x1000200000008?, 0x200010000?)
        /go/1.21.6/libexec/src/sync/once.go:74 +0x100
sync.(*Once).Do(...)
        /go/1.21.6/libexec/src/sync/once.go:65
github.com/quic-go/quic-go.(*baseServer).close(0x140002fb188?, {0x101a0e028?, 0x101fbd560?}, 0x70?)
        /quic-go/server.go:335 +0x64
github.com/quic-go/quic-go.(*baseServer).Close(...)
        /quic-go/server.go:330
github.com/quic-go/quic-go.(*Listener).Close(0x0?)
        /quic-go/server.go:138 +0x34

and


github.com/quic-go/quic-go.(*baseServer).close(0x1400038eae0?, {0x101a0e028?, 0x101fbc040?}, 0x0?)
        /quic-go/server.go:335 +0x64
github.com/quic-go/quic-go.(*Transport).close(0x14000398400, {0x101a0e028, 0x101fbc040})
        /quic-go/transport.go:333 +0x10c
```

I see that there are two orders that `mutex.Lock` can be called:

- baseServer.close -> server.mtx.Lock() -> closeServer() -> transport.mtx.Lock()
- transport.close -> transport.mtx.lock() -> baseServer.close() -> server.mtx.Lock()

When locking mutexes out of order, a deadlock can occur, which explains this deadlock (as far as I can tell): 

> Mutex lock ordering rule: Given a total ordering of all mutexes, a program is deadlock-free if each thread acquires its mutexes in order and releases them in reverse order.

marten-seemann:
Interesting. Want to submit a fix?

paralin:
@marten-seemann Perhaps this? It's a bit dirty but it works: 

```
From d143f249fb5ecf6e8ef27a812f2a0b33e4203941 Mon Sep 17 00:00:00 2001
From: Christian Stewart <christian@aperture.us>
Date: Wed, 24 Jan 2024 18:04:35 -0800
[...]

PR #4332, opened 2024-02-17
server: fix deadlock when closing concurrently with transport

sukunrt:
When the listener and transport are closed concurrently. We can get a deadlock

`Listener.Close` is in `Listener.CloseOnce.Do` block and calls `transport.closeServer` for `notifyOnClose`. `transport.closeServer` requires `Transport.mutex`

`Transport.Close` holds the `Transport.mutex` and calls `Listener.Close` and gets stuck at the `Listener.closeOnce.Do` block

### DIFF
--- a/server.go
+++ b/server.go
@@ -98,7 +98,7 @@ type baseServer struct {
 		protocol.Version,
 	) quicConn
 
-	closeOnce sync.Once
+	closeMx   sync.Mutex
 	errorChan chan struct{} // is closed when the server is closed
 	closeErr  error
 	running   chan struct{} // closed as soon as run() returns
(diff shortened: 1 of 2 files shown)


sukunrt:
Oh sorry. Yes. I didnt realise there was an open PR. Happy to close this.

marten-seemann:
#4266 is the issue, not a PR. Please keep this one open.

marten-seemann:
@sukunrt Did you have a way to reliably trigger this deadlock? If so, would it make sense to add this as a test case here?

sukunrt:
On my machine this test is consistently deadlocking on master.

marten-seemann (approved):

marten-seemann:
Thank you @sukunrt!

PR #4334, opened 2024-02-20
qlog/event.go: import the error structs from quic-go/qerr

haxkor:
Instead of importing the error structs from the quic-go package, we can be more specific and import them from the qerr package.
If you tried to import the quic-go/qlog package you got an cyclic import error.
This fixes the problem.


haxkor:
clumsy mistake, opened new one

Issue #437, opened 2017-02-23
Proposal for Stream Priorities


marten-seemann:
This has been a long-standing *TODO* in our code and now that the QUIC WG draft contains a section about stream prioritization (https://quicwg.github.io/base-drafts/draft-ietf-quic-transport.html#rfc.section.9.5) it's a good time to address this properly.

In HTTP/2 streams can have a priority between 1 and 256 (see http://httpwg.org/specs/rfc7540.html#rfc.section.5.3.2), and I can't think of any situation where one would need a more fine-grained control (or where it would make a significant difference on the wire).

## quic package

We add a method `SetPriority(weight uint8)` to the `Stream` interface. The weight is a value between 0 and 255, and adding 1 gives the priority of the stream. `SetPriority` may be called at any time, so it allows reprioritizing existing streams. The default priority is 1.
When two streams both have data to send, the number of packets containing stream data for each stream is proportional to its priority, i.e. if there stream 1 has priority 100 and stream 2 has priority 200, data will be transferred twice as fast for stream 2.

## h2quic package

The easiest solution is to give the header stream(s) (stream 3 for now, soon every stream 4n+3) should have the highest priority (256). Data streams have priority 1.

For a more sophisticated solution, we can think about using the H2 stream priority to reprioritize QUIC streams. We need to evaluate if this makes sense (i.e. how the Go H2 implementation currently handles prioritization and if this gives any benefits) and how to map the H2 priority to the QUIC priority  (should a data stream be allowed to have the same priority as a header stream?).

jared2501:
Hey @marten-seemann, I have a use case for quic that requires stream prioritization. I'm thinking of implementing the proposal mentioned above for quic. I have a couple quic(k) questions:

- Would you accept an implementation as a contribution?
- If so, it appears that in order to implement the `SetPriority` method on the `Stream` interface I'd need to make the `streamFramer.streamQueue` a priority queue instead of a list and make some changes to `streamFramer.PopStreamFrames` to pop frames in priority order. Does that sound about right?

marten-seemann:
Hi @jared2501, thanks for reaching out. We're always happy about contributions :)

[...]

PR #4335, opened 2024-02-20
qlog/event.go: import the error structs from quic-go/internal/qerr

haxkor:
Instead of importing the error structs from the quic-go package, we can be more specific and import them from the qerr package.
If you tried to import the quic-go/qlog package you got an cyclic import error.
This fixes the problem.

### DIFF
--- a/qlog/event.go
+++ b/qlog/event.go
@@ -7,8 +7,8 @@ import (
 	"net/netip"
 	"time"
 
-	"github.com/quic-go/quic-go"
 	"github.com/quic-go/quic-go/internal/protocol"
+	"github.com/quic-go/quic-go/internal/qerr"
 	"github.com/quic-go/quic-go/internal/utils"
 	"github.com/quic-go/quic-go/logging"
 
@@ -114,12 +114,12 @@ func (e eventConnectionClosed) IsNil() bool        { return false }
 
 func (e eventConnectionClosed) MarshalJSONObject(enc *gojay.Encoder) {
 	var (
-		statelessResetErr     *quic.StatelessResetError
-		handshakeTimeoutErr   *quic.HandshakeTimeoutError
-		idleTimeoutErr        *quic.IdleTimeoutError
-		applicationErr        *quic.ApplicationError
-		transportErr          *quic.TransportError
-		versionNegotiationErr *quic.VersionNegotiationError
+		statelessResetErr     *qerr.StatelessResetError
+		handshakeTimeoutErr   *qerr.HandshakeTimeoutError
+		idleTimeoutErr        *qerr.IdleTimeoutError
+		applicationErr        *qerr.ApplicationError
+		transportErr          *qerr.TransportError
+		versionNegotiationErr *qerr.VersionNegotiationError
 	)
 	switch {
 	case errors.As(e.e, &statelessResetErr):


haxkor:
Hi, my apologies I really should have been more precise in the description of the problem:
If you add a new module that utilizes /quic/qlog, and go on to import this package in the /quic package, you get an import loop.
I have prepared [ a minimal example: ](https://github.com/haxkor/quic-go/commit/94ec85cd89f2503c6d15bd3848230dd05ca9bf79)

In my program utilizing my fork of quic, this is the go build output
`package myprogram
        imports github.com/quic-go/quic-go
        imports github.com/quic-go/quic-go/new_module
        imports github.com/quic-go/quic-go/qlog
        imports github.com/quic-go/quic-go: import cycle not allowed
`
This stems from /quic/qlog/events.go importing the error structs from /quic

marten-seemann:
That’s a problem in your fork, not in quic-go. I don’t see why we should change quic-go in this case.

haxkor:
Right now its not a problem in quic-go, true. It would become a problem once you want to add some functionality involving qlog.
In qlog/events.go, do you think it is better to import the error structs from quic instead of the more specific (original) location?

marten-seemann:
> Right now its not a problem in quic-go, true. It would become a problem once you want to add some functionality involving qlog.

It really is not. qlog is working fine.

> In qlog/events.go, do you think it is better to import the error structs from quic instead of the more specific (original) location?

Yes. qlog is its own package, and there's no reason why it needs to depend on *anything* internal at all. It's just an implementation of the `logging.ConnectionTracer` interface.

haxkor:
> It's just an implementation of the `logging.ConnectionTracer` interface.
Thank you for this, I had not thought of it this way. I wanted to reuse much of qlogs code but that was not possible without my changes.
I guess I will have to copy the relevant into my new package and end up with duplicates.

marten-seemann:
Can you elaborate on the changes you want to make for qlog? Maybe that’s something that would be useful to add to the package in general?

haxkor:
So the big thing I am working on is a mechanism that prioritizes bidirectional streams over unidirectional streams.
For that I need a tracer that records when a Bi/Uni frame is placed in the ringbuffer. Further, I also want information about congestion control. 

I do not want to change qlog, but reuse it.
Because the qlogs are still nice to have, I want to reuse many parts of the connectionTracer implementation to keep the qlog functionality. But maybe this should be solved by having a multiplexed tracer. I havent really gotten behind how I could implement that, maybe I should be looking into this.

What could not be solved with a multiplexed tracer is the reuse of the writer.go or event.go.

Right now I circumvent this issue by adding a function to qlog/connection_tracer.go that returns a connectionTracer.
See the code [here](https://github.com/haxkor/quic-go/blob/balancer/qlog/connection_tracer.go)
This way I will be able to reuse the qlog tracer but have a call to my extra functionality when I need it.
I also added 2 new event types (FrameBuffer Read / Write) to qlog/event.go

So in short I dont really want to improve qlog but reuse it. I am very new to quic-go so I appreciate your feedback on my way of doing things, please let me know if there is a better approach to solve this. 


Btw, as I looked at at qlog/event.go again, I realised qlog does already depend on some internal packages:
`	"github.com/quic-go/quic-go/internal/protocol"
	"github.com/quic-go/quic-go/internal/utils"

marten-seemann:
I see, so you're forking the entire quic-go packet. That's definitely non-standard use of the library :)
You might be able to get around this with clever `replace` directives, but I'm not sure if that works in your use case.
That said, stream priorities is a very long-standing issue, see https://github.com/quic-go/quic-go/issues/437. Hopefully we can resolve at some point..

haxkor:
Thank you for pointing me to the existing, longstanding discussion.
Do I understand correctly that there is no code enabling stream prioritisation?

My approach is quite hacky and distinguishes only between Uni and Bidi streams, so it will only be a POC.

>clever replace directives

Well I just implemented the fix proposed here:)

PR #4339, opened 2024-03-01
feat(.github): run QUIC Interop Runner

mxinden:
On a pull request, run the QUIC Interop Runner testcases.

Runs the pull request as a client and server against all other available implementations.

Depends on https://github.com/quic-interop/quic-interop-runner/pull/356.

### DIFF
--- a/.github/workflows/build-interop-docker.yml
+++ b/.github/workflows/build-interop-docker.yml
@@ -12,13 +12,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
         with:
           platforms: linux/amd64,linux/arm64
       - name: Login to Docker Hub
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
@@ -34,7 +34,7 @@ jobs:
             echo 'tag=latest' | tee -a $GITHUB_OUTPUT;
             echo 'gitref=${{ github.sha }}' | tee -a $GITHUB_OUTPUT;
           fi
-      - uses: docker/build-push-action@v4
+      - uses: docker/build-push-action@v5
         with:
           context: "{{defaultContext}}:interop"
           platforms: linux/amd64,linux/arm64
--- a/.github/workflows/build-interop-docker.yml
+++ b/.github/workflows/build-interop-docker.yml
@@ -19,6 +19,7 @@ jobs:
         with:
           platforms: linux/amd64,linux/arm64
       - name: Login to Docker Hub
+        if: ${{ github.event_name != 'pull_request' }}
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
--- a/.github/workflows/build-interop-docker.yml
+++ b/.github/workflows/build-interop-docker.yml
@@ -51,3 +51,5 @@ jobs:
           name: 'quic-go-latest'
           image: ${{ steps.docker_build_and_push.outputs.imageID }}
           url: https://github.com/quic-go/quic-go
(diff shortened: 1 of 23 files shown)


mxinden:
> This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

Mind approving the workflow @marten-seemann? I expect I will have to iterate on this pull request a bit. Any way we can get around this restriction?

marten-seemann:
> Mind approving the workflow @marten-seemann? I expect I will have to iterate on this pull request a bit. Any way we can get around this restriction?

Sure! I don't think GitHub allows me to white-list you, since you're a first-time contributor. The only way to get around that would be to get another PR merged. Maybe you can find a typo in a comment or in the README somewhere? 🤡

marten-seemann:
Better solution: I can just invite you as a collaborator on the repo.

.github/workflows/build-interop-docker.yml
```diff
@@ -35,10 +36,20 @@ jobs:
...
+        id: docker_build_and_push
         with:
           context: "{{defaultContext}}:interop"
           platforms: linux/amd64,linux/arm64
-          push: true
+          push: ${{ github.event_name != 'pull_request' }}
           build-args: |
             GITREF=${{ steps.tag.outputs.gitref }}
           tags: martenseemann/quic-go-interop:${{ steps.tag.outputs.tag }}
+      - name: Run QUIC Interop tests
```
marten-seemann, line 47: You might want to make this a separate step:
1. Building the image is cheap (this is Go after all ;))
2. The run triggers are orthogonal anyway, so there's no condition where we'd build twice anyway.

mxinden: > You might want to make this a separate step:

Note that it is already a separate step. Are you suggesting to make it a separate job instead?

mxinden:
Making progress. Latest run:

```
+----------------+----------------------------------------------+----------------------------------------------+--------------------------------------------+
|                |                quic-go-latest                |                   quic-go                    |                   msquic                   |
+----------------+----------------------------------------------+----------------------------------------------+--------------------------------------------+
| quic-go-latest | ✓(H,DC,LR,C20,M,S,R,Z,3,B,U,A,L1,L2,C1,C2,6) | ✓(H,DC,LR,C20,M,S,R,Z,3,B,U,A,L1,L2,C1,C2,6) | ✓(H,DC,LR,C20,M,S,R,Z,B,U,A,L1,L2,C1,C2,6) |
|                |                   ?(E,V2)                    |                   ?(E,V2)                    |                 ?(3,E,V2)                  |
|                |                     ✕()                      |                     ✕()                      |                    ✕()                     |
+----------------+----------------------------------------------+----------------------------------------------+--------------------------------------------+
|    quic-go     | ✓(H,DC,LR,C20,M,S,R,Z,3,B,U,A,L1,L2,C1,C2,6) |                                              |                                            |
|                |                   ?(E,V2)                    |                                              |                                            |
|                |                     ✕()                      |                                              |                                            |
+----------------+----------------------------------------------+----------------------------------------------+--------------------------------------------+
|      neqo      |   ✓(H,DC,LR,C20,M,S,R,3,B,U,A,L2,C1,C2,6)    |                                              |                                            |
|                |                   ?(E,V2)                    |                                              |                                            |
|                |                   ✕(Z,L1)                    |                                              |                                            |
+----------------+----------------------------------------------+----------------------------------------------+
```

https://github.com/quic-go/quic-go/actions/runs/8143949752/job/22256953717?pr=4339

marten-seemann:
Very impressive!

I see that the runtime of this step is now 1.5h. Have you considered parallelizing it, similar to what the interop runner does? I realize that we then won't need the `--must-include` flag anymore (it's a useful addition nevertheless). Then we could run the entire test in just 10-15 minutes (maybe even faster?), which would make it more useful on PRs.

.github/workflows/build-interop-docker.yml
marten-seemann: Yes, separate job is what I meant, sorry for mixing up the terminology here. I think I'd even prefer a separate workflow, since it's (almost) orthogonal to this one.

mxinden:
https://github.com/mozilla/neqo/pull/1847/ adds parallelization to the Neqo Interop action, as Marten proposed above. I hope to get to port this to https://github.com/quic-interop/quic-interop-runner/pull/356 at some point and would then update this pull request.

Sorry for the delay.

PR #4356, opened 2024-03-09
quicvarint: use int instead of internal protocol.ByteCount type

marten-seemann:
Not sure if `int` or `uint64` would be better here.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,11 +5,21 @@ linters-settings:
   misspell:
     ignore-words:
       - ect
+  depguard:
+    rules:
+      quicvarint:
+        list-mode: strict
+        files:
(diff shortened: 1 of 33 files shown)


.golangci.yml
```diff
@@ -5,11 +5,21 @@ linters-settings:
...
       - ect
+  depguard:
+    rules:
+      quicvarint:
+        list-mode: strict
+        files:
+          - "**/github.com/quic-go/quic-go/quicvarint/*"
+          - "!$test"
+        allow:
+          - $gostd
```
marten-seemann, line 16: Adding a validation here. This package shouldn't need to use any packages outside of the standard library.

PR #4369, opened 2024-03-18
close connection when an abnormally large number of frames are queued

marten-seemann:
Under normal circumstances, we should be able to send out control frames right away, so we don't expect any queue to build up. To defend against resource exhaustion attacks, we limit the control frame queue to 16k elements.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -521,6 +521,9 @@ func (s *connection) run() error {
 
 runLoop:
 	for {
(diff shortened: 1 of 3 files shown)


marten-seemann:
Approved by @sukunrt and @MarcoPolo in https://github.com/quic-go/quic-go-ghsa-c33x-xqrf-c478/pull/2.

PR #4379, opened 2024-03-22
http3: don't modify any fields of the http.Request when doing 0-RTT

marten-seemann:
Depends on #4378.


sukunrt (approved):
Why is this required for 0RTT requests?

marten-seemann:
It should be possible to issue the same request twice.

Issue #3543, opened 2022-08-31
Is CONNECT-UDP RFC-9298 Supported?


magicalo:
Hi,

Does this support CONNECT-UDP and more over does it have UDP/TCP proxy capability that CONNECT-UDP and CONNECT Methods enable?

thx

marten-seemann:
If it was, it would be listed in the README. For proxying, see #3370.

magicalo:
So the answer is no.  Are there plans to support CONNECT-UDP so that I can use this code for a Proxy Server that will support clients that attempt to Proxy through to a  UDP resource behind the go Proxy server.
[client]==============[http/3 quic proxy]-----------------[udp resource] (via CONNECT-UDP)

RithvikChuppala:
@magicalo have you figured out how to proxy using QUIC-go?

Issue #3370, opened 2022-04-05
RoundTripper and Proxy


arcadi4n:
Hello, 

Thank you for your great project!

I wrote a client that will use the roundtripper and works fine. However, the computer on which I want to run the client has a system proxy configured. I tried something similar to how I would normally do it using the "net/http" library but I get an error that http3.roundTripper struct does't support the Proxy field.

Is there any way to do this?

marten-seemann:
What would the Proxy field do?

wireguard-dev:
any update?

pteichman:
A note for others in here: some of this can be done with the current API (though it's a little cumbersome). Dropping a little extra info in here in case it can inform an implementation:

My motivation is that I need to use HTTP3 to create a CONNECT style proxy to another host. The behavior is described in section 4.4 of RFC9114: the CONNECT is sent over HTTP3 (and quic) and the proxy creates a TCP connection to the requested host and port, which can be used via the quic stream from the CONNECT request. None of that is important from the HTTP3 client perspective, just a little background.

I can do this for stdlib http requests by setting the Proxy and ProxyConnectHeader (for auth) fields; with those set, `http.Transport` will automatically issue the CONNECT request as it dials.

With `quic-go/http3` you can do something similar by writing a new http.RoundTripper. Its RoundTrip needs to:

* Create an `http3.RoundTripper` instance that overrides Dial to connect to the proxy host
* Create the desired CONNECT request
* RoundTripOpt that request using the http3 RoundTripper; you must set DontCloseRequestStream
* On success, extract the HTTPStream from the CONNECT response body
* Create a stdlib `http.Transport` instance that overrides Dial to return that HTTPStream; I wrapped it in a struct that stubs out RemoteAddr and LocalAddr so it can be a `net.Conn`
* Use that transport to RoundTrip the request you want to send to the destination server

My use case would be satisfied by adding a stdlib compatible `Proxy` and `ProxyConnectHeader` field to `http3.RoundTripper`, but all of the above would need to be added with them; likely in its Dial.

Fundamentally though this is a way to round trip an HTTP2 or 1.1 request using the HTTP3 connection; it's more of a compatibility bridge between `net/http` and `quic-go/http3` than a new way to get an HTTP3 request through. CONNECT-UDP may be an answer for that, but that's outside of my use case.

paramaw:
hi @kokes, do you have any code samples of your implementation? I'm looking to do the same thing.

[...]

PR #4392, opened 2024-03-28
http3: support HTTP CONNECT and CONNECT-UDP (masque) protocols

max-b:
We'd like to propose these changes in order for us to support client http3 CONNECT and CONNECT-UDP.

We've rolled up a few changes into this single PR:
- Adding a `Proxy` method to the http3.RoundTripper
- Allowing access to the underlying `quic.Connection` from a `hijackableBody`
- Some slight adjustments to the handling of methods in the http3 `Client`/`RoundTripper`
- Adding support for the draft03 version of RFC 9297
  - We think this is a reasonable approach, in particular since it's the behavior of [quiche](https://github.com/cloudflare/quiche/blob/e71c51b8cd9aa3b78749fa7cb82f35699d23e7bf/quiche/src/h3/frame.rs#L243-L247) and [h2o](https://github.com/h2o/h2o/blob/7545f5fe55ccdc9a12bdeb246b5d7110f8bee225/lib/http3/common.c#L1145-L1150)

We're currently using quic-go along with these changes for our implementation of a MASQUE proxy client: https://github.com/Invisv-Privacy/masque/pull/4

I'm aware that we've not had any previous interaction w/ y'all so we're *very* open to any/all feedback/changes/etc.

I think that this would address the client sides of https://github.com/quic-go/quic-go/issues/3370 and https://github.com/quic-go/quic-go/issues/3543

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -15,6 +15,7 @@ import (
 // When a stream is taken over, it's the caller's responsibility to close the stream.
 type HTTPStreamer interface {
 	HTTPStream() Stream
+	HTTPConnection() quic.Connection
 }
 
 type StreamCreator interface {
@@ -56,6 +57,10 @@ func newRequestBody(str Stream) *body {
 func (r *body) HTTPStream() Stream {
 	r.wasHijacked = true
 	return r.str
+}
+
+func (r *body) HTTPConnection() quic.Connection {
+	return nil
 }
 
 func (r *body) wasStreamHijacked() bool {
@@ -133,4 +138,8 @@ func (r *hijackableBody) Close() error {
 
 func (r *hijackableBody) HTTPStream() Stream {
 	return r.str
+}
+
+func (r *hijackableBody) HTTPConnection() quic.Connection {
+	return r.conn
 }
--- a/http3/client.go
+++ b/http3/client.go
@@ -273,7 +273,7 @@ func (c *client) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Respon
 }
 
 func (c *client) roundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Response, error) {
-	if authorityAddr("https", hostnameFromRequest(req)) != c.hostname {
+	if !validProxyMethod(req.Method) && authorityAddr("https", hostnameFromRequest(req)) != c.hostname {
 		return nil, fmt.Errorf("http3 client BUG: RoundTripOpt called for the wrong client (expected %s, got %s)", c.hostname, req.Host)
 	}
 
@@ -459,7 +459,7 @@ func (c *client) doRequest(req *http.Request, conn quic.EarlyConnection, str qui
 	_, hasTransferEncoding := res.Header["Transfer-Encoding"]
 	isInformational := res.StatusCode >= 100 && res.StatusCode < 200
 	isNoContent := res.StatusCode == http.StatusNoContent
-	isSuccessfulConnect := req.Method == http.MethodConnect && res.StatusCode >= 200 && res.StatusCode < 300
+	isSuccessfulConnect := validProxyMethod(req.Method) && res.StatusCode >= 200 && res.StatusCode < 300
 	if !hasTransferEncoding && !isInformational && !isNoContent && !isSuccessfulConnect {
 		res.ContentLength = -1
 		if clens, ok := res.Header["Content-Length"]; ok && len(clens) == 1 {
--- a/http3/frames.go
+++ b/http3/frames.go
@@ -93,6 +93,10 @@ const (
 	settingExtendedConnect = 0x8
 	// HTTP Datagrams, RFC 9297
 	settingDatagram = 0x33
+	// For backwards compatibility with the draft datagram RFC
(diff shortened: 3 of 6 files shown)


marten-seemann:
Hi Max! Thank you for your PR!

quic-go will gain MASQUE capabilities soon(-ish). I just opened a tracking issue for that: https://github.com/quic-go/quic-go/issues/4393.

The way we’ll implement it will look a bit different though.
For example, we want to fully support HTTP datagrams (https://github.com/quic-go/quic-go/issues/3522), in a way where the HTTP/3 layer deals with prepending the Quarter Stream ID, and dispatches datagrams to their respective streams. I’ve been working towards that, and shipped a number of changes to the http3 package in the v0.42 release, in preparation for that.

Regarding backwards compatibility with old versions of the datagram draft, we won’t add that to quic-go, given that the RFC has shipped for quite a while now. If I understand correctly though, all you seem to need is a different code point in the SETTINGS. Does the current API already cover this? It should be possible to define additional settings code points using the `Other` map.

Regarding the `Proxy` addition to the `RoundTripper`, this looks reasonable to me at first glance (especially since the standard library `http.Transport` seems to do something similar, but I have to admit I haven’t thought about this a lot. I also see a `OnProxyConnectResponse` response, which I don’t understand what it’s used for yet. I’d prefer to continue this discussion on the issue though.

max-b:
Thanks so much for your interest and response! https://github.com/quic-go/quic-go/issues/4393 seems like an exciting direction for the project :)

> Regarding backwards compatibility with old versions of the datagram draft, we won’t add that to quic-go, given that the RFC has shipped for quite a while now. If I understand correctly though, all you seem to need is a different code point in the SETTINGS. Does the current API already cover this? It should be possible to define additional settings code points using the Other map.

I might not have the most full understanding of the api and how the `Other` map works, but because the draft code point has semantic meaning re: the frame settings I'm not sure the api supports our use case?

https://github.com/Invisv-Privacy/quic-go-upstream/blob/66efcd13532305e856773dd484adec8e86ee469a/http3/frames.go#L152-L160
https://github.com/Invisv-Privacy/quic-go-upstream/blob/66efcd13532305e856773dd484adec8e86ee469a/http3/client.go#L232-L239

> Regarding the Proxy addition to the RoundTripper, this looks reasonable to me at first glance (especially since the standard library http.Transport seems to do something similar, but I have to admit I haven’t thought about this a lot. 

We'd be happy to break this piece out into a separate PR if that would be helpful?

> I also see a OnProxyConnectResponse response, which I don’t understand what it’s used for yet. I’d prefer to continue this discussion on the issue though.

https://github.com/golang/go/issues/54299 and https://go-review.googlesource.com/c/go/+/447216 seem helpful. The functionality isn't quite parallel though as that method is actually responsible for the functionality of http/socks/etc proxying. It's not super clear to me how the `OnProxyConnectResponse` would fit in this case. To be honest, if you're trying to parallel golang's `net/http` api, the `Proxy` method might need to wait on you actually implementing the underlying functionality. 

Thanks!

marten-seemann:
> > Regarding backwards compatibility with old versions of the datagram draft, we won’t add that to quic-go, given that the RFC has shipped for quite a while now. If I understand correctly though, all you seem to need is a different code point in the SETTINGS. Does the current API already cover this? It should be possible to define additional settings code points using the Other map.
> 
> I might not have the most full understanding of the api and how the `Other` map works, but because the draft code point has semantic meaning re: the frame settings I'm not sure the api supports our use case?

You're right. You could negotiate the extension, but quic-go wouldn't recognize it as having been negotiated, and therefore reject datagrams.
How much do you care about the draft version? Do you expect to continue supporting for a long time?

marten-seemann:
> > Regarding the Proxy addition to the RoundTripper, this looks reasonable to me at first glance (especially since the standard library http.Transport seems to do something similar, but I have to admit I haven’t thought about this a lot.
> 
> We'd be happy to break this piece out into a separate PR if that would be helpful?

Yes, definitely! I'm not sure I fully understand this field (even in the standard library) yet: What happens if you return a socks5 URL? Doesn't this mean we now need to support socks5? What if http is returned? Do we now need to branch out to the standard library to do an HTTP request on TCP?
Maybe this relates to the larger question of whether this package should implement some kind of Happy Eyeballs v3 to dial both QUIC and TCP.
To be clear, this doesn't all need to be solved by your PR. I'm fine just adding proxying for HTTP/3 for now. Just want to know what the way forward is, before we commit to an API.

max-b:
> You're right. You could negotiate the extension, but quic-go wouldn't recognize it as having been negotiated, and therefore reject datagrams. How much do you care about the draft version? Do you expect to continue supporting for a long time?

We originally developed our client targeting h2o while it was still only implementing the draft. Since Fall of last year they've added support for the finalized rfc: https://github.com/h2o/h2o/pull/3278 We're currently only using the draft version, but in the future we do expect we'll eventually update to use the finalized spec, perhaps if/when h2o decides to drop draft support. We'd prefer to have the extra settings code point here and both quiche and h2o implement it this way, but it's obviously up to you and if you choose not to we will eventually move to the finalize rfc version.

marten-seemann:
CONNECT-UDP is now being worked on in https://github.com/quic-go/masque-go.

PR #4400, opened 2024-04-01
introduce a ConnectionTracingID type for the ConnectionTracingKey

marten-seemann:
This is a breaking API change. Unfortunately, context values are not strongly typed, so users will have have to take care to change their type assertions.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -59,6 +59,9 @@ var Err0RTTRejected = errors.New("0-RTT rejected")
 // as well as on the context passed to logging.Tracer.NewConnectionTracer.
 var ConnectionTracingKey = connTracingCtxKey{}
 
+// ConnectionTracingID is the type of the context value saved under the ConnectionTracingKey.
+type ConnectionTracingID uint64
+
 type connTracingCtxKey struct{}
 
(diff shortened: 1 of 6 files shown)


interface.go
```diff
@@ -59,6 +59,9 @@ var Err0RTTRejected = errors.New("0-RTT rejected")
 // as well as on the context passed to logging.Tracer.NewConnectionTracer.
 var ConnectionTracingKey = connTracingCtxKey{}
 
+// ConnectionTracingID is the type of the context value saved under the ConnectionTracingKey.
+type ConnectionTracingID uint64
```
marten-seemann, line 63: I'm not really happy with the name. Tracing is an overloaded term, and we already used it for event tracer.

Any suggestions for a better name?

sukunrt: I guess this is okay. 
Other ideas that come to my mind are:
```
ConnectionTrackingID
ConnectionSeqID (as the id is sequential)
ConnectionName (used as a name for this connection in tracking)
```

marten-seemann: Thank you!

PR #4407, opened 2024-04-03
Implement graceful shutdown for http3 servers.

WeidiDeng:
Fix [6195](https://github.com/caddyserver/caddy/issues/6195) for caddy.


http3/client.go
```diff
@@ -72,6 +72,10 @@ type client struct {
 	hostname string
 	conn     atomic.Pointer[quic.EarlyConnection]
 
+	receivedGoway atomic.Bool
```
francislavoie, line 75:
```suggestion
	receivedGoaway atomic.Bool
```

francislavoie (commented):
Small typo

WeidiDeng:
@marten-seemann Do you time to take a look at this draft?

http3/client.go
```diff
@@ -114,14 +119,16 @@ func newClient(hostname string, tlsConf *tls.Config, opts *roundTripperOpts, con
...
-		dialer:        dialer,
-		logger:        logger,
+		hostname:         authorityAddr("https", hostname),
+		tlsConf:          tlsConf,
+		requestWriter:    newRequestWriter(logger),
+		decoder:          qpack.NewDecoder(func(hf qpack.HeaderField) {}),
+		config:           conf,
+		opts:             opts,
+		dialer:           dialer,
+		receivedGoawayID: quic.StreamID(-4),
```
marten-seemann, line 129: This seems hacky. Introduce a protocol.InvalidStreamID instead?

http3/frames.go
```diff
@@ -184,3 +186,35 @@ func (f *settingsFrame) Append(b []byte) []byte {
 	}
 	return b
 }
+
+type goawayFrame struct {
+	ID quic.StreamID
+}
+
+func parseGoawayFrame(r io.Reader, l uint64) (*goawayFrame, error) {
+	if l > 8*(1<<10) {
```
marten-seemann, line 195: This doesn’t seem right.

http3/frames.go
```diff
@@ -184,3 +186,35 @@ func (f *settingsFrame) Append(b []byte) []byte {
...
+		return nil, err
+	}
+	frame := &goawayFrame{}
+	b := bytes.NewReader(buf)
+	id, err := quicvarint.Read(b)
+	if err != nil {
+		return nil, err
+	}
+	frame.ID = quic.StreamID(id)
+	return frame, nil
```
marten-seemann, line 212: Need to handle the case where there’s unconsumed data in the reader.

http3/server.go
```diff
@@ -217,7 +218,8 @@ type Server struct {
 	mutex     sync.RWMutex
 	listeners map[*QUICEarlyListener]listenerInfo
 
-	closed bool
+	closed      bool
+	connections map[*quic.Connection]func()
```
marten-seemann, line 222: Better: Use the tracing ID as a map key.

marten-seemann (commented):
Thank you for working on this @WeidiDeng!

I’ve been working on a big refactor of the http3 package (#4115, #4116, more coming), which at the end will enable HTTP datagram support, which will all be included in the next release. For proper datagram support, we’ll introduce some kind of connection and stream tracking (no PR yet), which would be helpful for this PR. I suggest we defer graceful shutdown until these PRs have landed.

http3/client.go
WeidiDeng: I got the stream ID from [rfc 9000](https://www.rfc-editor.org/rfc/rfc9000.html#name-stream-types-and-identifier). For different kinds of streams, the criteria for invalid id is different. I think a function that verifies ID would be better instead of a new type.

http3/frames.go
WeidiDeng: Just copied the length check from the setting frame parsing. I guess it should compare to the maximal length of a quic varint 😕 ?

http3/frames.go
WeidiDeng: I thought about that, but too lazy at that moment 😄 .

http3/server.go
WeidiDeng: But the methods on the `quic.Connection` needs to be called. To properly stop server from accepting new streams at the grace period's end. Malicious client may attempt to open new streams even if goaway is received. [Goaway itself doesn't close a connection](https://www.rfc-editor.org/rfc/rfc9114.html#name-goaway).

WeidiDeng:
@marten-seemann When do you think this patch is ready to be reviewed? I haven't updated it in a while, and several of those enhancements already landed in the latest release. If you think it's ok to add graceful shutdown, I'll update this patch against the latest release.

marten-seemann:
Sorry, I dropped the ball on this PR. Let's get this into the next release! I'll review it later.

http3/server.go
```diff
@@ -436,6 +447,33 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
+	s.mutex.Lock()
+	s.connections[&conn] = func() {
+		once.Do(func() {
+			mutex.Lock()
+			closed = true
+			b = (&goawayFrame{
+				ID: lastID + 4,
+			}).Append(b[:0])
+			str.Write(b)
+			mutex.Unlock()
```
marten-seemann, line 466: `str.Write` might block indefinitely. You shouldn't hold the mutex that long.

http3/server.go
marten-seemann: > Better: Use the tracing ID as a map key.

Update: the tracing ID has now been deprecated, so we shouldn't use that one.

http3/server.go
```diff
@@ -436,6 +447,33 @@ func (s *Server) handleConn(conn quic.Connection) error {
 	}).Append(b)
 	str.Write(b)
 
+	var (
+		once   sync.Once
+		mutex  sync.Mutex
+		closed bool
+		// client opened bidirectional stream ID starts at 0 and increase by 4
+		lastID = quic.StreamID(-4)
```
marten-seemann, line 455:
```suggestion
		lastID = protocol.InvalidStreamID
```

http3/server.go
```diff
@@ -618,11 +695,54 @@ func (s *Server) Close() error {
 	return err
 }
 
+func (s *Server) connectionsCount() int {
```
marten-seemann, line 698: The function name suggests that this function doesn't have any side effects, but it clearly does.

http3/server.go
```diff
@@ -454,6 +492,20 @@ func (s *Server) handleConn(conn quic.Connection) error {
 			}
 			return fmt.Errorf("accepting stream failed: %w", err)
 		}
+
+		var reject bool
+		mutex.Lock()
+		reject = closed
```
marten-seemann, line 498: `s.closed`?

http3/server.go
```diff
@@ -454,6 +492,20 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
+		if !reject {
+			lastID = str.StreamID()
```
marten-seemann, line 500: You don't need to hold the mutex here.

http3/client.go
```diff
@@ -83,6 +88,9 @@ func (c *SingleDestinationRoundTripper) init() {
 	c.decoder = qpack.NewDecoder(func(hf qpack.HeaderField) {})
 	c.requestWriter = newRequestWriter()
 	c.hconn = newConnection(c.Connection, c.EnableDatagrams, protocol.PerspectiveClient, c.Logger)
+	c.hconn.controlStrHandler = c.readControlStream
+	c.receivedGoawayID = quic.StreamID(-4)
```
marten-seemann, line 92:
```suggestion
	c.receivedGoawayID = protocol.InvalidStreamID
```

http3/client.go
```diff
@@ -208,6 +264,36 @@ func (c *SingleDestinationRoundTripper) roundTrip(req *http.Request) (*http.Resp
...
+	// the context may not be canceled if it's added after goaway frame is processed
+	if goawayID, ok := c.goawayReceived(); ok && goawayID <= id {
+		cancel(errGoaway)
+		str.CancelWrite(quic.StreamErrorCode(ErrCodeRequestCanceled))
+		str.CancelRead(quic.StreamErrorCode(ErrCodeRequestCanceled))
+		return nil, errGoaway
+	}
+
+	// Separate goroutine to prevent interference with request cancellation
+	go func() {
```
marten-seemann, line 289: I might be missing something, but this looks to me like you're canceling requests that are in flight when receiving the GOAWAY. But that's exactly what we should not do: Requests still in flight should complete as if nothing happened.

http3/frames.go
```diff
@@ -194,3 +195,41 @@ func (f *settingsFrame) Append(b []byte) []byte {
...
+
+type goawayFrame struct {
+	ID quic.StreamID
+}
+
+// max length of an encoded quic varint: quicvarint.Len(quicvarint.Max)
+const maxQuicVarintLen = 8
+
+func parseGoawayFrame(r io.Reader, l uint64) (*goawayFrame, error) {
+	if l > maxQuicVarintLen {
```
marten-seemann, line 207: Easier: just parse the varint, and check that the length matches the number of bytes parsed afterwards.

http3/frames.go
```diff
@@ -194,3 +195,41 @@ func (f *settingsFrame) Append(b []byte) []byte {
 	}
 	return b
 }
+
+type goawayFrame struct {
+	ID quic.StreamID
```
marten-seemann, line 200:
```suggestion
	StreamID quic.StreamID
```

http3/frames.go
```diff
@@ -194,3 +195,41 @@ func (f *settingsFrame) Append(b []byte) []byte {
...
+	}
+	frame.ID = quic.StreamID(id)
+	return frame, nil
+}
+
+func (f *goawayFrame) Append(b []byte) []byte {
+	b = quicvarint.Append(b, 0x7)
+	b = quicvarint.Append(b, uint64(quicvarint.Len(uint64(f.ID))))
+	b = quicvarint.Append(b, uint64(f.ID))
+	return b
```
marten-seemann, line 234: Nit:
```suggestion
	return quicvarint.Append(b, uint64(f.ID))
```

marten-seemann (commented):
Is there any way we can solve this without having a map of connections in the `Server`. That feels kind of heavy. We already have a few Go routines blocking on `AcceptStream`, maybe we can use one of those to do the work of sending the GOAWAY?

http3/frames.go
WeidiDeng: There is a check afterwards to see if data read matches the varint it encodes. This check is necessary.

marten-seemann: No, it is not. This code is just unnecessarily complicated. Use `quicvarint.Read`.

http3/server.go
WeidiDeng: The map is there to make sure all connection is closed before `CloseGracefully` returns. It's kept in a map because this way server can begin to send goaway frame this way instead of relying on `AcceptStream` to accept a stream and then write a goaway.

The alternative is to create a goroutine that listens for when the server is closed. But that's even heavier weight than the current method of tracking with map. I guess that's why stdlib uses maps to do the same.

http3/server.go
WeidiDeng: Is it fine to not close quic.Connection on the server side. Right now it's only closed on the server side when an error occurred during request handling.

http3/frames.go
WeidiDeng: A malformed goaway frame can have an arbitrary large length defined. And `quicvarint.Read` is used later.

marten-seemann: Just parse the two varints, then check that the length is consistent. Saves you 10 LOC and multiple allocs.

WeidiDeng: I used an `io.LimitedReader` and `bufio.Reader` to read from the frame body. Do you think it's ok now?

marten-seemann: Still too complicated.

1. Read varint (length)
2. Read varint (length)
3. Check that the bytes consumed by (2) match the value read in (1).

WeidiDeng: You have to wrap the reader with a `bufio.Reader` to be able to use Read varint. And to limit the number of bytes read a limited reader is needed, or else the stream may be corrupt due to more data read than needed.

WeidiDeng: Any feedbacks? @marten-seemann

marten-seemann: My previous comment still stands.

WeidiDeng: Show me some code snippets. I don't think I know how to think of this. Double read?

marten-seemann: Sure. I'll do this once the rest of the PR is in a shape where we could merge it.

http3/server.go
marten-seemann: Why don't you want to close it? That's the point of graceful *shutdown* after all, isn't it?

WeidiDeng: Sending goaway frame requires the server having some kinds of references to the quic connection. Closing it is not necessary as stdlib doesn't close either.

WeidiDeng: @marten-seemann  I'm thinking using a counter to track the number of accepted quic connections, and a context that will be canceled when shutdown is called. `context. AfterFunc` can be used to send goaway in this case. The server only knows about the number of the quic connections. What do you think of this method?

Also, perhaps I should implement server side graceful shutdown first, client side later. To keep changes small.

marten-seemann:
I refactored the tests in https://github.com/quic-go/quic-go/commit/e933c80ada58eba408722e912f0a9cf094f27ecf. Also, it looks like we should probably merge #4689 first. Does this make sense @WeidiDeng?

PR #4408, opened 2024-04-03
make CancelWrite a no-op after Close was called on the send stream

### LINKED ISSUE #4404: CancelWrite after Close should be a no-op
@marten-seemann Hey any update on this ? We are adding support for quic to Prysm and are running into this issue. It does appear this is also related to https://github.com/quic-go/quic-go/issues/4139 , but that appears to still be a RFC. 

Currently we ensure that all open streams are eventually reset so that they can be appropriately cleaned up. However with libp2p's Stream API and the necessity to support other multiplexers(yamux, mplex), resetting QUIC streams the same way unfortunately leads to data being lost with the remote peer unable to read the transmitted data when we initiate a reset. 

We could fix this by adding a special sleep for QUIC streams so that data is reliably sent out before we reset it, but would prefer to use a cleaner/more graceful solution.

_Originally posted by @nisdas in https://github.com/quic-go/quic-go/issues/3291#issuecomment-2031443539_

marten-seemann:
Fixes #4404.

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -381,7 +381,7 @@ func (s *sendStream) CancelWrite(errorCode StreamErrorCode) {
 // must be called after locking the mutex
 func (s *sendStream) cancelWriteImpl(errorCode qerr.StreamErrorCode, remote bool) {
 	s.mutex.Lock()
-	if s.cancelWriteErr != nil {
+	if s.cancelWriteErr != nil || s.finishedWriting {
 		s.mutex.Unlock()
 		return
 	}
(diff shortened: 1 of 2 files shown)


sukunrt (approved):
Are we dropping support for discarding written data on Reset? 

https://www.rfc-editor.org/rfc/rfc9000.html#section-3.1-9
>From any state that is one of "Ready", "Send", or "Data Sent", an application can signal that it wishes to abandon transmission of stream data.

marten-seemann:
Note: https://github.com/quic-go/quic-go/issues/4404#issuecomment-2041079139 calls into question if this PR is the right solution.

sukunrt (approved):

PR #4412, opened 2024-04-05
http3: sniff Content-Type when flushing the ResponseWriter

WeidiDeng:
When caddy log is enabled, it can be observed `Content-Type` sniffing no longer works for http3, but works for h1/2. This is caused because when logging is enabled, caddy wraps all response writers with `responseRecorder` which will call `WriteHeader` [first](https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/responsewriter.go#L169).

Stdlib will try set `Content-Type` header in the same logic, for [http1](https://github.com/golang/go/blob/master/src/net/http/server.go#L1439), `*chunkWriter` is wrapped by `*response` to write the actual data. [http2](https://github.com/golang/net/blob/master/http2/server.go#L2630) will set this header in the same way.

Because this header is set once, it won't be overwritten automatically


marten-seemann:
@WeidiDeng The large refactor for HTTP Datagram support (#4452) was merged. Do you want to update this PR?

http3/response_writer.go
```diff
@@ -102,24 +102,26 @@ func (w *responseWriter) WriteHeader(status int) {
...
-			// we shouldn't sniff the body.
-			hasTE := w.header.Get("Transfer-Encoding") != ""
-			hasCE := w.header.Get("Content-Encoding") != ""
-			if !hasCE && !haveType && !hasTE && len(p) > 0 {
-				w.header.Set("Content-Type", http.DetectContentType(p))
-			}
-		}
+		w.sniffContentType(p)
 		w.WriteHeader(http.StatusOK)
+		// 1xx status are not allowed to have a body, but they are temporary status
```
marten-seemann, line 124: This comment appears misplaced? You just wrote a 200 in the line before.

marten-seemann (commented):
This looks a lot cleaner now.

Do I understand correctly that the only thing that this PR changes is that the sniffing algorithm is now also run when flushing, whereas before it was only run on `Write`?

marten-seemann:
This PR is the only thing left for the v0.43 release.

http3/response_writer.go
WeidiDeng: I placed here by design. Before if a 1xx status is written, the content-type sniffing won't be run automatically because the status is considered unable to have a body and yet body is allowed anyway here. Though it looks strange here.

marten-seemann, line 124:
```suggestion
```

marten-seemann:
> Do I understand correctly that the only thing that this PR changes is that the sniffing algorithm is now also run when flushing, whereas before it was only run on `Write`?

@WeidiDeng Can you confirm this?

WeidiDeng:
@marten-seemann Yes, I confirm that's the change.

marten-seemann (approved):
Thank you @WeidiDeng!

PR #4419, opened 2024-04-08
fix documentation for CancelWrite after Close on the send stream

### LINKED ISSUE #4404: CancelWrite after Close should be a no-op

### LINKED ISSUE #4411: allow remote-triggered stream resets after closing a send stream
I think I understand why we made this decision back then. Take a look at the send stream states from RFC 9000:
image

It explicitly contains the transition `Data Sent` to `Reset Sent`. While it makes sense to have `CancelWrite` after `Close` a no-op (it's basically a misuse of the API, you might argue), a sender might receive a STOP_SENDING from the peer. In that case, it does make sense to send a RESET_STREAM and stop retransmitting the stream data.

_Originally posted by @marten-seemann in https://github.com/quic-go/quic-go/issues/4404#issuecomment-2035593544_

marten-seemann:
Fixes #4404. Closes #4411. Closes #4408. Closes #4413.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -124,7 +124,9 @@ type SendStream interface {
 	// CancelWrite aborts sending on this stream.
 	// Data already written, but not yet delivered to the peer is not guaranteed to be delivered reliably.
 	// Write will unblock immediately, and future calls to Write will fail.
-	// When called multiple times or after closing the stream it is a no-op.
+	// When called multiple times it is a no-op.
+	// When called after Close, it aborts delivery. Note that there is no guarantee if
+	// the peer will receive the FIN or the reset first.
 	CancelWrite(StreamErrorCode)
 	// The Context is canceled as soon as the write-side of the stream is closed.
 	// This happens when Close() or CancelWrite() is called, or when the peer


sukunrt (approved):

sukunrt:
We need #4408 too right? That handles the case where the stream close has been ACKed. In that case this `CancelWrite` should be a no-op

marten-seemann:
> We need #4408 too right? That handles the case where the stream close has been ACKed. In that case this `CancelWrite` should be a no-op

Debatable. Technically yes, but it seems to be such a corner case that is doesn't really matter.

PR #4421, opened 2024-04-09
http3: fix documentation

korzhao:


marten-seemann:
Thanks! This is already fixed in the new documentation page, which will launch in the next couple of weeks, so I'm not going to merge this PR.

PR #4430, opened 2024-04-13
make stream state transitions observable on SendStream and ReceiveStream

### LINKED ISSUE #4428: obtaining the stream state
HTTP Datagrams (#3522) require us to track the state of the underlying QUIC stream:
* It is not permitted to send DATAGRAMs for a stream whose send side is closed. We should reject calls to `SendDatagram` in our API.
* Conversely, when DATAGRAMs are received for a stream whose receive side is closed, they MUST be discarded.

My first idea how to implement this was tracking the state transitions by wrapping a `quic.Stream`. For example, we **know** that the send side is closed when `Closed` and `CancelWrite` are called. So far, so good. However, the send side can also be closed by receiving a STOP_SENDING frame. As a consumer of a `quic.Stream`, we'll only learn about this on the next call to `Write`.
But here's the problem: What if the application never writes again. Sure - it should not just walk away from the stream without either calling `Close` or `CancelWrite`, but there's no guarantee for that. At the QUIC layer, we'd garbage collect the stream, whereas on the HTTP/3 layer this stream would remain open indefinitely.

The same happens on the read side, if the application never calls `Read` after the RESET_STREAM has been received. Again, the application shouldn't do that (and risks leaking the stream if it does and no RESET_STREAM is received), but we also don't have a guarantee for that.

It would be nice if there was an API that could be used to obtain the stream state, avoid any synchronization issues. It would be even nicer if no `quic.Config` entry was required, since our HTTP/3 API allows using a QUIC connection.

marten-seemann:
Fixes #4428.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -104,10 +104,10 @@ type ReceiveStream interface {
 	// Even if Read times out, it may return n > 0, indicating that some data was successfully read.
 	// A zero value for t means Read will not time out.
 	SetReadDeadline(t time.Time) error
-	// OnStateChange sets a callback to be called when the stream transitions between stream states.
+	// OnStateTransition sets a callback to be called when the stream transitions between stream states.
 	// It is possible to set the callback after a transition has happened.
 	// In that case, it is immediately called for that transition.
-	OnStateChange(func(StreamState))
+	OnStateTransition(func(StreamTransition))
 }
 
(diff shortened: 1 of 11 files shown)


marten-seemann:
We'll go with #4445 and #4460 instead.

PR #4434, opened 2024-04-14
Http3 datagram

Jerry0666:
I add GetReqId, to assign an id in the http request header, and before client do request return, we can use this id to get the stream early, so we can send the http3 datagram before getting the http response.
This is the sample client code:
```
package main

import (
	"fmt"
	"net/http"

	"crypto/tls"

	"github.com/quic-go/quic-go"
)

func main() {
	roundTripper := &http3.RoundTripper{
		TLSClientConfig: &tls.Config{
			InsecureSkipVerify: true,
		},
		QuicConfig: &quic.Config{
			EnableDatagrams: true,
		},
		EnableDatagrams: true,
		ReqId:           0,
	}
	client := &http.Client{
		Transport: roundTripper,
	}
	defer roundTripper.Close()

	req, _ := http.NewRequest(http.MethodConnect, "https://127.0.0.1:30000/", nil)
	req.Proto = "connect-ip"
	roundTripper.AssignReqId(req)
	go client.Do(req)
	roundTripper.InitialMap()
	reqId, _ := http3.GetReqId(req)
	str := roundTripper.GetReqStream(reqId)
	fmt.Println("get the http3 stream")
	if str == nil {
		fmt.Println("receive stream is nil")
	}
	d, _ := str.Datagrammer()
	d.SendMessage([]byte("http3 datagram test..."))
	for {

	}

}
```


marten-seemann:
See #3522

PR #4436, opened 2024-04-15
Fix SendDatagram

nekohasekai:
SendDatagram returned wrong PeerMaxDatagramFrameSize value

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -2358,9 +2358,10 @@ func (s *connection) SendDatagram(p []byte) error {
 	}
 
 	f := &wire.DatagramFrame{DataLenPresent: true}
-	if protocol.ByteCount(len(p)) > f.MaxDataLen(s.peerParams.MaxDatagramFrameSize, s.version) {
+	maxDataLen := f.MaxDataLen(s.peerParams.MaxDatagramFrameSize, s.version)
+	if protocol.ByteCount(len(p)) > maxDataLen {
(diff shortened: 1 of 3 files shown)


nekohasekai:
I found that there is already a test there

https://github.com/quic-go/quic-go/blob/7b9d21fbe6b96aa4942703df89fb93acc1ea77a9/integrationtests/self/datagram_test.go#L66.

But MatchError always returns true because DatagramTooLargeError implements Is(error)

```go
func (matcher *MatchErrorMatcher) Match(actual any) (success bool, err error) {
                ...

		// first try the built-in errors.Is
		if errors.Is(actualErr, expected.(error)) {
			return true, nil
		}
```

https://github.com/quic-go/quic-go/blob/7b9d21fbe6b96aa4942703df89fb93acc1ea77a9/errors.go#L70-L73

nekohasekai:
It almost passed. There is a race, but it doesn't seem to be related to my changes.

marten-seemann (changes requested):
I don’t understand your changes regarding the errors. I also don’t see any test that’s testing your change.

marten-seemann:
That might be a separate issue. Please don't change the error implementation in this PR.

marten-seemann:
@nekohasekai Are you still working on this PR? Would be nice to get the fix into the v0.43 release.

PR #4445, opened 2024-04-18
delay completion of the send stream until the reset error was delivered

marten-seemann:
Part of #4428.

The application now needs to _somehow_ retrieve the error generated by receiving the STOP_SENDING frame (or, alternatively, `Close` or `CancelWrite`).

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -363,9 +363,24 @@ func (s *sendStream) getDataForWriting(f *wire.StreamFrame, maxBytes protocol.By
 }
 
 func (s *sendStream) isNewlyCompleted() bool {
-	completed := (s.finSent || (s.cancelWriteErr != nil && s.cancellationFlagged)) &&
-		s.numOutstandingFrames == 0 && len(s.retransmissionQueue) == 0
-	if completed && !s.completed {
+	if s.completed {
+		return false
+	}
+	// We need to keep the stream around until all frames have been sent and acknowledged.
+	if s.numOutstandingFrames > 0 || len(s.retransmissionQueue) > 0 {
+		return false
+	}
+	// The stream is completed if we sent the FIN.
+	if s.finSent {
+		s.completed = true
+		return true
+	}
+	// The stream is also completed if:
+	// 1. the application called CancelWrite, or
+	// 2. we received a STOP_SENDING, and
+	// 		* the application consumed the error via Write, or
+	//		* the application called CLsoe
+	if s.cancelWriteErr != nil && (s.cancellationFlagged || s.finishedWriting) {
 		s.completed = true
 		return true
 	}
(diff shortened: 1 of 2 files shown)


sukunrt:
I understand this is to avoid sending http datagrams *after* the stream has received STOP_SENDING. 

Can you elaborate on 
>The application now needs to somehow retrieve the error generated by receiving the STOP_SENDING frame

Is the expectation that users do an empty write on the stream to know if it errors? 

Is it better to provide a method to retrieve the state of the stream Or just an `IsCompleted` method on the stream? This should work on the send side since we want to know whether the stream is open when we are sending the http datagram.

sukunrt:
How does delaying garbage collection help here? 
If the stream has been closed for writing, it's alright if we've removed all references to it from the connection. Writes post garbage collection will error anyway.

marten-seemann:
> How does delaying garbage collection help here? If the stream has been closed for writing, it's alright if we've removed all references to it from the connection. Writes post garbage collection will error anyway.

That's correct. On the QUIC layer, we don't need the map entry anymore. However, the upper layer has no way of discovering that QUIC garbage collected the stream, so it's impossible to keep a map at the application layer (as we need to do in HTTP/3 for datagram support) that doesn't leak (iff the error code is never read), as described in #4428.

> Can you elaborate on
>
> > The application now needs to somehow retrieve the error generated by receiving the STOP_SENDING frame
>
> Is the expectation that users do an empty write on the stream to know if it errors?

No, that would be an awkward API indeed. For the send stream, the expectation is that you either:
1. repeatedly call `Write` until it errors, or
2. repeatedly call `Write` and call `Close` once you've written everything, or
3. at any point call `CancelWrite` if you want to abort sending

All the cases listed above lead to garbage collection of the stream.

---

The one thing you must not do is the following: Call `Write` (once or multiple times) with a `nil` error every time, and then just walk away.

Before this PR, in this case the stream would have been garbage collected as soon as a STOP_SENDING was received. With this PR, the PR won't be garbage collected (until you call `Write` again, or `Close`, or `CancelWrite`).

--- 

Does this make sense?

send_stream.go
```diff
@@ -379,14 +379,20 @@ func (s *sendStream) Close() error {
 		return nil
 	}
 	s.finishedWriting = true
-	if s.cancelWriteErr != nil {
-		s.mutex.Unlock()
-		return fmt.Errorf("close called for canceled stream %d", s.streamID)
-	}
-	s.ctxCancel(nil)
+	s.cancellationFlagged = true
+	cancelWriteErr := s.cancelWriteErr
```
sukunrt, line 383: I think it's better to only set `s.cancellationFlagged = true` if cancelWriteErr was not true. I don't think this leads to a bug anywhere. 
Something like: 

```
cancelWriteErr := s.cancelWriteErr
if cancelWriteErr != nil {
    s.cancellationFlagged = true
}
```

sukunrt (approved):

send_stream.go
marten-seemann: Done.

marten-seemann:
Thanks to @sukun for the super helpful review on this PR!

Issue #4396, opened 2024-03-28
Dialing fails with an INTERNAL_ERROR when ECN is enabled


arashpayan:
I found an issue with ECN on at least one configuration of Linux:

**Configuration**
Distribution: CentOS Linux
Version: 7
Kernel: 3.10.0-1160.el7.x86_64
CPU: Intel Xeon CPU E5-2683 v4 @ 2.10GHz

**Reproduction**
``` go
conn, err := quic.DialAddr(context.Background(), "some.address:port", &tls.Config{}, &quic.Config{})
```

**Expected**
`conn` is valid and `err` should be nil.

**Actual**
`err` is `INTERNAL_ERROR (local): write udp [::]:<SRCPORT>->XXX.XXX.XXX.XXX:DSTPORT: sendmsg: invalid argument`

If the program is run with `QUIC_GO_DISABLE_ECN=true` the dial succeeds.

**Notes**
Here's a qlog in case it helps: https://ara.sh/private/4250200fe9d3c5d44d3f60cf346c9c82bd0f53_client.qlog

I'm available to gather other info and try out other fixes as you may need.

marten-seemann:
We've had some reports of ECN not working on certain systems: https://github.com/quic-go/quic-go/issues/4178

> I'm available to gather other info and try out other fixes as you may need.

This would be greatly appreciated!!! I was unable to reproduce these failures, so if you could help us figure out a way to make ECN work on every system, that would be amazing!

First suspect would be the size of the cmsg, see https://github.com/AdguardTeam/AdGuardHome/issues/6422#issuecomment-1862093806. Could you change it and see if that fixes the problem?

arashpayan:
That was it! Changing `ecnIPv4DataLen` to 4 lets it connect. It's very odd that the man page says it's a single byte.

Would you like me to open a PR with the change?

marten-seemann:
We're in a bit of a conundrum here. We changed it to the 1 byte value in https://github.com/quic-go/quic-go/pull/4127, based on the discussion in https://github.com/AdguardTeam/AdGuardHome/issues/6335. Apparently there's a (different?) set of Linux machines that fails if we don't use the 1 byte value.

What we could do (and what I would really, really hate!) is build a fallback: first try one length, and then switch to the other length. I'm not ready to build such an ugly mechanism yet.

I'm wondering if there's any way we can find out which size to use. Any ideas?

Also looping in @ainar-g, do you have any opinion here?

[...]

PR #4447, opened 2024-04-19
fix UDP_SEGMENT check on some linux kernels

### LINKED ISSUE #4446: isGSOSupported incorrect value on some Android devices
I was debugging issues with slow (or completely stuck) streams on some Android devices that we are experiencing with quic-go in production on [anytype](http://github.com/anyproto/anytype-heart/pulls). After disabling UDP optimizations by using basicConn, the problems were resolved, so I began investigating different syscalls in `sys_conn_oob.go`.

`isGSOSupported` returns true on the problematic device, so I tried to disable it by setting the OS environment variable `QUIC_GO_DISABLE_GSO=0`, and it helped. According to the [Linux manual pages](https://man7.org/linux/man-pages/man7/udp.7.html), the `UDP_SEGMENT` socket option may return 0 if it is disabled. Currently, in the Linux implementation, [only the error is checked](https://github.com/quic-go/quic-go/blob/master/sys_conn_helper_linux.go#L67). This check only verifies whether the Linux kernel recognizes this socket option (kernels >=4.18 should recognize it). The Android device I tested has a kernel version of `4.14.180-perf+`, but it appears to recognize this socket option and returns 0.

requilence:
Check the `UDP_SEGMENT` socket option for 0 value(according to https://man7.org/linux/man-pages/man7/udp.7.html) inside `isGSOSupported` func on linux.

Fixes #4446

### DIFF
--- a/sys_conn_helper_linux.go
+++ b/sys_conn_helper_linux.go
@@ -62,13 +62,18 @@ func isGSOSupported(conn syscall.RawConn) bool {
 	if err == nil && disabled {
 		return false
 	}
-	var serr error
+	var (
+		serr error
+		sval int
+	)
 	if err := conn.Control(func(fd uintptr) {
-		_, serr = unix.GetsockoptInt(int(fd), unix.IPPROTO_UDP, unix.UDP_SEGMENT)
+		sval, serr = unix.GetsockoptInt(int(fd), unix.IPPROTO_UDP, unix.UDP_SEGMENT)
 	}); err != nil {
 		return false
 	}
-	return serr == nil
+	// 0 value means GSO is disabled
+	// according to https://man7.org/linux/man-pages/man7/udp.7.html
+	return serr == nil && sval != 0
 }
 
 func appendUDPSegmentSizeMsg(b []byte, size uint16) []byte {


marten-seemann:
I see the tests on Ubuntu failing here. I thought that these machines support GSOs though. Potential false positive?

requilence:
@marten-seemann Hm, yes, I think `0` may mean GSO is supported but disabled for the current connection. I think we need research more. So, as far as I understand the original idea was:
- check if system support `UDP_SEGMENT` socket option at all by checking error on `unix.GetsockoptInt`
- set it for the specific packet by appending `syscall.IPPROTO_UDP` msg
- in case we get error `unix.EIO` during `writePacket`. Which, according to the comment suppose to check, whatever tx checksums are enabled

Instead of checking whatever `val != 0` I tried this approach:
- Before getting socketOpt I set it. I put 1500 (because I saw the gso_size to be beyond 1500), but other values >0 give the same result.
```	
if err := conn.Control(func(fd uintptr) {
		gerr := unix.SetsockoptInt(int(fd), unix.IPPROTO_UDP, unix.UDP_SEGMENT, 1500)
		fmt.Printf("SetsockoptInt gsp: %v\n", gerr)
		sval, serr = unix.GetsockoptInt(int(fd), unix.IPPROTO_UDP, unix.UDP_SEGMENT)
}); err != nil {
		return false
}
```
- I got nil error on set/get and got the same 1500 on Get
- Now for all `sconn.writePacket` calls I get this error: `Write: write udp [::]:43377->192.168.184.136:7000: sendmsg: invalid argument`. Before manually setting the connection's `unix.IPPROTO_UDP` it return nil error.

requilence:
BTW, according to [this linux kernel commit](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-4.14.y&id=2e8de8576343ab540856082916bfb84d17288b08) and [v4.18 release notes](https://kernelnewbies.org/Linux_4.18), control message was added in v4.18.1. So it's probably a bug that setting it in 4.14.180 return no error.  Probably it may be a good idea to add a limitation based on linux kernel version >=4.18

marten-seemann:
> BTW, according to [this linux kernel commit](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-4.14.y&id=2e8de8576343ab540856082916bfb84d17288b08) and [v4.18 release notes](https://kernelnewbies.org/Linux_4.18), `UDP_SEGMENT` control message was added just in v4.18.1. So it's probably a bug that setting it in 4.14.180 return no error. Probably it may be a good idea to add a limitation based on linux kernel version >=4.18

Interesting. Do I understand correctly that we'd fix the problem by disabling GSO on v4.18, or is anything else required as well?

I've tried to avoid checking the kernel version, but it seems like there's a problem with ECN in older kernel versions as well (see https://github.com/quic-go/quic-go/issues/4396), so maybe it makes sense to implement a check like that.

requilence:
correct, need to disable GSO one linux kernel version less than 4.18. 
I was trying to find a way to detect the support in a more robust way, but looks like it is not achievable.

PR #4448, opened 2024-04-19
http3: reuse clients on RoundTripOpt context canceled

GeorgeMac:
👋 Unsure if this is really a fix or not, but wanted to share for your consideration.

I found that when requests were getting context canceled for whatever reason (e.g. client going away) that the roundtripper was abandoning the connection and dialing again. This PR adds a check to observe the context canceled error and not remove the client in this particular instance. So that the cached client can be used in a subsequent round trip.

I appreciate context canceled can be a bit of a blunt instrument, so I am unsure if this is going to work in the broader state machine, but thought I would open it up for your thoughts.

p.s. love the project. I am attempting to build a reverse tunnel with, which is where I observed this behaviour:
https://github.com/flipt-io/reverst/. In my situation I am proxying a HTTP request from one server onto the http3 client. When e.g. my inbound request is canceled (e.g. the client goes away) this causes the connect to be dropped, which is quite a pain.

Update: I have included this in the project mentioned above and now the connection reuse is much better. I was getting constant reconnections without this change.

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -165,7 +165,10 @@ func (r *RoundTripper) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.
 	defer cl.useCount.Add(-1)
 	rsp, err := cl.rt.RoundTripOpt(req, opt)
 	if err != nil {
-		r.removeClient(hostname)
+		if !errors.Is(err, context.Canceled) {
+			r.removeClient(hostname)
+		}
+
 		if isReused {
 			if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
 				return r.RoundTripOpt(req, opt)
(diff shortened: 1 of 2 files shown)


http3/roundtrip.go
```diff
@@ -165,7 +165,10 @@ func (r *RoundTripper) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.
 	defer cl.useCount.Add(-1)
 	rsp, err := cl.rt.RoundTripOpt(req, opt)
 	if err != nil {
-		r.removeClient(hostname)
+		if !errors.Is(err, context.Canceled) {
```
marten-seemann, line 168: Add a comment here why you're not removing the connection for context cancelations?

marten-seemann (approved):
Thank you for your PR and the thoughtful description.

The change lgtm, we want to remove connections that have already errored, not connections where the application decided to cancel (possibly immediately after dialing).

GeorgeMac:
Thank you @marten-seemann 🙏

marten-seemann (approved):
Thanks for adding the explanation!

PR #4449, opened 2024-04-19
http3: use a log/slog Logger for logging

### LINKED ISSUE #3548: http3: use the http.Server.ErrorLog if set
In a previous version, `v0.28` if I'm not mistaken, there was a possibility to set `http3.Server.Server.ErrorLog = log.New (...)`.  However now as the `http3.Server.Server` fields have been merged into `http3.Server`, the `ErrorLog` is missing.

There seems to be a private `logger util.Logger` field that can't be overriden, so perhaps making that public?

### LINKED ISSUE #4366: http3 server silently panics
With a std HTTP server I would expect the following code to print some indication that the goroutine handling the request panicked:

```go
package main

import (
        "flag"
        "net/http"
        "log"

)

func main() {
        cert := flag.String("crt", "", "Cert")
        key := flag.String("key", "", "Key")
        flag.Parse()

        http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
                panic("Hi there")
        })

        log.Fatal(http3.ListenAndServeQUIC(":443", *cert, *key, nil))
}
```

But http3 appears to just silently eat the panic. If this behavior is intentional, is there a way to surface panics so I am aware they're happening?

marten-seemann:
Fixes #4366. Closes #3548.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"sync"
 	"time"
 
 	"github.com/quic-go/quic-go"
 	"github.com/quic-go/quic-go/internal/protocol"
-	"github.com/quic-go/quic-go/internal/utils"
 	"github.com/quic-go/quic-go/quicvarint"
 
 	"github.com/quic-go/qpack"
@@ -61,11 +61,12 @@ type SingleDestinationRoundTripper struct {
 	// However, if the user explicitly requested gzip it is not automatically uncompressed.
 	DisableCompression bool
 
+	Logger *slog.Logger
+
 	initOnce      sync.Once
 	hconn         *connection
 	requestWriter *requestWriter
 	decoder       *qpack.Decoder
-	logger        utils.Logger
 }
 
 func (c *SingleDestinationRoundTripper) Start() Connection {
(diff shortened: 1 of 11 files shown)


http3/client.go
```diff
@@ -74,14 +75,15 @@ func (c *SingleDestinationRoundTripper) Start() Connection {
...
 	c.decoder = qpack.NewDecoder(func(hf qpack.HeaderField) {})
-	c.hconn = newConnection(c.Connection, c.EnableDatagrams, c.UniStreamHijacker, protocol.PerspectiveClient, c.logger)
+	c.hconn = newConnection(c.Connection, c.EnableDatagrams, c.UniStreamHijacker, protocol.PerspectiveClient, c.Logger)
 	// send the SETTINGs frame, using 0-RTT data, if possible
 	go func() {
 		if err := c.setupConn(c.Connection); err != nil {
-			c.logger.Debugf("Setting up connection failed: %s", err)
+			if c.Logger != nil {
+				c.Logger.Debug("setting up connection failed", "error", err)
+			}
```
rfwatson, line 86: Question: should we always fallback to `slog.Default()` if no logger is explicitly set?

rfwatson (approved):
Thanks for updating this :pray:

http3/client.go
marten-seemann: I'm not sure. I don't want to spam peoples' logs with debug messages, so I'm inclined to leave it as is far now. That said, I don't think we have a good / consistent logging philosophy at the moment, so I'd be happy to discuss and revise this decision in the future.

PR #4455, opened 2024-04-23
Increase send/receive buffers to 7MiB

### LINKED ISSUE #4426: Increase buffer sizes
This topic came up in the Syncthing forum: https://forum.syncthing.net/t/system-is-slow-on-windows-look-here/21925

I'm unsure if users really need to modify the registry or if the socket options are enough to tune the buffer sizes. After briefly skimming the code, I couldn't spot where we do this for Windows?

bt90:
Fixes #4426

### DIFF
--- a/internal/protocol/params.go
+++ b/internal/protocol/params.go
@@ -3,10 +3,10 @@ package protocol
 import "time"
 
 // DesiredReceiveBufferSize is the kernel UDP receive buffer size that we'd like to use.
-const DesiredReceiveBufferSize = (1 << 20) * 2 // 2 MB
+const DesiredReceiveBufferSize = (1 << 20) * 7 // 7 MB
 
 // DesiredSendBufferSize is the kernel UDP send buffer size that we'd like to use.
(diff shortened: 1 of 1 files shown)


marten-seemann:
> We also need to update the relevant section of the [wiki](https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes).
> 
> ## non-BSD
> ```
> sysctl -w net.core.rmem_max=7340032
> sysctl -w net.core.wmem_max=7340032
> ```
> 
> ## BSD
> ```
> sysctl -w kern.ipc.maxsockbuf=8441037
> ```
> 
> ```
> kern.ipc.maxsockbuf==8441037
> ```

Thanks. Updated on the new (unreleased) docs page.

marten-seemann (approved):

bt90:
@marten-seemann I just had a look at the updated wiki entry. Is there a reason why you use `7500000` instead of `7340032` for non-BSD? The defaults and most examples I could find online used values based on KiB/MiB instead of KB/MB.
No idea if this has any side effects or is just an aesthetic thing, though.

PR #4460, opened 2024-04-25
delay completion of the receive stream until the reset error was read

marten-seemann:
The other part of #4428. Needed for #3522.

With this change, a receive stream is garbage collected only once the reset error was actually read (in `Read`), or if `CancelRead` was called.

### DIFF
--- a/receive_stream.go
+++ b/receive_stream.go
@@ -41,9 +41,10 @@ type receiveStream struct {
 	// Note that for local cancellations, this doesn't necessarily mean that we know the final offset yet.
 	errorRead           bool
 	completed           bool // set once we've called streamSender.onStreamCompleted
+	cancelledRemotely   bool
+	cancelledLocally    bool
+	cancelErr           *StreamError
 	closeForShutdownErr error
-	cancelReadErr       error
-	resetRemotelyErr    *StreamError
 
 	readChan chan struct{}
 	readOnce chan struct{} // cap: 1, to protect against concurrent use of Read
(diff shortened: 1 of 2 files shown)


sukunrt:
Same comment as https://github.com/quic-go/quic-go/pull/4445#issuecomment-2076845424

Why does the QUIC layer need to keep a reference to the closed stream? Read after completion will error.

receive_stream.go
```diff
@@ -83,27 +86,45 @@ func (s *receiveStream) Read(p []byte) (int, error) {
...
 }
 
-func (s *receiveStream) readImpl(p []byte) (bool /*stream completed */, int, error) {
-	if s.finRead {
-		return false, 0, io.EOF
+func (s *receiveStream) isNewlyCompleted() bool {
+	// We're done with the stream once:
+	// 1. The application has consumed the io.EOF or the cancellation error
+	// 2. We know the final offset (for flow control accounting)
+	isNewlyCompleted := !s.completed && s.errorRead && s.finalOffset != protocol.MaxByteCount
```
sukunrt, line 109: We only want the error to be read if we didn't call CancelRead, right?

receive_stream.go
```diff
@@ -202,7 +222,8 @@ func (s *receiveStream) dequeueNextFrame() {
 
 func (s *receiveStream) CancelRead(errorCode StreamErrorCode) {
 	s.mutex.Lock()
-	completed := s.cancelReadImpl(errorCode)
+	s.cancelReadImpl(errorCode)
+	completed := s.isNewlyCompleted()
```
sukunrt, line 226: Does this work when CancelRead happens *JUST* after receiving a reset from remote?

sukunrt (approved):

Issue #3663, opened 2023-01-03
Tracking Issue: reduce allocations during the handshake


marten-seemann:
Using the benchmark test (#3662) we can now see how many allocations are happening during the handshake... and it's a lot.
This also shows up in allocation profiles from libp2p nodes.

```
❯ go test -run adfsdfas -bench BenchmarkHandsh -v
goos: darwin
goarch: arm64
pkg: github.com/lucas-clemente/quic-go/integrationtests/self
BenchmarkHandshake
BenchmarkHandshake-10               1598            743054 ns/op          217400 B/op       3024 allocs/op
PASS
```

marten-seemann:
Some standard library improvements:
* https://go-review.googlesource.com/c/crypto/+/480055
* https://go-review.googlesource.com/c/go/+/480095

PR #4481, opened 2024-05-03
wire: refactor header parsing to use quicvarint.Parse

marten-seemann:
Part of #3663.

Now that we have a `quicvarint.Parse` (#4475), we can use that when parsing Long Header packets:

```
name                           old time/op    new time/op    delta
ParseInitial/without_token-16     114ns ± 4%      43ns ± 2%  -62.60%  (p=0.000 n=10+10)
ParseInitial/with_token-16        121ns ± 2%      54ns ± 1%  -55.22%  (p=0.000 n=10+8)
ParseRetry-16                     120ns ± 2%      53ns ± 2%  -55.76%  (p=0.000 n=10+10)

name                           old alloc/op   new alloc/op   delta
ParseInitial/without_token-16      192B ± 0%       96B ± 0%  -50.00%  (p=0.000 n=10+10)
ParseInitial/with_token-16         224B ± 0%      128B ± 0%  -42.86%  (p=0.000 n=10+10)
ParseRetry-16                      240B ± 0%      144B ± 0%  -40.00%  (p=0.000 n=10+10)

name                           old allocs/op  new allocs/op  delta
ParseInitial/without_token-16      4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=10+10)
ParseInitial/with_token-16         5.00 ± 0%      2.00 ± 0%  -60.00%  (p=0.000 n=10+10)
ParseRetry-16                      5.00 ± 0%      2.00 ± 0%  -60.00%  (p=0.000 n=10+10)
```

### DIFF
--- a/internal/wire/header.go
+++ b/internal/wire/header.go
@@ -171,48 +171,44 @@ func parseHeader(b []byte) (*Header, error) {
 
 	h := &Header{typeByte: typeByte}
 	l, err := h.parseLongHeader(b[1:])
-	h.parsedLen = l + 1
+	h.parsedLen = protocol.ByteCount(l) + 1
 	return h, err
 }
 
-func (h *Header) parseLongHeader(b []byte) (l protocol.ByteCount, err error) {
+func (h *Header) parseLongHeader(b []byte) (int, error) {
+	startLen := len(b)
 	if len(b) < 5 {
 		return 0, io.EOF
 	}
 	h.Version = protocol.Version(binary.BigEndian.Uint32(b[:4]))
-	l = 4
 	if h.Version != 0 && h.typeByte&0x40 == 0 {
-		return l, errors.New("not a QUIC packet")
+		return startLen - len(b), errors.New("not a QUIC packet")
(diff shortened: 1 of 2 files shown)


internal/wire/header.go
```diff
@@ -237,38 +240,39 @@ func (h *Header) parseLongHeader(b *bytes.Reader) error {
...
-	pl, err := quicvarint.Read(b)
+	pl, n, err := quicvarint.Parse(b)
+	l += protocol.ByteCount(n)
 	if err != nil {
-		return err
+		return 0, err
 	}
 	h.Length = protocol.ByteCount(pl)
-	return nil
+	return l, nil
```
sukunrt, line 275: can we do `startLen - len(b)` here instead of incrementing l for every field?

sukunrt (approved):

internal/wire/header.go
marten-seemann: Good point. So much cleaner!

PR #4484, opened 2024-05-04
wire: use quicvarint.Parse when parsing frames

marten-seemann:
```
name                  old time/op    new time/op    delta
ParseStreamAndACK-16     290ns ± 6%     244ns ± 4%  -15.80%  (p=0.000 n=9+10)
ParseOtherFrames-16      217ns ± 3%     171ns ± 3%  -21.43%  (p=0.000 n=9+10)

name                  old alloc/op   new alloc/op   delta
ParseStreamAndACK-16    1.58kB ± 0%    1.58kB ± 0%     ~     (all equal)
ParseOtherFrames-16       224B ± 0%      224B ± 0%     ~     (all equal)

name                  old allocs/op  new allocs/op  delta
ParseStreamAndACK-16      2.00 ± 0%      2.00 ± 0%     ~     (all equal)
ParseOtherFrames-16       6.00 ± 0%      6.00 ± 0%     ~     (all equal)
```

### DIFF
--- a/internal/wire/connection_close_frame.go
+++ b/internal/wire/connection_close_frame.go
@@ -1,7 +1,6 @@
 package wire
 
 import (
-	"bytes"
 	"io"
 
 	"github.com/quic-go/quic-go/internal/protocol"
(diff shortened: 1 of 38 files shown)


marten-seemann:
Good thing we enabled ClusterFuzzLite PR fuzzing. It caught a pretty serious bug that would otherwise have gone unnoticed.

internal/wire/connection_close_frame.go
```diff
@@ -16,40 +15,38 @@ type ConnectionCloseFrame struct {
...
 
 	reasonPhrase := make([]byte, reasonPhraseLen)
-	if _, err := io.ReadFull(r, reasonPhrase); err != nil {
-		// this should never happen, since we already checked the reasonPhraseLen earlier
-		return nil, err
-	}
+	copy(reasonPhrase, b)
 	f.ReasonPhrase = string(reasonPhrase)
-	return f, nil
+	return f, startLen - len(b) + int(reasonPhraseLen), nil
```
sukunrt, line 49: Do we need to move b ahead here? `b[reasonPhraseLen:]` io.ReadFull would, right?

marten-seemann: It's a shortcut. `b` is only used locally in this function, so any change we make to it after reading the reason phrase (which is the last field of the frame) doesn't have any effect. That's why in the last line we do `startLen - len(b) + int(reasonPhraseLen)` instead of `startLen - len(b)`.

sukunrt (approved):

PR #4489, opened 2024-05-05
http3: use the stream context to detect when the send side is closed

marten-seemann:


http3/datagram.go
```diff
@@ -85,9 +85,9 @@ start:
 		d.mx.Unlock()
 		return data, nil
 	}
-	if d.receiveErr != nil {
+	if receiveErr := d.receiveErr; receiveErr != nil {
 		d.mx.Unlock()
-		return nil, d.receiveErr
+		return nil, receiveErr
```
marten-seemann, line 90: This fixes a race condition.

PR #4504, opened 2024-05-10
utils: speed up timer by embedding the time.Timer

marten-seemann:
Note a huge speedup, but we're resetting timers _a lot_ since we're using it to pace packets.
```
name      old time/op  new time/op  delta
Timer-16  77.9ns ± 4%  75.8ns ± 3%  -2.59%  (p=0.000 n=50+47)
```

### DIFF
--- a/internal/utils/timer.go
+++ b/internal/utils/timer.go
@@ -7,14 +7,14 @@ import (
 
 // A Timer wrapper that behaves correctly when resetting
 type Timer struct {
-	t        *time.Timer
+	t        time.Timer
 	read     bool
 	deadline time.Time
 }
 
 // NewTimer creates a new timer that is not set
 func NewTimer() *Timer {
-	return &Timer{t: time.NewTimer(time.Duration(math.MaxInt64))}
+	return &Timer{t: *time.NewTimer(time.Duration(math.MaxInt64))}
 }
 
 // Chan returns the channel of the wrapped timer
--- a/internal/utils/timer_test.go
+++ b/internal/utils/timer_test.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"testing"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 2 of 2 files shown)


marten-seemann:
Oops, looks like CI really hates this change. But why?

onee-only:
Seems like `(*time.Timer).Reset` does not work for timer as value for some reason.

Since the timer is initialized with `math.MaxInt64` duration, and is not reset due to the problem above,
the tests are being timed out waiting for receiving from the timer's channel.

I've tested behavior with code snippet below.

using timer as pointer:
```go
func main() {
	now := time.Now()
	t := time.NewTimer(time.Second)
	t.Reset(time.Second * 2)

	<-t.C
	log.Println(time.Since(now))
	// 2024/05/19 17:09:14 2.000662625s
}
```

using timer as value:
```go
func main() {
	now := time.Now()

	t := *time.NewTimer(time.Second)
	t.Reset(time.Second * 2)

	<-t.C
	log.Println(time.Since(now))
	// 2024/05/19 17:07:54 1.001052084s
}
```

I'm not sure if there is a workaround.

sukunrt:
Can you try embedding a pointer to the timer?

marten-seemann:
> Can you try embedding a pointer to the timer?

That's what we currently have, or am I misunderstanding what you're suggesting?

sukunrt:
You aren't. Embedding the timer is what we have currently. I think this is what's happening. 

You cannot copy a time.Timer{}. Currently, we create a new timer using `time.NewTimer` which returns a pointer to the created timer. Now we dereference this pointer and copy the underlying timer value into our struct. This also copies the `time.Timer` struct's underlying `runtimeTimer`. This copying is illegal. The way runtime tracks these timers is by their address which it stores when creating the timer with `time.NewTimer`, the address of our struct is different because we copied this.  

Now when we call `timer.Reset` on the copied timer the implementation goes through the reset steps, but this new value will never be used because the runtime isn't tracking this copied timer at all. Specifically this slice doesn't have the copied timer: 
https://github.com/golang/go/blob/go1.22.3/src/runtime/proc.go#L3974

marten-seemann:
@sukunrt Thanks for researching this! Your explanation makes sense.

PR #4514, opened 2024-05-14
initialize the MTU discoverer when processing the transport parameters

marten-seemann:
On the client side, we always use the configured packet size. This comes with the risk of failing the handshake if the path doesn't support this MTU. If the server sends a max_udp_payload_size that's smaller than this size, we can safely ignore this: Obviously, the server still processed the (fully padded) Initial packet, despite claiming that it wouldn't do so.

On the server side, there's no downside to using 1200 bytes until we received the client's transport parameters:
* If the first packet didn't contain the entire ClientHello, all we can do is ACK that packet. We don't need a lot of bytes for that.
* If it did, we will have processed the transport parameters and initialized the MTU discoverer.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -153,7 +153,7 @@ type connection struct {
 	unpacker      unpacker
 	frameParser   wire.FrameParser
 	packer        packer
-	mtuDiscoverer mtuDiscoverer // initialized when the handshake completes
+	mtuDiscoverer mtuDiscoverer // initialized when the transport parameters are received
 
 	maxPayloadSizeEstimate atomic.Uint32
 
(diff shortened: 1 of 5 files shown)


connection.go
```diff
@@ -153,7 +153,7 @@ type connection struct {
 	unpacker      unpacker
 	frameParser   wire.FrameParser
 	packer        packer
-	mtuDiscoverer mtuDiscoverer // initialized when the handshake completes
+	mtuDiscoverer mtuDiscoverer // initialized when the transport parameters are received
```
marten-seemann, line 156: Note that the original comment here was incorrect: We used to initialize the `mtuDiscoverer` with the connection, not on handshake completion.

PR #4522, opened 2024-05-14
http3: rename ListenAndServe to ListenAndServeTLS

### LINKED ISSUE #4521: The method `ListenAndServe` in package `http3` is not compatible with `http`
The method `ListenAndServe` in package `http3` is not compatible with the method in built-in package `http`
https://github.com/quic-go/quic-go/blob/2557fb98b2423ab1f0ae525509c68b51510422e0/http3/server.go#L673

The definitions in `http` are
```go
func ListenAndServe(addr string, handler Handler) error
func ListenAndServeTLS(addr, certFile, keyFile string, handler Handler) error
```

Maybe we should rename it to `ListenAndServeTLS` ?

ETZhangSX:
Close #4521

### DIFF
--- a/example/main.go
+++ b/example/main.go
@@ -163,7 +163,7 @@ func main() {
 		go func() {
 			var err error
 			if *tcp {
-				err = http3.ListenAndServe(bCap, certFile, keyFile, handler)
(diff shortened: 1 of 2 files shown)


marten-seemann (approved):
Thanks!

PR #4523, opened 2024-05-14
fix(http3): handle streamStateSendAndReceiveClosed in onStreamStateChange

### LINKED ISSUE #4513: http3: streams leaked in outgoingStreamsMap
I was just stress testing our little reverse proxy built on quic-go and found that memory increased linearly with requests.
I was performing around 50 rps, all of which were responding OK/200.

A bit of profiling has shown that old stream nums and their relevant streams are accumulating here:
- https://github.com/quic-go/quic-go/blob/master/http3/conn.go#L103
- https://github.com/quic-go/quic-go/blob/master/streams_map_outgoing.go#L20

Screenshot 2024-05-13 at 17 39 00

Here I was running around 50 rps at an instance of our project which is [proxying requests](https://github.com/flipt-io/reverst/blob/main/internal/server/server.go#L287) from one Go stdlib net/http server over a http3.RoundTripper.
The dip and plateau is where I stopped sending requests. The memory is just being held in the streams map now.

Screenshot 2024-05-13 at 17 50 39

Here is screenshot of a heap profile where we can see it accumulating in the map.

Looking at the relevant code, I have yet to see anything clear streams from the StreamMap types.
I tried calling DeleteStream explicitly once `isDone` (in the code [here](https://github.com/quic-go/quic-go/blob/master/http3/conn.go#L91-L93)) was true. But that led to errors on one side of the connection. So I figured I best open this back up to you guys for your thoughts.

Cheers!

GeorgeMac:
Fixes #4513

This updates the `onStreamStateChange` to handle the `streamStateSendAndReceiveClosed` state.
Prior to this, when the client failed to observe this, the datagram previously tracked in the streams map was left behind. This caused the memory leak observed in the issue above. Adding this case causes the entry to be deleted and memory in our project is stable.

### DIFF
--- a/http3/conn.go
+++ b/http3/conn.go
@@ -85,6 +85,8 @@ func (c *connection) onStreamStateChange(id quic.StreamID, state streamState, e 
 		isDone = d.SetReceiveError(e)
 	case streamStateSendClosed:
 		isDone = d.SetSendError(e)
+	case streamStateSendAndReceiveClosed:
+		isDone = true
 	default:
 		return
 	}
--- a/http3/conn.go
+++ b/http3/conn.go
@@ -70,29 +70,11 @@ func newConnection(
 	return c
 }
 
-func (c *connection) onStreamStateChange(id quic.StreamID, state streamState, e error) {
+func (c *connection) clearStream(id quic.StreamID) {
 	c.streamMx.Lock()
 	defer c.streamMx.Unlock()
 
-	d, ok := c.streams[id]
-	if !ok { // should never happen
-		return
-	}
-	var isDone bool
-	//nolint:exhaustive // These are all the cases we care about.
-	switch state {
-	case streamStateReceiveClosed:
-		isDone = d.SetReceiveError(e)
-	case streamStateSendClosed:
-		isDone = d.SetSendError(e)
-	case streamStateSendAndReceiveClosed:
-		isDone = true
-	default:
-		return
-	}
-	if isDone {
-		delete(c.streams, id)
-	}
+	delete(c.streams, id)
 }
 
 func (c *connection) openRequestStream(
@@ -110,7 +92,7 @@ func (c *connection) openRequestStream(
 	c.streamMx.Lock()
 	c.streams[str.StreamID()] = datagrams
 	c.streamMx.Unlock()
-	qstr := newStateTrackingStream(str, func(s streamState, e error) { c.onStreamStateChange(str.StreamID(), s, e) })
+	qstr := newStateTrackingStream(str, c, datagrams)
 	hstr := newStream(qstr, c, datagrams)
 	return newRequestStream(hstr, requestWriter, reqDone, c.decoder, disableCompression, maxHeaderBytes), nil
 }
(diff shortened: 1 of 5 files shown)


marten-seemann:
Hi @GeorgeMac! Thank you for putting in the effort to debug this! Is there any way to properly test this, either as a unit or an end-to-end test. This seems very subtle, and I'm concerned we'll break this again in the future.

I'm a bit confused about my own code now:
1. It seems unnecessary to do state tracking in the `datagrammer`. Since the `sendErr` and the `receiveErr` are only set from the outside, it seems like we could just remove the `isDone` return value on `SetReceiveError` and `SetSendError`, right?
2. I'm wondering if we should replace the `onStateChange` callback with two callbacks, one for closing the send and one for closing the receive side. This would allow us to get rid of the `streamState` enum altogether.

What do you think?

GeorgeMac:
> Hi @GeorgeMac! Thank you for putting in the effort to debug this! Is there any way to properly test this, either as a unit or an end-to-end test. This seems very subtle, and I'm concerned we'll break this again in the future.
> 
> I'm a bit confused about my own code now:
> 
>     It seems unnecessary to do state tracking in the datagrammer. Since the sendErr and the receiveErr are only set from the outside, it seems like we could just remove the isDone return value on SetReceiveError and SetSendError, right?
>     I'm wondering if we should replace the onStateChange callback with two callbacks, one for closing the send and one for closing the receive side. This would allow us to get rid of the streamState enum altogether.
> 
> What do you think?

I see both of what you're saying there @marten-seemann 👍 

Correct me if I am wrong (slowly catching up), the purpose of the state tracking system is twofold:
1. Free the datagrammer when both send and recv are closed.
2. Proxy closing of send and receive cancelling / errors from the stream onto the datagrammer (by way of `onStateChange` triggering `Set(Send|Receive)Error`)?

What you're suggesting makes sense. Another idea could be to remove the callbacks in favour of a couple interfaces to represent the datagrammer and removal of the datagrammer from the parent connection. Something like:

```go
type streamCloser interface {
  streamClose(quic.StreamID)
}

type sendRecieveError interface {
    SetSendError(error)
    SetRecieveError(error)
}

type stateTrackingStream struct {
    quic.Stream
    
    sendError error
    receiveError error
    
    errorer sendRecieveError
    closer streamCloser
}
```

Where the purpose of the stateTrackingStream becomes to continue proxying, but the moment the pair of send and recv error are non-nil, it calls `s.closer.streamClose(s.StreamID())`.

`streamClose` would be implemented on connection and would clean the entry from the map?

Happy to take a stab at this and write tests around it.

marten-seemann:
> Correct me if I am wrong (slowly catching up), the purpose of the state tracking system is twofold:
> 
> 1. Free the datagrammer when both send and recv are closed.
> 2. Proxy closing of send and receive cancelling / errors from the stream onto the datagrammer (by way of `onStateChange` triggering `Set(Send|Receive)Error`)?

Correct!

Your idea works as well. No strong preference either way. I usually find callbacks a tiny bit easier to test, but the difference is quite minimal.

GeorgeMac:
@marten-seemann I pushed up a take on what I described. See what you think. Happy to adjust or go back to callbacks.

MarcoPolo (approved):
This is great! nice work @GeorgeMac. It simplifies the logic. A nice improvement to this part of the code.

GeorgeMac:
Thank you @MarcoPolo 🙏

marten-seemann (approved):
Thank you @GeorgeMac!

GeorgeMac:
Thanks for the great feedback!

PR #4532, opened 2024-05-27
deprecate the ConnectionTracingKey

marten-seemann:
Applications can set their own tracing key using `Transport.ConnContext`.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -57,9 +57,11 @@ var Err0RTTRejected = errors.New("0-RTT rejected")
 // ConnectionTracingKey can be used to associate a ConnectionTracer with a Connection.
 // It is set on the Connection.Context() context,
 // as well as on the context passed to logging.Tracer.NewConnectionTracer.
+// Deprecated: Applications can set their own tracing key using Transport.ConnContext.
 var ConnectionTracingKey = connTracingCtxKey{}
 
 // ConnectionTracingID is the type of the context value saved under the ConnectionTracingKey.
+// Deprecated: Applications can set their own tracing key using Transport.ConnContext.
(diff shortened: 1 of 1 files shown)


marten-seemann:
cc @sukunrt and @MarcoPolo, this will affect your use case of quic-go

interface.go
```diff
@@ -57,9 +57,11 @@ var Err0RTTRejected = errors.New("0-RTT rejected")
 // ConnectionTracingKey can be used to associate a ConnectionTracer with a Connection.
 // It is set on the Connection.Context() context,
 // as well as on the context passed to logging.Tracer.NewConnectionTracer.
+// Deprecated: Applications can set their own tracing key using Transport.ConnContext.
 var ConnectionTracingKey = connTracingCtxKey{}
 
 // ConnectionTracingID is the type of the context value saved under the ConnectionTracingKey.
+// Deprecated: Applications can set their own tracing key using Transport.ConnContext.
 type ConnectionTracingID uint64
```
MarcoPolo, line 65: What is the type of the context value now?

MarcoPolo:
Sounds good. Would help me (and maybe others) if there was a simple migration path call out in the release notes. 🙏

interface.go
marten-seemann: For now, everything stays as is, it's just deprecated.

In the future, there won't be a `quic.ConnectionTracingKey` anymore. Instead, you will need to inject your own value on the context using `Transport.ConnContext` (see https://github.com/quic-go/docs/pull/60).

marten-seemann:
For sure, this will be called out in the release notes, and we'll link to the new section on the documentation page.

PR #4534, opened 2024-05-28
add some Prometheus metrics

### LINKED ISSUE #4077: provide Prometheus metrics

marten-seemann:
Fixes #4077.

TODO:
- [ ] add documentation for quic-go.net

These metrics are by no means comprehensive, but it's a starting point that we can iterate from.


metrics/pool.go
```diff
@@ -0,0 +1,27 @@
...
+
+func getStringSlice() *[]string {
+	s := stringPool.Get().(*[]string)
+	*s = (*s)[:0]
+	return s
+}
+
+func putStringSlice(s *[]string) {
+	if c := cap(*s); c < capacity {
+		panic(fmt.Sprintf("expected a string slice with capacity 8 or greater, got %d", c))
```
sukunrt, line 24:
```suggestion
		panic(fmt.Sprintf("expected a string slice with capacity 4 or greater, got %d", c))
```

metrics/tracer.go
```diff
@@ -0,0 +1,142 @@
...
+			case hdr.Type == protocol.PacketTypeInitial:
+				var ccf *logging.ConnectionCloseFrame
+				for _, f := range frames {
+					cc, ok := f.(*logging.ConnectionCloseFrame)
+					if ok {
+						ccf = cc
+						break
+					}
+				}
+				if ccf.IsApplicationError {
```
sukunrt, line 86: Do we need a `ccf != nil` here?

metrics/connection_tracer.go
```diff
@@ -0,0 +1,201 @@
...
+// DefaultTracer returns a callback that creates a metrics ConnectionTracer.
+// The ConnectionTracer returned can be set on the quic.Config for a new connection.
+// It should be reused across QUIC connections.
+func DefaultTracer() func(_ context.Context, p logging.Perspective, _ logging.ConnectionID) *logging.ConnectionTracer {
+	return DefaultTracerWithRegisterer(prometheus.DefaultRegisterer)
+}
+
+// DefaultTracerWithRegisterer returns a callback that creates a metrics ConnectionTracer
+// using a given Prometheus registerer.
+func DefaultTracerWithRegisterer(registerer prometheus.Registerer) func(_ context.Context, p logging.Perspective, _ logging.ConnectionID) *logging.ConnectionTracer {
```
sukunrt, line 63: You can register the metrics here once instead of registering them in the callback. The current method also works but all but the first registration will return AlreadyRegistrationError so we can just register them upfront here.

sukunrt (approved):

metrics/pool.go
marten-seemann: Good catch. I'll rephrase the message, so we can adjust this value without changing the string.

metrics/tracer.go
marten-seemann: Technically not, since the only reason we send an Initial packet before creating a connection is struct is to reject the connection attempt. We should document that though.

marten-seemann: Or rather, be more defensive here. It's not good to panic just to collect metrics.

metrics/connection_tracer.go
marten-seemann: `NewClientConnectionTracerWithRegisterer` and `NewServerConnectionTracerWithRegisterer` are public functions though, so we'll need to register in these code paths as well (and catch the `AlreadyRegisteredError`s there).

sukunrt: That's a good point.

PR #4536, opened 2024-05-28
pass a context to Transport.ConnContext

marten-seemann:
This context is cancelled when the QUIC connection is closed, or when the QUIC handshake fails. This allows the application to easily build and garbage collect a map of active connections.

TODO:
- [x] figure out https://github.com/quic-go/quic-go/pull/4536#discussion_r1621622647

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -90,6 +90,7 @@ type Transport struct {
 	VerifySourceAddress func(net.Addr) bool
 
 	// ConnContext is called when the server accepts a new connection.
+	// The context is closed when the connection is closed, or when the handshake fails for any reason.
 	// The context returned from the callback is used to derive every other context used during the
 	// lifetime of the connection:
 	// * the context passed to crypto/tls (and used on the tls.ClientHelloInfo)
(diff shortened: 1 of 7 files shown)


transport.go
```diff
@@ -90,14 +90,15 @@ type Transport struct {
...
+	// The context is closed when the connection is closed, or when the handshake fails for any reason.
 	// The context returned from the callback is used to derive every other context used during the
 	// lifetime of the connection:
 	// * the context passed to crypto/tls (and used on the tls.ClientHelloInfo)
 	// * the context used in Config.Tracer
 	// * the context returned from Connection.Context
 	// * the context returned from SendStream.Context
 	// It is not used for dialed connections.
-	ConnContext func() context.Context
+	ConnContext func(context.Context) context.Context
```
marten-seemann, line 101: There's another restriction here: The context returned MUST be derived from the context passed into `ConnContext`.

We can:
1. Document it.
2. Document and enforce it, by attaching an internal value to the context, and checking that it exists on the returned context (and panic if not).
3. Remove this requirement. This means we'll have to pass a second `context.CancelCauseFunc` to the connection struct.

sukunrt (approved):

transport.go
marten-seemann: > Can we derive a context from the context returned from `transport.ConnContext`?
> 
> I mean
> 
> ```
> ctx, cancel = context.WithCancelCause(s.connContext())
> ...
> conn = s.newConn(ctx, cancel ...)
> ```

I don't think that works, since we want the context we pass to `ConnContext` to be canceled when the connection is closed.

marten-seemann: Actually, we can do 3 without any major refactor. A `context.ContextCancelFunc` is just a fancy name for a `func(error)`, so we can just wrap the cancel func of the original and the returned context into one function.

PR #4544, opened 2024-06-01
use ringbuffer

### LINKED ISSUE #4515: http3: wasteful allocation of datagrammer and stream state tracker for non-datagram requests
We can probably save quite a few allocs here.

Qjawko:
This PR closes #4515 issue

### DIFF
--- a/http3/datagram.go
+++ b/http3/datagram.go
@@ -3,6 +3,8 @@ package http3
 import (
 	"context"
 	"sync"
+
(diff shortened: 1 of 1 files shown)


marten-seemann:
This PR does NOT resolve #4515. In fact, it exacerbates the problem quite significantly, since in addition to allocating the stream state tracking structs we now also preallocate memory for datagrams.

PR #4550, opened 2024-06-03
fix the server's 0-RTT rejection logic when using GetConfigForClient

marten-seemann:


integrationtests/self/zero_rtt_test.go
```diff
@@ -813,87 +813,93 @@ var _ = Describe("0-RTT", func() {
 		Entry("doesn't reject 0-RTT when the server's transport connection flow control limit increased", func(c *quic.Config, limit uint64) { c.InitialConnectionReceiveWindow = limit }),
 	)
 
-	for _, l := range []int{0, 15} {
-		connIDLen := l
```
marten-seemann, line 817: This didn't work.

PR #4555, opened 2024-06-05
metrics: simplify constructor for the default connection tracer

marten-seemann:
This seems more ergonomic, and it makes usage more similar to the `qlog.DefaultTracer`.


sukunrt (approved):
More ergonomic indeed!

PR #4563, opened 2024-06-10
http3: do not cache dial error

### LINKED ISSUE #4561: http3: RoundTripper is caching dialErr since v0.43.0
Since da410a7, r.getClient is caching any errors from r.dial (specifically a custom Dial func) unless Close is called manually.

In my use case, it fails and always returns an error after the client's network environment changes, which is common on mobile devices.

nekohasekai:
Looks like this fixes #4561. 

I'm not familiar with http3.RoundTripper, maybe there's a better solution.

### DIFF
--- a/http3/roundtrip.go
+++ b/http3/roundtrip.go
@@ -166,6 +166,7 @@ func (r *RoundTripper) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.
 	}
 
 	if cl.dialErr != nil {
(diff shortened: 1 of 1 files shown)


marten-seemann:
Thank you! Your change looks correct, but at the very least, we need an integration test to make sure we don't break this again in the future.

I wasn't allowed to push to this branch, so I opened a new PR that contains your commit in #4573.

Issue #4553, opened 2024-06-05
Allow applications to specify MaxPacketBufferSize


chungthuang:
Some applications might want to use a MaxPacketBufferSize that is less than 1452 bytes. For example, some networks might have middle boxes that adds extra encapsulation before reaching the server. To support this, maybe we can add a new option to [Config](https://pkg.go.dev/github.com/quic-go/quic-go#Config). [Then we can pass it to TransportParameters](https://github.com/quic-go/quic-go/blob/master/connection.go#L299) to advertise the value to the peer.

@marten-seemann I'm happy to open a PR, but I have some questions about other places using `MaxPacketBufferSize`. I see it is also used in [buffer_pool](https://github.com/quic-go/quic-go/blob/master/buffer_pool.go#L86) and [stream frames](https://github.com/quic-go/quic-go/blob/master/internal/wire/pool.go#L14). What is the implication if the advertised `maxUDPPayloadSize` is different from size of the buffer pool and stream frames?

marten-seemann:
This is already possible. You can disable DPLPMTUD and set the initial packet size.

chungthuang:
Can you point me to how to set the initial packet size when DPLPMTUD is disabled? The doc says `If unavailable or disabled, packets will be at most 1280 bytes in size.`

marten-seemann:
Looks like we didn't properly update the document the documentation when we introduced `Config.InitialPacketSize`...

joliveirinha:
@marten-seemann to be clear, we don't want to disable PMTU Discovery. 

To be clear, what we really want is for the server to advertise a maximum packet size that it is willing to receive. Today, this value is hardcoded as a constant of MaxPacketBufferSize.

The reason why we want to do this, is because we don't want clients to exceed that value when doing PMTU discovery because it is possible that during the QUIC connection lifetime, that the PMTU decreases due to packet re-routing associated to how our anycast network works. 

Since PMTU doesn't support today the capability of decreasing in such event of MTU changing over time, then we need to guarantee that the server advertises the maximum value that it can receive packets in the event that a change happens.

marten-seemann:
Got it, so #4558 doesn't fix your issue (#3955).

While I understand that maxing `max_udp_payload_size` configurable would allow you to better work around that issue, I'm don't we should introduce additional API surface to do that. I'd much rather get #3955 fixed once and for all, and have PMTUD work out of the box, without any need for intervention via `quic.Config` options.

[...]

PR #4570, opened 2024-06-20
Introduce a config option for the max UDP payload size transport parameter

chungthuang:
This helps address #4553 in the short term before we have #3955.

### DIFF
--- a/config.go
+++ b/config.go
@@ -45,6 +45,12 @@ func validateConfig(config *Config) error {
 	if config.InitialPacketSize > protocol.MaxPacketBufferSize {
 		config.InitialPacketSize = protocol.MaxPacketBufferSize
 	}
+	if config.MaxUDPPayloadSize > 0 && config.MaxUDPPayloadSize < protocol.MinInitialPacketSize {
+		config.MaxUDPPayloadSize = protocol.MinInitialPacketSize
+	}
+	if config.MaxUDPPayloadSize > protocol.MaxPacketBufferSize {
+		config.MaxUDPPayloadSize = protocol.MaxPacketBufferSize
+	}
 	// check that all QUIC versions are actually supported
 	for _, v := range config.Versions {
 		if !protocol.IsValidVersion(v) {
@@ -104,6 +110,10 @@ func populateConfig(config *Config) *Config {
 	if initialPacketSize == 0 {
 		initialPacketSize = protocol.InitialPacketSize
 	}
+	maxUDPPayloadSize := config.MaxUDPPayloadSize
+	if maxUDPPayloadSize == 0 {
+		maxUDPPayloadSize = protocol.DefaultMaxUDPPayloadSize
+	}
 
 	return &Config{
 		GetConfigForClient:             config.GetConfigForClient,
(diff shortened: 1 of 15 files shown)


config.go
```diff
@@ -45,6 +45,12 @@ func validateConfig(config *Config) error {
 	if config.InitialPacketSize > protocol.MaxPacketBufferSize {
 		config.InitialPacketSize = protocol.MaxPacketBufferSize
 	}
+	if config.MaxUDPPayloadSize > 0 && config.MaxUDPPayloadSize < protocol.MinInitialPacketSize {
+		config.MaxUDPPayloadSize = protocol.MinInitialPacketSize
```
chungthuang, line 49: I was thinking should this be MinInitialPacketSize - 40 (IPv6 header) - 8 (UDP) header?

marten-seemann:
Should this be an option on the `Transport` instead? We could then enforce correct peer behavior, i.e. drop packets that are too large: either with an explicit size check, or by passing an appropriately sized buffer to the `ReadFrom` calls.

chungthuang:
Hi @marten-seemann, I followed your suggestion to move the config option to transport and drop large packets in `transport.handlePacket`. Let me know if that is what you have in mind.

config.go
Wasem9926: wwasim838@gmail.com

transport.go
```diff
@@ -398,6 +408,15 @@ func (t *Transport) handlePacket(p receivedPacket) {
 		t.handleNonQUICPacket(p)
 		return
 	}
+	if len(p.data) > int(t.MaxUDPPayloadSize) {
```
marten-seemann, line 411: This is inconsistent with how we handle packets larger 1452 bytes. How much effort would it be to wire this up properly, such that we pass a correctly sized buffer to the kernel? Would that create any problems with GRO?

DevinCarr: Are you asking if we can pipe this config directly into areas like [sys_conn.go#L94](https://github.com/chungthuang/quic-go/blob/b73a4de7eaa87773d9da570442752470b051ace6/sys_conn.go#L94), or [send_stream.go#L211](https://github.com/chungthuang/quic-go/blob/723a188dde0db4989e988749991fd424b8a0daab/send_stream.go#L211)?
I don't think any of those areas need to change as long as we make sure that the transport parameter `MaxUDPPayloadSize` provided is strictly less than the `MaxPacketBufferSize`? I understand that the RFC expects to allow the transport parameter to be set to a maximum of 65527, but I don't think that would work anyways regarding how the internal stream frame pools are statically allocated to block sizes of 1452.

Could you also point to a specific areas of where handling packets larger than the 1452? I could be missing it completely.

Regarding the GRO comment, similarly, if we make sure that the transport parameter is strictly less than the `MaxPacketBufferSize` of 1452, this shouldn't be a problem?

marten-seemann: > Are you asking if we can pipe this config directly into areas like [sys_conn.go#L94](https://github.com/chungthuang/quic-go/blob/b73a4de7eaa87773d9da570442752470b051ace6/sys_conn.go#L94), or [send_stream.go#L211](https://github.com/chungthuang/quic-go/blob/723a188dde0db4989e988749991fd424b8a0daab/send_stream.go#L211)?

Definitely not in the send stream, but yes, in the connection. That way, the kernel would handle the size of the read buffer.

I'm wondering how different packet sizes are dealt with when using GRO: For GRO, we'll pass a 64k buffer to the kernel. Does this mean that the kernel could (in principle) return one jumbo packet, or is there any way to set a maximum packet size?

PR #4571, opened 2024-06-22
ci: use Go 1.23rc2 for unit and integration tests

marten-seemann:


marten-seemann:
Blocked on the resolution of https://github.com/golang/go/issues/68124.

PR #4579, opened 2024-06-27
introduce a StreamLimitReachedError for Connection.Open{Uni}Stream

marten-seemann:
Using a concrete type is preferable to relying on interpreting the return values from `net.Error.Timeout` and `net.Error.Temporary`, especially since the latter has been deprecated since Go 1.18.


streams_map.go
```diff
@@ -38,11 +38,21 @@ type streamOpenErr struct{ error }
...
+func (streamOpenErr) Timeout() bool   { return false }
+func (e streamOpenErr) Unwrap() error { return e.error }
 
-// errTooManyOpenStreams is used internally by the outgoing streams maps.
-var errTooManyOpenStreams = errors.New("too many open streams")
+func (e streamOpenErr) Temporary() bool {
+	// In older versions of quic-go, the stream limit error was documented to be a net.Error.Temporary.
+	// This function was since deprecated, but we keep the existing behavior.
+	return errors.Is(e, StreamLimitReachedError{})
+}
```
sukunrt, line 48: Can we add a test for this?

streams_map_outgoing.go
```diff
@@ -60,7 +60,7 @@ func (m *outgoingStreamsMap[T]) OpenStream() (T, error) {
 	// if there are OpenStreamSync calls waiting, return an error here
 	if len(m.openQueue) > 0 || m.nextStream > m.maxStream {
 		m.maybeSendBlockedFrame()
-		return *new(T), streamOpenErr{errTooManyOpenStreams}
+		return *new(T), streamOpenErr{&StreamLimitReachedError{}}
```
sukunrt, line 63: This should be 
`streamOpenErr{StreamLimitReachedError{}} // no pointer `

Otherwise `err.Temporary` wont be true.

sukunrt (approved):

streams_map_outgoing.go
marten-seemann: Or alternatively, the `errors.Is` test should test for the pointer.

streams_map.go
marten-seemann: Done.

PR #4580, opened 2024-06-30
refactor: unnecessary else block

lkeix:
## What

I found an unnecessary `else` code block in quic-go and deleted it.

## How can this PR be improved?

1. This change improves code readability.
2. This change may slightly improve performance by removing the unnecessary `else` block.


marten-seemann (changes requested):
As the comment says, this was copied from the standard library. I don't see any reason to make changes to this code path.

It's definitely not right that this improves performance.

PR #4581, opened 2024-07-05
http3: add client-side HTTP Trailer support

sudorandom:
This is a continuation of the work done in #2344 and will resolve _half_ of #2266. This adds support for reading trailers in the HTTP client sent from servers after the HTTP body is received. I've adapted these changes to work with the updated state of the repo.

I've tested with this with a ConnectRPC client configured with quic-go with a gRPC server and it appears to work well.


http3/client.go
```diff
@@ -208,8 +208,9 @@ func (c *SingleDestinationRoundTripper) roundTrip(req *http.Request) (*http.Resp
 		}
 	}
 
+	resp := &http.Response{}
```
sudorandom, line 211: I'm not sure if this is the best decision. In the previous Pr, `doRequest` used to call the code to read the body after the HTTP status code is received. However, since we're now calling `openRequestStream` I also created the `*http.Response` here and pass it into `c.hconn.openRequestStream()` and `c.doRequest` (which, in turn, impacts how `responseFromHeaders` works).

http3/http_stream.go
```diff
@@ -75,12 +84,33 @@ func (s *stream) Read(b []byte) (int, error) {
...
+
+				for _, trailer := range trailers {
+					if s.resp.Trailer == nil {
+						s.resp.Trailer = http.Header{}
+					}
+					s.resp.Trailer.Add(trailer.Name, trailer.Value)
+				}
+
+				// Trailer Frame is the last frame.
+				return 0, nil
```
sudorandom, line 113: There is questions in the previous PR about if this is the best decision. I'm also not 100% sure on this. We may instead want to set a flag to ensure we don't accept data frames after the trailer frame.

akshayjshah:
After quite a few years, I'm excited to see some progress on trailers support!

sudorandom:
@marten-seemann is it possible to get some attention on this PR at some point? At the very least, can you allow the CI workflows to run?

http3/headers.go
```diff
@@ -178,25 +178,23 @@ func hostnameFromURL(url *url.URL) string {
 	return ""
 }
 
-func responseFromHeaders(headerFields []qpack.HeaderField) (*http.Response, error) {
+func responseFromHeaders(res *http.Response, headerFields []qpack.HeaderField) error {
```
marten-seemann, line 181: Why is this change necessary at all?

This function name doesn't seem accurate anymore. It would probably also help to add a comment.

http3/http_stream.go
```diff
@@ -68,19 +77,45 @@ func (s *stream) Read(b []byte) (int, error) {
...
+					continue
+				}
+				if f.Length > s.maxHeaderBytes {
+					return 0, fmt.Errorf("HEADERS frame too large: %d bytes (max: %d)", f.Length, s.maxHeaderBytes)
+				}
+				if seenTrailerFrame {
+					return 0, errors.New("HEADERS frame received after trailers")
+				}
+				p := make([]byte, f.Length)
+				s.Stream.Read(p)
```
marten-seemann, line 105: This doesn't work. You'll need to use `io.ReadFull`. Also, please don't ignore the error.

http3/http_stream.go
```diff
@@ -68,19 +77,45 @@ func (s *stream) Read(b []byte) (int, error) {
...
+				trailers, err := s.decoder.DecodeFull(p)
```
marten-seemann, line 106: It seems unfortunate that we now have two places where we define the parsing of a HEADERS frame.

http3/http_stream.go
```diff
@@ -68,19 +77,45 @@ func (s *stream) Read(b []byte) (int, error) {
...
+				if err != nil {
+					// Ignoring invalid frame
```
marten-seemann, line 108: Why? This seems wrong.

marten-seemann (commented):
Thanks for this PR, and sorry that it took so long to review (IETF weeks are busy...).

I like it, since it looks pretty straightforward. I'd like to simplify the design a little bit though.

How much work would it be to add trailer support for the server side as well? I'd really like to have both (preferably in a separate PR), as this would allow us to write an end-to-end test. It has been my experience that things tend to break in unexpected and annoying ways if we don't test them end-to-end.

http3/headers.go
sudorandom: I rolled this change back. This was definitely evidence that I was going down the wrong path 😓

http3/http_stream.go
sudorandom: Switched to `io.ReadFull` and it now returns the error instead of ignoring it

http3/http_stream.go
sudorandom: Yeah, I agree. This was a holdover from the first PR that I didn't quite understand the reasoning behind. I've adjusted it to return the error instead of ignoring it and adjusted the associated test

http3/http_stream.go
```diff
@@ -47,21 +47,29 @@ type stream struct {
...
-		datagrams: datagrams,
+		Stream:          str,
+		conn:            conn,
+		buf:             make([]byte, 16),
+		datagrams:       datagrams,
+		onTrailersFrame: func(*headersFrame) (int, error) { return 0, nil },
 	}
 }
 
+func (s *stream) setOnTrailersFrame(onTrailersFrame func(*headersFrame) (int, error)) {
```
marten-seemann, line 69: I'm a bit confused by this logic here. Why do we need to set a callback here? Why can't we just have a `requestStream.ReadTrailer() (*http.Response, error)`?

sudorandom: Hm.. The reason for this is to be able to re-use the `stream` type for clients. It's my understanding that when clients read the body, `stream. Read()` is called consecutively until the body is completely exhausted. The `stream.Read()` function doesn't really know when it will yield a headers frame (which contains the trailers): Therefore, I figured that there's a few options:
1. This weird callback behavior that I have here
2. Breaking the Read() interface and potentially returning the trailers
3. Directly add what is needed to parse the trailers when creating `stream`, even though none of those are relevant to server-side uses of this type (this is what I had before)

Because of the behavior of not knowing when the trailers will be sent, I don't understand how I would make your suggestion work. If I were to put it in requestStream.

I'm going to take another stab at it without resorting to callbacks

sudorandom:
@marten-seemann I have a branch with server support for trailers that contains an integration test that tests both client and server behavior. It's also probably not 100% perfect. Would you prefer that I make a new PR now, wait for this PR to be merged or include it in this PR?

http3/http_stream.go
sudorandom: Now that I look at this more, the issue is specifically with responses that do not have a content-length. There are different behavior there, which is kind of annoying.

marten-seemann:
> @marten-seemann I have a branch with server support for trailers that contains an integration test that tests both client and server behavior. It's also probably not 100% perfect. Would you prefer that I make a new PR now, wait for this PR to be merged or include it in this PR?

A separate PR would be great!

marten-seemann (commented):
I believe it's possible to make the design significantly simpler. Take a look at https://github.com/quic-go/quic-go/commit/f9f9151fe87abb2c1914d4511237b906d7999189. What do you think?

sudorandom:
@marten-seemann I cherrypicked your changes and finished the work needed to make it fully work. I think it looks pretty good now.

http3/conn.go
```diff
@@ -112,8 +114,37 @@ func (c *connection) openRequestStream(
...
+	var rsp = new(http.Response)
+	hstr := newStream(qstr, c, datagrams, func(r io.Reader, l uint64) error {
+		hdr, err := c.parseTrailer(r, l, maxHeaderBytes)
+		if err != nil {
+			return err
+		}
+		rsp.Trailer = hdr
+		return nil
+	})
+	// TODO: pass this http.Response to newRequestStream and use it there
```
marten-seemann, line 126: This comment is outdated, isn't it?

http3/headers.go
```diff
@@ -178,25 +178,24 @@ func hostnameFromURL(url *url.URL) string {
 	return ""
 }
 
-func responseFromHeaders(headerFields []qpack.HeaderField) (*http.Response, error) {
+func updateResponseFromHeaders(rsp *http.Response, headerFields []qpack.HeaderField) error {
```
marten-seemann, line 181: I suggest adding some documentation here, since I was at first surprised by the signature of this function:
```suggestion
// updateResponseFromHeaders sets up http.Response as an HTTP/3 response,
// using the decoded qpack header filed.
// It is only called for the HTTP header (and not the HTTP trailer).
// It takes an http.Response as an argument to allow the caller to set the trailer later on.
func updateResponseFromHeaders(rsp *http.Response, headerFields []qpack.HeaderField) error {
```

marten-seemann (approved):
Thank you @sudorandom, this looks pretty clean now!

I added two comments regarding comments, could you please take a quick look? If you're ok with my suggestions, we can get this PR merged now.

http3/conn.go
```diff
@@ -112,8 +114,37 @@ func (c *connection) openRequestStream(
 	c.streams[str.StreamID()] = datagrams
 	c.streamMx.Unlock()
 	qstr := newStateTrackingStream(str, c, datagrams)
-	hstr := newStream(qstr, c, datagrams)
-	return newRequestStream(hstr, requestWriter, reqDone, c.decoder, disableCompression, maxHeaderBytes), nil
+	var rsp = new(http.Response)
```
marten-seemann, line 117:
```suggestion
	rsp := &http.Response{}
```

http3/conn.go
```diff
@@ -112,8 +114,37 @@ func (c *connection) openRequestStream(
...
+	rsp := &http.Response{}
+	hstr := newStream(qstr, c, datagrams, func(r io.Reader, l uint64) error {
+		hdr, err := c.parseTrailer(r, l, maxHeaderBytes)
+		if err != nil {
+			return err
+		}
+		rsp.Trailer = hdr
+		return nil
+	})
+	// TODO: pass this http.Response to newRequestStream and use it there
```
marten-seemann, line 126:
```suggestion
```

PR #4587, opened 2024-07-20
http3: implement server idle timeout support

rs:
This update introduces the ability for an HTTP/3 server to enforce an idle timeout on connections. This timeout will trigger when no new requests are received on a connection, irrespective of any PING frames received at the QUIC level.

### DIFF
--- a/http3/conn.go
+++ b/http3/conn.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/quic-go/quic-go"
 	"github.com/quic-go/quic-go/internal/protocol"
@@ -51,6 +52,9 @@ type connection struct {
 
 	settings         *Settings
 	receivedSettings chan struct{}
+
+	idleTimeout time.Duration
+	idleTimer   *time.Timer
 }
 
 func newConnection(
@@ -59,17 +63,27 @@ func newConnection(
 	enableDatagrams bool,
 	perspective protocol.Perspective,
 	logger *slog.Logger,
+	idleTimeout time.Duration,
 ) *connection {
-	return &connection{
+	c := &connection{
 		ctx:              ctx,
 		Connection:       quicConn,
 		perspective:      perspective,
 		logger:           logger,
+		idleTimeout:      idleTimeout,
 		enableDatagrams:  enableDatagrams,
 		decoder:          qpack.NewDecoder(func(hf qpack.HeaderField) {}),
 		receivedSettings: make(chan struct{}),
 		streams:          make(map[protocol.StreamID]*datagrammer),
 	}
+	if idleTimeout > 0 {
+		c.idleTimer = time.AfterFunc(idleTimeout, c.onIdleTimer)
+	}
+	return c
+}
+
+func (c *connection) onIdleTimer() {
+	c.CloseWithError(quic.ApplicationErrorCode(ErrCodeNoError), "idle timeout")
 }
 
 func (c *connection) clearStream(id quic.StreamID) {
@@ -77,6 +91,9 @@ func (c *connection) clearStream(id quic.StreamID) {
 	defer c.streamMx.Unlock()
 
 	delete(c.streams, id)
+	if c.idleTimeout > 0 && len(c.streams) == 0 {
+		c.idleTimer.Reset(c.idleTimeout)
+	}
 }
 
 func (c *connection) openRequestStream(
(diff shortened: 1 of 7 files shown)


marten-seemann:
Is there anything analogous in HTTP/1.1 or HTTP/2?

rs:
Both `http.Server` and `http2.Server` have the exact same feature.

rs:
@marten-seemann what's your take on this PR?

marten-seemann:
I'll review this after the IETF week.

http3/conn.go
```diff
@@ -109,12 +126,24 @@ func (c *connection) acceptStream(ctx context.Context) (quic.Stream, *datagramme
 		strID := str.StreamID()
 		c.streamMx.Lock()
 		c.streams[strID] = datagrams
+		if c.idleTimeout > 0 {
+			if len(c.streams) == 1 {
```
marten-seemann, line 130: I don't understand the logic here.

http3/server.go
```diff
@@ -198,6 +198,12 @@ type Server struct {
 	// In that case, the stream type will not be set.
 	UniStreamHijacker func(StreamType, quic.ConnectionTracingID, quic.ReceiveStream, error) (hijacked bool)
 
+	// IdleTimeout specifies how long until idle clients connection should be
+	// closed. Quic PING frames are not considered activity for the purposes of
```
marten-seemann, line 202: Maybe generalize this statement, as not only PING frames reset QUIC's idle timeout? You could say that "idle" refers only to the HTTP/3 layer.

marten-seemann (commented):
Thank you for this PR @rs!

Can you add an integration test as well?

http3/conn.go
rs: We stop the idle timer as soon as the pool is no longer empty (first stream added to the pool).

rs:
I'll give it a try. I'm not familiar with the test framework in use.

marten-seemann:
> I'll give it a try. I'm not familiar with the test framework in use.

Sorry for that. We're planning to move away from Ginkgo (https://github.com/quic-go/quic-go/issues/3652), but it's a lot of work...

What I'm imagining is a test in `integrationtests/self/http_test.go`, where the server applies a short idle timeout. We can then verify that the client's second request fails (or is sent on a newly established connection). lmk if you run into any issues.

rs:
@marten-seemann I added an integration test, but I’m a little perplexed. The client connection being closed by the server should not be surfaced to the client’s caller, in my opinion. Isn’t this a bug in the client implementation?

marten-seemann:
> @marten-seemann I added an integration test, but I’m a little perplexed. The client connection being closed by the server should not be surfaced to the client’s caller, in my opinion. Isn’t this a bug in the client implementation?

Good question. I think there might be a bug. We have some redialing logic here: https://github.com/quic-go/quic-go/blob/42f04d4e02205ee796d1fdec12bd2b46166e7cde/http3/roundtrip.go#L173-L187
But it doesn't take into account the HTTP/3 error at all. Maybe we need to allow automatic redialing on H3_NO_ERRORs? What do you think would be the right behavior?

rs:
I think any error or closing of the connection should be hidden from the client if it does not directly impact an existing stream.

I don’t think it should auto-redial on close/error. Otherwise, a long-running client would keep redialing connections to domains it interacted with once, without any guarantee it will interact again soon.

Rather, on stream creation, if the connection is unhealthy but still in the pool, it should then be redialed.

By the way, regarding the code you highlighted, I don’t understand how the connection would be closed if the returned error did not result in the connection being closed. Wouldn’t this lead to leaked connections being kept alive if keepalive is enabled at the QUIC level?

marten-seemann:
> I think any error or closing of the connection should be hidden from the client if it does not directly impact an existing stream.

That depends on the type of error. If it's a protocol violation, either at the QUIC or the HTTP layer, we should definitely surface this error. The only errors we should hide are errors that are expected to happen in normal operation, such as idle timeouts, stateless resets, etc.



> I don’t think it should auto-redial on close/error. Otherwise, a long-running client would keep redialing connections to domains it interacted with once, without any guarantee it will interact again soon.
> 
> Rather, on stream creation, if the connection is unhealthy but still in the pool, it should then be redialed.

That's how the code works. In the code I linked above, calling `RoundTrip` on a connection that has seen an idle timeout (for example), would return a `net.Error.Timeout` error, and we'd dial a new connection right away. All we need to do is extend this logic to also cover H3_NO_ERROR errors.

rs:
Sorry if I was not clear. Let me rephrase. By an error not impacting an existing stream, I mean an error that occurred before the stream was created. If the client discovers it later (i.e., at the next stream creation), it may retry without surfacing the error to the caller, but only if it is guaranteed that the request wasn’t written on the previous connection.

The existing condition to retry the request on `net.Timeout` is, in my opinion, dangerous. You have no way of knowing if the request was received by the server and then timed out, or if the connection had already timed out. This runs the risk of performing the request twice, which in some cases might be unexpected and dangerous if the request is not idempotent.

Regarding my question on the `removeClient`, can you please help me understand how the removed connection is ensured to be closed?

marten-seemann:
I think I agree with your argument. I opened #4612.

What do we still need to do here to get this PR across the finish line?

rs:
To me, the PR is good to go.

Let me know if I should open another issue or if there is another way to discuss `removeClient` behavior. I believe there is a potential issue here that is different from the issue you just opened. I might be missing something, so I’d love to pick your brain on this.

marten-seemann (approved):

marten-seemann:
> Let me know if I should open another issue or if there is another way to discuss `removeClient` behavior. I believe there is a potential issue here that is different from the issue you just opened. I might be missing something, so I’d love to pick your brain on this.

Please do!

PR #4601, opened 2024-07-28
handshake: generate CRYPTO_ERRORs for internal_error TLS alerts

### LINKED ISSUE #4582: tls.Config.GetConfigForClient error is sent as an INTERNAL_ERROR
This should probably be a CRYPTO_ERROR, but not sure which one. The Go standard library uses an a TLS `internal_error` alert, which arguably is not a lot more meaningful. At least it points to the right component though.

marten-seemann:
Fixes #4582.

### DIFF
--- a/internal/handshake/crypto_setup.go
+++ b/internal/handshake/crypto_setup.go
@@ -625,7 +625,7 @@ func (h *cryptoSetup) ConnectionState() ConnectionState {
 
 func wrapError(err error) error {
 	// alert 80 is an internal error
-	if alertErr := tls.AlertError(0); errors.As(err, &alertErr) && alertErr != 80 {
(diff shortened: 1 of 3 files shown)


internal/handshake/crypto_setup.go
```diff
@@ -625,7 +625,7 @@ func (h *cryptoSetup) ConnectionState() ConnectionState {
 
 func wrapError(err error) error {
 	// alert 80 is an internal error
```
sukunrt, line 627: stale comment?

sukunrt (approved):

internal/handshake/crypto_setup.go
marten-seemann: We don't. Removed.

PR #4605, opened 2024-07-31
mark stream completed when CancelWrite is called after receiving STOP_SENDING

sukunrt:


MarcoPolo:
Probably worth updating this comment: https://github.com/MarcoPolo/quic-go/blob/master/send_stream.go#L380-L382

Also no real preference here, but I'd suggest this diff: https://gist.github.com/MarcoPolo/70f81a52bf30236d76cdbad72c87db6c

send_stream.go
```diff
@@ -423,6 +423,12 @@ func (s *sendStream) cancelWriteImpl(errorCode qerr.StreamErrorCode, remote bool
 		s.cancellationFlagged = true
 	}
 	if s.cancelWriteErr != nil {
+		// The stream was reset already. Check if the stream can be garbage collected now.
+		// this happens when we call CancelWrite after receiving a STOP_SENDING
+		//
+		if s.isNewlyCompleted() {
+			s.sender.onStreamCompleted(s.streamID)
```
marten-seemann, line 430: This needs to be called without holding the lock.

marten-seemann (commented):
Can you elaborate on what problem this PR fixes?

MarcoPolo:
My understanding is that in the event that:
1. I receive a StopSending frame.
2. Then I call `.CancelWrite`

I will never call `.onStreamCompleted(s.streamID)`. This PR attempts to fix that.

send_stream_test.go
```diff
@@ -909,6 +909,19 @@ var _ = Describe("Send Stream", func() {
...
+				mockSender.EXPECT().queueControlFrame(&wire.ResetStreamFrame{
+					StreamID:  streamID,
+					ErrorCode: 101,
+				})
+				mockSender.EXPECT().onStreamCompleted(gomock.Any())
+				str.handleStopSendingFrame(&wire.StopSendingFrame{
+					StreamID:  streamID,
+					ErrorCode: 101,
+				})
+				str.CancelWrite(101)
```
marten-seemann, line 922: Better, since more restrictive:
```suggestion
				str.handleStopSendingFrame(&wire.StopSendingFrame{
					StreamID:  streamID,
					ErrorCode: 101,
				})
				mockSender.EXPECT().onStreamCompleted(gomock.Any())
				str.CancelWrite(101)
```

sukunrt:
@MarcoPolo using a slightly modified version of your patch. 

It'd be nice if we could structure this as having a defer block on every method we expect to complete the stream. 
```
defer func() {
    if s.isNewlyCompleted() {
        s.sender.onStreamCompleted(s.streamID)
    }
}()
```
Unfortunately, there are way too many edge cases for that and restructuring the whole thing is too much effort for this PR. If interested, we can pick that up separately.

send_stream.go
```diff
@@ -421,6 +421,16 @@ func (s *sendStream) cancelWriteImpl(errorCode qerr.StreamErrorCode, remote bool
 	s.mutex.Lock()
 	if !remote {
 		s.cancellationFlagged = true
+		if s.cancelWriteErr != nil {
+			s.mutex.Unlock()
+			// The user has called CancelWrite. If the previous cancellation was
+			// because of a STOP_SENDING, we don't need to flag the error to the
+			// user any more.
+			if s.isNewlyCompleted() {
```
marten-seemann, line 429: This must be called while holding the mutex. I know it's a bit tricky. That's why you often see this pattern here:
```go
completed := s.isNewlyCompleted()
s.mutex.Unlock()
if completed {
    s.sender.onStreamCompleted(s.streamID)
}
```

marten-seemann (changes requested):

send_stream.go
MarcoPolo: Done in 1f96f4f1

marten-seemann (approved):
Danke!

PR #4607, opened 2024-08-01
simplify frame queuing logic for 0-RTT rejection

marten-seemann:
Even though we currently don't do so, sending MAX_DATA, MAX_STREAM_DATA and MAX_STREAMS is allowed in 0-RTT, see https://datatracker.ietf.org/doc/html/rfc9000#frame-types.

### DIFF
--- a/framer.go
+++ b/framer.go
@@ -1,7 +1,7 @@
 package quic
 
 import (
-	"errors"
+	"slices"
 	"sync"
 
 	"github.com/quic-go/quic-go/internal/ackhandler"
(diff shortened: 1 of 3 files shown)


framer.go
```diff
@@ -171,28 +171,26 @@ func (f *framer) AppendStreamFrames(frames []ackhandler.StreamFrame, maxLen prot
...
 			continue
 		default:
 			f.controlFrames[j] = f.controlFrames[i]
 			j++
 		}
 	}
-	f.controlFrames = f.controlFrames[:j]
-	f.controlFrameMutex.Unlock()
-	return nil
+	f.controlFrames = slices.Delete(f.controlFrames, j, len(f.controlFrames))
```
marten-seemann, line 195: And another small win here: `slices.Delete` zeroes the deleted fields in the slice, allowing for quicker gc.

PR #3292, opened 2021-10-25
implement qlog JSONSEQ format, bump qlog version

marten-seemann:

PR #4609, opened 2024-08-01
implement qlog JSON-SEQ format, bump qlog version

marten-seemann:
Closes #3292.


marten-seemann:
JSON-SEQ is an absolute pain to deal with in the real world. I'm strongly opposed to using it, and to making it a standard. We should reevaluate this decision in the QUIC working group: https://github.com/quicwg/qlog/issues/434

PR #4610, opened 2024-08-02
queue stream-related control frames in the respective stream

### LINKED ISSUE #4271: don't use the control frame queue for stream-related frames
Stream-related frames (MAX_STREAM_DATA, RESET_STREAM, STOP_SENDING, STREAM_BLOCKED) could be enqueued with their respective stream instead of using the control frame queue. The control frame queue is an incredibly mindless tool: It doesn't remember any context of the frames, it just queues and dequeues them, oblivious to what's happening around.

Queueing the frames at their respective stream would allow us to optimize frame handling, e.g.:
* send out a fresh MAX_STREAM_DATA frame (i.e. with an updated offset) if one is lost
* not send out a MAX_STREAM_DATA frame if there are still a few (>= 2 maybe) in flight
* not send out control frames if the stream state changed (e.g. there's no need to send a STREAM_BLOCKED frame if the peer sent us a STOP_SENDING)
* (de-) prioritize sending of control frames for certain streams (this probably is a terrible idea)

It would also allow us to delay garbage collecting of the stream until all relevant control frames for that stream have been acknowledged. To be clear, we can garbage collect the stream if there are still MAX_STREAM_DATA frames outstanding, and we would now know that losing these frames would be a no-op.

marten-seemann:
Fixes #4271.

This is a massive PR, but I don't see any way to break it up into smaller parts.

Conceptually, it's simple though:
So far, we had the stream call `queueControlFrame` to queue frames like MAX_STREAM_DATA, STREAM_DATA_BLOCKED, RESET_STREAM and STOP_SENDING, which would enqueue the frame in the `framer` (or the `windowUpdateQueue`, for MAX_STREAM_DATA frames). This is bad for multiple reasons:
* It's hard to control the number of queued frames, if we're blocked on sending out frames (due to congestion control, for example). I don't see a straightforward way to exploit this, but we've seen similar attacks with other frame types earlier this year.
* Once enqueued, there's no way to remove a frame from the queue. For example, we don't need to send out a queued MAX_STREAM_DATA frame once we've read the EOF, or received a RESET_STREAM frame. This logic is not implemented in this PR to keep things simple(r), and will be done in a follow-up PR.

Instead, we do the following:
Whenever the stream has any control frame to send, we report this to the `framer`. The framer then remembers that this stream has something to send, and when the time comes to pack a packet, it queries the stream for the frame. Iff the stream still has stuff to send at this point, it generates the respective frames and they get packed.


window_update_queue.go
marten-seemann, line 1: Finally getting rid of some more gQUIC terminology.

mock_stream_control_frame_getter_test.go
marten-seemann, line 1: Not happy about adding yet another mock interface, but it is really convenient for testing.

framer.go
```diff
@@ -16,11 +16,20 @@ const (
 	maxControlFrames = 16 << 10
 )
 
+// This is the largest possible size of a stream-related control frame
+// (which is the RESET_STREAM frame).
+const minStreamControlFrameSize = 25
```
sukunrt, line 21: Should this be `maxStreamControlFrameSize` ?

framer.go
```diff
@@ -113,6 +149,14 @@ func (f *framer) AddActiveStream(id protocol.StreamID, str sendStreamI) {
 	f.mutex.Unlock()
 }
 
+func (f *framer) AddStreamWithControlFrames(id protocol.StreamID, str streamControlFrameGetter) {
+	f.controlFrameMutex.Lock()
+	if _, ok := f.streamsWithControlFrames[id]; !ok {
+		f.streamsWithControlFrames[id] = str
+	}
```
sukunrt, line 156: Should we log if the stream is present in the map already?

framer.go
```diff
@@ -16,11 +16,20 @@ const (
 	maxControlFrames = 16 << 10
 )
```
sukunrt, line 17: Can we remove this and `framer.queuedTooManyControlFrames`?

framer.go
```diff
@@ -82,6 +94,29 @@ func (f *framer) AppendControlFrames(frames []ackhandler.Frame, maxLen protocol.
...
+		length += fr.Frame.Length(v)
+		if hasMore {
+			// It is rare that a stream has more than one control frame to queue.
+			// We don't want to spawn another loop for just to cover that case.
+			goto start
+		}
+	}
+
 	for len(f.controlFrames) > 0 {
 		frame := f.controlFrames[len(f.controlFrames)-1]
```
sukunrt, line 121: As we pop from the back of the slice, would this mean that we send `MAX_STREAMS_FRAME` in reverse order higher before lower? This is not a protocol violation so it's fine. I'm just curious.

sukunrt (approved):

framer.go
marten-seemann: Not quite yet. There are still a few other frame types that use the control frame queue: 
* connection ID management: NEW_CONNECTION_ID, RETIRE_CONNECTION_ID
* stream limit: MAX_STREAMS, STREAMS_BLOCKED
* connection-level flow control: MAX_DATA and the corresponding BLOCKED frame
* some frames that are sent very rarely: NEW_TOKEN, HANDSHAKE_DONE

The plan is to make sending of these frames more explicit. This seems more tractable now that the streams don't use the frame queue anymore.

framer.go
marten-seemann: This is not indicative of any error. For example, the receive side of a stream could send a MAX_STREAM_DATA frame, while the send side of the wishes to send a RESET_STREAM frame.

There are also situations where the same side of the stream might enqueue itself multiple times: For example, the send side might want to send a BLOCKED frame, and the user calls `CancelWrite` right after that (which would make the stream register to send a RESET_STREAM frame).

framer.go
marten-seemann: Yes indeed :)

marten-seemann: Done.

framer.go
marten-seemann: Possibly! Another reason to rewrite this code path: We should only ever send a single MAX_STREAMS frame (the one with the highest value).

PR #4614, opened 2024-08-03
tighten verification logic for leftover data in crypto streams

marten-seemann:


marten-seemann:
This is not a very interesting fix. It was not exploitable in practice, since we'd immediately dequeue in-order CRYPTO data when receiving a CRYPTO frame. It is the basis for an interesting optimization though (#4615), and this PR has been folded into this PR.

PR #4616, opened 2024-08-03
remove trivial mock implementation of the cryptoStream interface

marten-seemann:


marten-seemann:
This change was removed (by force-push) from master, since it made the build fail.

PR #4630, opened 2024-08-08
http3: add HTTP Trailer support for servers

### LINKED ISSUE #2266: HTTP trailer support

sudorandom:
Closes #2344. Fixes #2266.

---

This is a continuation of the work done in https://github.com/quic-go/quic-go/pull/2344 and will resolve the server half of https://github.com/quic-go/quic-go/issues/2266. This adds support for writing trailers and a functional test which verifies the basic functionality of trailers for the quic-go http3 client and server. Because of that, this PR rely on #4581.

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -251,15 +252,6 @@ func (w *responseWriter) promoteTrailer() {
 	}
 }
 
-func strSliceContains(ss []string, s string) bool {
-	for _, v := range ss {
-		if v == s {
-			return true
-		}
-	}
-	return false
-}
-
 func (w *responseWriter) declareTrailer(k string) {
 	k = http.CanonicalHeaderKey(k)
 	if !httpguts.ValidTrailerHeader(k) {
@@ -267,7 +259,8 @@ func (w *responseWriter) declareTrailer(k string) {
 		w.logger.Debug("ignoring invalid trailer", slog.String("header", k))
 		return
 	}
-	if !strSliceContains(w.trailers, k) {
+
+	if !slices.Contains(w.trailers, k) {
 		w.trailers = append(w.trailers, k)
 	}
 }
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -269,7 +269,7 @@ func (w *responseWriter) Flush() {
 	}
 }
 
-// declareTrailer will add the given to the trailer list, while also validating that the trailer has a
+// declareTrailer adds a trailer to the trailer list, while also validating that the trailer has a
 // valid name.
 func (w *responseWriter) declareTrailer(k string) {
 	if !httpguts.ValidTrailerHeader(k) {
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"net/textproto"
 	"strconv"
 	"strings"
 	"time"
@@ -205,14 +206,25 @@ func (w *responseWriter) writeHeader(status int) error {
 
 	// Handle trailer fields
 	if vals, ok := w.header["Trailer"]; ok {
+		trailers := make([]string, 0, len(vals))
 		for _, val := range vals {
-			w.declareTrailer(val)
+			for _, trailer := range strings.Split(val, ",") {
+				// We need to convert to the canonical header key value here because this will be called when using
+				// headers.Add or headers.Set.
+				trailer = textproto.CanonicalMIMEHeaderKey(strings.TrimSpace(trailer))
+				w.declareTrailer(trailer)
+				trailers = append(trailers, trailer)
+			}
 		}
+		w.header["Trailer"] = trailers
 	}
-	w.promoteTrailers()
 
 	for k, v := range w.header {
 		if _, excluded := w.trailers[k]; excluded {
+			continue
+		}
+		// Ignore "Trailer:" prefixed headers
+		if strings.HasPrefix(k, http.TrailerPrefix) {
 			continue
 		}
 		for index := range v {
@@ -237,29 +249,22 @@ func (w *responseWriter) FlushError() error {
 	if _, err := w.doWrite(nil); err != nil {
 		return err
 	}
-	if !w.trailerWritten {
-		return w.writeTrailers()
-	}
 	return nil
+}
+
+func (w *responseWriter) flushFinal() {
+	w.Flush()
+	if !w.trailerWritten {
+		if err := w.writeTrailers(); err != nil {
+			w.logger.Debug("could not write trailers", "error", err)
+		}
+	}
 }
 
 func (w *responseWriter) Flush() {
 	if err := w.FlushError(); err != nil {
 		if w.logger != nil {
 			w.logger.Debug("could not flush to stream", "error", err)
-		}
-	}
-}
-
-// promoteTrailers will discover trailers prefixed with "Trailer:" and "promote" them to the list of
-// trailers. This should be called before writing headers (to avoid writing trailers in the headers list)
-// and before sending trailers, to ensure that we capture all of the trailers added between sending the
-// header and finally sending the trailers.
-func (w *responseWriter) promoteTrailers() {
-	for k := range w.header {
-		// Handle "Trailer:" prefix
-		if strings.HasPrefix(k, http.TrailerPrefix) {
-			w.declareTrailer(k)
 		}
 	}
 }
@@ -293,7 +298,14 @@ func (w *responseWriter) hasNonEmptyTrailers() bool {
 
 // writeTrailers will write trailers to the stream if there are any.
 func (w *responseWriter) writeTrailers() error {
-	w.promoteTrailers() // promote headers added via "Trailer:" convention as trailers
+	// promote headers added via "Trailer:" convention as trailers, these can be added after
+	// streaming the status/headers have been written.
+	for k := range w.header {
+		// Handle "Trailer:" prefix
+		if strings.HasPrefix(k, http.TrailerPrefix) {
+			w.declareTrailer(k)
+		}
+	}
 
 	if !w.hasNonEmptyTrailers() {
 		return nil
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -44,6 +44,7 @@ type responseWriter struct {
 	headerComplete bool  // set once WriteHeader is called with a status code >= 200
 	headerWritten  bool  // set once the response header has been serialized to the stream
 	isHead         bool
+	trailerWritten bool // set once the response trailers has been serialized to the stream
 
 	hijacked bool // set on HTTPStream is called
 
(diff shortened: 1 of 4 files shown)


http3/conn.go
```diff
@@ -112,8 +114,37 @@ func (c *connection) openRequestStream(
 	c.streams[str.StreamID()] = datagrams
 	c.streamMx.Unlock()
 	qstr := newStateTrackingStream(str, c, datagrams)
-	hstr := newStream(qstr, c, datagrams)
-	return newRequestStream(hstr, requestWriter, reqDone, c.decoder, disableCompression, maxHeaderBytes), nil
+	var rsp = new(http.Response)
```
marten-seemann, line 117: To make `gofumpt` happy:
```suggestion
	rsp := &http.Response{}
```

marten-seemann:
Oops, I added my fix `gofumpt` fix to the wrong PR.

marten-seemann:
@sudorandom Can you rebase this PR, now that #4581 was merged?

sudorandom:
@marten-seemann done!

http3/response_writer.go
```diff
@@ -232,6 +235,79 @@ func (w *responseWriter) Flush() {
...
+		if !strings.HasPrefix(k, http2.TrailerPrefix) {
+			continue
+		}
+		trailerKey := strings.TrimPrefix(k, http2.TrailerPrefix)
+		w.declareTrailer(trailerKey)
+		w.header[http.CanonicalHeaderKey(trailerKey)] = vv
+	}
+}
+
+func strSliceContains(ss []string, s string) bool {
```
marten-seemann, line 254: `slices.Contains`?

marten-seemann (commented):
What's the logic here? If I understand correctly, there are two different ways to declare trailers in Go. Are both implemented here?

http3/response_writer.go
sudorandom: I switched the implementation around to use a map instead! But if it's kept a list, obviously `slices.Contains` is the better method for this

sudorandom:
@marten-seemann okay, the gist is, there's a new call after `ServeHTTP` to `writeTrailers`. `writeTrailers` will send the trailers frame with trailers that it has collected. You are right, there are two methods of defining trailers.

The preferred way is to use a "Trailers" header with one value containing the name of the "header" that will actually be a trailer. Trailers defined like this are collected inside of `writeHeader`. Trailers defined this way need to be defined before `WriteHeader` is called by the http handler.

The less-preferred way is to add headers at any point of the handling a request flow with the prefix `Trailer:`. These trailers are collected in two places, in `writeHeader` and at the start of `writeTrailers`. It's needed in two places for these reasons:
- It's needed before writing our headers because we need to exclude these "headers" with names prefixed with`Trailer:` because they are not valid headers and aren't intended to be sent as headers.
- It's also needed right before writing trailers because up until then, the handler could decide to add trailers at any point after the header frame was sent.

Why on EARTH Go decided not to just use the Trailers attribute on the `http.Response` object on the server side is the biggest mystery to me.

I have an issue with the current code now, though. There are tests that are failing because I need the trailer code to be sent before sending the trailers, so I added a call to `Flush()`, but that is breaking other tests. I think there's probably a way around that though

http3/response_writer.go
```diff
@@ -200,7 +203,18 @@ func (w *responseWriter) writeHeader(status int) error {
...
+	// Handle trailer fields
+	if vals, ok := w.header["Trailer"]; ok {
+		for _, val := range vals {
+			w.declareTrailer(val)
+		}
+	}
+	w.promoteTrailers()
+
 	for k, v := range w.header {
+		if _, excluded := w.trailers[k]; excluded {
```
marten-seemann, line 215: Would it make sense to fold the code of `promoteTrailers` here? Then we only need to iterate over the header map once.

marten-seemann (commented):
Sorry for the delay in reviewing this PR, I first had to wrap my head around announced and unannounced headers, and how they're handled in Go.

This code looks pretty clean to me! I suggest making the integration test a little bit stricter, to make sure that we're mirroring Go's HTTP/1.1 behavior.

I used this [Gist](https://gist.github.com/marten-seemann/29a6a7fef4bdb11b78efad52b72e4970) to play around with header in HTTP/1.1, in case that it helps.

integrationtests/self/http_test.go
```diff
@@ -89,6 +89,16 @@ var _ = Describe("HTTP tests", func() {
 			w.Header().Set("X-RemoteAddr", r.RemoteAddr)
 			w.WriteHeader(http.StatusOK)
 		})
+		mux.HandleFunc("/trailers", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
sudorandom: Done. I mostly used gRPC because it was just a well-known use-case of trailers. I changed it to be more generic

integrationtests/self/http_test.go
```diff
@@ -89,6 +89,16 @@ var _ = Describe("HTTP tests", func() {
...
+			defer GinkgoRecover()
+			w.Header().Set("Trailer", "Grpc-Status")
+			w.Header().Add("trailer", "Grpc-Message")
+			w.WriteHeader(200)
+			io.WriteString(w, "Hello, World!\n") // don't check the error here. Stream may be reset.
+			defer w.Header().Set("Grpc-Status", "10")
+			defer w.Header().Set("Grpc-Message", "Message here")
+			defer w.Header().Set(http.TrailerPrefix+"Grpc-Details", "details")
```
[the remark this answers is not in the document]
sudorandom: 👍

integrationtests/self/http_test.go
```diff
@@ -1020,4 +1030,19 @@ var _ = Describe("HTTP tests", func() {
...
+		resp, err := client.Get(fmt.Sprintf("https://localhost:%d/trailers", port))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(resp.StatusCode).To(Equal(200))
+		body, err := io.ReadAll(gbytes.TimeoutReader(resp.Body, 3*time.Second))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(body)).To(Equal("Hello, World!\n"))
+		Expect(resp.Header).To(Not(HaveKey("Grpc-Status")))
+		Expect(resp.Header).To(Not(HaveKey("Grpc-Message")))
+		Expect(resp.Header).To(Not(HaveKey("Grpc-Details")))
+		Expect(resp.Header).To(Not(HaveKey("Trailer:Grpc-Details")))
```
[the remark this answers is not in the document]
sudorandom: I added this test. I didn't have this before because it definitely explodes on the client side because colon is definitely not a valid character for headers. In either case, it's still good to have an explicit check. 👍

http3/response_writer.go
sudorandom: Yeah, good thought. I just added this

http3/response_writer.go
```diff
@@ -220,8 +246,19 @@ func (w *responseWriter) FlushError() error {
...
 	}
-	_, err := w.doWrite(nil)
-	return err
+	if _, err := w.doWrite(nil); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (w *responseWriter) flushFinal() {
```
sudorandom, line 255: I added this function because developers should be able to call Flush() several times in a handler, so we can't actually send the trailers after each flush. This function is just intended to be called at the very end of the request, called by `http3.Server`.

http3/response_writer.go
```diff
@@ -200,7 +204,29 @@ func (w *responseWriter) writeHeader(status int) error {
 		return err
 	}
 
+	// Handle trailer fields
+	if vals, ok := w.header["Trailer"]; ok {
+		trailers := make([]string, 0, len(vals))
+		for _, val := range vals {
+			for _, trailer := range strings.Split(val, ",") {
```
sudorandom, line 211: I didn't realize you can comma-separate the trailer names here, thanks for the example of that @marten-seemann. This code is a bit uglier but it now handles this better.

integrationtests/self/http_test.go
```diff
@@ -1020,4 +1030,19 @@ var _ = Describe("HTTP tests", func() {
 			Expect(num0RTTPackets.Load()).To(BeNumerically(">", 0))
 		})
 	})
+
+	It("sends and receives trailers", func() {
+		resp, err := client.Get(fmt.Sprintf("https://localhost:%d/trailers", port))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(resp.StatusCode).To(Equal(200))
+		body, err := io.ReadAll(gbytes.TimeoutReader(resp.Body, 3*time.Second))
+		Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
sudorandom: Done! I did this by just asserting on the entire trailers map, so this should catch both cases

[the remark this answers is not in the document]
marten-seemann: What I was going for is the following: Before reading the body, the trailer _keys_ (of the trailers that were announced) should already be available. That's how the standard library HTTP/1.1 trailer implementation works. And it's potentially useful for the application to know which trailer values will be available later.

However, I noticed that this doesn't work yet on the client side. I suggest we do this in a follow-up PR.

http3/response_writer.go
```diff
@@ -220,8 +246,19 @@ func (w *responseWriter) FlushError() error {
 	if !w.headerComplete {
 		w.WriteHeader(http.StatusOK)
 	}
-	_, err := w.doWrite(nil)
-	return err
+	if _, err := w.doWrite(nil); err != nil {
+		return err
+	}
+	return nil
```
marten-seemann, line 252:
```suggestion
	_, err := w.doWrite(nil)
	return err
```

http3/response_writer.go
marten-seemann: Is this because it's valid to call `http.ResponseWriter.Flush` before sending the entire response body?

Maybe make the name more descriptive?
```suggestion
func (w *responseWriter) flushTrailers() {
```

marten-seemann (commented):
I think the trailer list handling is not quite correct, see my comment below. Other than this, this PR looks good to me.

integrationtests/self/http_test.go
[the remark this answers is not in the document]
marten-seemann: Tracked in #4654.

http3/response_writer.go
marten-seemann: @sudorandom Should we revert the `strings.Split` change introduced in cffa7c41ba2258d2a00a16fa8845dd06c566ac8b?

sudorandom: I think this part is still required so the server can know which headers are actually trailers to exclude them in the header frame and include them into the trailers frame. I already changed this code to avoid modifying the "Trailer" header but I do think this code is required to handle [this example](https://pkg.go.dev/net/http#example-ResponseWriter-Trailers) from the server side.

I looked a bit at the net/http implementation, and it also has to do something similar with splitting values from the Trailer header before sending the headers on the server side:
- https://github.com/golang/go/blob/master/src/net/http/response.go#L297
- https://github.com/golang/go/blob/master/src/net/http/transfer.go#L544
- https://github.com/golang/go/blob/master/src/net/http/transfer.go#L786

As a note, `foreachHeaderElement` actually handles comma-separated header values.

marten-seemann: You're right, we need to split the list to avoid sending out the value as a header.

Do we need to make any changes here to resolve https://github.com/quic-go/quic-go/pull/4656#discussion_r1743240176?

sudorandom: No, I can resolve it in #4656

marten-seemann:
I just ran this code (https://gist.github.com/marten-seemann/a525014fe9aca032d3059c08a8810076) and decrypted the trace with Wireshark. Everything looks correct to me:

image

marten-seemann (approved):
Thank you @sudorandom! Great work pushing this through!

Issue #4637, opened 2024-08-21
http3 RoundTripper does not support Close during Read, inconsistent with http/http2


mmmray:
*copied from https://github.com/quic-go/quic-go/issues/2143#issuecomment-2301652942*

Both h1/h2 transports handle Close+Read without crashing, but h3 does not. This inconsistency makes it difficult to handle all HTTP versions transparently in application code that already relies on this behavior. While not explicitly documented, `net.TCPConn` and some other networking APIs also support this usecase.

The workaround I have so far is to wrap `response.Body` in this struct:

```go

type h3Body struct {
	inner  io.ReadCloser
	lock   sync.Mutex
	cancel context.CancelFunc
}

func (c *h3Body) Read(b []byte) (int, error) {
	c.lock.Lock()
	defer c.lock.Unlock()

	return c.inner.Read(b)
}

func (c *h3Body) Close() error {
	c.cancel()

	c.lock.Lock()
	defer c.lock.Unlock()

	return c.inner.Close()
}

ctx, ctxCancel := context.WithCancel(context.Background())
req, err := http.NewRequestWithContext(ctx, ...)

...

response, err := client.Do(req)

...

body := h3Body { inner: response.Body, cancel: ctxCancel }
```

But clearly this comes with some overhead on Read. I can also only `cancel()` without `inner.Close` to avoid the mutex, but that seems like it might cause more bugs.

marten-seemann:
Your workaround code is completely useless for debugging this issue. Instead of posting your workaround, please post a **minimal** example that triggers the alleged bug.

mmmray:
@marten-seemann there is nothing to debug, and this is not a bug. This was an explicit API decision made by you 5 years ago, and I'd like you to reconsider that decision for the reasons stated above. This issue is an exact duplicate of https://github.com/quic-go/quic-go/issues/2143. Nevertheless, here is some demo code that triggers the issue for me:

```go
package main

import (
	"net/http"
    "sync"
    "context"
    "fmt"
    "io"
	"github.com/quic-go/quic-go/http3"
)
func main() {
	roundTripper := &http3.RoundTripper{}
	defer roundTripper.Close()
	hclient := &http.Client{
		Transport: roundTripper,
	}

    var wg sync.WaitGroup
    wg.Add(10000)

    for i := 0; i < 10000; i++ {
        go func() {
            req, err := http.NewRequestWithContext(
                context.Background(),
                "GET",
[...]

PR #2144, opened 2019-09-17
fix race condition in body.requestDone

fbzhong:
Fixes #2143 .

Issue #2143, opened 2019-09-17
race condition in http3.body.requestDone()


fbzhong:
I encountered a crash bug, call stack is the following:
```
panic: close of closed channel

goroutine 83814 [running]:
github.com/lucas-clemente/quic-go/http3.(*body).requestDone(...)
	github.com/lucas-clemente/quic-go/http3/body.go:85
github.com/lucas-clemente/quic-go/http3.(*body).Close(0x400128d9b0, 0x4001223140, 0x48400)
	github.com/lucas-clemente/quic-go/http3/body.go:94 +0x5c
client.(*DeadlineReader).Close(0x40005e0140, 0x216cd4ff97d2, 0x745c043800)
	client/deadline_readwriter.go:72 +0x60
```

body.go:85
https://github.com/lucas-clemente/quic-go/blob/07ee05b5a967d37650cb32109f9dc15d72e300f7/http3/body.go#L81-L87

The code of client/deadline_readwriter.go:72
```
func (r *DeadlineReader) Close() (err error) {
	select {
	case <-r.closed:
		return ErrClosedPipe
	default:
	}

	close(r.closed)

	if r.lastResult != nil {
		r.bufPool.Put(r.lastResult.buf)
		r.lastResult = nil
	}

	return r.rd.Close()
}
```

I take a look at the body.go. The `body.requestDone()` are called at two position:
1. `body.Close()`
2. 'body.Read()`
So when one goroutine call Read(), the other is calling Close(), and there is no lock inside `requestDone()`, crash will happen.

marten-seemann:
Thanks for finding this.

There are two things that are suboptimal in this code:
1. `Read` should probably check that the error is not `io.EOF` before calling `requestDone`.
2. `requestDone` should use a `sync.Once` instead of a `bool`.

fbzhong:
Ok, I will send a PR later soon.

fbzhong:
@marten-seemann I think we should call `requestDone` whenever error occurs. For a blocking reader, `io.EOF` means reader is at end, it's time to close it.

marten-seemann:
It actually doesn't make a big difference, since the caller will call `http.Response.Close()` anyway after reading the body.

marten-seemann:
Actually, thinking about this a bit more, I don't think this is a bug in quic-go after all. 

`Read` and `Close` shouldn't be called concurrently, so it's not quic-go's fault if you introduce a race condition like this.

fbzhong:
[...]

PR #4638, opened 2024-08-21
http3: Handle concurrent Body.Close more gracefully

mmmray:
Fix https://github.com/quic-go/quic-go/issues/4637 https://github.com/quic-go/quic-go/issues/2143

Re-apply https://github.com/quic-go/quic-go/pull/2144

Calling `response.Body.Close()` concurrently with `response.Body.Read()` (or concurrently with another `Close()`) is not defined behavior by `io.ReadCloser`, but too much code relies on the current behavior of `http2.Transport`. At least don't crash due to double-closed channel.

While it's still not guaranteed that the `Read` is cancelled, that part actually seems to work as well with this PR.

Since this is a race condition, a unittest for this seems too resource-intensive for the testsuite, but happy to write one.

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"sync"
 
 	"github.com/quic-go/quic-go"
 )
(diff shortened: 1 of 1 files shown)


mmmray:
I'm sorry, I didn't know this project has Google CLA. Due to personal reasons I cannot use a Google account to contribute to this project.

PR #4639, opened 2024-08-23
http3: reject pseudo header fields in trailers

marten-seemann:
@sudorandom Would you mind reviewing this PR?

### DIFF
--- a/http3/conn.go
+++ b/http3/conn.go
@@ -116,7 +116,7 @@ func (c *connection) openRequestStream(
 	qstr := newStateTrackingStream(str, c, datagrams)
 	rsp := &http.Response{}
 	hstr := newStream(qstr, c, datagrams, func(r io.Reader, l uint64) error {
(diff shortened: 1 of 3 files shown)


http3/conn.go
```diff
@@ -139,11 +139,7 @@ func (c *connection) parseTrailer(r io.Reader, l, maxHeaderBytes uint64) (http.H
 	if err != nil {
 		return nil, err
 	}
-	h := http.Header{}
-	for _, field := range fields {
-		h.Add(field.Name, field.Value)
-	}
-	return h, nil
+	return parseTrailers(fields)
```
sudorandom, line 142: NIT: Now that there's `parseTrailers`, the naming of `connect.parseTrailer` seems to not quite fit. Maybe that should be `connect.decodeTrailers` ?

sudorandom (approved):
Looks good!

http3/conn.go
marten-seemann: Done.

Issue #4622, opened 2024-08-04
http3: uploading a body without a predefined content-length cause the body to be skipped


Ousret:
Originally coming from https://github.com/traefik/traefik/issues/10185

We spotted an issue where streaming a body without a predefined content-length cause Traefik to ignore the body.
The question I am raising here is: "Is this a quic-go bug or rather a mistake in Traefik usage of quic-go? or worst, our implementation is faulty?"

Streaming a body using http3 without content-length should wait for the FIN bit in the given stream.

Please find attached the reproducer: [reproducer.zip](https://github.com/user-attachments/files/16429164/reproducer.zip)
Unfortunately I am unable to create a reproducer solely using (isolated) quic-go.

expected output:

![image](https://github.com/user-attachments/assets/0bef897d-b88c-4d41-b2a3-bdaba55194af)

regards,

marten-seemann:
I'm not really sure how your setup looks like, but if a request don't have a Content-Length, you'll need to read until the end of the stream. Are you saying that quic-go is incorrectly setting a Content-Length of 0?

Ousret:
OK, I will try to explain myself better.

First and foremost, I am not a go developer. I happen to have a QUIC/H3 implementation built for Python using aws-lc-rs crypto layer.

And for our CI, we happen to test our implementation against Traefik, which itself depend on quic-go to handle the QUIC layer, and of course the HTTP/3 state-machine on top of it. As far as I am concerned, we have zero evidence that the issue is coming from quic-go. We are seeking an expert to help us solve this case pending for a year now.

We noticed a bug in the following scenario:

POST /post (no content-length) --------> traefik http3 entrypoint ---(translated to http1.1)----> httpbin container

---

here are the header sent (excl special headers) on all cases

```
user-agent: niquests/3.7.2
accept-encoding: gzip, deflate
accept: */*
```

### HTTP/3

Here is what "httpbin" container receive in http1.1 behind a traefik router

```
'headers': {'Accept': ['*/*'],
             'Accept-Encoding': ['gzip, deflate'],
             'Content-Length': ['0'],
             'Host': ['httpbin.local:4443'],
             'User-Agent': ['niquests/3.7.2'],
             'X-Forwarded-For': ['172.24.0.1'],
             'X-Forwarded-Host': ['httpbin.local:4443'],
             'X-Forwarded-Port': ['4443'],
             'X-Forwarded-Proto': ['https'],
[...]

PR #4645, opened 2024-08-27
http3: set ContentLength to -1 if no Content-Length header is set

marten-seemann:
This applies to both the `http.Response` and the `http.Request`.

### DIFF
--- a/http3/http_stream.go
+++ b/http3/http_stream.go
@@ -238,11 +238,7 @@ func (s *requestStream) ReadResponse() (*http.Response, error) {
 
 	// Check that the server doesn't send more data in DATA frames than indicated by the Content-Length header (if set).
 	// See section 4.1.2 of RFC 9114.
-	contentLength := int64(-1)
-	if _, ok := res.Header["Content-Length"]; ok && res.ContentLength >= 0 {
-		contentLength = res.ContentLength
-	}
-	respBody := newResponseBody(s.stream, contentLength, s.reqDone)
+	respBody := newResponseBody(s.stream, res.ContentLength, s.reqDone)
 
 	// Rules for when to set Content-Length are defined in https://tools.ietf.org/html/rfc7230#section-3.3.2.
 	_, hasTransferEncoding := res.Header["Transfer-Encoding"]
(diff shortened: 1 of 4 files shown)


http3/http_stream.go
```diff
@@ -238,11 +238,7 @@ func (s *requestStream) ReadResponse() (*http.Response, error) {
...
 	// See section 4.1.2 of RFC 9114.
-	contentLength := int64(-1)
-	if _, ok := res.Header["Content-Length"]; ok && res.ContentLength >= 0 {
-		contentLength = res.ContentLength
-	}
-	respBody := newResponseBody(s.stream, contentLength, s.reqDone)
+	respBody := newResponseBody(s.stream, res.ContentLength, s.reqDone)
 
 	// Rules for when to set Content-Length are defined in https://tools.ietf.org/html/rfc7230#section-3.3.2.
 	_, hasTransferEncoding := res.Header["Transfer-Encoding"]
```
bt90, line 244: Shouldn't we ignore this header?

marten-seemann: Why?

bt90: https://github.com/quic-go/quic-go/issues/4622#issuecomment-2313198937

marten-seemann: @bt90 I think this you're correct. This was introduced by @WeidiDeng. @WeidiDeng, can you please check? I think the HTTP/3 RFC is pretty unambiguous with respect to the Transfer-Encoding. What was the motivation for adding this?

WeidiDeng: image

This change was not made by me. But I am pretty sure `Transfer-Encoding` is not necessary in http/2 & http/3 because of the data frame wire format. In caddy, we were discussing the meaning of the `Transfer-Encoding` for responses to range requests in the 1.23 [changelog](https://tip.golang.org/doc/go1.23). And I found out it's not possible for http2 & http3.

marten-seemann: I see, you just moved them in https://github.com/quic-go/quic-go/pull/4412. Sorry for the noise.

marten-seemann: Fix in #4646.

marten-seemann:
Added some more tests. This should be ready to merge once CI passes.

PR #4646, opened 2024-08-28
http3: reject the Transfer-Encoding header field

marten-seemann:
The use of Transfer-Encoding is forbidden in HTTP/3, see section 4.1 of RFC 9114.

@bt90 would you mind reviewing this PR?

### DIFF
--- a/http3/http_stream.go
+++ b/http3/http_stream.go
@@ -245,11 +245,10 @@ func (s *requestStream) ReadResponse() (*http.Response, error) {
 	respBody := newResponseBody(s.stream, contentLength, s.reqDone)
 
 	// Rules for when to set Content-Length are defined in https://tools.ietf.org/html/rfc7230#section-3.3.2.
-	_, hasTransferEncoding := res.Header["Transfer-Encoding"]
 	isInformational := res.StatusCode >= 100 && res.StatusCode < 200
 	isNoContent := res.StatusCode == http.StatusNoContent
 	isSuccessfulConnect := s.isConnect && res.StatusCode >= 200 && res.StatusCode < 300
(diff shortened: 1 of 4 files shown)


bt90 (approved):
Looks reasonable

http3/http_stream.go
```diff
@@ -245,11 +245,10 @@ func (s *requestStream) ReadResponse() (*http.Response, error) {
...
 	// Rules for when to set Content-Length are defined in https://tools.ietf.org/html/rfc7230#section-3.3.2.
-	_, hasTransferEncoding := res.Header["Transfer-Encoding"]
 	isInformational := res.StatusCode >= 100 && res.StatusCode < 200
 	isNoContent := res.StatusCode == http.StatusNoContent
 	isSuccessfulConnect := s.isConnect && res.StatusCode >= 200 && res.StatusCode < 300
-	if !hasTransferEncoding && !isInformational && !isNoContent && !isSuccessfulConnect {
+	if !isInformational && !isNoContent && !isSuccessfulConnect {
 		res.ContentLength = -1
 		if clens, ok := res.Header["Content-Length"]; ok && len(clens) == 1 {
 			if clen64, err := strconv.ParseInt(clens[0], 10, 64); err == nil {
```
bt90, line 254: Is there a reason why we're parsing the header a second time? Can't we use `contentLength` here?

bt90:
@marten-seemann i've created #4647 to track other headers we might want to reject

http3/http_stream.go
marten-seemann: No reason. But that's a fix for another PR :)

PR #4648, opened 2024-08-29
http3: avoid re-parsing of the Content-Length header for responses

marten-seemann:
There's a slight change in behavior here: For requests that don't have a body, we now still set the `http.Response.ContentLength` to `-1` (once we have #4645). I think this ok:
```
// ContentLength records the length of the associated content. The
// value -1 indicates that the length is unknown. Unless Request.Method
// is "HEAD", values >= 0 indicate that the given number of bytes may
// be read from Body.
ContentLength int
```

@bt90 @WeidiDeng What do you think?

### DIFF
--- a/http3/http_stream.go
+++ b/http3/http_stream.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strconv"
 
 	"github.com/quic-go/quic-go"
 	"github.com/quic-go/quic-go/internal/protocol"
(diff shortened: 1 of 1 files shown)


bt90:
Talking about the principle of least surprise: if there's no body, I'd be surprised if the length was marked as unknown. A missing body has a known length of 0 bytes. Or is there a reason why we need to distinguish between no body and an empty body?

marten-seemann:
It just means we can't remove the lines I'm removing here in this PR. I'll update it.

marten-seemann:
It's in [fd04f71](https://github.com/quic-go/quic-go/pull/4648/commits/fd04f710d7a1768012e2ca327f1609bc3e3787b3). It's kind of ugly.

bt90:
I haven't looked into it, but RFC7230 has largely been superseded by RFC9110.

https://datatracker.ietf.org/doc/html/rfc9110#name-content-length

bt90:
> It's in [fd04f71](https://github.com/quic-go/quic-go/pull/4648/commits/fd04f710d7a1768012e2ca327f1609bc3e3787b3). It's kind of ugly.

Ah, i see. So there are three cases:

* no `Content-Length` header: either the size is unknown or it's not allowed for this kind of response
* `Content-Length: 0` for an empty body
* `Content-Length: x` if a body of known length is present

I'd agree that the `-1` value is probably the cleanest solution.

marten-seemann:
> It's in [fd04f71](https://github.com/quic-go/quic-go/pull/4648/commits/fd04f710d7a1768012e2ca327f1609bc3e3787b3). It's kind of ugly.

This commit was slightly wrong, I messed up inverting the boolean expression. It's still not pretty, but it's not worse than what we had before, so I'm inclined to just merge it as is.

PR #4655, opened 2024-09-01
http3: reject connection-specific header fields, check value of TE

### LINKED ISSUE #4647: http3: reject unsupported headers
I think we're still allowing a few headers which are violating [RFC9113](https://datatracker.ietf.org/doc/html/rfc9113#name-connection-specific-header-) and [RFC9114](https://datatracker.ietf.org/doc/html/rfc9114#name-http-fields)

* `Connection`
* `TE` with any other value as `trailers`
* `Keep-Alive`
* `Proxy-Connection`
* `Upgrade`

According to the spec they should result in a stream error of type [H3_MESSAGE_ERROR](https://datatracker.ietf.org/doc/html/rfc9114#H3_MESSAGE_ERROR)

marten-seemann:
Fixes #4647.

### DIFF
--- a/http3/headers.go
+++ b/http3/headers.go
@@ -28,6 +28,15 @@ type header struct {
 	Headers http.Header
 }
 
+// connection-specific header fields must not be sent on HTTP/3
+var invalidHeaderFields = [...]string{
+	"connection",
+	"keep-alive",
+	"proxy-connection",
+	"transfer-encoding",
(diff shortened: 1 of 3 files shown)


marten-seemann:
@bt90 Could I ask you for a review of this PR?

bt90 (approved):

PR #4656, opened 2024-09-01
http3: initialize trailer map with empty trailer entries when parsing the header

### LINKED ISSUE #4654: http3: initialize trailer map with empty trailer entries when parsing the header
What I was going for is the following: Before reading the body, the trailer _keys_ (of the trailers that were announced) should already be available. That's how the standard library HTTP/1.1 trailer implementation works. And it's potentially useful for the application to know which trailer values will be available later.

However, I noticed that this doesn't work yet on the client side. I suggest we do this in a follow-up PR.

_Originally posted by @marten-seemann in https://github.com/quic-go/quic-go/pull/4630#discussion_r1739969283_

sudorandom:
This should resolve #4654

This RP changes the client logic to prepopulates `Response.Trailer` using the trailers defined in the "Trailer" header and then removes the trailer header, like net/http does.

This is based off of #4630, so only the last few commits are unique.

### DIFF
--- a/http3/headers.go
+++ b/http3/headers.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/textproto"
 	"net/url"
 	"strconv"
 	"strings"
@@ -222,6 +223,7 @@ func updateResponseFromHeaders(rsp *http.Response, headerFields []qpack.HeaderFi
 	rsp.Proto = "HTTP/3.0"
 	rsp.ProtoMajor = 3
 	rsp.Header = hdr.Headers
+	formatTrailers(rsp)
 	rsp.ContentLength = hdr.ContentLength
 
 	status, err := strconv.Atoi(hdr.Status)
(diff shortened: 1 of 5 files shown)


http3/headers.go
```diff
@@ -203,3 +205,21 @@ func updateResponseFromHeaders(rsp *http.Response, headerFields []qpack.HeaderFi
 	rsp.Status = hdr.Status + " " + http.StatusText(status)
 	return nil
 }
+
+// formatTrailers parses trailers header.
+// For example: {"Trailer": {"Trailer1, Trailer2", "Trailer3"}}
+// Should be converted to: {"Trailer": {"Trailer1", "Trailer2", "Trailer3"}}
+func formatTrailers(rsp *http.Response) {
```
marten-seemann, line 212:
```suggestion
// processTrailers initializes the rsp.Trailer map, and adds keys for every announced header value.
// The Trailer header is removed from the http.Response.Header map.
// It handles both duplicate as well as comma-separated values for the Trailer header.
// For example:
//     Trailer: Trailer1, Trailer2
//     Trailer: Trailer3
// Will result in a http.Response.Trailer map containing the keys "Trailer1", "Trailer2", "Trailer3".
func processTrailers(rsp *http.Response) {
```

marten-seemann (commented):
LGTM modulo one naming / comment suggestion.

integrationtests/self/http_test.go
```diff
@@ -1042,11 +1043,18 @@ var _ = Describe("HTTP tests", func() {
 		resp, err := client.Get(fmt.Sprintf("https://localhost:%d/trailers", port))
 		Expect(err).ToNot(HaveOccurred())
 		Expect(resp.StatusCode).To(Equal(200))
-		Expect(resp.Header.Values("Trailer")).To(Equal([]string{"AtEnd1, AtEnd2", "LAST"}))
+		Expect(resp.Header.Get("Trailer")).To(Equal(""))
```
marten-seemann: This doesn't match Go's standard library behavior, see the screenshot in https://github.com/quic-go/quic-go/pull/4630#discussion_r1739962302.

marten-seemann (changes requested):

integrationtests/self/http_test.go
sudorandom: Actually, I think it does. Over the wire, the trailers are sent but on the client side the "Trailer" header is processed and then removed from the list of headers before being given to the caller. See here: https://github.com/golang/go/blob/2707d42966f8985a6663c93e943b9a44b9399fca/src/net/http/transfer.go#L781

marten-seemann:
@sudorandom I just merged #4630. Can you rebase this PR?

sudorandom:
> @sudorandom I just merged #4630. Can you rebase this PR?

Yep, rebased!

integrationtests/self/http_test.go
marten-seemann: You're right, sorry for the confusion.

marten-seemann (approved):

PR #4659, opened 2024-09-03
timer: fix behavior for new go1.23 timer semantics

sukunrt:
New unbuffered timer introduced in go1.23 breaks the existing implementation specifically the part where the timer channel will not have any stale reads after `timer.Stop` returns.  

To verify
1.run kubo 0.30.0-rc2 for 10 minutes. 
2. Now try stopping the daemon with ctrl+c. It'll fail(not exit) because a few goroutines are stuck on https://github.com/quic-go/quic-go/blob/master/internal/utils/timer.go#L35. You can put a print statement there to verify that's what's exactly happening. 

From reading all the documentation, I expected this to work as it is because if we haven't read from the timer channel `timer.Stop` should return true. But that's not the case(possibly a golang bug?).

It can happen that `timer.Stop` returns false and we haven't received anything from the timer channel. This happens when the timer is [executed](https://github.com/golang/go/blob/master/src/runtime/time.go#L971) concurrently with [timer.Stop](https://github.com/golang/go/blob/master/src/runtime/time.go#L334). 
This is specifically the section that leads to this behavior: https://github.com/golang/go/blob/master/src/runtime/time.go#L1033-L1050.

### DIFF
--- a/internal/utils/timer.go
+++ b/internal/utils/timer.go
@@ -29,13 +29,17 @@ func (t *Timer) Reset(deadline time.Time) {
 		return
 	}
 
-	// We need to drain the timer if the value from its channel was not read yet.
-	// See https://groups.google.com/forum/#!topic/golang-dev/c9UUfASVPoU
-	if !t.t.Stop() && !t.read {
-		select {
-		case <-t.t.C:
-		default:
+	// check if asynctimerchan behavior is enabled
+	// cap(t.t.C) == 1 implies old go1.22 timers
+	if cap(t.t.C) == 1 {
+		// We need to drain the timer if the value from its channel was not read yet.
+		// See https://groups.google.com/forum/#!topic/golang-dev/c9UUfASVPoU
+		if !t.t.Stop() && !t.read {
+			<-t.t.C
 		}
+	} else {
+		// for new go1.23 unbuffered timers no value will be present in the channel after Stop returns
+		t.Stop()
 	}
 	if !deadline.IsZero() {
 		t.t.Reset(time.Until(deadline))
--- a/internal/utils/timer.go
+++ b/internal/utils/timer.go
@@ -32,7 +32,10 @@ func (t *Timer) Reset(deadline time.Time) {
 	// We need to drain the timer if the value from its channel was not read yet.
 	// See https://groups.google.com/forum/#!topic/golang-dev/c9UUfASVPoU
 	if !t.t.Stop() && !t.read {
-		<-t.t.C
+		select {
+		case <-t.t.C:
+		default:
+		}
 	}
 	if !deadline.IsZero() {
 		t.t.Reset(time.Until(deadline))


marten-seemann:
Was this called out in the Go 1.23 release notes? This seems to break the Go backwards compatibility guarantee, doesn't it?

sukunrt:
It's not clear to me: https://go.dev/doc/go1.23

> Second, the timer channel associated with a Timer or Ticker is now unbuffered, with capacity 0. The main effect of this change is that Go now guarantees that for any call to a Reset or Stop method, no stale values prepared before that call will be sent or received after the call. Earlier versions of Go used channels with a one-element buffer, making it difficult to use Reset and Stop correctly. A visible effect of this change is that len and cap of timer channels now returns 0 instead of 1, which may affect programs that poll the length to decide whether a receive on the timer channel will succeed. Such code should use a non-blocking receive instead.

Note: this behavior only happens when go.mod has go version as 1.23 and not otherwise. 

It's difficult to share a small repro for the behavior. In small tests, I'm unable to trigger this race, but with kubo this reliably fails.

marten-seemann:
> It's not clear to me: https://go.dev/doc/go1.23
> 
> > Second, the timer channel associated with a Timer or Ticker is now unbuffered, with capacity 0. The main effect of this change is that Go now guarantees that for any call to a Reset or Stop method, no stale values prepared before that call will be sent or received after the call. Earlier versions of Go used channels with a one-element buffer, making it difficult to use Reset and Stop correctly. A visible effect of this change is that len and cap of timer channels now returns 0 instead of 1, which may affect programs that poll the length to decide whether a receive on the timer channel will succeed. Such code should use a non-blocking receive instead.

Does this mean that we don't need to wrap the timer at all?

sukunrt:
Yes, we can remove the wrapper completely, but only when go.mod go version is go1.23

sukunrt:
Converting this to draft. The proposed solution is racy for go versions < 1.23

MarcoPolo:
> From reading all the documentation, I expected this to work as it is because if we haven't read from the timer channel timer.Stop should return true. But that's not the case(possibly a golang bug?).

Just a small note. The docs for [Stop](https://cs.opensource.google/go/go/+/go1.23.0:src/time/sleep.go;l=113) say: "For a chan-based timer created with NewTimer(d), as of Go 1.23, **any receive from t.C after Stop has returned is guaranteed to block** rather than receive a stale time value from before the Stop".

sukunrt:
Just after that line: 

>if the program has not received from t.C already **and the timer is
running**, Stop is guaranteed to return true.


In our case, the timer is not running but we haven't received anything either. So `t.read` is false and reading from the channel blocks forever. 

I'm confused how to make this work with both new and old timers.

marten-seemann:
After discussing this with @sukunrt offline, no fix will be needed once https://github.com/golang/go/issues/69312 is resolved. Until then, users can use `GODEBUG="asynctimerchan=1"`, or avoid using Go 1.23. We might add a note to our release notes for the next release.

Thank you @sukunrt for digging into this!

PR #4661, opened 2024-09-05
protocol: use the packet number decoding / encoding logic from the RFC

marten-seemann:
The current packet number decoding / encoding logic stems from (a very old version) of Chrome's QUIC implementation. I don't understand it, especially the test cases.

Instead, we should just switch to what RFC 9000 defines. This will unblock #4650.


sukunrt (approved):
:rocket:

PR #4666, opened 2024-09-08
fix calculation of idle timeout when the peer set max_idle_timeout to 0

### LINKED ISSUE #4644: Incorrect handling of the idle timeout parameter in v0.46.0
According to RFC 9000, the max_idle_timeout parameter of transport parameters is used to negotiate the idle timeout mechanism's duration. However, when it is set to 0, it indicates that idle timeout is not enabled.

> max_idle_timeout (0x01):  The maximum idle timeout is a value in milliseconds that is encoded as an integer; see (Section 10.1).       Idle timeout is disabled when both endpoints omit this transport parameter or specify a value of 0.

However, in version v0.46.0, if max_idle_timeout is set to 0 by client, it actually sets the idle timer time to 0, causing the connection to be immediately deemed idle and closed after the first stream is closed.

marten-seemann:
Fixes #4644.

In that case, the idle timeout should be our local value, not 0.

@Blankwonder Could you take a look at this PR please?

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -1768,8 +1768,9 @@ func (s *connection) applyTransportParameters() {
 	params := s.peerParams
 	// Our local idle timeout will always be > 0.
 	s.idleTimeout = s.config.MaxIdleTimeout
-	if s.idleTimeout > 0 && params.MaxIdleTimeout < s.idleTimeout {
-		s.idleTimeout = params.MaxIdleTimeout
+	// If the peer advertised an idle timeout, take the minimum of the values.
+	if params.MaxIdleTimeout > 0 {
+		s.idleTimeout = min(s.idleTimeout, params.MaxIdleTimeout)
 	}
 	s.keepAliveInterval = min(s.config.KeepAlivePeriod, min(s.idleTimeout/2, protocol.MaxKeepAliveInterval))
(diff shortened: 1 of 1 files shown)


Blankwonder (approved):
This should solve the problem and align with the description in the RFC. Hope to further support setting MaxIdleTimeout to 0 in the future to disable the idle timer.

marten-seemann:
> This should solve the problem and align with the description in the RFC. Hopefully, future support will allow setting MaxIdleTimeout to 0 to disable the idle timer.

Why not just use a super large value (24h or more)? Should have the same effect, right?

Blankwonder:
> > This should solve the problem and align with the description in the RFC. Hopefully, future support will allow setting MaxIdleTimeout to 0 to disable the idle timer.
> 
> Why not just use a super large value (24h or more)? Should have the same effect, right?

Indeed, but setting it to 0 might be more elegant and can also avoid the overhead of the timer (although it's almost negligible).

marten-seemann:
> Indeed, but setting it to 0 might be more elegant and can also avoid the overhead of the timer (although it's almost negligible).

There's a timer running anyway, for things like pacing, loss recovery etc. The only saving I see is a few bytes when sending the transport parameters.

PR #4668, opened 2024-09-10
transport: allow for custom non-QUIC packet detection

bokysan:
Existing QUIC packed detection is quite rudimentary. In some advanced cases, the user might want to have more control over exact detection of QUIC and Non-QUIC packets.

This commit simply allows a function to be defined on the Transport which overrides this simple handling. Should no such function be defined, the processing falls back to the previous method.

### DIFF
--- a/internal/ackhandler/interfaces.go
+++ b/internal/ackhandler/interfaces.go
@@ -49,5 +49,5 @@ type ReceivedPacketHandler interface {
 	DropPackets(protocol.EncryptionLevel)
 
 	GetAlarmTimeout() time.Time
-	GetAckFrame(encLevel protocol.EncryptionLevel, onlyIfQueued bool) *wire.AckFrame
+	GetAckFrame(_ protocol.EncryptionLevel, now time.Time, onlyIfQueued bool) *wire.AckFrame
 }
--- a/internal/ackhandler/received_packet_handler.go
+++ b/internal/ackhandler/received_packet_handler.go
@@ -87,7 +87,7 @@ func (h *receivedPacketHandler) GetAlarmTimeout() time.Time {
 	return h.appDataPackets.GetAlarmTimeout()
 }
 
-func (h *receivedPacketHandler) GetAckFrame(encLevel protocol.EncryptionLevel, onlyIfQueued bool) *wire.AckFrame {
+func (h *receivedPacketHandler) GetAckFrame(encLevel protocol.EncryptionLevel, now time.Time, onlyIfQueued bool) *wire.AckFrame {
 	//nolint:exhaustive // 0-RTT packets can't contain ACK frames.
 	switch encLevel {
 	case protocol.EncryptionInitial:
@@ -101,7 +101,7 @@ func (h *receivedPacketHandler) GetAckFrame(encLevel protocol.EncryptionLevel, o
 		}
 		return nil
 	case protocol.Encryption1RTT:
-		return h.appDataPackets.GetAckFrame(onlyIfQueued)
+		return h.appDataPackets.GetAckFrame(now, onlyIfQueued)
 	default:
 		// 0-RTT packets can't contain ACK frames
 		return nil
--- a/internal/ackhandler/received_packet_handler_test.go
+++ b/internal/ackhandler/received_packet_handler_test.go
@@ -17,7 +17,8 @@ func TestGenerateACKsForPacketNumberSpaces(t *testing.T) {
 	sentPackets := NewMockSentPacketTracker(ctrl)
 	handler := newReceivedPacketHandler(sentPackets, utils.DefaultLogger)
 
-	sendTime := time.Now().Add(-time.Second)
+	now := time.Now()
+	sendTime := now.Add(-time.Second)
 	sentPackets.EXPECT().GetLowestPacketNotConfirmedAcked().AnyTimes()
 	sentPackets.EXPECT().ReceivedPacket(protocol.EncryptionInitial, sendTime).Times(2)
 	sentPackets.EXPECT().ReceivedPacket(protocol.EncryptionHandshake, sendTime).Times(2)
@@ -31,7 +32,7 @@ func TestGenerateACKsForPacketNumberSpaces(t *testing.T) {
 	require.NoError(t, handler.ReceivedPacket(4, protocol.ECNCE, protocol.Encryption1RTT, sendTime, true))
 
 	// Initial
-	initialAck := handler.GetAckFrame(protocol.EncryptionInitial, true)
+	initialAck := handler.GetAckFrame(protocol.EncryptionInitial, now, true)
 	require.NotNil(t, initialAck)
 	require.Equal(t, []wire.AckRange{{Smallest: 2, Largest: 3}}, initialAck.AckRanges)
 	require.Zero(t, initialAck.DelayTime)
@@ -40,7 +41,7 @@ func TestGenerateACKsForPacketNumberSpaces(t *testing.T) {
 	require.Zero(t, initialAck.ECNCE)
 
 	// Handshake
-	handshakeAck := handler.GetAckFrame(protocol.EncryptionHandshake, true)
+	handshakeAck := handler.GetAckFrame(protocol.EncryptionHandshake, now, true)
 	require.NotNil(t, handshakeAck)
 	require.Equal(t, []wire.AckRange{{Smallest: 1, Largest: 2}}, handshakeAck.AckRanges)
 	require.Zero(t, handshakeAck.DelayTime)
@@ -49,10 +50,10 @@ func TestGenerateACKsForPacketNumberSpaces(t *testing.T) {
 	require.Zero(t, handshakeAck.ECNCE)
 
 	// 1-RTT
-	oneRTTAck := handler.GetAckFrame(protocol.Encryption1RTT, true)
+	oneRTTAck := handler.GetAckFrame(protocol.Encryption1RTT, now, true)
 	require.NotNil(t, oneRTTAck)
 	require.Equal(t, []wire.AckRange{{Smallest: 4, Largest: 5}}, oneRTTAck.AckRanges)
-	require.InDelta(t, time.Second, oneRTTAck.DelayTime, float64(50*time.Millisecond))
+	require.Equal(t, time.Second, oneRTTAck.DelayTime)
 	require.Zero(t, oneRTTAck.ECT0)
 	require.Zero(t, oneRTTAck.ECT1)
 	require.EqualValues(t, 2, oneRTTAck.ECNCE)
@@ -71,7 +72,7 @@ func TestReceive0RTTAnd1RTT(t *testing.T) {
 	require.NoError(t, handler.ReceivedPacket(2, protocol.ECNNon, protocol.Encryption0RTT, sendTime, true))
 	require.NoError(t, handler.ReceivedPacket(3, protocol.ECNNon, protocol.Encryption1RTT, sendTime, true))
 
-	ack := handler.GetAckFrame(protocol.Encryption1RTT, true)
+	ack := handler.GetAckFrame(protocol.Encryption1RTT, time.Now(), true)
 	require.NotNil(t, ack)
 	require.Equal(t, []wire.AckRange{{Smallest: 2, Largest: 3}}, ack.AckRanges)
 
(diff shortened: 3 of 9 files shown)


marten-seemann:
Thanks for the PR, but I'm not sure which issue it is supposed to solve? Why submit code before discussing the problem and possible solutions? I'm not a fan of adding additional API surface, unless there is a **very good** justification.

> Existing QUIC packed detection is quite rudimentary

You might want to expand on this, and give concrete examples where the current code falls short of your expectations.

bokysan:
Hey @marten-seemann,

I'm glad you asked. I'm currently working on a software which works similarly to [sslh](https://github.com/yrutschle/sslh). Basically it detects the type of client request and routes the request accordingly.

Now, for TCP connections this is easy to do in go. For UDP connections as well, but not if `quic.Transport` handles the whole UDP connection. Specifically, since the code inspects the first byte only, it will make some incorrect assumptions if the given packet is a QUIC packet or not. E.g. a lot of time it will assume that a DNS request is a QUIC packet, when in fact it is not.

Consequence is that `ReadNonQUICPacket` gets called sporadically.

I do understand that this is an edge case and that's why I did not want to change the code to much. To the best of my knowledge:
- this fix works
- all tests pass

An alternative would be to implement a proper QUIC packet detection inside the code and process really _just_ QUIC packets, but I felt this would be outside of the scope of the project.

marten-seemann:
Thank you for the explanation!

Could you share your custom implementation of `IsNonQUICPacket`? I'd like to get a feeling for how you'd use it. 

I'm wondering if there's something we could do in quic-go in order to improve our detection logic that would cover your use case.

> An alternative would be to implement a proper QUIC packet detection inside the code and process really _just_ QUIC packets, but I felt this would be outside of the scope of the project.

Yeah, the problem here that the QUIC `connection` struct handles packets async, and there's no obvious way to hand a packet back to the `Transport` if handling fails (and there are many code paths where handling can fail!).

transport.go
```diff
@@ -104,6 +104,10 @@ type Transport struct {
 	// Tracer.Close is called when the transport is closed.
 	Tracer *logging.Tracer
 
+	// IsNonQuicPacket is a custom function that can be set to override the detection of
+	// non-QUIC packets. If not set, default DefaultIsNonQuicPacket handler is used
+	IsNonQuicPacket func(packet []byte) bool
```
marten-seemann, line 109:
```suggestion
	IsNonQUICPacket func(packet []byte) bool
```

Same for `DefaultIsNonQUICPacket` below.

bokysan:
> I'm wondering if there's something we could do in quic-go in order to improve our detection logic that would cover your use case.

The code is a bit more complex, but basically loops through different protocols and tries to determine if the given array of bytes:
- is a specific protocol
- isn't a specific protocol
- there isn't enough data to make an informed decision

I wouldn't want to paste the whole detection logic here. It's outside the scope of this ticket, but, just an example, the code below is used to detect the DNS packet. I don't think it makes sense to add this kind of detection logic inside quic-go, but using something similar to detect QUIC packets only would perhaps make sense.

```go
package protocols

var DnsProtocol = &Protocol{
	Name:                  "DNS",
	Layer3ProtocolSupport: ProtocolStream | ProtocolPacket,
	Wrap:                  nil,
	MaxBytesNeeded:        14,
	IsProtocol: func(header []byte, protocolType Layer3ProtocolType) (bool, error) {
		if protocolType == ProtocolStream {
			// TCP requests will be prefixed with packet length, little endian
			if len(header) < 14 {
				return false, ErrNotEnoughData
			}

			size := uint16(header[0])<<8 | uint16(header[1])
			if size < 12 {
				return false, ErrProtocolNotDetected
			}

			header = header[2:]
		} else {
			if len(header) < 12 {
				return false, ErrNotEnoughData
			}
		}

		// https://stackoverflow.com/questions/7565300/identifying-dns-packets
		// Header section format
		//
		//The header contains the following fields:
		//
		//                                    1  1  1  1  1  1
		//      0  1  2  3  4  5  6  7  8  9  0  1  2  3  4  5
		//    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
		//    |                      ID                       |
		//    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
		//    |QR|   Opcode  |AA|TC|RD|RA|   Z    |   RCODE   |
		//    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
		//    |                    QDCOUNT                    |
		//    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
		//    |                    ANCOUNT                    |
		//    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
		//    |                    NSCOUNT                    |
		//    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
		//    |                    ARCOUNT                    |
		//    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+

		// A 16 bit identifier assigned by the program that generates any kind of query.  This identifier is copied
		// the corresponding reply and can be used by the requester to match up replies to outstanding queries.
		id := uint16(header[0])<<8 | uint16(header[1])
		id2 := uint16(header[3])<<8 | uint16(header[2])

		// A one bit field that specifies whether this message is a query (0), or a response (1).
		qr := (id2 & 0b0000_000_0_0_0_0_0000_1) > 0

		// A four bit field that specifies kind of query in this
		//  message. This value is set by the originator of a query
		//  and copied into the response.  The values are:
		//  0               a standard query (QUERY)
		//  1               an inverse query (IQUERY)
		//  2               a server status request (STATUS)
		//  3-15            reserved for future use
		opcode := uint8((id2 & 0b0000_000_0_0_0_0_1111_0) >> 1)

		// Authoritative Answer - this bit is valid in responses, and specifies that the responding name server is an
		// authority for the domain name in question section. Note that the contents of the answer section may have
		// multiple owner names because of aliases.  The AA bit corresponds to the name which matches the query name, or
		// the first owner name in the answer section.
		aa := (id2 & 0b0000_000_0_0_0_1_0000_0) > 0

		// TrunCation - specifies that this message was truncated due to length greater than that permitted on the
		// transmission channel.
		tc := (id2 & 0b0000_000_0_0_1_0_0000_0) > 0

		// Recursion Desired - this bit may be set in a query and is copied into the response.  If RD is set, it directs
		// the name server to pursue the query recursively. Recursive query support is optional.
		// rd := (id2 & 0b0000_000_0_1_0_0_0000_0) > 0

		// Recursion Available - this be is set or cleared in a response, and denotes whether recursive query support is
		// available in the name server.
		// ra := (id2 & 0b0000_000_1_0_0_0_0000_0) > 0

		// Reserved for future use.  Must be zero in all queries and responses.
		z := uint8((id2 & 0b0000_111_0_0_0_0_0000_0) >> 9)

		// Response code - this 4 bit field is set as part of responses.  The values have the following interpretation:
		//  0     No error condition
		//  1     Format error - The name server was unable to interpret the query.
		//  2     InternalDnsServer failure - The name server was unable to process this query due to a problem with the name
		//        server.
		//  3     Name Error - Meaningful only for responses from an authoritative name server, this code signifies that
		//        the domain name referenced in the query does not exist.
		//  4     Not Implemented - The name server does not support the requested kind of query.
		//  5     Refused - The name server refuses to perform the specified operation for policy reasons.  For example,
		//        a name server may not wish to provide the information to the particular requester, or a name server
		//         may not wish to perform a particular operation (e.g., zone transfer) for particular data.
		//  6-15  Reserved for future use.
		rcode := uint8((id2 & 0b1111_000_0_0_0_0_0000_0) >> 12)
		// An unsigned 16-bit integer specifying the number of entries in the question section.
		qdcount := uint16(header[4])<<8 | uint16(header[5])
		// an unsigned 16-bit integer specifying the number of resource records in the answer section.
		ancount := uint16(header[6])<<8 | uint16(header[7])
		//  an unsigned 16-bit integer specifying the number of name server resource records in the authority records section.
		nscount := uint16(header[8])<<8 | uint16(header[9])
		// an unsigned 16-bit integer specifying the number of resource records in the additional records section.
		arcount := uint16(header[10])<<8 | uint16(header[11])

		if id == 0x0000 {
			return false, ErrProtocolNotDetected
		}

		if opcode > 0x0002 {
			return false, ErrProtocolNotDetected
		}

		if aa {
			return false, ErrProtocolNotDetected
		}

		if protocolType == ProtocolStream && tc {
			return false, ErrProtocolNotDetected
		}

		if z != 0x000 {
			return false, ErrProtocolNotDetected
		}

		// https://stackoverflow.com/a/32292048/2642257
		// In any DNS packet the QDCOUNT field will be exactly one (0x0001).
		// Technically other values are allowed by the protocol, but in practise
		// they are never used.
		if qdcount != 0x0001 {
			return false, ErrProtocolNotDetected
		}

		// in a query (QR == 0) the ANCOUNT and NSCOUNT values will be exactly zero
		// (0x0000).
		if ancount != 0x0000 {
			return false, ErrProtocolNotDetected
		}
		if nscount != 0x0000 {
			return false, ErrProtocolNotDetected
		}

		// ARCOUNT will typically be 0, 1, or 2, depending on
		// whether EDNS0 (RFC 2671) and TSIG (RFC 2845) are being used.
		if arcount > 0x0002 {
			return false, ErrProtocolNotDetected
		}

		// RCODE will be typically zero in a query, but should never be more than 9
		// NOERROR   RCODE:0: DNS Query completed successfully
		// FORMERR   RCODE:1: DNS Query Format Error
		// SERVFAIL  RCODE:2: InternalDnsServer failed to complete the DNS request
		// NXDOMAIN  RCODE:3: Domain name does not exist.
		// NOTIMP    RCODE:4: Function not implemented
		// REFUSED   RCODE:5: The server refused to answer for the query
		// YXDOMAIN  RCODE:6: Name that should not exist, does exist
		// XRRSET    RCODE:7: RRset that should not exist, does exist
		// NOTAUTH   RCODE:8: InternalDnsServer not authoritative for the zone
		// NOTZONE   RCODE:9: Name not in zone
		if rcode > 0x000A {
			return false, ErrProtocolNotDetected
		}

		// A one bit field that specifies whether this message is a query (0), or a response (1).
		if !qr {
			return false, ErrProtocolNotDetected
		}

		return true, nil
	},
}
```

marten-seemann:
@bokysan Sorry for the delay here.

I'm wondering if we should adjust the API here. The original API was discussed in https://github.com/quic-go/quic-go/issues/3929.

Slightly modifying proposal (2) from that issue, here's what we could do:
Add a callback to the the `Transport`: `InterceptPacket(b []byte, net.Addr) (intercepted bool)`. 

If set, the `listen` Go routine would call this callback with every packet received. It's the applications responsibility to process the packet quickly and not block the run loop. If `intercepted` is `true`, the QUIC stack would skip this packet.

@bokysan What do you think?
@AudriusButkevicius Do you have any opinion on this?

bokysan:
`InterceptPacket` does seem like a good idea -- I'm happy to provide a patch.

bokysan:
@marten-seemann Would you then still leave the `IsPotentialQUICPacket` in place and queueing of non-QUIC packets? I mean, this would be a potentially breaking change.

marten-seemann:
It would be a breaking change, but that's what we have semver for 🙃 Would be nice to hear from @bt90 and @AudriusButkevicius, who I believe have been using this API.

We should probably have something like `DefaultIsNonQUICPacket` (maybe renamed to `IsProbablyQUICPacket` or something along those lines).

AudriusButkevicius:
Intercept hook is fine. Still allows us to achieve what we need.

marten-seemann:
> `InterceptPacket` does seem like a good idea -- I'm happy to provide a patch.

@bokysan That would be appreciated - preferably as a separate PR.

bokysan:
I'll need more info then. Should I just remove the queue and everything else? This initial PR was meant to be as little disruptive as possible.

marten-seemann:
Yes, that's the nice thing about the new API. We don't need the queue anymore, so this should simplify things a bit.

marten-seemann:
What happened here?

bokysan:
As agreed, closing this ticket and reopening with `InterceptPacket`. I'm just reusing the same repo.

marten-seemann:
Would be nice if you could use branches, and leave your old code here as it was.

bokysan:
Sorry. :-/

PR #4669, opened 2024-09-11
http3: send GOAWAY when server is shutting down

WeidiDeng:
Part of #153. Supersedes [4407](https://github.com/quic-go/quic-go/pull/4407). Only implements server goaway sending logic.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -715,7 +715,17 @@ func (s *Server) CloseGracefully(ctx context.Context) error {
 	if s.closeCancel != nil {
 		s.closeCancel()
 	}
+	// this server hasn't been used to serve QUIC connections
+	if s.closeDoneChan == nil {
+		s.closeDoneChan = make(chan struct{})
+	}
 	s.mutex.Unlock()
+
+	// fast check if there is no serve goroutine running
+	if s.connCount.Load() == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
+		close(s.closeDoneChan)
+		return s.Close()
+	}
 
 	select {
 	// all serve goroutine finished
--- a/http3/server.go
+++ b/http3/server.go
@@ -282,6 +282,16 @@ func (s *Server) ServeQUICConn(conn quic.Connection) error {
 		s.closeDoneChan = make(chan struct{})
 	}
 	s.mutex.Unlock()
+
+	s.connCount.Add(1)
+	defer func() {
+		s.mutex.RLock()
+		// close once when the server is closed and the connection count is zero
+		if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
+			close(s.closeDoneChan)
+		}
+		s.mutex.RUnlock()
+	}()
 	return s.handleConn(conn)
 }
 
@@ -312,6 +322,15 @@ func (s *Server) serveListener(ln QUICEarlyListener) error {
 			return err
 		}
 		go func() {
+			s.connCount.Add(1)
+			defer func() {
+				s.mutex.RLock()
+				// close once when the server is closed and the connection count is zero
+				if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
+					close(s.closeDoneChan)
+				}
+				s.mutex.RUnlock()
+			}()
 			if err := s.handleConn(conn); err != nil {
 				if s.Logger != nil {
 					s.Logger.Debug("handling connection failed", "error", err)
@@ -506,18 +525,6 @@ func (s *Server) handleConn(conn quic.Connection) error {
 		wg sync.WaitGroup
 	)
 
-	s.connCount.Add(1)
-	defer func() {
-		// some requests may be still running, wait for them to finish before decreasing the counter
-		wg.Wait()
-		s.mutex.RLock()
-		// close once when the server is closed and the connection count is zero
-		if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
-			close(s.closeDoneChan)
-		}
-		s.mutex.RUnlock()
-	}()
-
 	hconn := newConnection(
 		ctx,
 		conn,
@@ -541,6 +548,8 @@ func (s *Server) handleConn(conn quic.Connection) error {
 				// set a deadline to send the GOAWAY frame
 				ctrlStr.SetWriteDeadline(time.Now().Add(goawayTimeout))
 				ctrlStr.Write(b)
+				// some requests may be still running, wait for them to finish before returning
+				wg.Wait()
 				return http.ErrServerClosed
 			}
 
--- a/http3/server.go
+++ b/http3/server.go
@@ -334,10 +334,6 @@ func (s *Server) serveListener(ln QUICEarlyListener) error {
 				if s.Logger != nil {
 					s.Logger.Debug("handling connection failed", "error", err)
 				}
-				// server closed
-				if s.closeCtx.Err() != nil {
-					conn.CloseWithError(quic.ApplicationErrorCode(ErrCodeNoError), "")
-				}
 			}
 		}()
 	}
@@ -544,7 +540,14 @@ func (s *Server) handleConn(conn quic.Connection) error {
 	for {
 		str, datagrams, err := hconn.acceptStream(s.graceCtx)
 		if err != nil {
-			// server is closed, send GOAWAY frame and return, since those requests will be rejected anyway
+			// not gracefully closed, close the connection immediately
+			if s.closeCtx.Err() != nil {
+				conn.CloseWithError(quic.ApplicationErrorCode(ErrCodeNoError), "")
+				return http.ErrServerClosed
+			}
+
+			// gracefully closed, send GOAWAY frame and wait for requests to complete or grace period to end
+			// new requests will be rejected and shouldn't be sent
 			if s.graceCtx.Err() != nil {
 				b = (&goawayFrame{
 					StreamID: lastID + 4,
@@ -557,6 +560,8 @@ func (s *Server) handleConn(conn quic.Connection) error {
 				// some requests may be still running, wait for them to finish before returning
 				case <-handlingDoneChan:
 				case <-s.closeCtx.Done():
+					// close the connection after graceful period
+					conn.CloseWithError(quic.ApplicationErrorCode(ErrCodeNoError), "")
 				}
 				return http.ErrServerClosed
 			}
--- a/http3/server.go
+++ b/http3/server.go
@@ -707,6 +707,7 @@ func (s *Server) Close() error {
 	if s.closeCtx == nil {
 		return nil
 	}
+	s.closeCancel()
 
 	var err error
 	for ln := range s.listeners {
@@ -727,6 +728,7 @@ func (s *Server) CloseGracefully(ctx context.Context) error {
 		s.mutex.Unlock()
 		return nil
 	}
+	s.closeCancel()
 	s.mutex.Unlock()
 
 	// fast check if there is no serve goroutine running
--- a/http3/server.go
+++ b/http3/server.go
@@ -509,7 +509,7 @@ func (s *Server) handleConn(conn quic.Connection) error {
 	}()
 
 	// first valid ID is zero, subtract 4, so goaway frame id starts at 0
-	var lastID = quic.StreamID(-4)
+	lastID := quic.StreamID(-4)
 
 	hconn := newConnection(
 		ctx,
--- a/http3/server.go
+++ b/http3/server.go
@@ -274,6 +274,15 @@ func (s *Server) Serve(conn net.PacketConn) error {
 	return s.serveListener(ln)
 }
 
+func (s *Server) decreaseConnCount() {
+	s.mutex.RLock()
+	// close once when the server is closed and the connection count is zero
+	if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
+		close(s.closeDoneChan)
+	}
+	s.mutex.RUnlock()
+}
+
 // ServeQUICConn serves a single QUIC connection.
 func (s *Server) ServeQUICConn(conn quic.Connection) error {
 	s.mutex.Lock()
@@ -284,14 +293,7 @@ func (s *Server) ServeQUICConn(conn quic.Connection) error {
 	s.mutex.Unlock()
 
 	s.connCount.Add(1)
-	defer func() {
-		s.mutex.RLock()
-		// close once when the server is closed and the connection count is zero
-		if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
-			close(s.closeDoneChan)
-		}
-		s.mutex.RUnlock()
-	}()
+	defer s.decreaseConnCount()
 	return s.handleConn(conn)
 }
 
@@ -323,14 +325,7 @@ func (s *Server) serveListener(ln QUICEarlyListener) error {
 		}
 		go func() {
 			s.connCount.Add(1)
-			defer func() {
-				s.mutex.RLock()
-				// close once when the server is closed and the connection count is zero
-				if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
-					close(s.closeDoneChan)
-				}
-				s.mutex.RUnlock()
-			}()
+			defer s.decreaseConnCount()
 			if err := s.handleConn(conn); err != nil {
 				if s.Logger != nil {
 					s.Logger.Debug("handling connection failed", "error", err)
--- a/http3/server.go
+++ b/http3/server.go
@@ -334,6 +334,10 @@ func (s *Server) serveListener(ln QUICEarlyListener) error {
 				if s.Logger != nil {
 					s.Logger.Debug("handling connection failed", "error", err)
 				}
+				// server closed
+				if s.closeCtx.Err() != nil {
+					conn.CloseWithError(quic.ApplicationErrorCode(ErrCodeNoError), "")
+				}
 			}
 		}()
 	}
@@ -554,8 +558,6 @@ func (s *Server) handleConn(conn quic.Connection) error {
 				case <-handlingDoneChan:
 				case <-s.closeCtx.Done():
 				}
-				// close the connection
-				conn.CloseWithError(quic.ApplicationErrorCode(ErrCodeNoError), "")
 				return http.ErrServerClosed
 			}
 
--- a/http3/server.go
+++ b/http3/server.go
@@ -512,7 +512,7 @@ func (s *Server) handleConn(conn quic.Connection) error {
 		wg.Wait()
 		s.mutex.RLock()
 		// close once when the server is closed and the connection count is zero
-		if s.closed && s.connCount.Add(-1) == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
+		if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
 			close(s.closeDoneChan)
 		}
 		s.mutex.RUnlock()
--- a/http3/server.go
+++ b/http3/server.go
@@ -220,7 +220,7 @@ type Server struct {
 	closeCancel context.CancelFunc // cancels the closeCtx
 
 	closeDoneChan  chan struct{} // will be closed when the server is shutdown and connCount is zero
-	doneChanClosed atomic.Bool   // closeDoneChan will be closed once
+	doneChanClosed atomic.Bool   // makes sure closeDoneChan will be closed once
 
 	altSvcHeader string
 }
@@ -471,8 +471,9 @@ func (s *Server) removeListener(l *QUICEarlyListener) {
 }
 
 func (s *Server) handleConn(conn quic.Connection) error {
-	// send a SETTINGS frame
-	str, err := conn.OpenUniStream()
+	// open the control stream and send a SETTINGS frame, it's also used to send a GOAWAY frame later
+	// when the server is gracefully closed
+	ctrlStr, err := conn.OpenUniStream()
 	if err != nil {
 		return fmt.Errorf("opening the control stream failed: %w", err)
 	}
@@ -483,7 +484,7 @@ func (s *Server) handleConn(conn quic.Connection) error {
 		ExtendedConnect: true,
 		Other:           s.AdditionalSettings,
 	}).Append(b)
-	str.Write(b)
+	ctrlStr.Write(b)
 
 	ctx := conn.Context()
 	ctx = context.WithValue(ctx, ServerContextKey, s)
@@ -500,19 +501,15 @@ func (s *Server) handleConn(conn quic.Connection) error {
 	defer func() {
 		s.connCount.Add(-1)
 		s.mutex.RLock()
-		// close once
-		if s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
+		// close once when the server is closed and the connection count is zero
+		if s.closed && s.connCount.Load() == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
 			close(s.closeDoneChan)
 		}
 		s.mutex.RUnlock()
 	}()
 
-	var (
-		// first valid ID is zero, subtract 4, so goaway frame id starts at 0
-		lastID = quic.StreamID(-4)
-		// ctrlStr is the control stream, which is used to send GOAWAY frames.
-		ctrlStr = str
-	)
+	// first valid ID is zero, subtract 4, so goaway frame id starts at 0
+	var lastID = quic.StreamID(-4)
 
 	hconn := newConnection(
 		ctx,
@@ -530,10 +527,12 @@ func (s *Server) handleConn(conn quic.Connection) error {
 		str, datagrams, err := hconn.acceptStream(s.closeCtx)
 		if err != nil {
 			// server is closed, send GOAWAY frame and return, since those requests will be rejected anyway
-			if errors.Is(err, context.Canceled) {
+			if s.closeCtx.Err() != nil {
 				b = (&goawayFrame{
 					StreamID: lastID + 4,
 				}).Append(b[:0])
+				// set a deadline to send the GOAWAY frame
+				ctrlStr.SetWriteDeadline(time.Now().Add(5 * time.Second))
 				ctrlStr.Write(b)
 				return http.ErrServerClosed
 			}
@@ -694,8 +693,9 @@ func (s *Server) Close() error {
 	defer s.mutex.Unlock()
 
 	s.closed = true
-	if s.closeCancel != nil {
-		s.closeCancel()
+	// server is never used
+	if s.closeCtx == nil {
+		return nil
 	}
 
 	var err error
@@ -712,18 +712,15 @@ func (s *Server) Close() error {
 func (s *Server) CloseGracefully(ctx context.Context) error {
 	s.mutex.Lock()
 	s.closed = true
-	if s.closeCancel != nil {
-		s.closeCancel()
-	}
-	// this server hasn't been used to serve QUIC connections
-	if s.closeDoneChan == nil {
-		s.closeDoneChan = make(chan struct{})
+	// server is never used
+	if s.closeCtx == nil {
+		s.mutex.Unlock()
+		return nil
 	}
 	s.mutex.Unlock()
 
 	// fast check if there is no serve goroutine running
-	if s.connCount.Load() == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
-		close(s.closeDoneChan)
+	if s.connCount.Load() == 0 {
 		return s.Close()
 	}
 
--- a/http3/server.go
+++ b/http3/server.go
@@ -46,6 +46,8 @@ const (
 	streamTypeQPACKDecoderStream = 3
 )
 
+const goawayTimeout = 5 * time.Second
+
 // A QUICEarlyListener listens for incoming QUIC connections.
 type QUICEarlyListener interface {
 	Accept(context.Context) (quic.EarlyConnection, error)
@@ -508,10 +510,9 @@ func (s *Server) handleConn(conn quic.Connection) error {
 	defer func() {
 		// some requests may be still running, wait for them to finish before decreasing the counter
 		wg.Wait()
-		s.connCount.Add(-1)
 		s.mutex.RLock()
 		// close once when the server is closed and the connection count is zero
-		if s.closed && s.connCount.Load() == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
+		if s.closed && s.connCount.Add(-1) == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
 			close(s.closeDoneChan)
 		}
 		s.mutex.RUnlock()
@@ -538,7 +539,7 @@ func (s *Server) handleConn(conn quic.Connection) error {
 					StreamID: lastID + 4,
 				}).Append(b[:0])
 				// set a deadline to send the GOAWAY frame
-				ctrlStr.SetWriteDeadline(time.Now().Add(5 * time.Second))
+				ctrlStr.SetWriteDeadline(time.Now().Add(goawayTimeout))
 				ctrlStr.Write(b)
 				return http.ErrServerClosed
 			}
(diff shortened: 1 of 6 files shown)


WeidiDeng:
@marten-seemann This is the new patch that handles graceful shutdown for servers.

http3/server.go
```diff
@@ -479,6 +492,24 @@ func (s *Server) handleConn(conn quic.Connection) error {
 		}
 	}
 
+	s.connCount.Add(1)
+
+	var (
+		lastID atomic.Int64
+	)
```
marten-seemann, line 499:
```suggestion
	var lastID atomic.Int64
```

http3/server.go
```diff
@@ -479,6 +492,24 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
+	// first valid ID is zero, subtract 4, so goaway frame id starts at 0
+	lastID.Store(-4)
+	// run in a separate goroutine
+	stop := context.AfterFunc(s.closeCtx, func() {
+		b = (&goawayFrame{
+			StreamID: quic.StreamID(lastID.Load()) + 4,
+		}).Append(b[:0])
+		str.Write(b)
+	})
```
marten-seemann, line 508: Every function is going to get run a separate Go routine, leading a large number of spawned Go routines when the server is shut down.

What do you think of doing this in the the `for` loop that accepts streams? We'd pass the `closeCtx` to `hconn.acceptStream`, and introduce a separate check when this function returns an error.

http3/server.go
```diff
@@ -658,11 +704,46 @@ func (s *Server) Close() error {
...
+			return s.Close()
+		}
+		select {
+		case <-ctx.Done():
+			s.Close()
+			return ctx.Err()
+		case <-timer.C:
+			timer.Reset(nextPollInterval())
+		}
+	}
```
marten-seemann, line 746: We could also create a channel that is closed once the number reaches 0. Every Go routine decrementing the counter would check the value of the counter, and close the channel once it's 0. This would allow us to rewrite this as a simple `select` statement, selecting from this channel, and a timer channel.

marten-seemann (commented):
Thanks for splitting this up. This makes reviewing a lot easier.

Just to confirm, is your plan to work on the client-side behavior once this PR is merged?

http3/server.go
WeidiDeng: Done, and simpler I think

http3/server.go
WeidiDeng: Done

http3/server.go
```diff
@@ -479,6 +496,24 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
+			close(s.closeDoneChan)
+		}
+		s.mutex.RUnlock()
+	}()
+
+	var (
+		// first valid ID is zero, subtract 4, so goaway frame id starts at 0
+		lastID = quic.StreamID(-4)
+		// ctrlStr is the control stream, which is used to send GOAWAY frames.
+		ctrlStr = str
```
marten-seemann, line 514: Nit: name the variable `ctrlStr` right when opening the control stream.

http3/server.go
```diff
@@ -492,14 +527,25 @@ func (s *Server) handleConn(conn quic.Connection) error {
 	// Process all requests immediately.
 	// It's the client's responsibility to decide which requests are eligible for 0-RTT.
 	for {
-		str, datagrams, err := hconn.acceptStream(context.Background())
+		str, datagrams, err := hconn.acceptStream(s.closeCtx)
 		if err != nil {
+			// server is closed, send GOAWAY frame and return, since those requests will be rejected anyway
+			if errors.Is(err, context.Canceled) {
```
marten-seemann, line 533: Probably more error-prone, in case we decide to use a different (or a derived) context here at some point in the future:
```suggestion
			if s.closeCtx.Err() != nil {
```

http3/server.go
```diff
@@ -492,14 +527,25 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
+				b = (&goawayFrame{
+					StreamID: lastID + 4,
+				}).Append(b[:0])
+				ctrlStr.Write(b)
```
marten-seemann, line 537: What happens if we're flow control blocked? Should we set a write deadline (5s maybe?), so we can guarantee that this function returns even if the peer is messing with our flow control window?

http3/server.go
```diff
@@ -213,7 +214,13 @@ type Server struct {
...
 	listeners map[*QUICEarlyListener]listenerInfo
 
-	closed bool
+	closed      bool
+	connCount   atomic.Int64       // number of connections
+	closeCtx    context.Context    // canceled when the server is closed
+	closeCancel context.CancelFunc // cancels the closeCtx
+
+	closeDoneChan  chan struct{} // will be closed when the server is shutdown and connCount is zero
+	doneChanClosed atomic.Bool   // closeDoneChan will be closed once
```
marten-seemann, line 223: This comment seems incomplete.

http3/server.go
```diff
@@ -660,9 +709,33 @@ func (s *Server) Close() error {
...
+		s.closeCancel()
+	}
+	// this server hasn't been used to serve QUIC connections
+	if s.closeDoneChan == nil {
+		s.closeDoneChan = make(chan struct{})
+	}
+	s.mutex.Unlock()
+
+	// fast check if there is no serve goroutine running
+	if s.connCount.Load() == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
```
marten-seemann, line 725: Can we just return immediately here, or am I missing something?

WeidiDeng:
Yeah, client side handling will be done in the next patch.

http3/server.go
```diff
@@ -660,9 +709,33 @@ func (s *Server) Close() error {
...
-	// TODO: implement
-	return nil
+func (s *Server) CloseGracefully(ctx context.Context) error {
+	s.mutex.Lock()
+	s.closed = true
+	if s.closeCancel != nil {
+		s.closeCancel()
+	}
+	// this server hasn't been used to serve QUIC connections
+	if s.closeDoneChan == nil {
```
marten-seemann, line 719: Can we return immediately?

http3/server.go
```diff
@@ -479,6 +496,24 @@ func (s *Server) handleConn(conn quic.Connection) error {
 		}
 	}
 
+	s.connCount.Add(1)
+	defer func() {
+		s.connCount.Add(-1)
+		s.mutex.RLock()
+		// close once
+		if s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
```
marten-seemann, line 504: We only want to execute this when the last connection is closed, right?

WeidiDeng: oops, forgot to check connection count 😅

WeidiDeng:
@marten-seemann Do you have time to take a look?

http3/server.go
```diff
@@ -479,6 +497,26 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
 	}
 
+	var (
+		// first valid ID is zero, subtract 4, so goaway frame id starts at 0
+		lastID = quic.StreamID(-4)
+		// wait until all request goroutines are done
+		wg sync.WaitGroup
+	)
+
+	s.connCount.Add(1)
```
marten-seemann, line 507: I don't think this is quite right. `sync.WaitGroup` requires you to call `Add` and `Wait` from the same Go routine.

This means that we need to make sure that `handleConn` is not called after the shutdown has been initiated.

http3/server.go
```diff
@@ -479,6 +497,26 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
+	defer func() {
+		// some requests may be still running, wait for them to finish before decreasing the counter
+		wg.Wait()
+		s.connCount.Add(-1)
+		s.mutex.RLock()
+		// close once when the server is closed and the connection count is zero
+		if s.closed && s.connCount.Load() == 0 && s.doneChanClosed.CompareAndSwap(false, true) {
```
marten-seemann, line 514: This is racy. Instead, you need to take the return value from `Add`.

http3/server.go
```diff
@@ -492,15 +530,32 @@ func (s *Server) handleConn(conn quic.Connection) error {
...
-		str, datagrams, err := hconn.acceptStream(context.Background())
+		str, datagrams, err := hconn.acceptStream(s.closeCtx)
 		if err != nil {
+			// server is closed, send GOAWAY frame and return, since those requests will be rejected anyway
+			if s.closeCtx.Err() != nil {
+				b = (&goawayFrame{
+					StreamID: lastID + 4,
+				}).Append(b[:0])
+				// set a deadline to send the GOAWAY frame
+				ctrlStr.SetWriteDeadline(time.Now().Add(5 * time.Second))
```
marten-seemann, line 541: Nit: Make this a constant at the beginning of the file? Makes it look less like a magic value.

http3/server.go
```diff
@@ -660,9 +720,31 @@ func (s *Server) Close() error {
...
+	// fast check if there is no serve goroutine running
+	if s.connCount.Load() == 0 {
+		return s.Close()
+	}
+
+	select {
+	// all serve goroutine finished
+	case <-s.closeDoneChan:
+		return s.Close()
+	// context canceled
```
marten-seemann, line 743: Don't need a comment for the obvious.
```suggestion
```

http3/server.go
```diff
@@ -660,9 +720,31 @@ func (s *Server) Close() error {
...
+	s.mutex.Unlock()
+
+	// fast check if there is no serve goroutine running
+	if s.connCount.Load() == 0 {
+		return s.Close()
+	}
+
+	select {
+	// all serve goroutine finished
+	case <-s.closeDoneChan:
```
marten-seemann, line 741:
```suggestion
	case <-s.closeDoneChan: // all connections were closed
```

marten-seemann (commented):
I think we'll have to be a bit more careful with the connection counting `WaitGroup`. Other than that, this PR looks good to me.

Can you add a few test cases though?

I think this already makes for a nice integration test: Call `CloseGracefully`, and check that this function doesn't return until the client decides to close the connection. The client doesn't need to understand GOAWAY for that yet, you can just manually trigger closing of the connection.

marten-seemann:
> @marten-seemann Do you have time to take a look?

Sorry, I don't get a notification when you push changes! Please feel free to ping me, or to re-request a review, so that I get notified once you've made changes.

http3/server.go
WeidiDeng: The document only says `Note that calls with a positive delta that occur when the counter is zero must happen before a Wait.` And deferred function does run in the same goroutine as the surrounding function.

The counter for wait ground is incremented just before creating `handleRequest` goroutine, and that's the correct usage.

http3/server.go
```diff
@@ -213,7 +214,13 @@ type Server struct {
...
 	listeners map[*QUICEarlyListener]listenerInfo
 
-	closed bool
+	closed      bool
+	connCount   atomic.Int64       // number of connections
+	closeCtx    context.Context    // canceled when the server is closed
+	closeCancel context.CancelFunc // cancels the closeCtx
+
+	closeDoneChan  chan struct{} // will be closed when the server is shutdown and connCount is zero
+	doneChanClosed atomic.Bool   // makes sure closeDoneChan will be closed once
```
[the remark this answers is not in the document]
WeidiDeng: Not necessary. The difference between `sync.Once` and `atomic.Bool` for making sure a function runs only once is that, `sync.Once` makes sure after `Do` is called, the function is done, where `atomic.Bool` has no such guarantee.

I don't think `sync.Once` is necessary in this case. As long as the channel is closed, the graceful shutdown can proceed.

http3/server.go
marten-seemann: I'm wondering if we should just make `handleConn` block until all streams have finished, and then do the connection counting logic in the function that calls `handleConn`. What do you think?

WeidiDeng: I guess it can be done.

http3/server.go
```diff
@@ -297,6 +322,15 @@ func (s *Server) serveListener(ln QUICEarlyListener) error {
...
 		go func() {
+			s.connCount.Add(1)
+			defer func() {
+				s.mutex.RLock()
+				// close once when the server is closed and the connection count is zero
+				if s.connCount.Add(-1) == 0 && s.closed && s.doneChanClosed.CompareAndSwap(false, true) {
+					close(s.closeDoneChan)
+				}
+				s.mutex.RUnlock()
+			}()
```
marten-seemann, line 333: You're repeating a lot of code here. Let's just create a new method to avoid that.

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1069,118 @@ var _ = Describe("HTTP tests", func() {
...
+		// makes two requests, one fast and one slow, both are expected to finish successfully
+		wg.Add(1)
+		go func() {
+			resp, err := client.Get(fmt.Sprintf("https://localhost:%d/fast", port))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(resp.StatusCode).To(Equal(200))
+			body, err := io.ReadAll(resp.Body)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(body).To(Equal(PRDataLong))
+			wg.Done()
```
marten-seemann, line 1104: defer this at the beginning of the function.
Or better, use a `done` channel, so you can have an assertion that the client returns (using `Eventually`).

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1069,118 @@ var _ = Describe("HTTP tests", func() {
 			"Unannounced": {"Surprise!"},
 		})))
 	})
+
+	It("graceful shutdown successfully with no timeout", func() {
```
marten-seemann, line 1073: I'm not sure I understand what this test actually tests. Let's make this more intuitive.

1. Remove `/fast`.
2. Remove the timer from `/slow`. Instead, block on a channel.
3. Call `CloseGracefully`. Make sure that it blocks.
4. Close the channel.
5. Assert that `CloseGracefully` returns.

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1069,118 @@ var _ = Describe("HTTP tests", func() {
...
+			<-fastChan
+			<-slowChan
+			err := server.CloseGracefully(context.Background())
+			Expect(err).ToNot(HaveOccurred())
+			wg.Done()
+		}()
+		wg.Wait()
+	})
+
+	It("graceful shutdown successfully with timeout", func() {
```
marten-seemann, line 1128: Here the `/fast` actually makes sense.

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1069,118 @@ var _ = Describe("HTTP tests", func() {
...
+		mux.HandleFunc("/slow", func(w http.ResponseWriter, r *http.Request) {
+			close(slowChan)
+			ticker := time.NewTicker(time.Second)
+			defer ticker.Stop()
+			chunkSize := len(PRDataLong) / 10
+			for i := range 10 {
+				<-ticker.C
+				w.Write(PRDataLong[i*chunkSize : (i+1)*chunkSize])
+			}
+		})
```
marten-seemann, line 1149: Again, this should just block on something. Maybe `r.Body.Read`, so you can actually assert that the underlying connection is closed?

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1069,118 @@ var _ = Describe("HTTP tests", func() {
...
+		}()
+		wg.Wait()
+	})
+
+	It("graceful shutdown successfully with timeout", func() {
+		// server will begin shutdown after receiving two requests
+		var (
+			fastChan = make(chan struct{})
+			slowChan = make(chan struct{})
+			wg       sync.WaitGroup
```
marten-seemann, line 1133: Avoid `sync.WaitGroup`. It might block indefinitely when something goes wrong, which is very annoying to debug. Instead, use a channel and `Eventually`, so you can actually fail the test.

marten-seemann (commented):
Code looks good modulo one nit.

I added a couple of suggestions to make the integration test more useful.

Can you add some unit tests for the GOAWAY frame parsing as well please?

integrationtests/self/http_test.go
WeidiDeng: You can't close a channel when `CloseGracefully` is blocked, it has to be done before that. Http handler has no way of knowing whether a graceful shutdown started.

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1068,105 @@ var _ = Describe("HTTP tests", func() {
...
+			serverDoneChan      = make(chan struct{})
+		)
+
+		mux.HandleFunc("/slow", func(w http.ResponseWriter, r *http.Request) {
+			close(startShutdownChan)
+			<-shutdownStartedChan
+			chunkSize := len(PRDataLong) / 10
+			for i := range 10 {
+				w.Write(PRDataLong[i*chunkSize : (i+1)*chunkSize])
+			}
```
marten-seemann, line 1086: Why is there a loop here?

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1068,105 @@ var _ = Describe("HTTP tests", func() {
...
+		var (
+			startShutdownChan   = make(chan struct{})
+			shutdownStartedChan = make(chan struct{})
+			clientDoneChan      = make(chan struct{})
+			serverDoneChan      = make(chan struct{})
+		)
+
+		mux.HandleFunc("/slow", func(w http.ResponseWriter, r *http.Request) {
+			close(startShutdownChan)
+			<-shutdownStartedChan
```
marten-seemann, line 1082: This channel doesn't do anything, as it's immediately closed in the `CloseGracefully` Go routine.

What you should do instead is use a `time.Sleep(250*time.Millisecond)` here. That will (in practice) guarantee that the `Write` is run after `CloseGracefully`.

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1068,105 @@ var _ = Describe("HTTP tests", func() {
...
+
+		go func() {
+			<-startShutdownChan
+			close(shutdownStartedChan)
+			err := server.CloseGracefully(context.Background())
+			Expect(err).ToNot(HaveOccurred())
+			close(serverDoneChan)
+		}()
+		Eventually(clientDoneChan, "5s").Should(BeClosed())
+		Eventually(serverDoneChan, "5s").Should(BeClosed())
```
marten-seemann, line 1106:
```suggestion
		Eventually(clientDoneChan, 5*time.Second).Should(BeClosed())
		Eventually(serverDoneChan, 5*time.Second).Should(BeClosed())
```

integrationtests/self/http_test.go
marten-seemann: This tests takes a whopping 20s! That's a dealbreaker. Please make sure that it finishes in significantly less than half a second.

integrationtests/self/http_test.go
```diff
@@ -1068,4 +1068,105 @@ var _ = Describe("HTTP tests", func() {
...
+			fastChan       = make(chan struct{})
+			slowChan       = make(chan struct{})
+			fastDoneChan   = make(chan struct{})
+			slowDoneChan   = make(chan struct{})
+			serverDoneChan = make(chan struct{})
+		)
+
+		mux.HandleFunc("/fast", func(w http.ResponseWriter, r *http.Request) {
+			close(fastChan)
+			w.Write(PRDataLong)
```
marten-seemann, line 1121: Write less data.

marten-seemann (changes requested):
The integration test is taking way too long, see my comment there.

Looks like the server also needs some unit tests.

integrationtests/self/http_test.go
WeidiDeng: By reducing `MaxIdleTime`, the time to finish a test can be reduced to a degree, but not that small. It can't be reduced to less than half a second.

marten-seemann: Of course it can, and I already suggested you how to do it:
1. Call `CloseGracefully`.
2. After 150ms, cancel the context.
3. After 300ms, assert that the request has returned.

WeidiDeng: What do you think that I wrap the `http.ResponseWriter` that will extend the deadline everytime data is written. It will also embed a client context that will be cancelled in case the deadline exceeded.

I feel like the with timeout test should have some tolerance for a longer test. After all, it's simulating a long request interrupted. I'll keep it under 10 seconds.

marten-seemann: > What do you think that I wrap the http.ResponseWriter that will extend the deadline everytime data is written. It will also embed a client context that will be cancelled in case the deadline exceeded.

I don't understand what this would test.

> I feel like the with timeout test should have some tolerance for a longer test. After all, it's simulating a long request interrupted. I'll keep it under 10 seconds.

The entire HTTP test suite, all 35 integration tests, runs in less than 2s. It is not reasonable to add a single test that runs 5x as long.

WeidiDeng: > > What do you think that I wrap the http.ResponseWriter that will extend the deadline everytime data is written. It will also embed a client context that will be cancelled in case the deadline exceeded.
> 
> I don't understand what this would test.
> 
> >

Not a test, but the implementation detail that's used to test. Because when the server is closed and ongoing requests are supposedly interrupted, quic needs some time to react to the loss of connectivity. Using context and deadline we can stop when no progress is made.



> > I feel like the with timeout test should have some tolerance for a longer test. After all, it's simulating a long request interrupted. I'll keep it under 10 seconds.
> 
> The entire HTTP test suite, all 35 integration tests, runs in less than 2s. It is not reasonable to add a single test that runs 5x as long.

How about 2.5x as long as a compromise. I could try for lower but that just magic numbers from then on.

marten-seemann: > Because when the server is closed and ongoing requests are supposedly interrupted, quic needs some time to react to the loss of connectivity. Using context and deadline we can stop when no progress is made.

It doesn't. We're not testing a connection timeout. When the context of `CloseGracefully` is canceled, the server is supposed to send a CONNECTION_CLOSE, which immediately closes the connection. 300ms, as I suggested earlier, should be **plenty** of time to test this.

WeidiDeng: The current implementation doesn't close those quic connections. The only thing it closes is the underlying datagram socket. Hence the use of context.

I suppose it can be added here? The signature needs to be changed.

image

WeidiDeng: But then it only sends a CONNECTION_CLOSE, when all serve routine is done. It should send CONNECTION_CLOSE when the context passed to `CloseGracefully` is done.

WeidiDeng: Perhaps we need to store quic connections in a map?

marten-seemann: > Perhaps we need to store quic connections in a map?

There's probably a better solution by canceling a context, closing a channel or something along those lines.

WeidiDeng: By using channel, a new goroutine is required. Also the main closing call needs to know if these `CONNECTION_CLOSE`s finish sending,

WeidiDeng: In the first draft I'll use map to store these connections. Will come back later to see the alternative implementations.

WeidiDeng: Perhaps, I can use another context [here](https://github.com/quic-go/quic-go/blob/master/http3/server.go#L490) that will be responsible for closing the connection. This context is tied with another context associated with server that will be canceled when `CONNECTION_CLOSE` should be sent. Yet another counter and channel. Any idea why the context passed isn't used though?

WeidiDeng: @marten-seemann Is it safe to use another context [here](https://github.com/quic-go/quic-go/blob/master/http3/conn.go#L181)? It's shared by both server and client. If no map is used, this is the best place I can think of.

marten-seemann: You could pass a context to `HandleUnidirectionalStreams`, right?

WeidiDeng: No, you can't

marten-seemann: What's with the snarky answers?

Do you want this PR to get merged or not? If you're not interested in resolving the remaining issues, shall we just close this PR and not waste any more of both of our time?

WeidiDeng: I'm not snarky. The [function](https://github.com/quic-go/quic-go/blob/master/http3/conn.go#L173) signature is `HandleUnidirectionalStreams(hijack func(StreamType, quic.ConnectionTracingID, quic.ReceiveStream, error) (hijacked bool))`. There is no way to pass a context.

WeidiDeng: Can I use some of the codes [here](https://github.com/quic-go/quic-go/pull/4689) in this branch instead? Some of the changes conflict.

marten-seemann: I suggest we merge #4689 first. This is a bug.

WeidiDeng:
@marten-seemann I incorporated your changes in this branch.

marten-seemann:
I merged #4689. Can you rebase this PR please?

WeidiDeng:
@marten-seemann There is some race going on. If I close the connection immediately like you did. `graceful shutdown successfully with no timeout` will fail because although the request handling is done. The data written isn't delivered yet to the peer (I think, from `SendStream` documentation, `CancelWrite` specifically), causing the response to fail. How do I deal with this situation? That the data is delivered to the peer. Do I need to use the channel in the client in the tests to simulates this situation?

marten-seemann:
> @marten-seemann There is some race going on. If I close the connection immediately like you did. `graceful shutdown successfully with no timeout` will fail because although the request handling is done. The data written isn't delivered yet to the peer (I think, from `SendStream` documentation, `CancelWrite` specifically), causing the response to fail. How do I deal with this situation? That the data is delivered to the peer. Do I need to use the channel in the client in the tests to simulate this situation?

Interesting. I believe this is because all data has been sent (and maybe even received), but not read yet by the client. And since CONNECTION_CLOSE is immediate, the client will return the CONNECTION_CLOSE error to the application to the application.

Unfortunately, the RFC is not entirely clear about this, but I think the correct solution is to wait for the client to close the connection. On the server side, we'd then only close the connection once the context is closed.

WeidiDeng:
Currently I check for both graceful and non graceful shutdown, and only close the connection if non graceful. It helps in this case. But maybe I can close the connection client side once client side goaway handling is implemented.

marten-seemann:
Yes, it's fine if with this PR, `CloseGracefully` will always block until the context is cancelled. That's expected since the clients are not cooperating yet.

WeidiDeng:
@marten-seemann It's ready for review now. The failing macos tests seem to be temporary.

http3/server.go
```diff
@@ -497,19 +538,51 @@ func (s *Server) handleConn(serverCtx context.Context, conn quic.Connection) err
...
+					StreamID: lastID + 4,
+				}).Append(b[:0])
+				// set a deadline to send the GOAWAY frame
+				ctrlStr.SetWriteDeadline(time.Now().Add(goawayTimeout))
+				ctrlStr.Write(b)
+
+				select {
+				// some requests may be still running, wait for them to finish before returning
+				// do nothing let quic to deliver the data the peer
+				case <-handlingDoneChan:
```
marten-seemann, line 562: We can't return early here. We still need to close the connection if the client handle the GOAWAY frame.

http3/server.go
```diff
@@ -497,19 +538,51 @@ func (s *Server) handleConn(serverCtx context.Context, conn quic.Connection) err
...
 			var appErr *quic.ApplicationError
 			if errors.As(err, &appErr) && appErr.ErrorCode == quic.ApplicationErrorCode(ErrCodeNoError) {
 				return nil
 			}
 			return fmt.Errorf("accepting stream failed: %w", err)
 		}
-		go s.handleRequest(hconn, str, datagrams, hconn.decoder)
+
+		lastID = str.StreamID()
+		runningRequests.Add(1)
```
marten-seemann, line 578: I don't think we need this tracking here, since it's the client's responsibility to close the connection after receiving GOAWAY and processing in-flight requests.

integrationtests/self/http_test.go
marten-seemann, line 1: Can we use the tests I suggested in e933c80ada58eba408722e912f0a9cf094f27ecf?

http3/server.go
WeidiDeng: GOAWAY needs to contain the streamID by rfc.

http3/server.go
WeidiDeng: > @marten-seemann There is some race going on. If I close the connection immediately like you did. `graceful shutdown successfully with no timeout` will fail because although the request handling is done. The data written isn't delivered yet to the peer (I think, from `SendStream` documentation, `CancelWrite` specifically), causing the response to fail. How do I deal with this situation? That the data is delivered to the peer. Do I need to use the channel in the client in the tests to simulate this situation?

That will cause this problem.

http3/server.go
marten-seemann: Yes, we need `lastID`, but we don't need `runningRequests`.

http3/server.go
marten-seemann: It doesn't. Just don't close the connection before cancelation of the context.

http3/server.go
WeidiDeng: The stream id is used by client to know which request should be retried, because it's not handled by the server.

http3/server.go
WeidiDeng: If client side reading can be changed to read the available data first, and ignore the connection error because the data is already there and fully read. Then it can be changed.

http3/server.go
marten-seemann: Yes I know, but there's no point in tracking the number of running requests. It's the client's responsibility to close the connection when it's done, as we did in https://github.com/quic-go/quic-go/pull/4669#issuecomment-2403747650.

The server only closes the connection once the context is cancelled, and at that point, we don't care how many requests are active.

http3/server.go
marten-seemann: That's not how QUIC works. But as discussed in https://github.com/quic-go/quic-go/pull/4669#issuecomment-2403747650, it's the client who closes the connection.

WeidiDeng: Do you mean the context of the quic.Connection? So the server will close the connection when either it's forcibly shut down or when the client closes?

http3/server.go
WeidiDeng: As above, the context here means the context of the quic.Connection?

http3/server.go
marten-seemann: We don't need any context here. We just need to close the connect when the close context is cancelled.

WeidiDeng: Just wait for the `closeCtx` to be done, then?

http3/server.go
WeidiDeng: Server is tracking the number of requests so that closeCtx will be canceled appropriately

marten-seemann: If you use a `context.Background`, things might block indefinitely. That's expected. Don't do that.

WeidiDeng: That's a normal use of graceful shutdown, which means to wait for all requests to finish indefinitely. stdlib does this too, albeit with polling.

marten-seemann: And you can do that, but if you're willing to wait indefinitely, you can't complain that you'll wait indefinitely if one of the clients misbehaves...

In practice, you'll *always* apply a timeout.

WeidiDeng: But it should behave correctly when the server is willing to wait. And the server indeed finishes handling the requests.

marten-seemann: If you're willing to wait forever, the correct behavior is to wait for the client to close the connection.

WeidiDeng: Per stdlib [documents](https://pkg.go.dev/net/http#Server.Shutdown):

>Shutdown gracefully shuts down the server without interrupting any active connections. Shutdown works by first closing all open listeners, then closing all idle connections, and then waiting indefinitely for connections to return to idle and then shut down.

When all requests associated with a quic.Connection are done, it's in the idle state.

WeidiDeng: There is no need to wait for client to close the connection, as newer requests, which shouldn't be sent, will not be handled anymore.

marten-seemann: I don't see the problem. We already established that the server can't close connections for HTTP/3 without disrupting the client's response processing.

This is exactly what RFC 9114 specifies as well. Let's not violate the RFC.

WeidiDeng: The original problem is that waiting only for closeCtx will lead to deadlock if the grace close context will never be canceled.

The connection can be closed either all requests are done or grace period is over.

WeidiDeng: > I don't see the problem. We already established that the server can't close connections for HTTP/3 without disrupting the client's response processing.
> 
> This is exactly what RFC 9114 specifies as well. Let's not violate the RFC.

If the response can't be guaranteed to succeed, let's just comment out that test. Just let the server close the connection as all requests finish.

marten-seemann: No, we're not going to disable a test that is correct in order to keep incorrect code around.

I'm getting tired of this. I've said this 5 times now, and you keep ignoring me. I don't know how we should make progress like this.

marten-seemann:
Thank you for your work on this PR @WeidiDeng! I feel like we can make progress faster if I push this across the finish line, so I'm going to close this PR, and will continue the work in #4691.

PR #4673, opened 2024-09-13
stream: don't cancel stream after shutdown

sukunrt:
This ensures that `stream.Write` and `stream.Read` return the error code
from connection close, if the stream was closed as a result of
connection close.


marten-seemann (approved):
Nice find!

Really appreciate the meaningful test cases as well!

PR #4685, opened 2024-09-30
utils: fix timer in Go 1.23

ainar-g:
See https://tip.golang.org/doc/go1.23#timer-changes.


ainar-g:
For context, we've been seeing various services getting stuck or timing out after moving to Go 1.23.  The goroutine profiles always point to this line being the culprit.

PR #4689, opened 2024-10-08
http3: immediately close all connections on Server.Close

marten-seemann:
cc @WeidiDeng

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -267,7 +267,7 @@ func (s *Server) Serve(conn net.PacketConn) error {
 
 // ServeQUICConn serves a single QUIC connection.
 func (s *Server) ServeQUICConn(conn quic.Connection) error {
-	return s.handleConn(conn)
+	return s.handleConn(context.Background(), conn)
 }
 
 // ServeListener serves an existing QUIC listener.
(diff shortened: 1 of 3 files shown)


http3/server.go
```diff
@@ -267,7 +267,7 @@ func (s *Server) Serve(conn net.PacketConn) error {
 
 // ServeQUICConn serves a single QUIC connection.
 func (s *Server) ServeQUICConn(conn quic.Connection) error {
-	return s.handleConn(conn)
+	return s.handleConn(context.Background(), conn)
```
WeidiDeng, line 270: I don't think this is necessary. It will cause the quic connection to be shutdown. Users should coordinate this with the server shutdown, or not call shutdown at all.

marten-seemann:
This matches the behavior of HTTP/1.1: https://gist.github.com/marten-seemann/daaeb2949cdebd68d6c93381f6589c17. Closing the server kills the request.

WeidiDeng:
I tested your patch, but the response is successful. The test is inside the integration test. The OS is windows 10 amd64.

image

image

Any idea why?

WeidiDeng:
`server_test` is using a mock quic stream, but `http_test` is using a real quic stream. This shouldn't how the protocol work, right?

marten-seemann:
> I tested your patch, but the response is successful. The test is inside the integration test. The OS is windows 10 amd64.
> 
> image
> 
> image
> 
> Any idea why?

Because it needs #4689.

WeidiDeng:
I added some of your changes though.

image

image

The first line is the debug output

WeidiDeng:
I see what's going on here. But I fixed it already.

marten-seemann:
@WeidiDeng Let's merge this PR, then you can rebase #4669.

WeidiDeng:
I actually [removed the new context parameter](https://github.com/quic-go/quic-go/pull/4669/files#diff-415989f1994f2599ac390daa6df5e43c8970139c46fc4cda1c1c91e2457207f0R498). I feel like my change just overwrites yours.

marten-seemann:
> I actually [removed the new context parameter](https://github.com/quic-go/quic-go/pull/4669/files#diff-415989f1994f2599ac390daa6df5e43c8970139c46fc4cda1c1c91e2457207f0R498). I feel like my change just overwrites yours.

We can deal with this on your PR. This PR is correct in and of itself, and we need a very similar behavior for graceful shutdown (once the context expires).

PR #4691, opened 2024-10-10
http3: send GOAWAY when server is shutting down

marten-seemann:
Continuation of #4669. Part of #153.


marten-seemann:
This should be ready to go now.

Issue #4705, opened 2024-10-18
http3: race-condition triggers panic when closing server


SlIdE42:
commit 04ecd301c0fbf8b51625683b4a76707ff8c57b9d (pull request #4691) introduce a new race condition into function `decreaseConnCount()` from `http3/server.go`:

https://github.com/quic-go/quic-go/blob/eaa879f32f141d455519c8666736b2ede15307e6/http3/server.go#L289

when heavy load, channel s.connHandlingDone could be close more than once and panic:

```
panic: close of closed channel

goroutine 49295 [running]:
[github.com/quic-go/quic-go/http3.(*Server).decreaseConnCount(...)](http://github.com/quic-go/quic-go/http3.(*Server).decreaseConnCount(...))
	[github.com/quic-go/quic-go@v0.48.0/http3/server.go:289](http://github.com/quic-go/quic-go@v0.48.0/http3/server.go:289)
[github.com/quic-go/quic-go/http3.(*Server).serveListener.func1](http://github.com/quic-go/quic-go/http3.(*Server).serveListener.func1)()
	[github.com/quic-go/quic-go@v0.48.0/http3/server.go:340](http://github.com/quic-go/quic-go@v0.48.0/http3/server.go:340) +0x140
created by [github.com/quic-go/quic-go/http3.(*Server).serveListener](http://github.com/quic-go/quic-go/http3.(*Server).serveListener) in goroutine 82
	[github.com/quic-go/quic-go@v0.48.0/http3/server.go:333](http://github.com/quic-go/quic-go@v0.48.0/http3/server.go:333) +0x2c
```

PR #4707, opened 2024-10-19
http3: fix graceful server shutdown

WeidiDeng:
Fix [6643](https://github.com/caddyserver/caddy/issues/6643#event-14725985001). Also close the underlying quic listeners when there is no quic connection to clean them up.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -763,7 +763,7 @@ func (s *Server) Shutdown(ctx context.Context) error {
 	s.mutex.Unlock()
 
 	if s.connCount.Load() == 0 {
-		return nil
+		return s.Close()
 	}
 	select {
 	case <-s.connHandlingDone: // all connections were closed
--- a/http3/server.go
+++ b/http3/server.go
@@ -222,6 +222,7 @@ type Server struct {
 	graceCtx         context.Context    // canceled when the server is closed or gracefully closed
 	graceCancel      context.CancelFunc // cancels the graceCtx
 	connCount        atomic.Int64
+	connDoneOnce     atomic.Bool
 	connHandlingDone chan struct{}
 
 	altSvcHeader string
@@ -285,7 +286,7 @@ func (s *Server) init() {
 }
 
 func (s *Server) decreaseConnCount() {
-	if s.connCount.Add(-1) == 0 {
+	if s.connCount.Add(-1) == 0 && s.graceCtx.Err() != nil && s.connDoneOnce.CompareAndSwap(false, true) {
 		close(s.connHandlingDone)
 	}
 }
--- a/http3/server.go
+++ b/http3/server.go
@@ -222,7 +222,6 @@ type Server struct {
 	graceCtx         context.Context    // canceled when the server is closed or gracefully closed
 	graceCancel      context.CancelFunc // cancels the graceCtx
 	connCount        atomic.Int64
-	connDoneOnce     atomic.Bool
 	connHandlingDone chan struct{}
 
 	altSvcHeader string
@@ -286,7 +285,7 @@ func (s *Server) init() {
 }
 
 func (s *Server) decreaseConnCount() {
-	if s.connCount.Add(-1) == 0 && s.graceCtx.Err() != nil && s.connDoneOnce.CompareAndSwap(false, true) {
+	if s.connCount.Add(-1) == 0 && s.graceCtx.Err() != nil {
 		close(s.connHandlingDone)
 	}
 }


marten-seemann:
This fix is wrong / lazy. We shouldn't be adding new connections once the server is closed.

WeidiDeng:
> This fix is wrong / lazy. We shouldn't be adding new connections once the server is closed.

The server is not closing at that moment. Why do you think it's adding new connections once the server is closed? I linked a issue from caddy, when the user is running caddy normally without stopping, not [this issue](https://github.com/quic-go/quic-go/issues/4705).

Yes, you shouldn't add new connections after shutdown starts. But I'm talking about when the server is running normally.

WeidiDeng:
I changed the title to better reflect what is being fixed.

marten-seemann:
#4705 tracks the issue. This PR claims to fix the issue, and it does, in a certain sense of "fixing".

The problem is that the connection counter goes up and down when connections are added / are closed. When the server is closed, and it hits 0, the channel is closed.

Now the problem is that we don't properly stop adding new connections when the server is closed, which is why the counter might hit 0 twice. This is the race condition we need to fix.

This PR "fixes" the problem, but not by addressing the race condition, but by making sure that the channel is only closed once. This is why I called it wrong / lazy before. I don't believe that there's anything in this PR that we can use for the actual fix, hence I closed it.

I'm planning to work on a proper fix next week.

francislavoie:
> This is why I called it wrong / lazy before.

Okay. I just want to make sure there's no bad blood here, because I'm hearing frustration from @WeidiDeng about his attempts to collaborate. I think we all have the same goals!

> I'm planning to work on a proper fix next week.

Okay, in that case I think we should probably revert Caddy to 0.47.0 for the time being (to unblock ourselves from making a further Caddy beta release), since 0.48.0 has this problem for now.

marten-seemann:
> Okay. I just want to make sure there's no bad blood here, because I'm hearing frustration from @WeidiDeng about his attempts to collaborate. I think we all have the same goals!

Sorry, bad choice of words!

I took another look at this PR, I'm wondering if all we need to do is this check:
```
if s.connCount.Add(-1) == 0 && s.graceCtx.Err() != nil {
```

Maybe there's no race condition in the first place? Then we don't need the atomic. @WeidiDeng, what do you think?

marten-seemann:
I see, you already made the change. I reopened the PR. Sorry again!

Have you tested if this new version actually fixes the problem? Would be great to get confirmation before I cut a patch release.

marten-seemann (approved):

francislavoie:
It was tested by a user here https://github.com/caddyserver/caddy/issues/6643#issuecomment-2423758401

marten-seemann:
> It was tested by a user here [caddyserver/caddy#6643 (comment)](https://github.com/caddyserver/caddy/issues/6643#issuecomment-2423758401)

Is that the version with or without the atomic? Because if there is a race condition in the connection counting logic somewhere, we might still run into the panic (because the counter can reach 0 multiple times after closing).

francislavoie:
Ah, that test was before that commit. We can ask them to try it again.

WeidiDeng:
@marten-seemann The original title of the pr is `Fix shutdown`, I was not very considerate when choosing the title of this pr, I guess that confused you so you think I was attempting to fix the double close channel when the server shuts down, which I kinda see why.

I was fixing the issue when caddy crashes when it's running normally without any restart or reload, after all, caddy will only shutdown http3 servers during these periods.

I later changed the title and remove the atomic guarding since I figured that caused you to misunderstand my intention.

Regarding the tests, I was able to reload caddy successfully using http3 client to send requests to caddy. http3.client will stall when caddy is reloading because they're not able to handle goaway frames yet. In a test environment using actual browsers, there is no problem either. But that's just my experience.

Speaking of goaway frame client-side handling, @marten-seemann do you still want a pr to be opened?

When using http3.Server directly in my program, I can create the race when `ServeQUICConn` is called directly, I wasn't able to create the race when using quic listeners.

WeidiDeng:
I noticed another problem with my current implementation though. If caddy is idle, during http3 shutdown, h3 shutdown will block here.

https://github.com/quic-go/quic-go/blob/6af2b1a478211ebf98125b996e13f6d994e6ba3e/http3/server.go#L746

This is because I added to close all quic listeners when the conn count is zero: this here is necessary for caddy to release resources

```
if s.connCount.Load() == 0 {
	return s.Close()
```

I can choose to close `connHandlingDone` here, but I feel this would be a race with connection counting with very exact timing,

```
if s.connCount.Add(-1) == 0 && s.graceCtx.Err() != nil {
```

As the order of execution may be like, :

1. `s.connCount.Add(-1) == 0` during `decreaseConnCount`,
2. `s.connCount.Load() == 0` during `Shutdown`,
3. `s.graceCtx.Err() != nil` during `decreaseConnCount`, 

In this order the channel will be closed twice, though I can't create this problem during testing. @marten-seemann If you have a good idea without using locks, please let me know.

marten-seemann:
> I noticed another problem with my current implementation though. If caddy is idle, during http3 shutdown, h3 shutdown will block here.
> 
> https://github.com/quic-go/quic-go/blob/6af2b1a478211ebf98125b996e13f6d994e6ba3e/http3/server.go#L746
> 
> This is because I added to close all quic listeners when the conn count is zero: this here is necessary for caddy to release resources
> 
> ```
> if s.connCount.Load() == 0 {
> 	return s.Close()
> ```

Can you elaborate how this can happen? Once we reach this check, we should've stopped handling new connections, no?

WeidiDeng:
The proposed change is to unblock the channel later, if the server succesfully added quic listeners but wasn't receiving new quic connections, or the server became idle because all connections were done and newer ones are not coming:

```
if s.connCount.Load() == 0 {
	close(s.connHandlingDone)
	return s.Close()
```

When the server did accept new connections, we stopped adding new connections here, but when finishing handling old connections, the counter will eventually go down to zero and close the channel.

Between different instances of `decreaseConnCount `, the channel will be closed once, but there can be race when channel is closed in `Shutdown` and `decreaseConnCount` at the exact moment when the counter is decreased and the check for shutdown, described in

1. `s.connCount.Add(-1) == 0` during `decreaseConnCount`,
2. `s.connCount.Load() == 0` during `Shutdown`,
3. `s.graceCtx.Err() != nil` during `decreaseConnCount`,

WeidiDeng:
Never mind, I just need to close the quic listeners then. There is no need to risk the race.

JeDaYoshi:
Yup, was about to come here to confirm that this fixes the issue for me.

marten-seemann:
Thank you @WeidiDeng for the fix, and thanks to everyone for testing!

> Speaking of goaway frame client-side handling, @marten-seemann do you still want a pr to be opened?

Yes, that would be great! Let's discuss the implementation strategy on #153 first though.

marten-seemann:
I have no idea what's wrong with stupid @codecov (again!). This PR literally doesn't change test coverage at all, let alone by hundreds of lines, as a 0.56% change would imply.

PR #4743, opened 2024-12-02
wait until handshaking connections have terminated when closing server

marten-seemann:
This should resolve the painful race condition we're seeing in the integration tests

### DIFF
--- a/server.go
+++ b/server.go
@@ -114,6 +114,7 @@ type baseServer struct {
 	invalidTokenQueue       chan rejectedPacket
 	connectionRefusedQueue  chan rejectedPacket
 	retryQueue              chan rejectedPacket
+	handshakingCount        sync.WaitGroup
 
 	verifySourceAddress func(net.Addr) bool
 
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -713,8 +716,9 @@ func (s *baseServer) handleInitialImpl(p receivedPacket, hdr *wire.Header) error
 		delete(s.zeroRTTQueues, hdr.DestConnectionID)
 	}
 
-	go conn.run()
+	s.handshakingCount.Add(1)
 	go func() {
```
sukunrt, line 720: This goroutine can push a connection on the conn queue right when the server is closing. Is that a problem?

sukunrt (approved):

server.go
sukunrt: Isn't there a race here? 
```Accept ...
	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	case conn := <-s.connQueue:
		return conn, nil
	case <-s.errorChan:
		return nil, s.closeErr
	}
```

sukunrt: connQueue has a connection and errorChan is closed.

marten-seemann: That is true. It's independent from this PR though. I'll open an issue.

marten-seemann: #4744

PR #4437, opened 2024-04-15
http3: process 1xx status codes

mchtech:
Fix #4403
copy the impletation from the go standard library and add tests.

PR #4749, opened 2024-12-06
http3: add client trace support

### LINKED ISSUE #3342: httptrace support
Recently, I was using http3 for server development, and then I tested it with K6. I found that K6 does not support http3, Then I joined guic to K6. Quic can't support httptrace, and then I can't do a very detailed performance test. Please support it
PR here:

https://github.com/grafana/k6/pull/2407

### LINKED ISSUE #4748: support for the capability of httptrace.ClientTrace
Hi，
I need to obtain the time spent in each stage of the HTTP request process, for example, DNS resolution time and connection establishment time. However, quic-go does not support the capabilities of httptrace. 

Currently, I am attempting to add these features. But when implementing the `GotConn` function, the net.Conn interface needs to be provided. However, neither `quic.EarlyConnection` nor `quic.Stream` implements this interface. My current thought is to create a new struct named `fakeConn` that combines Stream and EarlyConnection. 

Below is my implementation:

```go
// fakeConn is a wrapper around quic.EarlyConnection and quic.Stream,
// because quic connection or stream doesn't implement net.Conn.
type fakeConn struct {
	conn   quic.EarlyConnection
	stream quic.Stream
}

func (c *fakeConn) Close() error                       { return c.stream.Close() }
func (c *fakeConn) Read(p []byte) (int, error)         { return c.stream.Read(p) }
func (c *fakeConn) Write(p []byte) (int, error)        { return c.stream.Write(p) }
func (c *fakeConn) SetDeadline(t time.Time) error      { return c.stream.SetDeadline(t) }
func (c *fakeConn) SetReadDeadline(t time.Time) error  { return c.stream.SetReadDeadline(t) }
func (c *fakeConn) SetWriteDeadline(t time.Time) error { return c.stream.SetWriteDeadline(t) }
func (c *fakeConn) RemoteAddr() net.Addr               { return c.conn.RemoteAddr() }
func (c *fakeConn) LocalAddr() net.Addr                { return c.conn.LocalAddr() }

func traceGotConn(trace *httptrace.ClientTrace, conn quic.EarlyConnection, reused bool) {
	if trace != nil && trace.GotConn != nil {
		stream, _ := conn.OpenStream()
		trace.GotConn(httptrace.GotConnInfo{
			Conn:   &fakeConn{conn: conn, stream: stream},
			Reused: reused,
		})
	}
}
```

Does anyone have any excellent suggestions for implementing GotConn?

lRoccoon:
Since the QUIC protocol's connection establishment involves TLS handshake, I think ConnectDone and TLSHandshakeStart hooks should be called after transport dial is completed, and TLSHandshakeDone after HandshakeComplete.

Notice: Wait100Continue not implemented as quic-go doesn't support handling 1xx responses. 

Fixes #4748. Fixes #3342.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -294,10 +294,13 @@ func (c *ClientConn) sendRequestBody(str Stream, body io.ReadCloser, contentLeng
 }
 
 func (c *ClientConn) doRequest(req *http.Request, str *requestStream) (*http.Response, error) {
+	trace := httptrace.ContextClientTrace(req.Context())
 	if err := str.SendRequestHeader(req); err != nil {
+		traceWroteRequest(trace, err)
 		return nil, err
 	}
 	if req.Body == nil {
+		traceWroteRequest(trace, nil)
 		str.Close()
 	} else {
 		// send the request body asynchronously
@@ -308,7 +311,9 @@ func (c *ClientConn) doRequest(req *http.Request, str *requestStream) (*http.Res
 			if req.ContentLength > 0 {
 				contentLength = req.ContentLength
 			}
-			if err := c.sendRequestBody(str, req.Body, contentLength); err != nil {
+			err := c.sendRequestBody(str, req.Body, contentLength)
+			traceWroteRequest(trace, err)
+			if err != nil {
 				if c.logger != nil {
 					c.logger.Debug("error writing request", "error", err)
 				}
@@ -318,7 +323,6 @@ func (c *ClientConn) doRequest(req *http.Request, str *requestStream) (*http.Res
 	}
 
 	// copy from net/http: support 1xx responses
-	trace := httptrace.ContextClientTrace(req.Context())
 	num1xx := 0               // number of informational 1xx headers received
 	const max1xxResponses = 5 // arbitrary bound on number of informational responses
 
@@ -338,10 +342,9 @@ func (c *ClientConn) doRequest(req *http.Request, str *requestStream) (*http.Res
 			if num1xx > max1xxResponses {
 				return nil, errors.New("http: too many 1xx informational responses")
 			}
-			if trace != nil && trace.Got1xxResponse != nil {
-				if err := trace.Got1xxResponse(resCode, textproto.MIMEHeader(res.Header)); err != nil {
-					return nil, err
-				}
+			traceGot1xxResponse(trace, resCode, textproto.MIMEHeader(res.Header))
+			if resCode == 100 {
+				traceGot100Continue(trace)
 			}
 			continue
 		}
--- a/http3/http_stream.go
+++ b/http3/http_stream.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptrace"
 
 	"github.com/quic-go/quic-go"
 	"github.com/quic-go/quic-go/internal/protocol"
@@ -147,10 +148,12 @@ type requestStream struct {
 	reqDone            chan<- struct{}
 	disableCompression bool
 	response           *http.Response
+	trace              *httptrace.ClientTrace
 
 	sentRequest   bool
 	requestedGzip bool
 	isConnect     bool
+	firstByte     bool
 }
 
 var _ RequestStream = &requestStream{}
@@ -163,6 +166,7 @@ func newRequestStream(
 	disableCompression bool,
 	maxHeaderBytes uint64,
 	rsp *http.Response,
+	trace *httptrace.ClientTrace,
 ) *requestStream {
 	return &requestStream{
 		stream:             str,
@@ -172,6 +176,7 @@ func newRequestStream(
 		disableCompression: disableCompression,
 		maxHeaderBytes:     maxHeaderBytes,
 		response:           rsp,
+		trace:              trace,
 	}
 }
 
@@ -197,8 +202,12 @@ func (s *requestStream) SendRequestHeader(req *http.Request) error {
 
 func (s *requestStream) ReadResponse() (*http.Response, error) {
 	fp := &frameParser{
-		r:    s.Stream,
 		conn: s.conn,
+		r: &tracingReader{
+			Reader: s.Stream,
+			first:  &s.firstByte,
+			trace:  s.trace,
+		},
 	}
 	frame, err := fp.ParseNext()
 	if err != nil {
@@ -267,4 +276,19 @@ func (s *stream) SendDatagram(b []byte) error {
 func (s *stream) ReceiveDatagram(ctx context.Context) ([]byte, error) {
 	// TODO: reject if datagrams are not negotiated (yet)
 	return s.datagrams.Receive(ctx)
+}
+
+type tracingReader struct {
+	io.Reader
+	first *bool
+	trace *httptrace.ClientTrace
+}
+
+func (r *tracingReader) Read(b []byte) (int, error) {
+	n, err := r.Reader.Read(b)
+	if n > 0 && r.first != nil && !*r.first {
+		traceGotFirstResponseByte(r.trace)
+		*r.first = true
+	}
+	return n, err
 }
--- a/http3/ip_addr.go
+++ b/http3/ip_addr.go
@@ -1,0 +1,48 @@
+package http3
+
+import (
+	"net"
+	"strings"
+)
+
+// An addrList represents a list of network endpoint addresses.
+// Copy from [net.addrList] and change type from [net.Addr] to [net.IPAddr]
+type addrList []net.IPAddr
+
+// isIPv4 reports whether addr contains an IPv4 address.
+func isIPv4(addr net.IPAddr) bool {
+	return addr.IP.To4() != nil
+}
+
+// isNotIPv4 reports whether addr does not contain an IPv4 address.
+func isNotIPv4(addr net.IPAddr) bool { return !isIPv4(addr) }
+
+// forResolve returns the most appropriate address in address for
+// a call to ResolveTCPAddr, ResolveUDPAddr, or ResolveIPAddr.
+// IPv4 is preferred, unless addr contains an IPv6 literal.
+func (addrs addrList) forResolve(network, addr string) net.IPAddr {
+	var want6 bool
+	switch network {
+	case "ip":
+		// IPv6 literal (addr does NOT contain a port)
+		want6 = strings.ContainsRune(addr, ':')
+	case "tcp", "udp":
+		// IPv6 literal. (addr contains a port, so look for '[')
+		want6 = strings.ContainsRune(addr, '[')
+	}
+	if want6 {
+		return addrs.first(isNotIPv4)
+	}
+	return addrs.first(isIPv4)
+}
+
+// first returns the first address which satisfies strategy, or if
+// none do, then the first address of any kind.
+func (addrs addrList) first(strategy func(net.IPAddr) bool) net.IPAddr {
+	for _, addr := range addrs {
+		if strategy(addr) {
+			return addr
+		}
+	}
+	return addrs[0]
+}
--- a/http3/request_writer.go
+++ b/http3/request_writer.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/http/httptrace"
 	"strconv"
 	"strings"
 	"sync"
@@ -42,8 +43,12 @@ func (w *requestWriter) WriteRequestHeader(str quic.Stream, req *http.Request, g
 	if err := w.writeHeaders(buf, req, gzip); err != nil {
 		return err
 	}
-	_, err := str.Write(buf.Bytes())
-	return err
+	if _, err := str.Write(buf.Bytes()); err != nil {
+		return err
+	}
+	trace := httptrace.ContextClientTrace(req.Context())
+	traceWroteHeaders(trace)
+	return nil
 }
 
 func (w *requestWriter) writeHeaders(wr io.Writer, req *http.Request, gzip bool) error {
@@ -198,16 +203,16 @@ func (w *requestWriter) encodeHeaders(req *http.Request, addGzipHeader bool, tra
 	// 	return errRequestHeaderListSize
 	// }
 
-	// trace := httptrace.ContextClientTrace(req.Context())
-	// traceHeaders := traceHasWroteHeaderField(trace)
+	trace := httptrace.ContextClientTrace(req.Context())
+	traceHeaders := traceHasWroteHeaderField(trace)
 
 	// Header list size is ok. Write the headers.
 	enumerateHeaders(func(name, value string) {
 		name = strings.ToLower(name)
 		w.encoder.WriteField(qpack.HeaderField{Name: name, Value: value})
-		// if traceHeaders {
-		// 	traceWroteHeaderField(trace, name, value)
-		// }
+		if traceHeaders {
+			traceWroteHeaderField(trace, name, value)
+		}
 	})
 
 	return nil
--- a/http3/transport.go
+++ b/http3/transport.go
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"net/http/httptrace"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -197,7 +198,9 @@ func (t *Transport) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Res
 		return nil, fmt.Errorf("http3: invalid method %q", req.Method)
 	}
 
+	trace := httptrace.ContextClientTrace(req.Context())
 	hostname := authorityAddr(hostnameFromURL(req.URL))
+	traceGetConn(trace, hostname)
 	cl, isReused, err := t.getClient(req.Context(), hostname, opt.OnlyCachedConn)
 	if err != nil {
 		return nil, err
@@ -213,6 +216,7 @@ func (t *Transport) RoundTripOpt(req *http.Request, opt RoundTripOpt) (*http.Res
 		t.removeClient(hostname)
 		return nil, cl.dialErr
 	}
+	traceGotConn(trace, cl.conn, isReused)
 	defer cl.useCount.Add(-1)
 	rsp, err := cl.rt.RoundTrip(req)
 	if err != nil {
@@ -313,19 +317,48 @@ func (t *Transport) dial(ctx context.Context, hostname string) (quic.EarlyConnec
 			t.transport = &quic.Transport{Conn: udpConn}
 		}
 		dial = func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error) {
-			udpAddr, err := net.ResolveUDPAddr("udp", addr)
+			network := "udp"
+			udpAddr, err := t.resolveUDPAddr(ctx, network, addr)
 			if err != nil {
 				return nil, err
 			}
-			return t.transport.DialEarly(ctx, udpAddr, tlsCfg, cfg)
+			trace := httptrace.ContextClientTrace(ctx)
+			traceConnectStart(trace, network, udpAddr.String())
+			traceTLSHandshakeStart(trace)
+			conn, err := t.transport.DialEarly(ctx, udpAddr, tlsCfg, cfg)
+			var state tls.ConnectionState
+			if conn != nil {
+				state = conn.ConnectionState().TLS
+			}
+			traceTLSHandshakeDone(trace, state, err)
+			traceConnectDone(trace, network, udpAddr.String(), err)
+			return conn, err
 		}
 	}
-
 	conn, err := dial(ctx, hostname, tlsConf, t.QUICConfig)
 	if err != nil {
 		return nil, nil, err
 	}
 	return conn, t.newClient(conn), nil
+}
+
+func (t *Transport) resolveUDPAddr(ctx context.Context, network, addr string) (*net.UDPAddr, error) {
+	host, portStr, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, err
+	}
+	port, err := net.LookupPort(network, portStr)
+	if err != nil {
+		return nil, err
+	}
+	resolver := net.DefaultResolver
+	ipAddrs, err := resolver.LookupIPAddr(ctx, host)
+	if err != nil {
+		return nil, err
+	}
+	addrs := addrList(ipAddrs)
+	ip := addrs.forResolve(network, addr)
+	return &net.UDPAddr{IP: ip.IP, Port: port, Zone: ip.Zone}, nil
 }
 
 func (t *Transport) removeClient(hostname string) {
--- a/http3/conn.go
+++ b/http3/conn.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"net/http/httptrace"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -123,7 +124,8 @@ func (c *connection) openRequestStream(
 		rsp.Trailer = hdr
 		return nil
 	})
-	return newRequestStream(hstr, requestWriter, reqDone, c.decoder, disableCompression, maxHeaderBytes, rsp), nil
+	trace := httptrace.ContextClientTrace(ctx)
+	return newRequestStream(hstr, requestWriter, reqDone, c.decoder, disableCompression, maxHeaderBytes, rsp, trace), nil
 }
 
 func (c *connection) decodeTrailers(r io.Reader, l, maxHeaderBytes uint64) (http.Header, error) {
--- a/http3/http_stream_test.go
+++ b/http3/http_stream_test.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"net/http/httptrace"
 
 	mockquic "github.com/quic-go/quic-go/internal/mocks/quic"
 	"github.com/quic-go/quic-go/internal/protocol"
@@ -165,6 +166,7 @@ var _ = Describe("Request Stream", func() {
 			true,
 			math.MaxUint64,
 			&http.Response{},
+			&httptrace.ClientTrace{},
 		)
 	})
 
--- a/http3/trace.go
+++ b/http3/trace.go
@@ -1,0 +1,105 @@
+package http3
+
+import (
+	"crypto/tls"
+	"net"
+	"net/http/httptrace"
+	"net/textproto"
+	"time"
+
+	"github.com/quic-go/quic-go"
+)
+
+func traceGetConn(trace *httptrace.ClientTrace, hostPort string) {
+	if trace != nil && trace.GetConn != nil {
+		trace.GetConn(hostPort)
+	}
+}
+
+// fakeConn is a wrapper for quic.EarlyConnection
+// because the quic connection does not implement net.Conn.
+type fakeConn struct {
+	conn quic.EarlyConnection
+}
+
+func (c *fakeConn) Close() error                       { panic("connection operation prohibited") }
+func (c *fakeConn) Read(p []byte) (int, error)         { panic("connection operation prohibited") }
+func (c *fakeConn) Write(p []byte) (int, error)        { panic("connection operation prohibited") }
+func (c *fakeConn) SetDeadline(t time.Time) error      { panic("connection operation prohibited") }
+func (c *fakeConn) SetReadDeadline(t time.Time) error  { panic("connection operation prohibited") }
+func (c *fakeConn) SetWriteDeadline(t time.Time) error { panic("connection operation prohibited") }
+func (c *fakeConn) RemoteAddr() net.Addr               { return c.conn.RemoteAddr() }
+func (c *fakeConn) LocalAddr() net.Addr                { return c.conn.LocalAddr() }
+
+func traceGotConn(trace *httptrace.ClientTrace, conn quic.EarlyConnection, reused bool) {
+	if trace != nil && trace.GotConn != nil {
+		trace.GotConn(httptrace.GotConnInfo{
+			Conn:   &fakeConn{conn: conn},
+			Reused: reused,
+		})
+	}
+}
+
+func traceGotFirstResponseByte(trace *httptrace.ClientTrace) {
+	if trace != nil && trace.GotFirstResponseByte != nil {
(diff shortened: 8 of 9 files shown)


http3/transport.go
```diff
@@ -96,6 +97,8 @@ type Transport struct {
 	// However, if the user explicitly requested gzip it is not automatically uncompressed.
 	DisableCompression bool
 
+	Resolver *net.Resolver
```
marten-seemann, line 100: Do we need this? Why does the `http.Transport` not have a field like this?

marten-seemann (commented):
This will also need unit and integration tests.

http3/transport.go
marten-seemann: `http.Transport` does not embed a `net.Dialer`. Instead it has `Dial` callback, which can be used to inject a `net.Dialer`.

But then `http3.Transport` has such a callback as well.

marten-seemann:
> Notice: Wait100Continue not implemented as quic-go doesn't support handling 1xx responses.

We do! Have a look at the HTTP integration tests, there are two test cases for that.

http3/transport.go
lRoccoon: ok, I'll change it.

lRoccoon:
> > Notice: Wait100Continue not implemented as quic-go doesn't support handling 1xx responses.
> 
> We do! Have a look at the HTTP integration tests, there are two test cases for that.

ok, let me give it a try.

lRoccoon:
> > Notice: Wait100Continue not implemented as quic-go doesn't support handling 1xx responses.
> 
> We do! Have a look at the HTTP integration tests, there are two test cases for that.

Wait100Continue is called if the Request specified `Expect: 100-continue`. I looked through the srouce code but couldn't find the relevant location. Can you help confirm this?

marten-seemann:
Support for 1xx was added in #4437. Does that help?

lRoccoon:
> Support for 1xx was added in #4437. Does that help?

[Expect: 100-continue ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect)

Expect: 100-continue is a request header that indicates the client will only send the request body after receiving an HTTP 100 response from the server. I did not find the relevant source code.

marten-seemann:
@mchtech Could you help us out here? Do I understand correctly that we have support for 1xx status codes, but not for the 100-continue header?

integrationtests/self/http_test.go
```diff
@@ -957,3 +957,118 @@ func TestHTTPStreamer(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, PRData, repl)
 }
+
+func TestHTTPClientTrace(t *testing.T) {
```
marten-seemann, line 961: Would it be useful to test the order of the events? You could pass them into an event channel.

integrationtests/self/http_test.go
```diff
@@ -957,3 +957,118 @@ func TestHTTPStreamer(t *testing.T) {
...
+	tlsHandshakeStart := false
+	tlsHandshakeDone := false
+	wroteHeaderField := false
+	wroteHeaders := false
+	wait100Continue := false
+	wroteRequest := false
+	trace := httptrace.ClientTrace{
+		GetConn: func(hostPort string) {
+			getConn = true
+			require.Equal(t, hostPort, fmt.Sprintf("localhost:%d", port))
```
marten-seemann, line 988: It's invalid to use `testing.T` concurrently.

marten-seemann (commented):
What's the status of this PR? Is it ready for review or is there more work to be done?

integrationtests/self/http_test.go
lRoccoon: good idea!

lRoccoon: fixed

integrationtests/self/http_test.go
lRoccoon: fixed

lRoccoon:
> What's the status of this PR? Is it ready for review or is there more work to be done?

PR is ready

lRoccoon:
@marten-seemann Is it possible to merge this pull request?

http3/transport.go
```diff
@@ -258,7 +263,10 @@ func (t *Transport) getClient(ctx context.Context, hostname string, onlyCached b
 		go func() {
 			defer close(cl.dialing)
 			defer cancel()
+			traceConnectStart(trace, "udp", hostname)
 			conn, rt, err := t.dial(ctx, hostname)
+			traceConnectDone(trace, "udp", hostname, err)
+			traceTLSHandshakeStart(trace)
```
marten-seemann, line 269: This is not correct. The TLS handshake is part of the QUIC handshake. You probably want to exclude address resolution though.

http3/client.go
```diff
@@ -196,9 +196,12 @@ func (c *ClientConn) roundTrip(req *http.Request) (*http.Response, error) {
 		// wait for the handshake to complete
 		earlyConn, ok := c.Connection.(quic.EarlyConnection)
 		if ok {
+			trace := httptrace.ContextClientTrace(req.Context())
 			select {
 			case <-earlyConn.HandshakeComplete():
+				traceTLSHandshakeDone(trace, earlyConn.ConnectionState().TLS, nil)
```
marten-seemann, line 202: This is not correct. Not every request executes the TLS handshakes (namely, if a connection is reused).

http3/client.go
```diff
@@ -196,9 +196,12 @@ func (c *ClientConn) roundTrip(req *http.Request) (*http.Response, error) {
...
 			case <-req.Context().Done():
+				traceTLSHandshakeDone(trace, earlyConn.ConnectionState().TLS, req.Context().Err())
```
marten-seemann, line 204: This is not correct. A context cancellation is not equivalent to a TLS handshake failure.

http3/client.go
```diff
@@ -343,6 +350,9 @@ func (c *ClientConn) doRequest(req *http.Request, str *requestStream) (*http.Res
 					return nil, err
```
marten-seemann, line 350: This should be refactored to call `traceGot1xxResponse`.

http3/transport.go
```diff
@@ -258,7 +263,10 @@ func (t *Transport) getClient(ctx context.Context, hostname string, onlyCached b
 		go func() {
 			defer close(cl.dialing)
 			defer cancel()
+			traceConnectStart(trace, "udp", hostname)
```
marten-seemann, line 266: Should this distinguish between "udp4" and "udp6"? What does the standard library implementation do? Do they use "tcp", or "tcp4" / "tcp6"?

integrationtests/self/http_test.go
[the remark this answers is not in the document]
lRoccoon: ok

http3/request_writer.go
```diff
@@ -201,13 +204,16 @@ func (w *requestWriter) encodeHeaders(req *http.Request, addGzipHeader bool, tra
 	// trace := httptrace.ContextClientTrace(req.Context())
 	// traceHeaders := traceHasWroteHeaderField(trace)
```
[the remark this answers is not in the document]
lRoccoon: ok

http3/transport.go
```diff
@@ -313,10 +322,23 @@ func (t *Transport) dial(ctx context.Context, hostname string) (quic.EarlyConnec
...
+				return nil, err
+			}
+			resolver := net.DefaultResolver
+			ipAddrs, err := resolver.LookupIPAddr(ctx, host)
+			if err != nil {
+				return nil, err
+			}
+			addrs := addrList(ipAddrs)
+			ip := addrs.forResolve(network, addr)
+			udpAddr := &net.UDPAddr{IP: ip.IP, Port: port, Zone: ip.Zone}
```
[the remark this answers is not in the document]
lRoccoon: The functions `DNSStart` and `DNSDone` are called by `nettrace.Trace`, which requires adding `ctx` to the dns resolve in order to be called correctly. 

https://github.com/golang/go/blob/master/src/net/http/httptrace/trace.go#L48
https://github.com/golang/go/blob/master/src/net/lookup.go#L313

http3/http_stream.go
```diff
@@ -211,6 +216,10 @@ func (s *requestStream) ReadResponse() (*http.Response, error) {
 		s.conn.CloseWithError(quic.ApplicationErrorCode(ErrCodeFrameUnexpected), "expected first frame to be a HEADERS frame")
 		return nil, errors.New("http3: expected first frame to be a HEADERS frame")
 	}
+	if !s.firstByte {
+		traceGotFirstResponseByte(s.trace)
```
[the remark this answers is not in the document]
lRoccoon: I initially hoped to place it in the `frameParser.ParseNext` method, but the scope of changes felt a bit large, and there was no way to carry down the `ctx`. Later, I chose to do it when reading the first header frame.

http3/request_writer.go
```diff
@@ -43,6 +44,8 @@ func (w *requestWriter) WriteRequestHeader(str quic.Stream, req *http.Request, g
 		return err
 	}
 	_, err := str.Write(buf.Bytes())
+	trace := httptrace.ContextClientTrace(req.Context())
+	traceWroteHeaders(trace)
```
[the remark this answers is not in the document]
lRoccoon: Yes, I'll change it.

http3/client.go
lRoccoon: ok

http3/transport.go
lRoccoon: The standard library supports `tcp`, `tcp4`, and `tcp6`; this information is carried down from `net.Dial`. However, quic-go does not follow this approach, so using the default `udp` means it will automatically handle ipv4 and ipv6.

http3/transport.go
lRoccoon: I also have some doubts about this part, because the QUIC handshake includes the TLS handshake, which is completely different from TCP + TLS. I referred to the implementation of curl, and the TLS handshake time is generally around 1-2 ms. I understand that this should not include network overhead; it is just a local processing delay. 

Here, I would also like to hear your suggestions.

lRoccoon: I looked at the implementation of the standard library, and I understand that the time taken by the `sysDialer.dialSingle` function corresponding to `ConnectStart` and `ConnectDone` can be related to the time taken by the `quic.Transport.dial` function.

http3/ip_addr.go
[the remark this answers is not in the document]
lRoccoon: The logic for handling DNS resolution can also be placed in `trace.go`, but I think this part is relatively independent and has been separated into its own file.

http3/http_stream.go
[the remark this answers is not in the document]
marten-seemann: You can wrap the stream's `Read` method.

http3/transport.go
marten-seemann: The TLS handshake should only be traced when we're actually establishing a new QUIC connection. That means that in the case of a reused connection, we don't call it at all.

http3/transport.go
```diff
@@ -276,6 +282,7 @@ func (t *Transport) getClient(ctx context.Context, hostname string, onlyCached b
 		}
 		select {
 		case <-cl.conn.HandshakeComplete():
+			traceTLSHandshakeDone(trace, cl.conn.ConnectionState().TLS, nil)
```
marten-seemann, line 285: This shouldn't be called for reused connections.

http3/transport.go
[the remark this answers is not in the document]
marten-seemann: I really, really dislike the complexity here. There must be a better solution.

How does the standard library HTTP/2 implementation handle this?

transport.go
```diff
@@ -204,7 +205,18 @@ func (t *Transport) DialEarly(ctx context.Context, addr net.Addr, tlsConf *tls.C
 	return t.dial(ctx, addr, "", tlsConf, conf, true)
 }
 
-func (t *Transport) dial(ctx context.Context, addr net.Addr, host string, tlsConf *tls.Config, conf *Config, use0RTT bool) (EarlyConnection, error) {
+func (t *Transport) dial(ctx context.Context, addr net.Addr, host string, tlsConf *tls.Config, conf *Config, use0RTT bool) (c EarlyConnection, err error) {
```
marten-seemann, line 208: Nit:
```suggestion
func (t *Transport) dial(ctx context.Context, addr net.Addr, host string, tlsConf *tls.Config, conf *Config, use0RTT bool) (_ EarlyConnection, err error) {
```

marten-seemann:
@lRoccoon Are you planning to address my most recent comments? Would love to get this merged soon!

http3/transport.go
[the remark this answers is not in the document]
lRoccoon: http2 use `net.Dialer.DialContext`, but quic-go use `WriteTo`

lRoccoon:
> @lRoccoon Are you planning to address my most recent comments? Would love to get this merged soon!

I am working on the part of connection multiplexing that does not call the TLS hooks, which may take some more time.

transport.go
lRoccoon: fixed

http3/transport.go
lRoccoon: removed

lRoccoon:
@marten-seemann I deleted the implementations of TLSHandshakeStart and TLSHandshakeDone. After reading the code, I found that it is impossible to accurately distinguish between QUIC connection establishment and TLS handshake; these two processes are intertwined.

marten-seemann:
It's not impossible, they are just happening at the same time. Therefore, you should emit `TLSHandshakeStart` right when dialing a new connection.

http3/http_stream.go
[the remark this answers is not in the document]
lRoccoon: I tried to add a new Reader to accurately call `GotFirstResponseByte`, but there are many references to `quic.Stream` in `http3.stream`, and the scope of changes would be quite large. I don't think it's worth making such significant changes just for tracing. 

I also re-read the code for http2 and found that h2 calls it after reading the complete header frame, so I believe this handling is reasonable.

https://github.com/golang/go/blob/master/src/net/http/h2_bundle.go#L9992

http3/transport.go
lRoccoon: I gave up. In QUIC, it seems that the connection establishment and TLS handshake cannot be accurately separated; they are both included in the QUIC handshake process.

http3/http_stream.go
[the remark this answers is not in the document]
marten-seemann: `quic.Stream` is an interface, so you can easily wrap it:
```go
type wrappedStream struct {
    quic.Stream
}

func (s *wrappedStream) Read(b []byte) (int, error) {
     // check if this is the first read
     // ...
}
```

> I also re-read the code for http2 and found that h2 calls it after reading the complete header frame, so I believe this handling is reasonable.
>
> https://github.com/golang/go/blob/master/src/net/http/h2_bundle.go#L9992

You might have noticed there's a TODO there.

lRoccoon:
> It's not impossible, they are just happening at the same time. Therefore, you should emit `TLSHandshakeStart` right when dialing a new connection.

In the traditional TCP timing process, it is generally believed that the total time is divided into DNS + Connection + TLS + Request Send + Waiting + Content Download. If Connection includes TLS, then the calculation of the total time may be inaccurate.

https://blog.cloudflare.com/a-question-of-timing/

http3/transport.go
marten-seemann: You shouldn't separate it. Just emit the event at the same time you emit the `ConnectStart` event. That actually is **exactly** the right thing to do.

Same goes for `ConnectDone` and `TLSHandshakeDone` (be careful of 0-RTT though).

http3/http_stream.go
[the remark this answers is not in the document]
lRoccoon: Let me try.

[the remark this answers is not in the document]
lRoccoon: fixed, I encapsulated a reader in ReadResponse.

lRoccoon:
> > It's not impossible, they are just happening at the same time. Therefore, you should emit `TLSHandshakeStart` right when dialing a new connection.
> 
> In the traditional TCP timing process, it is generally believed that the total time is divided into DNS + Connection + TLS + Request Send + Waiting + Content Download. If Connection includes TLS, then the calculation of the total time may be inaccurate.
> 
> After careful consideration, under the QUIC protocol, to implement something similar to traditional TCP timing, either Connection or TLS can be chosen, but not both.
> 
> https://blog.cloudflare.com/a-question-of-timing/

@marten-seemann Can you help take a look at this issue? Many testing tools, such as [httpstat](https://github.com/davecheney/httpstat), rely on this logic. If the Connect and TLS handshake stages overlap, it can cause errors in these testing tools.

marten-seemann:
> @marten-seemann Can you help take a look at this issue? Many testing tools, such as [httpstat](https://github.com/davecheney/httpstat), rely on this logic. If the Connect and TLS handshake stages overlap, it can cause errors in these testing tools.

That would be a bug in these tools then. If they want to support HTTP/3, they need to deal with the fact QUIC runs connection establishment and the TLS handshake at the same time. Fundamentally, there's no way around this: this is how QUIC was designed, and for very good reason (we save 1 RTT during connection establishment!).

I'm strongly opposed to emitting wrong information in order to work around bugs in other libraries.

lRoccoon:
> That would be a bug in these tools then. If they want to support HTTP/3, they need to deal with the fact QUIC runs connection establishment and the TLS handshake at the same time. Fundamentally, there's no way around this: this is how QUIC was designed, and for very good reason (we save 1 RTT during connection establishment!).
> 
> I'm strongly opposed to emitting wrong information in order to work around bugs in other libraries.
@marten-seemann Okk，I‘ll modify it according to the correct implementation.

lRoccoon:
@marten-seemann I have readjusted the logic of the TLS handshake hooks, please help confirm it.

http3/http_stream.go
```diff
@@ -268,3 +278,23 @@ func (s *stream) ReceiveDatagram(ctx context.Context) ([]byte, error) {
 	// TODO: reject if datagrams are not negotiated (yet)
 	return s.datagrams.Receive(ctx)
 }
+
+type warpReader struct {
```
marten-seemann, line 282:
```suggestion
type tracingReader struct {
```

http3/http_stream.go
```diff
@@ -268,3 +278,23 @@ func (s *stream) ReceiveDatagram(ctx context.Context) ([]byte, error) {
...
+	io.Reader
+	first bool
+	cb    func()
```
marten-seemann, line 285: I'd prefer to have the trace here:
```suggestion
	trace *httptrace.ClientTrace
```

http3/http_stream.go
```diff
@@ -268,3 +278,23 @@ func (s *stream) ReceiveDatagram(ctx context.Context) ([]byte, error) {
...
+		return r.Reader.Read(b)
+	}
+	n1, err := r.Reader.Read(b[:1])
+	if err != nil {
+		return 0, err
+	}
+	r.first = true
+	r.cb()
+	n2, err := r.Reader.Read(b[1:])
+	return n1 + n2, err
```
marten-seemann, line 299: This can be simplified. An `io.Reader` will return `n < len(b)` if less than `len(b)` bytes are available:
```suggestion
	n, err := r.Reader.Read(b)
	if n > 0 && !r.first {
		traceGotFirstResponseByte(s.trace)
		r.first = true
	}
	return n, err
}
```

http3/http_stream.go
```diff
@@ -147,10 +148,12 @@ type requestStream struct {
 	reqDone            chan<- struct{}
 	disableCompression bool
 	response           *http.Response
+	trace              *httptrace.ClientTrace
 
 	sentRequest   bool
 	requestedGzip bool
 	isConnect     bool
+	firstByte     bool
```
marten-seemann, line 156: I believe you don't need this anymore, since this is now tracked by the wrapper reader, right?

transport.go
```diff
@@ -204,7 +205,28 @@ func (t *Transport) DialEarly(ctx context.Context, addr net.Addr, tlsConf *tls.C
...
+	trace := httptrace.ContextClientTrace(ctx)
+	if trace != nil {
+		if trace.ConnectStart != nil {
+			trace.ConnectStart(addr.Network(), addr.String())
+		}
+		if trace.TLSHandshakeStart != nil {
+			trace.TLSHandshakeStart()
+		}
```
marten-seemann, line 216: For consistency, this should be a `traceConnectStart` and `traceTLSHandshakeStart` in trace.go, so you can avoid the `if` statements here.

transport.go
```diff
@@ -204,7 +205,28 @@ func (t *Transport) DialEarly(ctx context.Context, addr net.Addr, tlsConf *tls.C
 	return t.dial(ctx, addr, "", tlsConf, conf, true)
 }
 
-func (t *Transport) dial(ctx context.Context, addr net.Addr, host string, tlsConf *tls.Config, conf *Config, use0RTT bool) (EarlyConnection, error) {
+func (t *Transport) dial(ctx context.Context, addr net.Addr, host string, tlsConf *tls.Config, conf *Config, use0RTT bool) (c EarlyConnection, err error) {
+	trace := httptrace.ContextClientTrace(ctx)
+	if trace != nil {
```
marten-seemann, line 210: Move this directly above the call to `dial`. No need to trace before the setup calls.

transport.go
```diff
@@ -204,7 +205,28 @@ func (t *Transport) DialEarly(ctx context.Context, addr net.Addr, tlsConf *tls.C
...
+		if trace.ConnectStart != nil {
+			trace.ConnectStart(addr.Network(), addr.String())
+		}
+		if trace.TLSHandshakeStart != nil {
+			trace.TLSHandshakeStart()
+		}
+		defer func() {
```
marten-seemann, line 217: Nit: no need to defer. Just move this directly after the dial call.

transport.go
```diff
@@ -204,7 +205,28 @@ func (t *Transport) DialEarly(ctx context.Context, addr net.Addr, tlsConf *tls.C
...
+			if trace.TLSHandshakeDone != nil {
```
marten-seemann, line 218: Same here. The `if` should be moved to trace.go for consistency.

transport.go
```diff
@@ -204,7 +205,28 @@ func (t *Transport) DialEarly(ctx context.Context, addr net.Addr, tlsConf *tls.C
...
+				state := tls.ConnectionState{}
+				if c != nil {
```
marten-seemann, line 220: I believe this will become simpler when you don't `defer`, since `c != nil` if `err == nil`.

marten-seemann (commented):
Yes, the handshake tracing looks correct now! Thank you for making the changes!

http3/http_stream.go
lRoccoon: It still needs to be retained temporarily because `ReadResponse` will be called multiple times, and each call will new `tracingReader`. Therefore, this information needs to be preserved in the `requestStream`.

http3/http_stream.go
lRoccoon: fixed

http3/http_stream.go
lRoccoon: fixed

http3/http_stream.go
lRoccoon: fixed

lRoccoon:
@marten-seemann The code has been revised based on the comments.

transport.go
marten-seemann, line 1: I must have missed this in my last review. This can't be in the quic package, this needs to happen in the http3 package.

marten-seemann (changes requested):

transport.go
lRoccoon: @marten-seemann  This part is more challenging because Connect in http2 is implemented through `nettrace`, and it seems impossible to achieve in http3 package.

marten-seemann: We can't pollute the quic package with **anything** HTTP-related. There's a clear separation between the layers.

lRoccoon: @marten-seemann I discovered that `quic` also has a `tracer`. Can we implement Connect and TLS Hooks here and provide this capability to http3 through the tracer?

marten-seemann: We could, be we don't need to. We have everything we need in the http3 package: https://github.com/quic-go/quic-go/blob/ea6d198c3d71e2b4b865b886342455fd297debef/http3/transport.go#L288-L329

lRoccoon: The `dial` includes DNS+Connect, making it impossible to distinguish precisely, especially since the `Transport.Dial` function allows for customization.

marten-seemann: Should we say that if the user specifies a `Transport.Dial`, it's their responsibility to trace `DNSStart`, `DNSDone`, `ConnectStart`, `ConnectDone`, `TLSHandshakeStart` and `TLSHandshakeDone`?

Fundamentally, we can't know what the user's callback actually does. As an extreme example, a user could return an existing connection from an application-level connection pool, so there might not even be any DNS resolution, TLS handshake and connection establishment.

lRoccoon: Indeed, it is implemented by the user themselves, but we should provide a reusable implementation to help users more easily achieve custom `Dial` support for httptrace.

marten-seemann: In the default implementation we should make the correct tracer calls, of course.

lRoccoon: Add `Transport.DefaultDial` method

marten-seemann: > Add `Transport.DefaultDial` method

No! Don't add any new method.
Just do DNS resolution in the http3 package, and trace the TLS and Connect events just before / after you call quic.Transport.Dial

lRoccoon: The `dial` is too long, so I moved the parsing logic into a separate function.

http3/http_stream.go
marten-seemann: Ok. Not a big fan of this, but I see your point. We might be able to refactor this by not initializing a new frame parser, but this can be done in a future PR. I'll open an issue for that.

marten-seemann (commented):
One small issue with the handshake done event. Other than this, this PR looks pretty good to me.

@lRoccoon Can you resolve this one? Then we should be ready to get this merged.

http3/transport.go
```diff
@@ -313,21 +317,50 @@ func (t *Transport) dial(ctx context.Context, hostname string) (quic.EarlyConnec
...
-			return t.transport.DialEarly(ctx, udpAddr, tlsCfg, cfg)
+			trace := httptrace.ContextClientTrace(ctx)
+			traceConnectStart(trace, network, udpAddr.String())
+			traceTLSHandshakeStart(trace)
+			conn, err := t.transport.DialEarly(ctx, udpAddr, tlsCfg, cfg)
+			var state tls.ConnectionState
+			if conn != nil {
+				state = conn.ConnectionState().TLS
+			}
+			traceTLSHandshakeDone(trace, state, err)
```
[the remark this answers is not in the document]
lRoccoon: okk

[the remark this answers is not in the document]
lRoccoon: @marten-seemann I tried to make some modifications, but found that the call to ConnectDone occurs after receiving the Response. I understand this might be related to 0-RTT, where data exchange can be completed simultaneously with the handshake when the request and response is small. Here's my implementation code; please help confirm if it meets expectations.

```go
dial = func(ctx context.Context, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlyConnection, error) {
	network := "udp"
	udpAddr, err := t.resolveUDPAddr(ctx, network, addr)
	if err != nil {
		return nil, err
	}
	trace := httptrace.ContextClientTrace(ctx)
	traceConnectStart(trace, network, udpAddr.String())
	traceTLSHandshakeStart(trace)
	conn, err := t.transport.DialEarly(ctx, udpAddr, tlsCfg, cfg)
	var state tls.ConnectionState
	if conn != nil {
		state = conn.ConnectionState().TLS
	}
	if trace != nil {
		go func() {
			select {
			case <-conn.HandshakeComplete():
				traceTLSHandshakeDone(trace, state, err)
				traceConnectDone(trace, network, udpAddr.String(), err)
			default:
			}
		}()
	}
	return conn, err
}
```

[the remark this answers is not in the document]
marten-seemann: Now this is definitely wrong. Connect needs to be traced before HandshakeDone.

> but found that the call to ConnectDone occurs after receiving the Response

I'm wondering if the Goroutine is the right solution to begin with: We know (from the mechanics of the TLS handshake and how keys become available that we'll only be able to *read* the server's response once the handshake is complete.

The ideal solution would be to trace HandshakeDone just before we read the first response sent by the server. Obviously, we only need to do that if we actually dialed a new connection.

Any ideas?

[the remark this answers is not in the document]
marten-seemann: I'm wondering if we can use a `*sync.Once` in the client, which is only set if HandshakeDone needs to be traced. Then in `clientConn.roundTrip`, we can trace the HandshakeDone from that `sync.Once`, thereby guaranteeing that it will only be emitted once per connection.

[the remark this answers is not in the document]
lRoccoon: I think when establishing a connection in 0-RTT, is it only the absence of `TLSHandshakeStart` and `TLSHandshakeDone`, and others should be consistent with 1-RTT? 

Then corresponding tracers should be provided at the QUIC level to achieve differentiation. Currently, it should be difficult to handle at the h3 level.

I tried to add `sync.Once` in `http3.ClientConn`, but this can only solve the problem that `TLSHandshakeDone` is not called in 0-RTT, and cannot achieve not calling `TLSHandshakeStart`. The implementation is rather strange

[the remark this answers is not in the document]
marten-seemann: > Then corresponding tracers should be provided at the QUIC level to achieve differentiation. Currently, it should be difficult to handle at the h3 level.

I don't think this is necessary. There's really nothing that the QUIC layer knows that the HTTP/3 layer doesn't know, so we can (and should!) implement everything in HTTP/3.

> I think when establishing a connection in 0-RTT, is it only the absence of TLSHandshakeStart and TLSHandshakeDone, and others should be consistent with 1-RTT?

The 0-RTT handshake still runs a TLS handshake. It's just happening in parallel with the first requests being sent to the server. So we still need the TLS events.

marten-seemann (approved):
Ok, let's get this PR in. There's a lot of good stuff here.

I've opened https://github.com/quic-go/quic-go/issues/4870 to document that 0-RTT is currently traced incorrectly, and given some hints on how we can resolve this in the future.

marten-seemann:
Thank you for all your hard work @lRoccoon!

PR #4752, opened 2024-12-11
update dependencies

phoenix6936:
update dependencies
@marten-seemann


marten-seemann:
There’s no good reason to do this whatsoever.

marten-seemann:
I don't see what we'd gain from this. Dependabot is just creating a massive amount of noise, distracting us from actual work.

PR #4762, opened 2024-12-18
drain server's accept queue before returning ErrClosed from Accept

marten-seemann:


server.go
```diff
@@ -328,6 +328,13 @@ func (s *baseServer) Accept(ctx context.Context) (Connection, error) {
...
 
 func (s *baseServer) accept(ctx context.Context) (quicConn, error) {
+	// first drain the queue...
+	select {
+	case conn := <-s.connQueue:
+		return conn, nil
+	default:
+	}
+	// ...only then return the error
 	select {
```
sukunrt, line 338: goroutines Waiting here will still have racy behavior right? 

if context Close and push to queue happen at the same time, then either of the select arms will execute.

I believe we need: 
```
	// first drain the queue...
	select {
	case conn := <-s.connQueue:
		return conn, nil
	default:
	}
	// ...only then return the error
	select {
	case <-ctx.Done():
	        // first drain the queue...
	        select {
	        case conn := <-s.connQueue:
		    return conn, nil
	        default:
	        }
		return nil, ctx.Err()
	case conn := <-s.connQueue:
		return conn, nil
	case <-s.errorChan:
	          // first drain the queue...
	        select {
	        case conn := <-s.connQueue:
		    return conn, nil
	        default:
	        }
		return nil, s.closeErr
	}
```

sukunrt (commented):
LGTM

PR #4773, opened 2024-12-19
drain receivedPackets buffer on connection close

MarcoPolo:
If packets were queued up in our receivedPackets queue when we closed the connection we would fail reuse them. Eventually the GC would have to clean this as trash.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -496,6 +496,22 @@ func (s *connection) run() error {
 	var closeErr closeError
 	defer func() { s.ctxCancel(closeErr.err) }()
 
+	defer func() {
+		// Drain queued packets that will never be processed.
+		for {
+			select {
+			case p, ok := <-s.receivedPackets:
+				if !ok {
+					return
+				}
+				p.buffer.Decrement()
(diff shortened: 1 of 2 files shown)


marten-seemann (approved):

connection.go
```diff
@@ -1650,6 +1650,22 @@ func (s *connection) handleCloseError(closeErr *closeError) {
 		s.tracer.ClosedConnection(e)
 	}
 
+	defer func() {
```
marten-seemann, line 1653: Should this happen at the end of the run loop?

MarcoPolo: That's where I had it originally, but moved it here to match the rest of cleanup code. I'll move it since I don't have a strong preference.

MarcoPolo: Remember that this needs to happen after we remove ourselves from the packet handler. I've moved this to a defer in the run method.

MarcoPolo: Feel free to move this somewhere else

marten-seemann (approved):
Thank you @MarcoPolo!

PR #4787, opened 2024-12-23
add a BLOCKED frame packetization test

marten-seemann:
For #4771. Still missing: the actual fix, but at least we already have a test case.


marten-seemann:
This test was added in #4801 and #4845.

PR #4798, opened 2024-12-26
http3: allow concurrent calls to Body.Close

RPRX:
Re-apply https://github.com/quic-go/quic-go/pull/4638

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -119,7 +119,6 @@ func (r *hijackableBody) Read(b []byte) (int, error) {
 
 func (r *hijackableBody) requestDone() {
 	if r.reqDone != nil {
-		// To make sure we won't close r.reqDone more than once.
 		r.reqDoneOnce.Do(func() {
 			close(r.reqDone)
 		})
--- a/http3/body.go
+++ b/http3/body.go
@@ -96,8 +96,8 @@ type hijackableBody struct {
 	// only set for the http.Response
 	// The channel is closed when the user is done with this response:
 	// either when Read() errors, or when Close() is called.
-	reqDone       chan<- struct{}
-	reqDoneClosed sync.Once
+	reqDone     chan<- struct{}
+	reqDoneOnce sync.Once
 }
 
 var _ io.ReadCloser = &hijackableBody{}
@@ -119,7 +119,8 @@ func (r *hijackableBody) Read(b []byte) (int, error) {
 
 func (r *hijackableBody) requestDone() {
 	if r.reqDone != nil {
-		r.reqDoneClosed.Do(func() {
+		// To make sure we won't close r.reqDone more than once.
+		r.reqDoneOnce.Do(func() {
 			close(r.reqDone)
 		})
 	}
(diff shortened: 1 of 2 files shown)


marten-seemann:
What is "more graceful" about this? Please describe the bug this is supposed to fix (or even better: open an issue for that).

Please also add a test case.

RPRX:
> What is "more graceful" about this? Please describe the bug this is supposed to fix (or even better: open an issue for that).

If the requestDone() is called concurrently, **the program may panic, because "close of closed channel"**: https://github.com/XTLS/Xray-core/issues/4148

http3/body.go
```diff
@@ -96,7 +97,7 @@ type hijackableBody struct {
 	// The channel is closed when the user is done with this response:
 	// either when Read() errors, or when Close() is called.
 	reqDone       chan<- struct{}
-	reqDoneClosed bool
+	reqDoneClosed sync.Once
```
marten-seemann, line 100:
```suggestion
	reqDoneOnce sync.Once
```

marten-seemann (commented):
Please also add a test case.

RPRX:
> Please also add a test case.

I'm not sure if we can make real concurrent calls that must call "close(r.reqDone)" concurrently.

Please feel free to edit the code / the commit message directly.

RPRX:
And I'm not sure if `if r.reqDone == nil {` should be deleted.

marten-seemann:
> I'm not sure if we can make real concurrent calls that must call "close(r.reqDone)" concurrently.

Sure, you can spawn 2 Go routines that both call `Close`. While that's not guaranteed to hit the race condition every time, at least from time to time it will.

RPRX:
I've added a line of comment, so that developers will handle these code carefully. I think it is more useful than a random test.

marten-seemann:
This comment doesn't add any value. Please remove it. And please add a test case, as I've requested for the third time now. It is useful, and this PR will not be merged without a test.

RPRX:
Well, I've add a test case, though "that's not guaranteed to hit the race condition every time". Please review it.

marten-seemann (approved):

http3/body_test.go
```diff
@@ -54,6 +54,15 @@ var _ = Describe("Response Body", func() {
 		Expect(rb.Close()).To(Succeed())
 	})
 
+	It("allows concurrent calls to Close", func() {
+		str := mockquic.NewMockStream(mockCtrl)
+		rb := newResponseBody(&stream{Stream: str}, -1, reqDone)
+		str.EXPECT().CancelRead(quic.StreamErrorCode(ErrCodeRequestCanceled)).MaxTimes(2)
+		go Expect(rb.Close()).To(Succeed())
```
marten-seemann, line 61:
```suggestion
		go func() {
			defer GinkgoRecover()
			Expect(rb.Close()).To(Succeed())
		}()
```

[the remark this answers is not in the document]
marten-seemann: @RPRX Can you apply this? For some reason, GitHub doesn't allow me to push to the branch.

RPRX:
@marten-seemann It's ready.

marten-seemann (approved):
Thank you for your contribution @RPRX!

PR #4804, opened 2024-12-27
fix accounting for lost RESET_STREAM frames in the send stream

### LINKED ISSUE #4803: send stream doesn't account for lost RESET_STREAM frames
The stream has a `numOutstandingFrames` counter, and only considers itself closed once this counter hits 0, and a few other conditions are met: for example the stream needs to have sent the FIN, or have been reset.

When a STREAM frames is lost, this number is reduced (which is correct):
https://github.com/quic-go/quic-go/blob/f6e7789854f034a592ba6f12bfbd84b639ef06e9/send_stream.go#L556-L565

However, when a RESET_STREAM frame is lost, it's not:
https://github.com/quic-go/quic-go/blob/f6e7789854f034a592ba6f12bfbd84b639ef06e9/send_stream.go#L592-L597

I suspect that this is incorrect. The effect is that the stream never marks itself for garbage collection, and we don't increase the peer's stream credit when the stream it closed. This might be the cause for some connection stalls that have been report, first and foremost #4712.

marten-seemann:
Fixes #4803.

~~This PR won't be merged before #4799, as I want to have decent tests for this, and I'm not going to write any more Ginkgo tests.~~

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -592,6 +592,7 @@ func (s *sendStreamResetStreamHandler) OnAcked(wire.Frame) {
 func (s *sendStreamResetStreamHandler) OnLost(wire.Frame) {
 	s.mutex.Lock()
 	s.queuedResetStreamFrame = true
+	s.numOutstandingFrames--
 	s.mutex.Unlock()
 	s.sender.onHasStreamControlFrame(s.streamID, (*sendStream)(s))
 }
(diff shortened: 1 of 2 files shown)


marten-seemann:
I just merged #4799, so this PR is ready to go 🚀

MarcoPolo (commented):
lgtm

PR #4844, opened 2025-01-07
reliably queue MAX_DATA frames

marten-seemann:
We need to make sure that a MAX_DATA frame actually triggers the packing of a new packet (with which it can then be sent out). Otherwise, reading from a stream might get the MAX_DATA frame queued, but if no packet is sent, it's never sent out to the peer, potentially leading to a deadlock of the connection.

### DIFF
--- a/receive_stream.go
+++ b/receive_stream.go
@@ -91,15 +91,18 @@ func (s *receiveStream) Read(p []byte) (int, error) {
 	defer func() { <-s.readOnce }()
 
 	s.mutex.Lock()
-	queuedNewControlFrame, n, err := s.readImpl(p)
+	queuedStreamWindowUpdate, queuedConnWindowUpdate, n, err := s.readImpl(p)
 	completed := s.isNewlyCompleted()
 	s.mutex.Unlock()
 
 	if completed {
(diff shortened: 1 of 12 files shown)


receive_stream.go
```diff
@@ -205,9 +207,13 @@ func (s *receiveStream) readImpl(p []byte) (bool, int, error) {
 		// when a RESET_STREAM was received, the flow controller was already
 		// informed about the final byteOffset for this stream
 		if !s.cancelledRemotely {
-			if queueMaxStreamData := s.flowController.AddBytesRead(protocol.ByteCount(m)); queueMaxStreamData {
+			hasStream, hasConn := s.flowController.AddBytesRead(protocol.ByteCount(m))
```
MarcoPolo, line 210: nit: may as well assign these to your vars `hasXWindowUpdate`

MarcoPolo (commented):
lgtm

receive_stream.go
marten-seemann: Depends on how the loop executes / returns. I think it should be fine, but I wanted to be very explicit about not setting value from `true` back to `false`.

PR #4846, opened 2025-01-07
drain server's accept queue before returning ErrClosed from Accept

### LINKED ISSUE #4744: race condition when draining server's accept queue
Isn't there a race here? 
```go 
func (s *baseServer) accept(ctx context.Context) (quicConn, error) {
	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	case conn := <-s.connQueue:
		return conn, nil
	case <-s.errorChan:
		return nil, s.closeErr
	}
}

```

_Originally posted by @sukunrt in https://github.com/quic-go/quic-go/pull/4743#discussion_r1866115876_

marten-seemann:
Alternative to #4762, which hopefully doesn't contain the race condition that @sukunrt identified there.

Fixes #4744. Closes #4762.

### DIFF
--- a/server.go
+++ b/server.go
@@ -105,10 +105,18 @@ type baseServer struct {
 		protocol.Version,
 	) quicConn
 
-	closeMx   sync.Mutex
-	errorChan chan struct{} // is closed when the server is closed
-	closeErr  error
-	running   chan struct{} // closed as soon as run() returns
+	closeMx sync.Mutex
+	// errorChan is closed when Close is called. This has two effects:
+	// 1. it cancels handshakes that are still in flight (using CONNECTION_REFUSED) errors
+	// 2. it stops handling of packets passed to this server
+	errorChan chan struct{}
+	// acceptChan is closed when Close returns.
+	// This only happens once all handshake in flight have either completed and canceled.
+	// Calls to Accept will first drain the queue of connections that have completed the handshake,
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -328,20 +337,25 @@ func (s *baseServer) Accept(ctx context.Context) (Connection, error) {
...
+	// ...only then return the error
 	select {
 	case <-ctx.Done():
 		return nil, ctx.Err()
 	case conn := <-s.connQueue:
 		return conn, nil
-	case <-s.errorChan:
+	case <-s.stopAccepting:
 		return nil, s.closeErr
 	}
```
sukunrt, line 354: it should be 

```
func (s *baseServer) accept(ctx context.Context) (quicConn, error) {
    select {
        case <-s.stopAccepting:
            select {
                case conn := <-s.connQueue:
                    return conn, nil
                default:
            }
            return nil, s.closeErr
        case conn :=  <-s.connQueue:
            return conn, nil
        case <-ctx.Done():
            return nil, ctx.Err()
}
```

Conider this sequence of events
1. conn queue is empty
2. A go routine is waiting on the second Accept select block
3. sever.Close() called
4. A handshake completes and pushes to s.connQueue
5. s.stopAccepting is closed.
6. The goroutine in _2_ can now either return the connection or s.closeErr

sukunrt (approved):

PR #4851, opened 2025-01-08
remove validation enforcing one Transport per net.PacketConn

marten-seemann:
It is invalid to use a `net.PacketConn` in multiple `Transport`s. However, the validation logic is causing pain when using wrapped `net.PacketConn`s. It was introduce to guard against incorrect uses of the API when the `Transport` was introduced, but this is probably less relevant now than it was back then.


MarcoPolo (approved):
Nice to see the large deletions 🗑️

Issue #4850, opened 2025-01-08
smarter connection ID selection logic


marten-seemann:
We currently rotate connection IDs after a fixed number of packets was sent:
https://github.com/quic-go/quic-go/blob/c42f8456ab2f493d9ea6af982115e204946d2045/conn_id_manager.go#L198-L202

While this logic is fine to exercise the mechanism every once in a while, it doesn't really do anything beyond that. Specifically, it doesn't reduce linkability between the old and the new CID. In order to do so, we should rotate CIDs after a period of quiescence (maybe a few seconds by default?).

PR #4852, opened 2025-01-08
fix memory leak on connection ID rotation when closing connection

marten-seemann:
In rare instances, the connection ID manager might switch to a new connection ID when sending the packet containing the CONNECTION_CLOSE frame. The connection ID manager removes the active stateless reset token from the packet handler map when it is closed, so we need to make sure that this happens last, otherwise the packet handler will keep a pointer to the closed connection indefinitely.

Thanks to @MarcoPolo for debugging and reporting this issue.

Unfortunately, I don't see a good way to test this for now, since a fresh connection (as we use in tests), comes with a connection ID that doesn't have a stateless reset token. #4850 might make this easier to test at some point.

### DIFF
--- a/conn_id_manager.go
+++ b/conn_id_manager.go
@@ -35,6 +35,8 @@ type connIDManager struct {
 	addStatelessResetToken    func(protocol.StatelessResetToken)
 	removeStatelessResetToken func(protocol.StatelessResetToken)
 	queueControlFrame         func(wire.Frame)
+
+	closed bool
 }
 
 func newConnIDManager(
@@ -148,6 +150,7 @@ func (h *connIDManager) addConnectionID(seq uint64, connID protocol.ConnectionID
 }
 
 func (h *connIDManager) updateConnectionID() {
+	h.assertNotClosed()
 	h.queueControlFrame(&wire.RetireConnectionIDFrame{
 		SequenceNumber: h.activeSequenceNumber,
 	})
@@ -166,6 +169,7 @@ func (h *connIDManager) updateConnectionID() {
 }
 
 func (h *connIDManager) Close() {
+	h.closed = true
 	if h.activeStatelessResetToken != nil {
 		h.removeStatelessResetToken(*h.activeStatelessResetToken)
 	}
@@ -182,6 +186,7 @@ func (h *connIDManager) ChangeInitialConnID(newConnID protocol.ConnectionID) {
 
 // is called when the server provides a stateless reset token in the transport parameters
 func (h *connIDManager) SetStatelessResetToken(token protocol.StatelessResetToken) {
+	h.assertNotClosed()
 	if h.activeSequenceNumber != 0 {
 		panic("expected first connection ID to have sequence number 0")
 	}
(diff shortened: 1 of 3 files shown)


MarcoPolo (changes requested):
This doesn't fix the issue. Did you mean to add a `defer`?

I think we should track whether the connIDManager is closed, and panic if we try to do something on a closed connIDManager, as I do in this patch: https://github.com/quic-go/quic-go/commit/8b9c8cb2f667. That will make it obvious when we do the wrong thing. It also makes it possible to test, since you test the invariant that the `connIDManager` is not closed when calling `PackApplicationClose`. I do this in that patch as well.

With the panic, we can see the stack trace of the failure and see that this is related to packing the connection close:
```
github.com/quic-go/quic-go.(*connIDManager).Get(0x102958140?)
        /Users/marco/pl/marten-seemann/quic-go/conn_id_manager.go:216 +0x9c
github.com/quic-go/quic-go.(*packetPacker).packConnectionClose(0x140005b8410, 0x1, 0x1, 0x0, {0x0, 0x0}, 0x500, 0x1)
        /Users/marco/pl/marten-seemann/quic-go/packet_packer.go:253 +0x500
github.com/quic-go/quic-go.(*packetPacker).PackApplicationClose(0x102a132a0?, 0x14002497460?, 0x102516600?, 0xb8eb58?)
        /Users/marco/pl/marten-seemann/quic-go/packet_packer.go:185 +0x3c
github.com/quic-go/quic-go.(*connection).sendConnectionClose(0x140031a5408, {0x102a132a0, 0x14002497460})
        /Users/marco/pl/marten-seemann/quic-go/connection.go:2148 +0x140
github.com/quic-go/quic-go.(*connection).handleCloseError(0x140031a5408, 0x14002d9fd68)
        /Users/marco/pl/marten-seemann/quic-go/connection.go:1668 +0x37c
github.com/quic-go/quic-go.(*connection).run(0x140031a5408)
        /Users/marco/pl/marten-seemann/quic-go/connection.go:645 +0x82c
github.com/quic-go/quic-go.(*client).dial.func1()
        /Users/marco/pl/marten-seemann/quic-go/client.go:214 +0x40
created by github.com/quic-go/quic-go.(*client).dial in goroutine 7977
        /Users/marco/pl/marten-seemann/quic-go/client.go:213 +0x47c
```

I generally don't like to panic if we can avoid it, but if this happens it's a serious issue and hard to debug.

marten-seemann:
> This doesn't fix the issue. Did you mean to add a `defer`?

Yes 🤦🏻‍♂️

Added in aeea67c4c.

I still need to think about panicking. Not really a fan of that, but I see your point.

MarcoPolo:
I think it's also okay if we didn't panic and change the return signatures instead to also return an error. Seems like a more annoying change.

MarcoPolo (approved):
Looks good. Would be nice to have this checked by tests :)

p-shahi:
@marten-seemann - we're waiting on this PR to land w/ a new quic-go patch release. Do you know the eta?

PR #4859, opened 2025-01-11
refactor client dialing logic, move it into the Transport

marten-seemann:
After the `Transport` refactor, there's little justification for keeping the `client` around.

Part of #3652.


marten-seemann:
Tests on Windows are - for once - not a bug / peculiarity of Windows, but an actual bug. What a nice surprise.

Will need to rebase once #4872 is merged.

PR #4878, opened 2025-01-15
reject NEW_CONNECTION_ID frames when using zero-length connection IDs

marten-seemann:
The RFC says:
> An endpoint that selects a zero-length connection ID during the handshake cannot issue a new connection ID. A zero-length Destination Connection ID field is used in all packets sent toward such an endpoint over any network path.


mxinden (approved):
:+1: for the record, this is how I understand RFC 9000 as well.

marten-seemann:
Thanks for checking @mxinden!

PR #4882, opened 2025-01-17
preserve error from cancelled SendStream during cancellation and closing

### LINKED ISSUE #4808: calling Close on a SendStream after cancelation should not change the error returned by Write
This applies to both `CancelWrite` and to receiving a STOP_SENDING frame.

marten-seemann:
Fixes #4808.

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -37,10 +37,10 @@ type sendStream struct {
 
 	writeOffset protocol.ByteCount
 
-	cancelWriteErr      *StreamError
-	closeForShutdownErr error
-
-	queuedResetStreamFrame bool
+	// finalError is the error that is returned by Write.
+	// It can either be a cancellation error or the shutdown error.
+	finalError             error
+	queuedResetStreamFrame *wire.ResetStreamFrame
 
 	finishedWriting bool // set once Close() is called
 	finSent         bool // set when a STREAM_FRAME with FIN bit has been sent
@@ -48,6 +48,8 @@ type sendStream struct {
 	// This can happen because the application called CancelWrite,
 	// or because Write returned the error (for remote cancellations).
 	cancellationFlagged bool
+	cancelled           bool // both local and remote cancellations
+	closedForShutdown   bool // set by closeForShutdown
 	completed           bool // set when this stream has been reported to the streamSender as completed
 
 	dataForWriting []byte // during a Write() call, this slice is the part of p that still needs to be sent out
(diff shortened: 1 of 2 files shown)


send_stream.go
```diff
@@ -512,7 +515,8 @@ func (s *sendStream) SetWriteDeadline(t time.Time) error {
 // The peer will NOT be informed about this: the stream is closed without sending a FIN or RST.
 func (s *sendStream) closeForShutdown(err error) {
 	s.mutex.Lock()
-	s.closeForShutdownErr = err
+	s.closedForShutdown = true
+	s.finalError = err
```
sukunrt, line 519: will this override errors on streams already reset? I'm guessing it'll only be called on open streams.

sukunrt (approved):

send_stream.go
```diff
@@ -426,13 +427,13 @@ func (s *sendStream) CancelWrite(errorCode StreamErrorCode) {
 
 func (s *sendStream) cancelWriteImpl(errorCode qerr.StreamErrorCode, remote bool) {
```
sukunrt, line 428: cancel after close will still overwrite the error.

send_stream.go
marten-seemann: Actually, you're making a good point here. If the stream is cancelled, it remains in the streams map until the RESET_STREAM frame has been acknowledged (usually 1 RTT).

We actually need to make sure to not replace the error here. I'll also see if I can add a test for that.

send_stream.go
marten-seemann: Yes, that's a feature, not a bug :)

You can cancel a stream after calling closing it. This will actually send a RESET_STREAM frame, and addition to the FIN that might already have been sent out.

The reason for this is that this cancels retransmissions for STREAM frames that might be still in flight.

marten-seemann: Let me add some more documentation for that, this is kind of unexpected.

marten-seemann: Done in 1c23ade.

send_stream.go
marten-seemann: Done in 3227966.

send_stream.go
sukunrt: >The reason for this is that this cancels retransmissions for STREAM frames that might be still in flight.

This only applies if the FIN hasn't been ACKed, right?

marten-seemann: It doesn't. There could be outstanding data before the FIN that you might want to cancel.

sukunrt (approved):

PR #4892, opened 2025-01-20
use RandomNonce in the init to allow fipsonly

juliens:
In the next Go release, FIPS improvements will be included. One of these is the FIPS-only mode, which triggers a panic if non-FIPS-compliant code is used.

When using quic-go, during the initialization of the handshake/retry process, there is a non-FIPS-compliant operation that causes a panic in FIPS-only mode.

I’m aware there are other issues related to FIPS-only mode in the code. However, the goal of this PR is to provide users with the ability to disable quic-go in FIPS mode. Without this PR, since the issue occurs during initialization, it’s not possible to disable it.

### DIFF
--- a/internal/handshake/retry.go
+++ b/internal/handshake/retry.go
@@ -25,7 +25,7 @@ func initAEAD(key [16]byte) cipher.AEAD {
 	if err != nil {
 		panic(err)
 	}
-	aead, err := cipher.NewGCM(aes)
+	aead, err := cipher.NewGCMWithRandomNonce(aes)
 	if err != nil {
 		panic(err)
 	}


marten-seemann:
I have to admit, I haven't looked at FIPS mode in Go (neither the old one nor the new one) in detail.

Using `NewGCMWithRandomNonce` for Retry seems wrong though, since Retry uses a fixed nonce, as defined in section 5.8 of RFC 9001.

Can you elaborate why quic-go is not FIPS-compliant? How is FIPS-only mode enabled? Can we just return an error in `Transport.init`?

juliens:
`cipher.NewGCMWithRandomNonce` is only available in Go 1.24, so we either need to wait for Go 1.24 or find an alternative solution

juliens:
Thank you @marten-seemann  for the clarification regarding the RFC. I wasn’t aware of the details in Section 5.8, and I now see how this directly conflicts with the FIPS requirements.

At the moment, I don’t see a clear solution to reconcile this. Our current goal is to allow disabling quic-go in a FIPS context. However, this isn’t fully possible since the issue arises in the init, which doesn’t provide an opportunity for us to conditionally disable it.

I’m not sure what the best approach would be here. Perhaps using build tags combined with a conditional check in the init function to determine if FIPS mode is required? What do you think?

marten-seemann:
> However, this isn’t fully possible since the issue arises in the init, which doesn’t provide an opportunity for us to conditionally disable it.

Got it. Now we could move to a lazy initialization (maybe using `sync.Once`), but I'd like to understand FIPS mode better.

Would it make sense to open an issue in the standard library to clarify how FIPS mode is supposed to interact with QUIC?

juliens:
> Would it make sense to open an issue in the standard library to clarify how FIPS mode is supposed to interact with QUIC?
Yes, I think that makes sense. We could gain valuable insights from those working on FIPS implementation in Go.

> Got it. Now we could move to a lazy initialization (maybe using sync.Once), but I'd like to understand FIPS mode better.
This can resolve the current issue, allowing us to implement a more robust fix once we receive clarification from the FIPS team. Do you want me to create a PR?

Thank you very much for your responsiveness!

marten-seemann:
I'd prefer to first open the issue and hear what the Go team thinks. I'd be surprised if it using QUIC was impossible when using FIPS mode (as QUIC _requires_ you to use a fixed nonce).

Could you open the issue and post the link here?

juliens:
I suggest opening a regular issue here (in the quic-go repository) instead of this “incorrect” PR. Then, I can create an issue in the Go repository and link the two issues.

marten-seemann:
Even better!

PR #4896, opened 2025-01-21
Fix tools.go

matthewmueller:
Hey folks, getting some annoying errors in my editor from this package:

CleanShot 2025-01-20 at 22 22 40@2x

Here's the full error:

> error while importing github.com/quic-go/quic-go: import "github.com/onsi/ginkgo/v2/ginkgo" is a program, not an importable package

This PR fixes this issue by importing paths that have .go files and aren't a main package.

This should be a no-op since it'll still download the tools via `go mod tidy`.

### DIFF
--- a/integrationtests/gomodvendor/go.mod
+++ b/integrationtests/gomodvendor/go.mod
@@ -8,6 +8,7 @@ toolchain go1.22.4
 require github.com/quic-go/quic-go v0.21.0
 
 require (
+	github.com/go-logr/logr v1.2.4 // indirect
 	github.com/go-task/slim-sprig v0.0.0-20230315185526-52ccab3ef572 // indirect
 	github.com/google/pprof v0.0.0-20210407192527-94a9f03dee38 // indirect
 	github.com/onsi/ginkgo/v2 v2.9.5 // indirect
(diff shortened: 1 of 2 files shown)


marten-seemann:
Why doesn't this error occur on CI?

matthewmueller:
> Why doesn't this error occur on CI?

I think because CI isn't running with `-tags=tools`

marten-seemann:
What's the command you're running with `-tags=tools`? Can we run this command on CI as well, to make sure this bug isn't reintroduced at some later point?

marten-seemann:
Looks like the linting workflow is unhappy, something doesn't seem to work with `go mod vendor`.

matthewmueller:
> What's the command you're running with -tags=tools? Can we run this command on CI as well, to make sure this bug isn't reintroduced at some later point?

So you could repro this with:

```go
go test -tags=tools .
tools.go:6:2: import "github.com/onsi/ginkgo/v2/ginkgo" is a program, not an importable package
```

But no one really does this. I just realized that in my VSCode settings I have: 

```
"go.buildTags": "tools"
```

Which instructs gopls to also type-check `tools.go`

This was to silence another error:

```
No packages found for open file $GOPATH/quic-go/tools.go.
This file may be excluded due to its build tags; try adding "-tags=<build tag>" to your gopls "buildFlags" configuration
```

> Looks like the linting workflow is unhappy, something doesn't seem to work with go mod vendor.

Just fixed it.

marten-seemann:
> But no one really does this, neither does CI. I just realized that in my VSCode settings I have:
> 
> ```
> "go.buildTags": "tools"
> ```

Why though? This seems like a highly unusual configuration, so maybe we don't need to do anything here?

PR #4911, opened 2025-01-22
proxy: simplify packet sorting logic using the slices package

marten-seemann:

### DIFF
--- a/integrationtests/tools/proxy/proxy.go
+++ b/integrationtests/tools/proxy/proxy.go
@@ -2,7 +2,7 @@ package quicproxy
 
 import (
 	"net"
-	"sort"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


integrationtests/tools/proxy/proxy.go
```diff
@@ -65,7 +59,15 @@ func (q *queue) Add(e packetEntry) {
...
-			sort.Stable(q.Packets)
+			slices.SortStableFunc(q.Packets, func(a, b packetEntry) int {
+				if a.Time.Before(b.Time) {
+					return -1
+				}
+				if a.Time.After(b.Time) {
+					return 1
+				}
+				return 0
+			})
```
marten-seemann, line 70: Actually, we don't need to sort at all. This slice is always sorted. We just need to insert the new element at the correct index.

Issue #4894, opened 2025-01-20
FIPS Compliance Issues with Go 1.24


juliens:
In the upcoming Go version 1.24, there are several improvements related to FIPS 140-3 compliance https://github.com/golang/go/issues/69536. This issue is to track the problems we are encountering when using Go 1.24 with the `fips-only` mode (`GODEBUG="fips140=only"`), and to provide an update on the progress.

We are facing an issue related to the use of fixed nonce. 
According to section 5.8 of RFC 9001, quic requires a fixed nonce for the handshake retry, but this conflicts with FIPS, which prohibits the use of fixed nonces. 
This issue occurs during the initialization of the handshake retry, which prevents the import of the quic library in FIPS-only mode.

Moreover, this problem is not limited to the handshake retry; there are other places in the code where this issue arises. When running tests with `fips-only` mode enabled, multiple panic errors are raised with the following message:
```
panic: crypto/cipher: use of GCM with arbitrary IVs is not allowed in FIPS 140-only mode, use NewGCMWithRandomNonce
```

marten-seemann:
As mentioned in https://github.com/quic-go/quic-go/pull/4892#issuecomment-2601856164, we could move to a lazy initialization of the retry cipher, but arguably, this would just be a workaround: initializing state is exactly what `init` was designed for.

I'd like to understand how FIPS mode is supposed to work with QUIC at all, since using a fixed nonce is required by the RFC (and doesn't pose a security risk in this case). Maybe Go's verification logic is too strict?

marten-seemann:
We should probably block the v0.49 on the resolution of this issue, particularly on https://github.com/golang/go/issues/69536#issuecomment-2606165418. If we actually need to implement the workaround mentioned above, it would be nice to ship that as part of v0.49.0.

marten-seemann:
My current interpretation of the issue is that the `panic` in `cipher.NewGCM` is too restrictive, and that the proper fix would be to remove it (relax it?) in the Go standard library. @juliens Do you agree?

juliens:
I think your interpretation is likely correct. While it may not be entirely confirmed, it aligns with the current observations and seems like a reasonable assumption. To ensure full FIPS compliance for Quic-Go, though, we may need to address other aspects beyond just the retry case (maybe all links to the `cipher.NewGCM`).

[...]

PR #4916, opened 2025-01-23
handshake: lazily create the AEAD used for Retry

juliens:
This PR changes the init for retry to a sync.Once, like this, ciphers are only init if you use it.

Linked to #4894


internal/handshake/retry.go
```diff
@@ -11,11 +11,12 @@ import (
...
 var (
-	retryAEADv1 cipher.AEAD // used for QUIC v1 (RFC 9000)
-	retryAEADv2 cipher.AEAD // used for QUIC v2 (RFC 9369)
+	initRetryOnce sync.Once
+	retryAEADv1   cipher.AEAD // used for QUIC v1 (RFC 9000)
+	retryAEADv2   cipher.AEAD // used for QUIC v2 (RFC 9369)
 )
 
-func init() {
+func initRetry() {
```
marten-seemann, line 19: Please  add a comment here, explaining why this isn't done in `init`. It's probably also a good idea to link to your issue.

internal/handshake/retry.go
```diff
@@ -41,6 +45,8 @@ var (
 
 // GetRetryIntegrityTag calculates the integrity tag on a Retry packet
 func GetRetryIntegrityTag(retry []byte, origDestConnID protocol.ConnectionID, version protocol.Version) *[16]byte {
+	initRetryOnce.Do(initRetry)
+
 	retryMutex.Lock()
```
tmthrgd, line 50: Just a drive-by comment, but you already have a global `sync.Mutex` here so you don't actually need a seperate `sync.Once`.

juliens: you mean something like this:
```golang
// GetRetryIntegrityTag calculates the integrity tag on a Retry packet
func GetRetryIntegrityTag(retry []byte, origDestConnID protocol.ConnectionID, version protocol.Version) *[16]byte {
	retryMutex.Lock()
	defer retryMutex.Unlock()

	retryBuf.WriteByte(uint8(origDestConnID.Len()))
	retryBuf.Write(origDestConnID.Bytes())
	retryBuf.Write(retry)
	defer retryBuf.Reset()

	var tag [16]byte
	var sealed []byte
	if version == protocol.Version2 {
		// Instead of using a traditional init function, we use the mutex and "init" in the code to ensure the
		// cipher creation is only performed when the retry feature is actually used.
		// For more details, see: https://github.com/quic-go/quic-go/issues/4894
		if retryAEADv2 == nil {
			retryAEADv2 = initAEAD([16]byte{0x8f, 0xb4, 0xb0, 0x1b, 0x56, 0xac, 0x48, 0xe2, 0x60, 0xfb, 0xcb, 0xce, 0xad, 0x7c, 0xcc, 0x92})
		}

		sealed = retryAEADv2.Seal(tag[:0], retryNonceV2[:], nil, retryBuf.Bytes())
	} else {
		// Instead of using a traditional init function, we use the mutex and "init" in the code to ensure the
		// cipher creation is only performed when the retry feature is actually used.
		// For more details, see: https://github.com/quic-go/quic-go/issues/4894
		if retryAEADv1 == nil {
			retryAEADv1 = initAEAD([16]byte{0xbe, 0x0c, 0x69, 0x0b, 0x9f, 0x66, 0x57, 0x5a, 0x1d, 0x76, 0x6b, 0x54, 0xe3, 0x68, 0xc8, 0x4e})
		}

		sealed = retryAEADv1.Seal(tag[:0], retryNonceV1[:], nil, retryBuf.Bytes())
	}
	if len(sealed) != 16 {
		panic(fmt.Sprintf("unexpected Retry integrity tag length: %d", len(sealed)))
	}
	return &tag
}

```

marten-seemann: Thank you for the drive-by review @tmthrgd!

The code looks good to me, and maybe we don't even need the comment anymore that I asked for earlier? Or maybe move it above the variable declaration?

marten-seemann: @juliens Can you commit these changes? Would like to get the release out asap.

marten-seemann: Just pushed the changes. Let's get that release out! :)

PR #4918, opened 2025-01-23
ackhandler: remove error return value from OnLossDetectionTimeout

marten-seemann:
This removes two bug checks, one of which was completely useless, but the other one could potentially have been useful. Not sure how I feel about that.

Depends on #4917.


marten-seemann:
Stupid GitHub automatically closed this PR. How hard can it be to fix this bug? Seriously 🤬

PR #4922, opened 2025-01-25
proxy: add function to simulate NAT rebinding

marten-seemann:
Needed for #234.


1121437613:
Hi, I'd like to know how to use this feature? I'm encountering an issue that might be caused by NAT rebinding.I can't draw a conclusion because I can't simulate a NAT timeout to produce NAT rebinding.Of course, this issue is unrelated to quic-go; I'm just checking if the new features added can help me.

marten-seemann:
@1121437613 First of all, you shouldn't. The proxy is not intended for external use. Its **only** use case is for testing of quic-go.

Usage is super simple, just call `SwitchConn`.

Issue #2482, opened 2020-04-10
implement the ack-frequency extensions


marten-seemann:
Wire format changes of this extensions are now tied to the QUIC drafts, so we can safely implement and deploy this extension.

https://janaiyengar.github.io/ack-frequency/draft-iyengar-quic-delayed-ack.html

vikulin:
Any updates on this?

vikulin:
Adding https://datatracker.ietf.org/doc/html/draft-iyengar-quic-delayed-ack-00

marten-seemann:
That's not the correct link: https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency

We won't implement this before until this document has made significant progress at the IETF (possibly WGLC?). Even then, we'd need a test suite able to demonstrate the benefit of this extension.

vikulin:
> That's not the correct link: https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency
> 
> We won't implement this before until this document has made significant progress at the IETF (possibly WGLC?). Even then, we'd need a test suite able to demonstrate the benefit of this extension.

yes, you are right. Can you please point out what ack frequency is currently applied for quic-go? I found this line but I am not sure it is correct:
https://github.com/lucas-clemente/quic-go/blob/master/internal/ackhandler/sent_packet_handler.go#L732

marten-seemann:
quic-go implements the acknowledgement strategy described in RFC 9002.

vikulin:
> > That's not the correct link: https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency
> > We won't implement this before until this document has made significant progress at the IETF (possibly WGLC?). Even then, we'd need a test suite able to demonstrate the benefit of this extension.
> 
> yes, you are right. Can you please point out what ack frequency is currently applied for quic-go? I found this line but I am not sure it is correct: https://github.com/lucas-clemente/quic-go/blob/master/internal/ackhandler/sent_packet_handler.go#L732

I found following: 
```
A receiver SHOULD send an ACK frame after receiving at least two ack-eliciting packets
```
@marten-seemann could you please point out where the rule is implemented in quic-go code? This is only what I am asking for.

marten-seemann:
https://github.com/lucas-clemente/quic-go/blob/master/internal/ackhandler/received_packet_tracker.go

larryliu2018:
Hi Marten Seemann
[...]

PR #4928, opened 2025-01-26
use a ringbuffer to store received packets in the connection

marten-seemann:
Reduces memory usage by replacing a fixed-capacity channel (256) with a ring buffer (initial capacity: 8). The buffer grows to 256 only when many packets are enqueued.

In the future, we will use this for a very cheap test if there are packets received that need processing (which we prioritize over sending out new packets).

The amount of data allocated during the handshake is reduced by roughly 1/4:
```
name          old time/op    new time/op    delta
Handshake-16     630µs ± 1%     631µs ± 5%     ~     (p=0.829 n=8+10)

name          old alloc/op   new alloc/op   delta
Handshake-16     264kB ± 0%     200kB ± 0%  -24.10%  (p=0.000 n=9+9)

name          old allocs/op  new allocs/op  delta
Handshake-16     2.54k ± 0%     2.54k ± 0%     ~     (p=0.837 n=10+10)

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -19,6 +19,7 @@ import (
 	"github.com/quic-go/quic-go/internal/protocol"
 	"github.com/quic-go/quic-go/internal/qerr"
 	"github.com/quic-go/quic-go/internal/utils"
+	"github.com/quic-go/quic-go/internal/utils/ringbuffer"
 	"github.com/quic-go/quic-go/internal/wire"
 	"github.com/quic-go/quic-go/logging"
 )
@@ -154,8 +155,10 @@ type connection struct {
 	oneRTTStream        *cryptoStream // only set for the server
 	cryptoStreamHandler cryptoStreamHandler
 
-	receivedPackets  chan receivedPacket
-	sendingScheduled chan struct{}
+	notifyReceivedPacket chan struct{}
+	sendingScheduled     chan struct{}
+	receivedPacketMx     sync.Mutex
+	receivedPackets      ringbuffer.RingBuffer[receivedPacket]
 
 	// closeChan is used to notify the run loop that it should terminate
 	closeChan chan struct{}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: connection.go [hunk 1]
```diff
@@ -784,32 +784,45 @@ func (s *connection) handleHandshakeConfirmed(now time.Time) error {
...
-receiveLoop:
+	s.receivedPacketMx.Lock()
+	numPackets := s.receivedPackets.Len()
+	if numPackets == 0 {
+		s.receivedPacketMx.Unlock()
+		return false, nil
+	}
+
+	var hasMorePackets bool
 	for i := 0; i < numPackets; i++ {
```
sukunrt, line 799: We can keep reading here. Why stop at `numPackets?` The loop exits depending on `hasMorePackets`

connection.go
```diff
@@ -478,7 +481,8 @@ func (s *connection) preSetup() {
 		s.perspective,
 	)
 	s.framer = newFramer(s.connFlowController)
-	s.receivedPackets = make(chan receivedPacket, protocol.MaxConnUnprocessedPackets)
+	s.receivedPackets.Init(8)
```
sukunrt, line 484: Should we initialize it with a bigger capacity? 
The struct is about 100 bytes. We can have 100 in 10 kB.

connection.go
```diff
@@ -2030,7 +2054,10 @@ func (s *connection) sendPacketsWithGSO(now time.Time) error {
...
 
 		// Prioritize receiving of packets over sending out more packets.
-		if len(s.receivedPackets) > 0 {
+		s.receivedPacketMx.Lock()
+		hasPackets := !s.receivedPackets.Empty()
+		s.receivedPacketMx.Unlock()
+		if hasPackets {
 			s.pacingDeadline = deadlineSendImmediately
 			return nil
 		}
```
sukunrt, line 2063: This isn't introduced in this change. But the comment reads the opposite of what we're doing:

 If we have packets to read, send packets immediately.

sukunrt (approved):

connection.go
marten-seemann: We need to send out an ACK every now and then. Currently, that's every 256 packets. We should probably introduce a smaller constant for that, or even some dynamic logic (maybe when we add support for ACK_FREQUENCY: https://github.com/quic-go/quic-go/issues/2482).

connection.go
marten-seemann: Growing it is cheap, it's just a slice allocation. I remember seeing numbers from one of the mobile OS vendors that the median QUIC connection sends ~10 packets. In total.

connection.go
marten-seemann: Will be done in #4927.

PR #4931, opened 2025-01-26
congestion: limit pacing events to 64 per RTT to reduce timer frequency

marten-seemann:
Currently, the minimum pacing delay is 1ms. If the bandwidth allows sending more frequently, we instead release a batch of packets into the network every 1ms.

Additionally, we now also limit the number of pacing events to 64 per RTT, i.e. apply a minimum pacing delay of 1/64 RTT. This allows us to arm the pacing timer less frequently, while still achieving most of the benefits of packet pacing.


marten-seemann:
Something is wrong with this PR. It completely kills transfer performance.

PR #4940, opened 2025-01-27
ackhandler: implement timer logic for path probe packets

marten-seemann:
Part of #234.

Path probe packets are treated differently from regular packets: The new path might have a vastly different RTT than the original path.

Path probe packets are declared lost 1s after they are sent. This value can be reduced, once implement proper retransmission logic for lost path probes.


marten-seemann:
The loss detection timeout currently doesn't declare probe packets lost.

marten-seemann:
> The loss detection timeout currently doesn't declare probe packets lost.

Fixed in 5cc8435.

PR #4955, opened 2025-02-14
use clickable links in documentation, fix list styling

mark-rushakoff:
1. Use doc links (the square brackets around identifiers) to make clickable links on pkg.go.dev (https://go.dev/doc/comment#doclinks)
2. Indent lists so that they are rendered properly -- at which point go fmt updates the bullet character to a dash (https://go.dev/doc/comment#lists)
3. Add a couple missing punctuation marks

These all will improve the experience of browsing the quic-go docs on pkg.go.dev.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -44,19 +44,22 @@ type TokenStore interface {
 }
 
 // Err0RTTRejected is the returned from:
-// * Open{Uni}Stream{Sync}
-// * Accept{Uni}Stream
-// * Stream.Read and Stream.Write
+//   - Open{Uni}Stream{Sync}
+//   - Accept{Uni}Stream
+//   - Stream.Read and Stream.Write
+//
 // when the server rejects a 0-RTT connection attempt.
 var Err0RTTRejected = errors.New("0-RTT rejected")
 
-// ConnectionTracingKey can be used to associate a ConnectionTracer with a Connection.
+// ConnectionTracingKey can be used to associate a [logging.ConnectionTracer] with a [Connection].
 // It is set on the Connection.Context() context,
 // as well as on the context passed to logging.Tracer.NewConnectionTracer.
+//
 // Deprecated: Applications can set their own tracing key using Transport.ConnContext.
 var ConnectionTracingKey = connTracingCtxKey{}
 
 // ConnectionTracingID is the type of the context value saved under the ConnectionTracingKey.
+//
 // Deprecated: Applications can set their own tracing key using Transport.ConnContext.
 type ConnectionTracingID uint64
 
@@ -66,9 +69,9 @@ type connTracingCtxKey struct{}
 // context returned by tls.Config.ClientInfo.Context.
 var QUICVersionContextKey = handshake.QUICVersionContextKey
 
-// Stream is the interface implemented by QUIC streams
-// In addition to the errors listed on the Connection,
-// calls to stream functions can return a StreamError if the stream is canceled.
+// Stream is the interface implemented by QUIC streams.
+// In addition to the errors listed on the [Connection],
+// calls to stream functions can return a [StreamError] if the stream is canceled.
 type Stream interface {
 	ReceiveStream
 	SendStream
@@ -141,12 +144,12 @@ type SendStream interface {
 
 // A Connection is a QUIC connection between two peers.
 // Calls to the connection (and to streams) can return the following types of errors:
-// * ApplicationError: for errors triggered by the application running on top of QUIC
-// * TransportError: for errors triggered by the QUIC transport (in many cases a misbehaving peer)
-// * IdleTimeoutError: when the peer goes away unexpectedly (this is a net.Error timeout error)
-// * HandshakeTimeoutError: when the cryptographic handshake takes too long (this is a net.Error timeout error)
-// * StatelessResetError: when we receive a stateless reset
-// * VersionNegotiationError: returned by the client, when there's no version overlap between the peers
+//   - [ApplicationError]: for errors triggered by the application running on top of QUIC
+//   - [TransportError]: for errors triggered by the QUIC transport (in many cases a misbehaving peer)
+//   - [IdleTimeoutError]: when the peer goes away unexpectedly (this is a [net.Error] timeout error)
+//   - [HandshakeTimeoutError]: when the cryptographic handshake takes too long (this is a net.Error timeout error)
+//   - [StatelessResetError]: when we receive a stateless reset
+//   - [VersionNegotiationError]: returned by the client, when there's no version overlap between the peers
 type Connection interface {
 	// AcceptStream returns the next stream opened by the peer, blocking until one is available.
 	// If the connection was closed due to a timeout, the error satisfies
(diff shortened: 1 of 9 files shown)


interface.go
```diff
@@ -44,20 +44,23 @@ type TokenStore interface {
...
 // when the server rejects a 0-RTT connection attempt.
 var Err0RTTRejected = errors.New("0-RTT rejected")
 
-// ConnectionTracingKey can be used to associate a ConnectionTracer with a Connection.
+// Deprecated: Applications can set their own tracing key using Transport.ConnContext.
+//
+// ConnectionTracingKey can be used to associate a [*logging.ConnectionTracer] with a [Connection].
 // It is set on the Connection.Context() context,
 // as well as on the context passed to logging.Tracer.NewConnectionTracer.
-// Deprecated: Applications can set their own tracing key using Transport.ConnContext.
```
marten-seemann, line 56: Why reorder the deprecation warning?

interface.go
```diff
@@ -44,20 +44,23 @@ type TokenStore interface {
 }
 
 // Err0RTTRejected is the returned from:
-// * Open{Uni}Stream{Sync}
-// * Accept{Uni}Stream
-// * Stream.Read and Stream.Write
+//   - Open{Uni}Stream{Sync}
+//   - Accept{Uni}Stream
+//   - Stream.Read and Stream.Write
```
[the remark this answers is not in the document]
mark-rushakoff: It does, but the list has to be indented first.

> A list is a span of indented or blank lines (which would otherwise be a code block, as described in the next section) in which the first indented line begins with a bullet list marker or a numbered list marker.

> Gofmt reformats bullet lists to use a dash as the bullet marker, two spaces of indentation before the dash, and four spaces of indentation for continuation lines.

The quic-go lists before this change are therefore rendered like this:

Screenshot 2025-02-15 at 9 25 21 AM

Instead of [e.g. from the standard library](https://pkg.go.dev/crypto/rand@go1.24.0#pkg-variables)

Screenshot 2025-02-15 at 9 26 47 AM

interface.go
mark-rushakoff: It was late and I accidentally left that detail out of the commit message.

https://go.dev/doc/comment#deprecations

> Paragraphs starting with Deprecated: are treated as deprecation notices. Some tools will warn when deprecated identifiers are used. [pkg.go.dev](https://pkg.go.dev/) will hide their docs by default.

quic-go currently renders like this:

Screenshot 2025-02-15 at 9 29 12 AM

----

Proper deprecations render like this:

Table of contents:
Screenshot 2025-02-15 at 9 29 49 AM

Hidden by default:
Screenshot 2025-02-15 at 9 30 11 AM

When shown:
Screenshot 2025-02-15 at 9 31 08 AM

mark-rushakoff: On re-read I could have kept it at the bottom but put a newline before it, so that it started the paragraph. I looked at some other deprecation somewhere that put it as the first line of the whole godoc block.

I'll update the PR to keep the deprecation at the end.

interface.go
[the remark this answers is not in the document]
marten-seemann: This is much better!

Should we link these functions as well?

interface.go
```diff
@@ -44,19 +44,22 @@ type TokenStore interface {
...
+//
 // when the server rejects a 0-RTT connection attempt.
 var Err0RTTRejected = errors.New("0-RTT rejected")
 
-// ConnectionTracingKey can be used to associate a ConnectionTracer with a Connection.
+// ConnectionTracingKey can be used to associate a [*logging.ConnectionTracer] with a [Connection].
```
marten-seemann, line 54:
```suggestion
// ConnectionTracingKey can be used to associate a [logging.ConnectionTracer] with a [Connection].
```

server.go
```diff
@@ -146,11 +146,11 @@ func (l *Listener) Accept(ctx context.Context) (Connection, error) {
...
 
 // Close closes the listener.
-// Accept will return ErrServerClosed as soon as all connections in the accept queue have been accepted.
+// Accept will return [ErrServerClosed] as soon as all connections in the accept queue have been accepted.
 // QUIC handshakes that are still in flight will be rejected with a CONNECTION_REFUSED error.
 // The effect of closing the listener depends on how it was created:
-// * if it was created using Transport.Listen, already established connections will be unaffected
-// * if it was created using the Listen convenience method, all established connection will be closed immediately
+//   - if it was created using Transport.Listen, already established connections will be unaffected
+//   - if it was created using the Listen convenience method, all established connection will be closed immediately
```
marten-seemann, line 153: Link the functions?

server.go
```diff
@@ -230,7 +230,7 @@ func listenUDP(addr string) (*net.UDPConn, error) {
 // Furthermore, it must define an application control (using NextProtos).
 // The quic.Config may be nil, in that case the default values will be used.
```
marten-seemann, line 231: Link the `Config`?

transport.go
```diff
@@ -171,7 +171,7 @@ func (t *Transport) Listen(tlsConf *tls.Config, conf *Config) (*Listener, error)
 
 // ListenEarly starts listening for incoming QUIC connections.
 // There can only be a single listener on any net.PacketConn.
-// Listen may only be called again after the current Listener was closed.
+// ListenEarly may only be called again after the current [Listener] was closed.
```
marten-seemann, line 174: This applies to both `Listener` and `EarlyListener`.
```suggestion
// ListenEarly may only be called again after the current listener was closed.
```

transport.go
```diff
@@ -160,7 +160,7 @@ type Transport struct {
 
 // Listen starts listening for incoming QUIC connections.
 // There can only be a single listener on any net.PacketConn.
-// Listen may only be called again after the current Listener was closed.
+// Listen may only be called again after the current [Listener] was closed.
```
marten-seemann, line 163: This applies to both `Listener` and `EarlyListener`.
```suggestion
// Listen may only be called again after the current listener was closed.
```

sys_conn.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 // OOBCapablePacketConn is a connection that allows the reading of ECN bits from the IP header.
-// If the PacketConn passed to Dial or Listen satisfies this interface, quic-go will use it.
+// If the PacketConn passed to [Dial] or [Listen] satisfies this interface, quic-go will use it.
```
marten-seemann, line 17: Unrelated to this PR, we should have changed this a long time ago.
```suggestion
// If the PacketConn passed to the [Transport] satisfies this interface, quic-go will use it.
```

interface.go
```diff
@@ -234,7 +237,7 @@ type TokenGeneratorKey = handshake.TokenProtectorKey
 // as they are allowed by RFC 8999.
 type ConnectionID = protocol.ConnectionID
 
-// ConnectionIDFromBytes interprets b as a Connection ID. It panics if b is
+// ConnectionIDFromBytes interprets b as a ConnectionID. It panics if b is
```
marten-seemann, line 240: Link?

marten-seemann:
@mark-rushakoff Could you address the review comments? Would love to get this PR in!

mark-rushakoff:
> @mark-rushakoff Could you address the review comments? Would love to get this PR in!

Sure thing, I've set some time aside to finish this on Monday (tomorrow).

interface.go
mark-rushakoff: I usually don't link the types when they are part of the function or method signature, since it's right there. But that's just my personal preference -- the standard library often does link the type.

marten-seemann:
Thank you @mark-rushakoff! Took me a while to get this in, but this will now be included in the v0.51 release.

Issue #4325, opened 2024-02-10
GSO support on Windows


marten-seemann:
Apparently it's possible to use GSO on Windows. For reference (and code points), quinn just [added support](https://github.com/quinn-rs/quinn/commit/33f6d89cf47fbd13083a465d6b044ada1b6099d2).

I don't particularly care about Windows too much, so I probably won't be working on this issue any time soon. If you're interested, please feel free to reach out or the send a PR.

Samsonnyyeet:
Hey @marten-seemann 

I'm interested in helping add GSO support for Windows. From my shallow inspection of the source code, and the linked quinn pr, I noticed that quinn sets the UDP_SEND_MSG_SIZE option. This flag is largely responsible for handling GSO support. 

Reference:
[Windows USO documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-segmentation-offload-uso-)

Implementing GSO support for Windows would require handling proper instantiation of this flag, and the creation of `sys_conn_helper_windows.go` file based on the structure of the linux implementation. 

(I'd appreciate feedback/advice on my approach!)

marten-seemann:
Your approach sounds reasonable. Are you developing on Windows? Will you be able to test your code, to make sure that it _actually_ sends out multiple QUIC packets per syscall?

Samsonnyyeet:
Hey! Really sorry for the delay. As it happen, my network card doesn't actually support GSO. I'm currently in the process of temporarily acquiring a laptop which supports GSO for the development of this feature.

> Are you developing on Windows?
Yep! I found your PRs which implemented GSO for linux and those were quite helpful to understand the general process of going about things.

marten-seemann:
Thanks for the update! Looking forward to your PR!

Samsonnyyeet:
Hey! FINALLY got my hands on a laptop with offloading support 😎.

I had a quick question. Please correct me if I'm wrong, but the windows specific implementation is using only basicConn. The linux implementation is relying on the `oobConn` struct built on top of `OOBCapablePacketConn` to implement GSO, and other functionalities. How should I structure the GSO support for windows? I assume making changes to the basicConn (to add a GSO support boolean) is not a good idea. Should I make another struct just for this on top of `OOBCapablePacketConn` similar to `oobConn` but specifically for windows?

marten-seemann:
> Hey! FINALLY got my hands on a laptop with offloading support 😎.

That's great!

[...]

PR #4962, opened 2025-02-17
Switching Windows to use `oobConn`

Samsonnyyeet:
### Which issue does this address?
#4325

### Purpose
This draft PR showcases the current progress on implementing `oobConn` support for Windows. Since this is my first time working on something like this, I recognize that the implementation is still a work in progress. There are several open issues and uncertainties, which I have marked with `TO DO:` comments.

Most of the code is largely based on the existing Linux implementation, with minimal changes to maintain clarity for existing contributors. I would sincerely appreciate any feedback or guidance on my approach, as well as suggestions for improvements.

### Major Changes and Issues
- Implemented cmsg support from scratch, as I couldn’t find existing support in the Go libraries.
- Using a workaround for ReadBatch, as it is unsupported on Windows.
- ECN support is still in progress; I am encountering several issues getting it to function correctly.
- Unable to find a Windows equivalent for Linux’s forced receive and send buffer settings.
- No `isGSOError` function implemented yet.
- Current tests need significant improvement, both in structure and coverage.

Any feedback or guidance would be greatly appreciated!

### DIFF
--- a/buffer_pool.go
+++ b/buffer_pool.go
@@ -12,6 +12,8 @@ type packetBuffer struct {
 	// refCount counts how many packets Data is used in.
 	// It doesn't support concurrent use.
 	// It is > 1 when used for coalesced packet.
+	// More than one packet's data can be in the same buffer?
+	// So more than one packet will reference this buffer?
 	refCount int
 }
 
--- a/buffer_pool.go
+++ b/buffer_pool.go
@@ -12,8 +12,6 @@ type packetBuffer struct {
 	// refCount counts how many packets Data is used in.
 	// It doesn't support concurrent use.
 	// It is > 1 when used for coalesced packet.
-	// More than one packet's data can be in the same buffer?
-	// So more than one packet will reference this buffer?
 	refCount int
 }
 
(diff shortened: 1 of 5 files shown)


Samsonnyyeet:
So I have fixed a ton of the issues with the first iteration. Still need to work on tests.

I have simplified the `oobConn` structure and entirely removed batched reads. ECN support has been fixed as well (correct control messages are now being generated). There was an oversight with the Control Message struct, which has been rectified. Also fixed a lot of the wrong socket options. 

Current issues/doubts:
- Windows expects c_int as the data type for the body of the ecn control messages. Do I continue using int32 or do I import and use cgo.
- WritePacket requires uint16 for gsoSize, which I am currently just converting to uint32 as expected by windows.
- Which options do I use for forceSetReceiveBuffer and forceSetSendBuffer?
- Still need to implement isECNEnabled (totally overlooked this 😭, will try to finish this asap)
- What tests should I focus on?
- Any other comments/advice please!

Samsonnyyeet:
Hey @marten-seemann !
Would love to get your opinion on the current structure of things.

PR #4966, opened 2025-02-23
add a Config flag to disable connection migration

marten-seemann:
For #234.


marten-seemann:
This is only needed in very special routing setups, namely if:
1. the server uses a multicast address, AND
2. routes packets based on the 4-tuple (and not based on QUIC connection IDs)

While this is a valid thing to do, this is also a super fringe use case, and it's really hard to avoid the impression that it's somehow possible to prevent the client from performing connection migration... which is totally not what this option does.

PR #4993, opened 2025-03-16
http3: reject duplicate pseudo headers

### LINKED ISSUE #4992: http3: duplicate pseudo-headers not detected
quic-go/http3 does not reject requests with duplicate pseudo-headers.

e.g.: a request with both `:method: GET` and `:method POST` is not rejected. the last occurring value is treated as the actual request-method.

a request like this should be treated as malformed, since this could cause trouble in reverse-proxies. however, in rfc 9114 this is not stated explicitly, see section 4.3.1: 

> All HTTP/3 requests MUST include exactly one value for the :method, :scheme, and :path pseudo-header fields, unless the request is a CONNECT request.

functionality should probably be implemented here:
https://github.com/quic-go/quic-go/blob/9ca74c70493a7268a8866495ab4dd8ec849f9111/http3/headers.go#L53-L75

pittgi:
fixes #4992


marten-seemann (approved):
Very nice. This looks clean!

PR #5008, opened 2025-03-26
fix size check when enqueuing egress datagram

### LINKED ISSUE #4983: Large egress datagrams are silently being dropped
**Summary**

Egress datagrams pass through a queue, with different size checks before entering the queue and after leaving the queue. This has the unfortunate effect that a given datagram can first be considered as having a valid size, and the second size check arriving at a different conclusion.

**Issue Description**

Here's the size check before entering the queue:
https://github.com/quic-go/quic-go/blob/24acc54ef10777e086574760c54ec27fefb32a4e/connection.go#L2588

`maxDataLen` is `1452` in my case.

When retrieving datagrams from the queue, a different size check is being applied:
https://github.com/quic-go/quic-go/blob/24acc54ef10777e086574760c54ec27fefb32a4e/packet_packer.go#L657

`maxFrameSize` is `1429` in my case.

Seemingly, all datagrams with sizes in the range 1430-1452 will cause execution to enter into the `else if`, where they are silently being dropped.

**Additional considerations**

Setting `maxDataLen` equal to `maxFrameSize` is only a partial solution because of additional bytes required in the packet:
https://github.com/quic-go/quic-go/blob/24acc54ef10777e086574760c54ec27fefb32a4e/packet_packer.go#L656
https://github.com/quic-go/quic-go/blob/24acc54ef10777e086574760c54ec27fefb32a4e/internal/wire/datagram_frame.go#L79

mvd-ows:
This fixes #4983 .

Because the conservative size check is now being correctly applied, we lose a few bytes on the max datagram that can be sent through. On the positive side, behavior is now more consistent and can be reasoned about.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -294,7 +294,7 @@ var newConnection = func(
 		s.tracer,
 		s.logger,
 	)
-	s.currentMTUEstimate.Store(uint32(estimateMaxPayloadSize(protocol.ByteCount(s.config.InitialPacketSize))))
+	s.currentMTUEstimate.Store(uint32(s.config.InitialPacketSize))
 	statelessResetToken := statelessResetter.GetStatelessResetToken(srcConnID)
 	params := &wire.TransportParameters{
 		InitialMaxStreamDataBidiLocal:   protocol.ByteCount(s.config.InitialStreamReceiveWindow),
@@ -410,7 +410,7 @@ var newClientConnection = func(
 		s.tracer,
 		s.logger,
 	)
-	s.currentMTUEstimate.Store(uint32(estimateMaxPayloadSize(protocol.ByteCount(s.config.InitialPacketSize))))
+	s.currentMTUEstimate.Store(uint32(s.config.InitialPacketSize))
 	oneRTTStream := newCryptoStream()
 	params := &wire.TransportParameters{
 		InitialMaxStreamDataBidiRemote: protocol.ByteCount(s.config.InitialStreamReceiveWindow),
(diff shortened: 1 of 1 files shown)


marten-seemann:
I don't understand what's going on here. Can you please explain?

mvd-ows:
Certainly -

`currentMTUEstimate` was being used incorrectly. The initial assignment tried to account for overhead and all subsequent assignments did not. What's more, the value of `currentMTUEstimate` is updated right away because the "MTU discoverer", even when PMTUD is disabled, will echo back `InitialPacketSize` that is was initialized with.

So effectively, `estimateMaxPayloadSize()` was not really being used, or only for a very brief period of time.

After correcting the initialization of `currentMTUEstimate`, the smallest possible change inside `SendDatagram()` would be to just ensure that `estimateMaxPayloadSize()` is being applied. But in the general case `currentMTUEstimate` will be smaller than `s.peerParams.MaxDatagramFrameSize`, so `maxDataLen` as it was previously being computed would not account for additional bytes required in the frame header. The value reported in `DatagramTooLargeError` would then also be slightly off.

Because there are two size limits to consider, I chose to make their computations more explicit, hoping that it would be clear enough without additional comments.

marten-seemann:
I'm still confused, both about the issue and the fix. Can you add an integration test that 1. shows the issue before the fix and 2. makes sure we don't break this in the future?

marten-seemann:
@mvd-ows Are you still working on this?

mvd-ows:
Yes, but other things took priority lately. I should be able to look at this again in about 10 days, and will then add the integration test.

PR #5017, opened 2025-03-31
Fix TestDatagramLoss failures due to a too eager context timeout

### LINKED ISSUE #4784: flaky TestDatagramLoss integration test
```
 === RUN   TestDatagramLoss
    datagram_test.go:230: dropped 15 incoming and 8 outgoing out of 208 packets
    datagram_test.go:233: server received 84 out of 100 sent datagrams
    datagram_test.go:235: client received 81 out of 100 sent datagrams
    datagram_test.go:236: 
        	Error Trace:	/Users/runner/work/quic-go/quic-go/integrationtests/self/datagram_test.go:236
        	Error:      	Max difference between 92 and 81 allowed is 5, but difference was 11
        	Test:       	TestDatagramLoss
        	Messages:   	datagrams received by the client
--- FAIL: TestDatagramLoss (0.33s)

### PR DESCRIPTION
**Pull request by Lekensteyn**
While datagram payloads are sent with intervals of 0.5ms (total 50ms), the TLS handshake alone takes 1 RTT (10ms + crypto overhead). On top of that, the send queue is currently limited to 32 packets and can block while waiting for the buffer to clear (presumably by receiving ACKs). Raise the previous timeout of 100ms to avoid early timeouts which could result in detecting too few received packets.

Improve the test to check for the exact number of expected DATAGRAM frames by disabling DPLPMTUD probes. These easily stood out as they are much larger than packets with DATAGRAM frames.

Fixes #4784
___
I noticed these test failures in #5009, hopefully the root cause (too low timeouts) is correctly identified.

### DIFF
--- a/integrationtests/self/datagram_test.go
+++ b/integrationtests/self/datagram_test.go
@@ -131,7 +131,7 @@ func TestDatagramLoss(t *testing.T) {
 	server, err := quic.Listen(
 		newUDPConnLocalhost(t),
 		getTLSConfig(),
-		getQuicConfig(&quic.Config{EnableDatagrams: true}),
+		getQuicConfig(&quic.Config{DisablePathMTUDiscovery: true, EnableDatagrams: true}),
 	)
 	require.NoError(t, err)
 	defer server.Close()
@@ -140,6 +140,7 @@ func TestDatagramLoss(t *testing.T) {
 	proxy := &quicproxy.Proxy{
 		Conn:       newUDPConnLocalhost(t),
 		ServerAddr: server.Addr().(*net.UDPAddr),
+		// Drop about 10% of Short Header packets with DATAGRAM frames
 		DropPacket: func(dir quicproxy.Direction, _, _ net.Addr, packet []byte) bool {
 			if wire.IsLongHeaderPacket(packet[0]) { // don't drop Long Header packets
 				return false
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Lekensteyn** [comment]
One failing test on Windows seems unrelated to this change:
```
--- FAIL: TestSentPacketHandlerPathProbe (0.00s)
    sent_packet_handler_test.go:1182: 
        	Error Trace:	D:/a/quic-go/quic-go/internal/ackhandler/sent_packet_handler_test.go:1182
        	Error:      	Not equal: 
        	            	expected: []protocol.PacketNumber{5, 6}
        	            	actual  : []protocol.PacketNumber{6}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,2 @@
        	            	-([]protocol.PacketNumber) (len=2) {
        	            	- (protocol.PacketNumber) 5,
        	            	+([]protocol.PacketNumber) (len=1) {
        	            	  (protocol.PacketNumber) 6
        	Test:       	TestSentPacketHandlerPathProbe
FAIL
```

To investigate this issue, I visualized the packet capture in Wireshark (`udp.stream==0`) using the TLS key log writer:
```diff
diff --git a/integrationtests/self/self_test.go b/integrationtests/self/self_test.go
index 40479c67..7201666c 100644
--- a/integrationtests/self/self_test.go
+++ b/integrationtests/self/self_test.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"log"
 	"math/rand"
 	"net"
 	"os"
@@ -77,6 +78,13 @@ func init() {
 		}},
 		NextProtos: []string{alpn},
 	}
+	if keyLogPath := os.Getenv("SSLKEYLOGFILE"); keyLogPath != "" {
+		keyLogFile, err := os.OpenFile(keyLogPath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
+		if err != nil {
+			log.Fatalf("Failed to open SSLKEYLOGFILE(%s): %s", keyLogPath, err)
+		}
+		tlsConfig.KeyLogWriter = keyLogFile
+	}
 	tlsConfLongChain, err := tools.GenerateTLSConfigWithLongCertChain(ca, caPrivateKey)
 	if err != nil {
 		panic(err)
```

There is also an `interop.GetSSLKeyLog` which does something similar, except that it truncates existing files instead of appending to them. It is also not clear if that function from `./interop/utils` is appropriate for use in the integrationtests package.

Lekensteyn:
Rebased and fixed the issue reference, this should fix #4784.

marten-seemann (commented):
I don't understand this fix. We're already skipping over packets smaller than `datagramSize`. There shouldn't be any other packets in that size range.

Lekensteyn:
The problem is not too small packets, but too large ones containing PADDING + PING frames. That added two additional frames from top of my head. This is not the primary fix.

Increasing the timeout prevents the variance in the counting packets with DATAGRAM frames. That's the actual fix for flaky tests.

Lekensteyn:
> Should we just disabled DPLPMTUD?

Thanks for the suggestion, that avoids some magic numbers. Applied your suggestion and rebased on current master. Hopefully we can get all tests consistently to green :D

marten-seemann (approved):
Thank you!

PR #5028, opened 2025-04-12
simplify path probe PING frame packing logic

### LINKED ISSUE #4830: probe packet PING packing logic can be simplified
It seems unnecessary to add a PING to the retransmission queue, just to dequeue it from the packet packer when packing the probe packer:
https://github.com/quic-go/quic-go/blob/ea6d198c3d71e2b4b865b886342455fd297debef/connection.go#L2080-L2087

Instead, the packet packer could just generate a fresh PING frame if there's no actual data to be sent. We don't need to involve the retransmission queue at all.

marten-seemann:
Fixes #4830.

This is more than just a refactor: Previously, if the probe packet was lost, we'd retransmit the PING frame. This is clearly not desirable (QUIC's PTO logic takes care of sending out further PTO packets, if necessary).

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -2319,29 +2319,25 @@ func (s *connection) sendProbePacket(sendMode ackhandler.SendMode, now time.Time
 	// Queue probe packets until we actually send out a packet,
 	// or until there are no more packets to queue.
 	var packet *coalescedPacket
-	for {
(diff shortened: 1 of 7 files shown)


marten-seemann:
Still needs test for the Initial / Handshake packing logic.

PR #5037, opened 2025-04-13
fix potential deadlock when OpenStreamSync context is cancelled

### LINKED ISSUE #5035: deadlock when the OpenStreamSync context is canceled
This only happens if a MAX_STREAMS frame is received at exactly the right time. Reproducer:

```go
func TestStreamsMapOutgoingReproduce(t *testing.T) {
	m := newOutgoingStreamsMap(
		protocol.StreamTypeBidi,
		func(n protocol.StreamNum) *mockGenericStream { return &mockGenericStream{num: n} },
		func(f wire.Frame) {},
	)

	ctx, cancel := context.WithCancel(context.Background())
	go func() {
		m.OpenStreamSync(ctx)
	}()

	time.Sleep(scaleDuration(10 * time.Millisecond))
	cancel()

	strChan := make(chan *mockGenericStream, 1)
	go func() {
		str, err := m.OpenStreamSync(context.Background())
		fmt.Println("stream", str, "error", err)
		strChan <- str
	}()

	m.SetMaxStream(1)
	select {
	case <-strChan:
	case <-time.After(time.Second):
		t.Fatal("timeout")
	}
}
```

marten-seemann:
Fixes #5035. ~~Depends on #5036.~~

### DIFF
--- a/streams_map_outgoing.go
+++ b/streams_map_outgoing.go
@@ -89,6 +89,9 @@ func (m *outgoingStreamsMap[T]) OpenStreamSync(ctx context.Context) (T, error) {
 			m.openQueue = slices.DeleteFunc(m.openQueue, func(c chan struct{}) bool {
 				return c == waitChan
 			})
+			// If we just received a MAX_STREAMS frame, this might have been the next stream
+			// that could be opened. Make sure we unblock the next OpenStreamSync call.
+			m.maybeUnblockOpenSync()
 			return *new(T), ctx.Err()
 		case <-waitChan:
 		}
@@ -103,7 +106,7 @@ func (m *outgoingStreamsMap[T]) OpenStreamSync(ctx context.Context) (T, error) {
 		}
 		str := m.openStream()
 		m.openQueue = m.openQueue[1:]
-		m.unblockOpenSync()
+		m.maybeUnblockOpenSync()
 		return str, nil
 	}
 }
@@ -173,7 +176,7 @@ func (m *outgoingStreamsMap[T]) SetMaxStream(num protocol.StreamNum) {
 	if m.maxStream < m.nextStream-1+protocol.StreamNum(len(m.openQueue)) {
 		m.maybeSendBlockedFrame()
 	}
-	m.unblockOpenSync()
+	m.maybeUnblockOpenSync()
 }
 
 // UpdateSendWindow is called when the peer's transport parameters are received.
@@ -188,7 +191,7 @@ func (m *outgoingStreamsMap[T]) UpdateSendWindow(limit protocol.ByteCount) {
 }
 
 // unblockOpenSync unblocks the next OpenStreamSync go-routine to open a new stream
-func (m *outgoingStreamsMap[T]) unblockOpenSync() {
+func (m *outgoingStreamsMap[T]) maybeUnblockOpenSync() {
 	if len(m.openQueue) == 0 {
 		return
 	}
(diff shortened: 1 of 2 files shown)


streams_map_outgoing.go
```diff
@@ -89,6 +89,9 @@ func (m *outgoingStreamsMap[T]) OpenStreamSync(ctx context.Context) (T, error) {
 			m.openQueue = slices.DeleteFunc(m.openQueue, func(c chan struct{}) bool {
 				return c == waitChan
 			})
+			// If we just received a MAX_STREAMS frame, this might have been the next stream
+			// that could be opened. Make sure we unblock the next OpenStreamSync call.
+			m.maybeUnblockOpenSync()
```
sukunrt, line 94:
```suggestion
			// If we just received a MAX_STREAMS frame, this might have been the next stream
			// that could be opened. Make sure we unblock the next OpenStreamSync call.
			// Not checking for a value in `waitChan` here as unblocking two calls is fine. 
			m.maybeUnblockOpenSync()
```

streams_map_outgoing.go
```diff
@@ -188,7 +191,7 @@ func (m *outgoingStreamsMap[T]) UpdateSendWindow(limit protocol.ByteCount) {
 }
 
 // unblockOpenSync unblocks the next OpenStreamSync go-routine to open a new stream
-func (m *outgoingStreamsMap[T]) unblockOpenSync() {
+func (m *outgoingStreamsMap[T]) maybeUnblockOpenSync() {
 	if len(m.openQueue) == 0 {
 		return
 	}
```
sukunrt, line 197: In the method `CloseWithError` 
```
func (m *outgoingStreamsMap[T]) CloseWithError(err error) {
	m.mutex.Lock()
	m.closeErr = err
	for _, str := range m.streams {
		str.closeForShutdown(err)
	}
	for _, c := range m.openQueue {
		if c != nil {
			close(c)
		}
	}
	m.mutex.Unlock()
}
```
we should set openQueue to nil, or at least clear it. Otherwise if there's a context cancellation concurrent with closing, it might cause a panic when  `maybeOpenStream` tries to push a value to the first element in `openQueue`

sukunrt, line 197: alternatively, cleck for `m.closeErr` here.

sukunrt (changes requested):

streams_map_outgoing.go
marten-seemann: That's the fix I chose to go with.

marten-seemann: Setting `openQueue` to `nil` is sufficient.

streams_map_outgoing_test.go
```diff
@@ -534,3 +538,47 @@ var _ = Describe("Streams Map (outgoing)", func() {
...
+		func(n protocol.StreamNum) *mockGenericStream { return &mockGenericStream{num: n} },
+		func(f wire.Frame) {},
+	)
+
+	const num = 100
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	errChan := make(chan error, num)
+	for range num {
```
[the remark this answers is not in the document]
marten-seemann: Oh boy, when is GitHub finally fixing this. @Copilot, this is getting annoying.

marten-seemann:
@sukunrt Thank you for the in-depth review! It was actually quite easy to trigger the panic you found in a unit test, which I am committing here (already in standard Go / require syntax, not Ginkgo). Without your suggested fix, this test reliably panics.

sukunrt (approved):

Issue #4473, opened 2024-05-01
Go 1.24: use math/rand/v2


marten-seemann:
~math/rand/v2 will be introduced in Go 1.23, which means we'll be able to use if once we update to Go 1.24.~

math/rand/v2 was introduced in Go 1.22, which means we'll be able to use if once we update to Go 1.23.

j178:
Hasn't it already been landed in Go 1.22? Here's the reference: https://go.dev/doc/go1.22#math_rand_v2

marten-seemann:
Thank you @j178, looks like I mixed things up. We'll be able to update the code later this year then :)

marten-seemann:
Looks like they removed the `Read` method from the `Rand` struct. This will make it very hard to migrate to math/rand/v2. How are we supposed to deterministically get random bytes?

marten-seemann:
Because cryptographic random number generators are good at generating deterministic random bytes?

sukunrt:
I think the solution is to use the ChaCha8 RNG: https://pkg.go.dev/math/rand/v2#ChaCha8.Read
See the discussion here: https://github.com/golang/go/issues/67059

Unfortunately, `ChaCha8.Read` was added in 1.23

marten-seemann:
It would be nice if this was a general property of every random number implementation in `math/rand/v2`, but I guess we can live with using `ChaCha8` everywhere.
Alternatively, we could also build our own `Read` wrapper and fill a byte slice with repeated calls to `Uint32` 🤡.

> Unfortunately, ChaCha8.Read was added in 1.23

There's no rush. We can just wait until Go 1.24.

PR #5046, opened 2025-04-14
protocol: use math/rand/v2 to generate greased versions

marten-seemann:
Part of #4473.

### DIFF
--- a/internal/protocol/version.go
+++ b/internal/protocol/version.go
@@ -1,13 +1,12 @@
 package protocol
 
 import (
+	"crypto/rand"
 	"encoding/binary"
 	"fmt"
 	"math"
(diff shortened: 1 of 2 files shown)


internal/protocol/version.go
```diff
@@ -105,7 +113,7 @@ func generateReservedVersion() Version {
 func GetGreasedVersions(supported []Version) []Version {
 	versionNegotiationMx.Lock()
 	defer versionNegotiationMx.Unlock()
-	randPos := rand.Intn(len(supported) + 1)
+	randPos := versionNegotiationRand.IntN(len(supported) + 1)
```
marten-seemann, line 116: This now also removes mutex contention around the global `rand` instance.

Issue #4958, opened 2025-02-15
FIPS 140 compatibility


hxtk:
# Problem

Any program using this package will panic immediately if run with `GODEBUG=fips140=only` because of the use of `crypto/cipher.NewGCM` in the package `github.com/quic-go/quic-go/internal/handshake`.

# Background

Golang 1.24 was recently released with the new `GODEBUG` setting, `fips140`. This setting has three modes: `off`, `on`, and `only`. In the third mode, non-FIPS-compliant cryptographic primitives immediately panic. This includes `crypto/cipher.NewGCM` because FIPS 140-3 does not in general allow arbitrary nonces, and with that primitive it is impossible for the FIPS module to attest that they were generated in a way that conforms to the specification. In order to be allowed with `GODEBUG=fips140=only`, a program must use the newly-defined `crypto/cipher.NewGCMWithRandomNonce`, which implicitly generates the required random nonce value and includes it as part of the overhead size of the AEAD primitive.

# Discussion

I don't have a simple suggestion here. I'm not even sure to what extent it's possible. In many cases this package uses arbitrary nonces that the FIPS mode setting is designed to forbid.

I can see they're used in the retry package as required by the RFCs, and I'm not sure how to convince the Go FIPS module to perform the required operation:
- https://github.com/quic-go/quic-go/blob/v0.49.0/internal/handshake/retry.go#L55
- https://www.rfc-editor.org/rfc/rfc9001#section-5.8
- https://datatracker.ietf.org/doc/html/rfc9369#name-retry-integrity-tag

I can also see sequence numbers are used as nonces (xor'd with an initial IV) in the cipher suite, and that, too, appears to be required by the RFC:
- https://github.com/quic-go/quic-go/blob/v0.49.0/internal/handshake/cipher_suite.go#L86
- https://www.rfc-editor.org/rfc/rfc9001#section-5.8

From what I'm reading so far, I'm not entirely sure it's even possible to get this working this side of the FIPS module boundary. It may have to be supported upstream by Golang.

marten-seemann:
Is this a duplicate of #4894?

hxtk:
Looks like it. I'll close this issue.

PR #5048, opened 2025-04-17
Fix tests in FIPS 140 mode

Lekensteyn:
Two easy test suite improvements to avoid blowing up when running in FIPS mode.

Hopefully this also lays down the basis for future NewGCM improvements per https://github.com/quic-go/quic-go/issues/4894#issuecomment-2793761639.
___
 * Replace RSA-1024 and Ed25519 test certs by P-256
    
    In FIPS mode, signing with RSA keys of 1024 bits and use of Ed25519 are
    not permitted, replace them by NIST P-256. This can also a little bit
    faster which might matter for fuzzing, but it is really insignificant.
    This fixes an explosion of errors while running:
    
        GODEBUG=fips140=on go test ./...
    
    This also fixes failures with `import _ "crypto/tls/fipsonly"` present:
    
        GOEXPERIMENT=boringcrypto go test ./...

 * Fix failing tests in FIPS mode due to TLS_CHACHA20_POLY1305_SHA256
    
    Skip disallowed cipher suite ChaCha20Poly1305 in FIPS mode. Tested with
    `GOEXPERIMENT=boringcrypto` (with `import _ "crypto/tls/fipsonly"`), and
    `GODEBUG=fips140=on`.
    
    `GODEBUG=fips140=only` still explodes for various reasons, including use
    of SHA-1 for fuzzing corpus filenames, and use of NewGCM as reported in
    https://github.com/quic-go/quic-go/issues/4894#issuecomment-2793761639

### DIFF
--- a/example/echo/echo.go
+++ b/example/echo/echo.go
@@ -2,8 +2,9 @@ package main
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/rsa"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
@@ -97,7 +98,7 @@ func (w loggingWriter) Write(b []byte) (int, error) {
 
 // Setup a bare-bones TLS config for the server
 func generateTLSConfig() *tls.Config {
-	key, err := rsa.GenerateKey(rand.Reader, 1024)
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 	if err != nil {
 		panic(err)
 	}
@@ -106,7 +107,11 @@ func generateTLSConfig() *tls.Config {
 	if err != nil {
 		panic(err)
 	}
-	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(key)})
+	privBytes, err := x509.MarshalPKCS8PrivateKey(key)
+	if err != nil {
+		panic(err)
+	}
+	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: privBytes})
 	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: certDER})
 
 	tlsCert, err := tls.X509KeyPair(certPEM, keyPEM)
--- a/fuzzing/handshake/fuzz.go
+++ b/fuzzing/handshake/fuzz.go
@@ -2,8 +2,9 @@ package handshake
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/rsa"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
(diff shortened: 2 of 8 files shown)


Lekensteyn:
Unit tests (macos, Go 1.23.x) failed due to a flaky test I think:
```
--- FAIL: TestPathManagerOutgoingRetransmissions (0.45s)
    path_manager_outgoing_test.go:170: 
        	Error Trace:	/Users/runner/work/quic-go/quic-go/path_manager_outgoing_test.go:170
        	Error:      	"451.720417ms" is not less than "450ms"
        	Test:       	TestPathManagerOutgoingRetransmissions
```

marten-seemann:
What's the point of this? quic-go can't be used in FIPS-140 mode anyway, so why should we care if the test suite can?

Lekensteyn:
quic-go can *currently* not be used in FIPS 140 mode, but the community appears to be interested in it (#4894, #4958). Fixing the tests would be the first step in making progress, and the changes involved are pretty small.

I see the following practical benefits:
- AES-GCM nonce reuse is fatal. It should not occur, but in the future we could rely on the (currently) internal `NewGCMWithCounterNonce` to prevent this completely. See [this comment](https://github.com/quic-go/quic-go/issues/4894#issuecomment-2793761639) linked from the commit message, and https://github.com/golang/go/issues/73110
- Use of RSA 1024 bits is questionable from a security perspective, even if limited to the test suite. Test suite snippets could serve as documentation too. Signing+verification of P-256 is slightly faster compared to RSA 1024 or Ed25519. 

The first small patch (moving away from RSA 1024) alone fixes 99% of the test failures. The second patch (skipping ChaCha20Poly1305) fixes two remaining test failures.

Personally I am interested in helping others, and making some small changes for FIPS 140 seemed a good start. If it causes issues later, feel free to ping me.

marten-seemann:
I'm aware that there seems to be some interest, and it might (or might not) happen at some point.

However, I really don't want to introduce additional complexity for a half-hearted attempt to make quic-go's test suite FIPS-140 compliant. **If** we decide to pursue full FIPS-140 compliance, making the tests compliant might make sense (not even sure if it's required though), but until then, I don't see a lot value whatsoever in going half (or let's be real, 5%) of the way.

Regarding your proposal to move away from RSA, that totally makes sense. These code paths are very old... I'd be happy to review a PR that switches to ed25519 (which I believe is what most people should be using these days).

Lekensteyn:
Would you be interested in contributions towards FIPS 140 compliance? It would be demotivating for me to investigate and carefully create patches towards it only for it to be rejected. Right now I believe it won't take much more to make tests pass in FIPS mode, I can do that extra work to demonstrate it if you like.

I agree that ed25519 would generally be nice to use, but even ECDSA with P-256 is already better than RSA. With the additional benefit that it does not actively work against FIPS 140 compliance.

marten-seemann:
> Would you be interested in contributions towards FIPS 140 compliance? 

Not really, to be honest. Getting quic-go FIPS-140 compliant is a project with a number of moving pieces, and there's no benefit in doing it halfway (or even 95% of the way). There's also no benefit in doing it without getting the actual CMVP validation, if I understand correctly. This is pretty low on my priority list at the moment (although that might change if somebody wants to fund this work stream). @Lekensteyn We have quite a few other issues you could work on that would have higher impact than working on FIPS-140.

> I agree that ed25519 would generally be nice to use, but even ECDSA with P-256 is already better than RSA.

#5050 switches to ed25519.

PR #5065, opened 2025-04-19
store the RTT in the token, not in the session ticket

### LINKED ISSUE #4158: save RTT in the token, not in the session ticket
The token is bound to the path, the session ticket isn’t. We can also save the cwnd, and for exactly the same reason, it should be saved in the token and not the ticket.

tanghaowillow:
This partially fixes #4158

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -238,10 +238,10 @@ var newConnection = func(
 	tlsConf *tls.Config,
 	tokenGenerator *handshake.TokenGenerator,
 	clientAddressValidated bool,
+	rtt time.Duration,
 	tracer *logging.ConnectionTracer,
 	logger utils.Logger,
 	v protocol.Version,
-	rtt time.Duration,
 ) quicConn {
 	s := &connection{
 		ctx:                 ctx,
--- a/connection.go
+++ b/connection.go
@@ -463,6 +463,7 @@ var newClientConnection = func(
 	if s.config.TokenStore != nil {
 		if token := s.config.TokenStore.Pop(s.tokenStoreKey); token != nil {
 			s.packer.SetToken(token.data)
+			s.rttStats.SetInitialRTT(token.rtt)
 		}
 	}
 	return s
@@ -1742,7 +1743,7 @@ func (s *connection) handleNewTokenFrame(frame *wire.NewTokenFrame) error {
 		}
 	}
 	if s.config.TokenStore != nil {
-		s.config.TokenStore.Put(s.tokenStoreKey, &ClientToken{data: frame.Token})
+		s.config.TokenStore.Put(s.tokenStoreKey, &ClientToken{data: frame.Token, rtt: s.rttStats.SmoothedRTT()})
 	}
 	return nil
 }
--- a/connection.go
+++ b/connection.go
@@ -285,6 +285,7 @@ var newConnection = func(
 		connIDGenerator,
 	)
 	s.preSetup()
+	s.rttStats.SetInitialRTT(RTT)
 	s.sentPacketHandler, s.receivedPacketHandler = ackhandler.NewAckHandler(
 		0,
 		protocol.ByteCount(s.config.InitialPacketSize),
--- a/connection.go
+++ b/connection.go
@@ -241,6 +241,7 @@ var newConnection = func(
 	tracer *logging.ConnectionTracer,
 	logger utils.Logger,
 	v protocol.Version,
+	RTT time.Duration,
 ) quicConn {
 	s := &connection{
 		ctx:                 ctx,
@@ -829,7 +830,7 @@ func (s *connection) handleHandshakeComplete(now time.Time) error {
 			s.queueControlFrame(s.oneRTTStream.PopCryptoFrame(protocol.MaxPostHandshakeCryptoFrameSize))
 		}
 	}
-	token, err := s.tokenGenerator.NewToken(s.conn.RemoteAddr())
+	token, err := s.tokenGenerator.NewToken(s.conn.RemoteAddr(), s.rttStats.SmoothedRTT())
 	if err != nil {
 		return err
 	}
--- a/connection.go
+++ b/connection.go
@@ -285,7 +285,7 @@ var newConnection = func(
 		connIDGenerator,
 	)
 	s.preSetup()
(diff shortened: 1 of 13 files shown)


connection.go
```diff
@@ -511,6 +511,10 @@ func (s *connection) preSetup() {
 	s.connState.Version = s.version
 }
 
+func (s *connection) setInitalCongestionStats(RTT time.Duration) {
+	s.rttStats.SetInitialRTT(RTT)
+}
```
marten-seemann, line 516: Why is this not on the connection constructor?

tanghaowillow: It can be. Using an additional function looks more explicit to me. What do you think?

marten-seemann: If it's only used once (which it is here), it should definitely be on the constructor. A method implies that it's potentially called multiple times over the lifetime of the object.

tanghaowillow: Ok, I have moved it to the constructor.

connection.go
```diff
@@ -241,6 +241,7 @@ var newConnection = func(
 	tracer *logging.ConnectionTracer,
 	logger utils.Logger,
 	v protocol.Version,
+	RTT time.Duration,
```
marten-seemann, line 244:
```suggestion
	rtt time.Duration,
```

internal/handshake/token_generator.go
```diff
@@ -35,6 +36,7 @@ type token struct {
 	IsRetryToken             bool
 	RemoteAddr               []byte
 	Timestamp                int64
+	RTT                      int64 // in ms
```
marten-seemann, line 39: Millisecond is too low-resolution for some circumstances (e.g. data center traffic). In the session ticket, we used microsecond resolution. Any reason to change it?

server.go
```diff
@@ -647,6 +648,12 @@ func (s *baseServer) handleInitialImpl(p receivedPacket, hdr *wire.Header) error
 		return nil
 	}
 
+	// restore RTT from token
+	var RTT time.Duration
```
marten-seemann, line 652:
```suggestion
	var rtt time.Duration
```

server.go
```diff
@@ -647,6 +648,12 @@ func (s *baseServer) handleInitialImpl(p receivedPacket, hdr *wire.Header) error
...
+	if (token != nil) && !token.IsRetryToken {
```
marten-seemann, line 653:
```suggestion
	if token != nil && !token.IsRetryToken {
```

fuzzing/tokens/fuzz.go
```diff
@@ -57,7 +57,7 @@ func newToken(tg *handshake.TokenGenerator, data []byte) int {
 		}
 	}
 	start := time.Now()
-	encrypted, err := tg.NewToken(addr)
```
marten-seemann, line 60: You need to reslice `data` here, so that you don't use the same field for IP and RTT.

marten-seemann (commented):
I might be missing something, but where is the RTT that's passed to the connection actually used?

tanghaowillow:
@marten-seemann Thanks for the careful review. 
One problem in my mind is that how should we do this for client side. The client should not use the session state to save and restore the RTT according to the reason mentioned in #4158. But does the quic spec allow a client to add custom data when caching a token (and restore the data later)?

marten-seemann:
> One problem in my mind is that how should we do this for client side.

We can totally do this. Haven't looked into this in detail, but we could add an `rtt` field to the `ClientToken`. This is orthogonal to this PR though, isn't it? The current code doesn't restore the RTT on the client side?

tanghaowillow:
Yes, the current change is only for server side, and the client side still saves RTT by appending it to the received session ticket.  I'll make the change for client side as well in this PR then.

connection.go
marten-seemann: This still needs to be fixed, variables should be lower case.

fuzzing/tokens/fuzz.go
marten-seemann: This too.

internal/handshake/token_generator.go
```diff
@@ -20,6 +20,7 @@ type Token struct {
 	IsRetryToken      bool
 	SentTime          time.Time
 	encodedRemoteAddr []byte
+	RTT               time.Duration
```
marten-seemann, line 23:
```suggestion
	// only set for tokens sent in NEW_TOKEN frames
	RTT               time.Duration
```

internal/handshake/token_generator.go
```diff
@@ -69,10 +71,11 @@ func (g *TokenGenerator) NewRetryToken(
 }
 
 // NewToken generates a new token to be sent in a NEW_TOKEN frame
-func (g *TokenGenerator) NewToken(raddr net.Addr) ([]byte, error) {
+func (g *TokenGenerator) NewToken(raddr net.Addr, RTT time.Duration) ([]byte, error) {
```
marten-seemann, line 74:
```suggestion
func (g *TokenGenerator) NewToken(raddr net.Addr, rtt time.Duration) ([]byte, error) {
```

internal/handshake/token_generator.go
```diff
@@ -107,6 +110,8 @@ func (g *TokenGenerator) DecodeToken(encrypted []byte) (*Token, error) {
 	if t.IsRetryToken {
 		token.OriginalDestConnectionID = protocol.ParseConnectionID(t.OriginalDestConnectionID)
 		token.RetrySrcConnectionID = protocol.ParseConnectionID(t.RetrySrcConnectionID)
+	} else {
+		token.RTT = time.Duration(t.RTT * int64(time.Microsecond))
```
marten-seemann, line 114: I assume you mean
```suggestion
		token.RTT = time.Duration(t.RTT) * time.Microsecond
```

marten-seemann:
@tanghaowillow Could you incorporate my feedback into this PR? Would love to get this in for the next release!

tanghaowillow:
Sure. I don't have my device for coding with me now and have some troubles with github website operations.  I'll take care of this on this weekend if no hurry.

marten-seemann:
That's fine, I've been planning the release for some time next week.

Issue #5088, opened 2025-04-28
interface index cmsg broken on s390x


Zxilly:
```
uname -a
Linux ci 6.8.0-58-generic #60-Ubuntu SMP Fri Mar 14 17:31:55 UTC 2025 s390x s390x s390x GNU/Linux
```

```
sudo strace -f -e trace=network -p 9800
strace: Process 9800 attached with 7 threads
[pid  9803] recvmmsg(3, [{msg_hdr={msg_name={sa_family=AF_INET, sin_port=htons(43387), sin_addr=inet_addr("IPMasked")}, msg_namelen=28 => 16, msg_iov=[{iov_base="\311\0\0\0\1\t\211\350\230\26\250\25\275\214\276\0\0D\3558#Iz\376\24\227|\254&\27\t\245"..., iov_len=1452}], msg_iovlen=1, msg_control=[{cmsg_len=28, cmsg_level=SOL_IP, cmsg_type=IP_PKTINFO, cmsg_data={ipi_ifindex=if_nametoindex("enc1000"), ipi_spec_dst=inet_addr("IPMasked"), ipi_addr=inet_addr("IPMasked")}}, {cmsg_len=17, cmsg_level=SOL_IP, cmsg_type=IP_TOS, cmsg_data=[0]}], msg_controllen=56, msg_flags=0}, msg_len=1280}], 8, 0, NULL) = 1
[pid  9803] recvmmsg(3, 0xc000390a00, 8, 0, NULL) = -1 EAGAIN (Resource temporarily unavailable)
[pid  9802] sendmsg(3, {msg_name={sa_family=AF_INET, sin_port=htons(43387), sin_addr=inet_addr("IPMasked")}, msg_namelen=16, msg_iov=[{iov_base="\307\0\0\0\1\0\4d\v#\37\0@u#\210j\271G\211?G,[;\377>:~\220z\356"..., iov_len=1280}], msg_iovlen=1, msg_control=[{cmsg_len=28, cmsg_level=SOL_IP, cmsg_type=IP_PKTINFO, cmsg_data={ipi_ifindex=33554432, ipi_spec_dst=inet_addr("IPMasked"), ipi_addr=inet_addr("0.0.0.0")}}], msg_controllen=32, msg_flags=0}, 0) = -1 ENODEV (No such device)
[...]

PR #5094, opened 2025-04-29
fix: parse ifindex from packet correctly

Zxilly:
Updates #5088 

Now `ipi_ifindex` can be properly parsed from the packet.

I guess this probably won't affect `darwin`, since it seems they only build `small-endian` devices today, but it's decided at compile time anyway and doesn't introduce additional overhead.

### DIFF
--- a/sys_conn_helper_darwin.go
+++ b/sys_conn_helper_darwin.go
@@ -30,7 +30,7 @@ func parseIPv4PktInfo(body []byte) (ip netip.Addr, ifIndex uint32, ok bool) {
 	if len(body) != 12 {
 		return netip.Addr{}, 0, false
 	}
-	return netip.AddrFrom4(*(*[4]byte)(body[8:12])), binary.LittleEndian.Uint32(body), true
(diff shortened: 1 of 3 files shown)


Zxilly:
Yes, s390x runs on big-endian

Zxilly:
I don't think this fixes the ecn issue on the s390x, it's a basic fix and we need more work.

marten-seemann:
> I don't think this fixes the ecn issue on the s390x, it's a basic fix and we need more work.

I edited your original issue, so now it's about the interface index.

Would you be interested in contributing a fix for ECN as well? Do you have any suggestion how we could test this in CI (not required, but in my experience, things that are not getting tested in CI tend to break sooner or later)?

marten-seemann:
> I know very little about ECNs, so I'm not quite sure what error occurred at this point.

Step 1 would be making sure that the cmsg doesn't return any error on the `sendmsg` call. Step 2 would be verifying that packets sent have the ECT0 set. This part is really easy: you can see this in Wireshark.

Zxilly:
We can merge this PR for now, at least this allows quic-go to run properly without ECN enabled, I'll try to understand some ECN related stuff later.

marten-seemann (approved):

PR #5107, opened 2025-05-02
split SNI and ECH extensions in the ClientHello

### LINKED ISSUE #4755: scramble the ClientHello
As an easy way to confuse traffic-inspecting middleboxes, we can cut the CRYPTO frame containing the ClientHello into multiple pieces, and shuffle their order (potentially interleaved with PADDING frames).

If there's an easy way to determine the position of the SNI, we can make sure that we always cut the SNI in (at least) two pieces as well.

And if PQ is enabled (and the ClientHello stretches across 2 packets anyway), we can make sure these two pieces are sent in two different packets.

marten-seemann:
Fixes #4755.


marten-seemann:
Looks like there's a flaky test:
```
 --- FAIL: TestInitialCryptoStreamClient (0.00s)
    crypto_stream_test.go:175: 
        	Error Trace:	D:/a/quic-go/quic-go/crypto_stream_test.go:175
        	Error:      	Should not be: 308
        	Test:       	TestInitialCryptoStreamClient

marten-seemann:
> Looks like there's a flaky test:
> 
> ```
>  --- FAIL: TestInitialCryptoStreamClient (0.00s)
>     crypto_stream_test.go:175: 
>         	Error Trace:	D:/a/quic-go/quic-go/crypto_stream_test.go:175
>         	Error:      	Should not be: 308
>         	Test:       	TestInitialCryptoStreamClient
> ```

This has been fixed. It came from the scrambling logic of the ClientHello: if the SNI ends up at the very end of the message, there's nothing to scramble.

marten-seemann:
Why is this annoying for QUIC servers?

marten-seemann:
For now it does, but don't rely on it. This is not part of the official API and we might remove it at some point.

Issue #5013, opened 2025-03-30
Need a way to set some state while rate limiting


sukunrt:
I'm limiting concurrent quic connections from the same IP in go-libp2p. The way we implement the limiting is to create a resource manager scope that tracks all the connection resource usage throughout the lifetime of the connection. 

I want to: 
1. On receiving ClientHello, create a scope and reject the connection without handshaking if there are too many connections from the same IP. 
2. On accepting a ClientHello, when I eventually get the quic connection from `Server.Accept` retrieve the scope created in 1. 

I can use `Config.GetConfigForClient` to rate limit based on IP, but there's no way to set store the scope that I can later get back after the handshake. 
I can use `Server.ConnContext` to store the scope in the context, but the `ConnContext` callback has no info about remote IP.   

If we add the connection context to `ClientHelloInfo` it'll be enough for my usecase. This change is also backwards compatible. See #5014

marten-seemann:
Can you elaborate what information the application would store and retrieve at which point during the handshake?

sukunrt:
I want to store the resource manager scope and retrieve it later when the handshake completes using `conn.Context().Value("scope")`

My proposed solution doesn't work as I need to modify the context used for the connection with: 
`context.WithValue(ctx, "scope", scope)`

If the ConnContext were something like http server, I could do: 
```
httpServer.ConnContext = func (ctx context.Context, conn net.Conn) {
    scope, err := rcmgr.OpenConnection(conn.RemoteAddr())
    if err != nil {
        conn.Close() // drop the connection rate limited
        return ctx
    }
    ctx = context.WithValue(ctx, "scope", scope)
    context.AfterFunc(ctx, func() {scope.Done()}) // arrange for scope to be closed whenever conn closes.  
    return ctx
}
```

Issue #3862, opened 2023-05-28
call GetConfigForClient after processing the ClientHello


marten-seemann:
As pointed out by @mcluseau in https://github.com/quic-go/quic-go/pull/3807#issuecomment-1537111815, the current API doesn’t allow setting transport parameters based on the APLN chosen during the TLS handshake.

This is because we call `GetConfigForClient` when creating the `connection` struct, which happens when we receive of the first Initial packet from the client. However, at that point we don’t necessarily possess the entire ClientHello yet (it might be split across multiple packets, for example).

We should call `GetConfigForClient` after _processing_ the ClientHello (which will call the `tls.Config.GetConfigForClient`).This means that we’ll need to hold off on initializing certain components needed by the `connection`. This will create a bit of complexity, but should be solvable.

The more interesting consequence is that we’ll need to rethink which values can be on the `Config`. For example, `InitialRTT` and `HandshakeTimeout` need to be moved somewhere else (on the `Transport` or a `ListenerConfig`?), since they’re needed right away when we start processing packets from the client.

marten-seemann:
As the issue and the label says, this issue is blocked at the moment.

marten-seemann:
Update regarding `HandshakeIdleTimeout`: It seems valuable to be able to set an idle timeout based on the IP address. It would therefore be nice to keep it on the `Config`.

One option is to use the outer `Config`'s `HandshakeIdleTimeout`, and update the value once the `GetConfigForClient` callback has been called.

marten-seemann:
Next question: It's not clear what to do about the `ConnectionTracer`. Not logging the first packet(s) is not a good option. However, without calling the callback, we don't know if a `ConnectionTracer` would be set for that connection. This means we'd have to cache the tracing events, just in case the callback returns a config that has tracing enabled.

This might be fine, as this is most likely not more than a few packets, if at all. On the other hand, it's an annoying bit of extra complexity.

vitorpmoreira:
Hello!

I've noticed that I get a "valid" connection and no error after dialing to a server which includes a custom peer certificate verification. I can even open a stream without any errors but it then fails when trying to writing to that stream (given that the server issued an error in that custom verification).
Does this have something to do with this behavior? It appears that the dial does not wait for the whole handshake, just for the first packet (which is sent by the server after validating the peer's certificate?).

Please take this with a grain of salt, I may have just said a bunch of non-sense (if so, I kindly ask you to enlighten me)!
Thank you!

[...]

PR #5122, opened 2025-05-05
pass ClientInfo to ConnContext, allow rejecting of handshakes

sukunrt:
This allows users to set some state related to the client's remote address in `ConnContext`.

This also refuses connection early if the returned context is cancelled.

Fixes: #5013 

See https://github.com/libp2p/go-libp2p/pull/3283/ for the dependant go-libp2p change.

### DIFF
--- a/integrationtests/self/handshake_context_test.go
+++ b/integrationtests/self/handshake_context_test.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"net"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -50,8 +52,8 @@ func TestConnContextOnServerSide(t *testing.T) {
 
 	tr := &quic.Transport{
 		Conn: newUDPConnLocalhost(t),
-		ConnContext: func(ctx context.Context) context.Context {
-			return context.WithValue(ctx, "foo", "bar") //nolint:staticcheck
+		ConnContext: func(ctx context.Context, _ *quic.ClientInfo) (context.Context, error) {
+			return context.WithValue(ctx, "foo", "bar"), nil //nolint:staticcheck
 		},
 	}
 	defer tr.Close()
@@ -132,11 +134,53 @@ func TestConnContextOnServerSide(t *testing.T) {
 	checkContext(tlsGetCertificateContextChan, false)
 }
 
+func TestConnContext(t *testing.T) {
+	var shouldReject atomic.Bool
+	tr := &quic.Transport{
+		Conn: newUDPConnLocalhost(t),
+		ConnContext: func(ctx context.Context, ci *quic.ClientInfo) (context.Context, error) {
+			if shouldReject.Load() {
+				return nil, errors.New("rejecting connection")
+			}
+			return context.WithValue(ctx, "addr", ci.RemoteAddr), nil //nolint:staticcheck
+		},
+	}
+	defer tr.Close()
+
+	server, err := tr.Listen(
+		getTLSConfig(),
+		getQuicConfig(nil),
+	)
+	require.NoError(t, err)
+	defer server.Close()
+	t.Run("ClientInfo", func(t *testing.T) {
+		shouldReject.Store(false)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+		defer cancel()
+		pc := newUDPConnLocalhost(t)
+		c, err := quic.Dial(ctx, pc, server.Addr(), getTLSClientConfig(), getQuicConfig(nil))
+		require.NoError(t, err)
+		defer c.CloseWithError(0, "")
+
+		conn, err := server.Accept(ctx)
+		require.NoError(t, err)
+		require.Equal(t, pc.LocalAddr().String(), conn.Context().Value("addr").(net.Addr).String())
+		conn.CloseWithError(0, "")
+	})
+	t.Run("Reject", func(t *testing.T) {
+		shouldReject.Store(true)
+		_, err := quic.Dial(context.Background(), newUDPConnLocalhost(t), server.Addr(), getTLSClientConfig(), getQuicConfig(nil))
+		require.Error(t, err)
+	})
+}
+
 // Users are not supposed to return a fresh context from ConnContext, but we should handle it gracefully.
 func TestConnContextFreshContext(t *testing.T) {
 	tr := &quic.Transport{
-		Conn:        newUDPConnLocalhost(t),
-		ConnContext: func(ctx context.Context) context.Context { return context.Background() },
+		Conn: newUDPConnLocalhost(t),
+		ConnContext: func(ctx context.Context, _ *quic.ClientInfo) (context.Context, error) {
+			return context.Background(), nil
+		},
 	}
 	defer tr.Close()
 	server, err := tr.Listen(getTLSConfig(), getQuicConfig(nil))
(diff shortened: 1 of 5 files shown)


transport.go
```diff
@@ -115,7 +115,8 @@ type Transport struct {
 	// implementation of this callback (negating its return value).
 	VerifySourceAddress func(net.Addr) bool
 
-	// ConnContext is called when the server accepts a new connection.
+	// ConnContext is called when the server accepts a new connection. To reject a connection, return a
+	// cancelled context.
```
marten-seemann, line 119: I've never seen this pattern before. Is there any precedent?

sukunrt: I couldn't find any. 

Do you want to change this to return `context.Context, error`.

marten-seemann:
Here are a few thoughts:
* Rejecting connections early (**before** calling the `connection` constructor) seems like the right way to do this.
* At some point, we want to delay calling `GetConfigForClient` until the entire ClientHello (which will commonly span 2 packets with PQ key shares) has been received and processed, in order to allow the application to apply different QUIC transport parameters based on the ALPN chosen (see https://github.com/quic-go/quic-go/issues/3862).
* Returning an `error` (as suggested in https://github.com/quic-go/quic-go/pull/5122#discussion_r2073640592) seems more idiomatic than returning a canceled context
* I'm not sure if I'm a fan of passing `ClientInfo` to both callbacks. No particular reason, just doesn't feel 100% right.

sukunrt:
>I'm not sure if I'm a fan of passing ClientInfo to both callbacks. No particular reason, just doesn't feel 100% right.

I think this is because now there are two ways to rate limit. Error from `GetConfigForClient` or error from `ConnContext`. 

I'd argue this is still directionally correct:
1. When we move `GetConfigForClient` to after parsing the tls ClientHello, we'll change this struct to include ALPNs, and other tls ClientHello info. So it'll be meaningfully different.
2. This method allows us to set some state in ConnContext dependent on the remote address(the entire reason why we are doing this change).

marten-seemann:
> When we move `GetConfigForClient` to after parsing the tls ClientHello, we'll change this struct to include ALPNs, and other tls ClientHello info. So it'll be meaningfully different.

That's a good point. If we don't want the application to keep a map of connection for which `tls.Config.GetConfigForClient` has been called (but `quic.Config.GetConfigForClient` not yet), we'll have to pass some more information to our callback.

Passing the ALPN seems like a reasonable suggestion, because that's probably the primary use case: different application protocols might have very different requirements regarding stream and flow control limits, as well as supported QUIC extensions.

At the same time, it also feels rather limiting. We support this one use case, but not even a tiny bit more (totally made-up example: what if I want to enable [Extended Key Updates](https://datatracker.ietf.org/doc/draft-rosomakho-quic-extended-key-update/), but only if a certain cipher suite / curve is in use?)

marten-seemann:
> Passing the ALPN seems like a reasonable suggestion, because that's probably the primary use case: different application protocols might have very different requirements regarding stream and flow control limits, as well as supported QUIC extensions.
> 
> At the same time, it also feels rather limiting. We support this one use case, but not even a tiny bit more (totally made-up example: what if I want to enable [Extended Key Updates](https://datatracker.ietf.org/doc/draft-rosomakho-quic-extended-key-update/), but only if a certain cipher suite / curve is in use?)

There's a very simple solution: use the `tls.ConnectionState`. Why didn't I think of this yesterday? 🤦🏻‍♂️

marten-seemann (commented):
I'm wondering if we should pass `ClientInfo` instead of `*ClientInfo` to the callback (both of them actually). This might save some allocs.

No need to do it in this PR, if at all, this should be done in a follow-up for both callbacks.

integrationtests/self/handshake_context_test.go
```diff
@@ -132,11 +134,53 @@ func TestConnContextOnServerSide(t *testing.T) {
 	checkContext(tlsGetCertificateContextChan, false)
 }
 
+func TestConnContext(t *testing.T) {
```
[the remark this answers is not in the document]
sukunrt: fixed

integrationtests/self/handshake_context_test.go
```diff
@@ -132,11 +133,53 @@ func TestConnContextOnServerSide(t *testing.T) {
 	checkContext(tlsGetCertificateContextChan, false)
 }
 
+func TestConnContextRejection(t *testing.T) {
+	testConnContextRejection(t, true)
+	testConnContextRejection(t, false)
```
marten-seemann, line 138:
```suggestion
	t.Run("rejecting", func(t *testing.T) {
		testConnContextRejection(t, true)
	})
	t.Run("not rejecting", func(t *testing.T) {
		testConnContextRejection(t, false)
	})
```

integrationtests/self/handshake_context_test.go
```diff
@@ -132,11 +133,53 @@ func TestConnContextOnServerSide(t *testing.T) {
...
+	defer server.Close()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	pc := newUDPConnLocalhost(t)
+	c, err := quic.Dial(ctx, pc, server.Addr(), getTLSClientConfig(), getQuicConfig(nil))
+	if reject {
+		require.Error(t, err)
+		return
+	}
+	require.NoError(t, err)
```
marten-seemann, line 167: We should expect an error in case the connection is getting rejected, shouldn't we?

sukunrt: made this an explicit `CONNECTION_REFUSED` check

integrationtests/self/handshake_context_test.go
```diff
@@ -132,11 +133,57 @@ func TestConnContextOnServerSide(t *testing.T) {
...
+		getQuicConfig(nil),
+	)
+	require.NoError(t, err)
+	defer server.Close()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	pc := newUDPConnLocalhost(t)
+	c, err := quic.Dial(ctx, pc, server.Addr(), getTLSClientConfig(), getQuicConfig(nil))
+	if reject {
+		require.ErrorContains(t, err, "CONNECTION_REFUSED")
```
marten-seemann, line 168:
```suggestion
		var terr *quic.TransportError
		require.ErrorAs(t, err, &terr)
		require.Equal(t, quic.ConnectionRefused, terr.ErrorCode)
```

marten-seemann, line 168:
```suggestion
		require.ErrorIs(t, err, &quic.TransportError{Remote: true, ErrorCode: quic.ConnectionRefused})
```

integrationtests/self/handshake_context_test.go
marten-seemann: I must have missed that, sorry.

integrationtests/self/handshake_context_test.go
```diff
@@ -132,11 +134,53 @@ func TestConnContextOnServerSide(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, pc.LocalAddr().String(), conn.Context().Value("addr").(net.Addr).String())
+		conn.CloseWithError(0, "")
+	})
+	t.Run("Reject", func(t *testing.T) {
+		shouldReject.Store(true)
+		_, err := quic.Dial(context.Background(), newUDPConnLocalhost(t), server.Addr(), getTLSClientConfig(), getQuicConfig(nil))
+		require.Error(t, err)
+	})
+}
```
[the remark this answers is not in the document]
marten-seemann: Yes, thank you!

marten-seemann (approved):

PR #5124, opened 2025-05-06
http3: Add a check for existing certs in TLSConfig before using certFile & keyFile

Northernside:
This PR updates the receiver method `ListenAndServeTLS` to use existing certs from `TLSConfig.Certificates` and ignore empty `certFile` & `certKey` parameters <- this behavior is the same as in [`net/http`](https://github.com/golang/go/blob/master/src/net/http/server.go#L3511-L3519).

Not sure if the previous behavior was intended by the authors of quic-go/http3 but I think this would count as something positive towards "feature parity".

I'm not a licensing expert, but I tried to check if the almost 1:1 "copy" of the codeblock is fine between the BSD-style and MIT license and as to my understanding, it should be. Please correct me if I'm wrong.

### DIFF
--- a/http3/server.go
+++ b/http3/server.go
@@ -204,14 +204,19 @@ func (s *Server) ListenAndServe() error {
 // If s.Addr is blank, ":https" is used.
 func (s *Server) ListenAndServeTLS(certFile, keyFile string) error {
 	var err error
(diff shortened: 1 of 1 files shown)


marten-seemann:
Why not just use `ListenAndServe` in that case?

Issue #3373, opened 2022-04-07
Consider moving away from unmaintained gojay


imsodin:
Gojay has not seen any activity in a long time, and one of it's dependencies isn't available at the given source path anymore, causing build failures:  
https://github.com/francoispqt/gojay/issues/128  
https://github.com/syncthing/syncthing/issues/8256

Looking at a recent comparison (https://medium.com/geekculture/my-golang-json-evaluation-20a9ca6ef79c) it looks like the standard `encoding/json` library is pretty decent for encoding (and there's only encoding in the usecase here, right?). Otherwise https://github.com/goccy/go-json or https://github.com/json-iterator/go look like good candidates.

dmanlfc:
@marten-seemann can you update accordingly?

```
go: github.com/lucas-clemente/quic-go@v0.25.0 requires
	github.com/francoispqt/gojay@v1.2.13 requires
	cloud.google.com/go@v0.37.0 requires
	go.opencensus.io@v0.18.0 requires
	git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999: invalid version: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /x86_64/host/usr/share/go-path/pkg/mod/cache/vcs/f6f367fb9d5656665ab3c2c1c3b2a44c2ffc79f6b70612f5864d7ceedcc7b9ea: exit status 128:
	fatal: repository 'https://git.apache.org/thrift.git/' not found

```

marten-seemann:
What is there to update? We're already using the most recent gojay version.
Wasn't the Go module proxy intended to prevent this from happening?

imsodin:
I don't think there's anything to update. And I don't understand under what circumstances that build issue is happening (doesn't for me). Maybe there's some build modes where said proxy isn't used? Anyway my request is more about dropping an unmaintained library (gojay), that also adds unnecessary dependencies - which as a nice side-effect will also fix this build error.

dmanlfc:
@marten-seemann the thrift source site is outdated

marten-seemann:
@imsodin, sorry, I was only replying to @dmanlfc yesterday.

> Looking at a recent comparison (https://medium.com/geekculture/my-golang-json-evaluation-20a9ca6ef79c) it looks like the standard encoding/json library is pretty decent for encoding (and there's only encoding in the usecase here, right?).

[...]

PR #5125, opened 2025-05-07
Fork minimal gojay to remove dependencies

klauspost:
gojay has a large number of [dependencies](https://github.com/francoispqt/gojay/blob/master/go.mod), mostly used for benchmarks.

After fork only 2 dependencies are left. It *seems* like upstream is pretty dead, hence the fork as a pragmatic solution.

While [this article](https://huntedlabs.com/the-russian-open-source-project-that-we-cant-live-without/) may overstate the risk the dependency can easily be removed.

If you prefer another solution feel free to do what you see fit.


marten-seemann:
I'm sympathetic to removing gojay (we've had an issue for this for years now: https://github.com/quic-go/quic-go/issues/3373), but I'm not sure I want to fork it.

That said, I'm a bit hesitant to add 30k LOC - even though it's just a fork. I think a more satisfying solution would be either switching to another JSON library (or maybe `encoding/json/v2`, haven't checked on the status of that project for a long time), or maybe even rolling our own encoder (how hard can it be 😬). The problem is that qlogging needs to be blazing fast if it should be even remotely useful in debugging transfers happening at gigabit speeds.

klauspost:
> That said, I'm a bit hesitant to add 30k LOC  - even though it's just a fork

I completely get that. That was also my initial though, but that changes needed for that appeared to be too substantial for me to commit to in a reasonable timeframe. So this was mainly a stop-gap solution, since I didn't expect a re-write to be in a reasonable timeframe.

I also get that there is no direct codepath to easyjson or most other deps, so it doesn't look like it would pick up init()'s from that package, which of course reduces the problem surface.

My perspective: We (MinIO) currently have an indirect dependency to the "problematic code". The only path is through this package. We would avoid a *lot* of questions if we didn't have that dependency. So for us it would be a lot better to not have it.

klauspost:
I listed the changes I did in the OP in case you would like to do it yourself 

In my own projects, I would personally also be more comfortable doing it myself.

klauspost:
@marten-seemann Yeah, this is clearly a stop-gap before #3373 hits. I guess the Q is more how long a stop-gap would remain. 

This change is a relatively "lightweight" change - I would more say it is "+30K LOC in repo, -300K LOC in dependencies" (didn't count). 

I fully trust you will do what makes sense. Thanks for looking at it!

xplshn:
> I'm sympathetic to removing gojay (we've had an issue for this for years now: #3373), but I'm not sure I want to fork it.
> 
> That said, I'm a bit hesitant to add 30k LOC - even though it's just a fork. I think a more satisfying solution would be either switching to another JSON library (or maybe `encoding/json/v2`, haven't checked on the status of that project for a long time), or maybe even rolling our own encoder (how hard can it be 😬). The problem is that qlogging needs to be blazing fast if it should be even remotely useful in debugging transfers happening at gigabit speeds.

What about https://github.com/goccy/go-json? It is pretty optimized, fast, and its pure Go. It's also compatible with the std's encoding/json.

marten-seemann:
> > I'm sympathetic to removing gojay (we've had an issue for this for years now: #3373), but I'm not sure I want to fork it.
> > That said, I'm a bit hesitant to add 30k LOC - even though it's just a fork. I think a more satisfying solution would be either switching to another JSON library (or maybe `encoding/json/v2`, haven't checked on the status of that project for a long time), or maybe even rolling our own encoder (how hard can it be 😬). The problem is that qlogging needs to be blazing fast if it should be even remotely useful in debugging transfers happening at gigabit speeds.
> 
> What about https://github.com/goccy/go-json? It is pretty optimized, fast, and its pure Go. It's also compatible with the std's encoding/json.

Please keep the discussion about alternative JSON encoding libraries on #3373. Compatibility with `encoding/json` is neither required nor desired, since reflections are slow. Whatever we go with needs to be at least as fast as gojay *in our particular use case*.

klauspost:
> github.com/quic-go/gojay

@marten-seemann Sounds good to me! 👍🏼

klauspost:
👍🏼 Will open a PR with cleanup and one here to do the replace.

@marten-seemann Would you prefer I do a `replace github.com/francoispqt/gojay v1.2.13 => github.com/quic-go/gojay` or should I change the imports?   (I have no opinion)

marten-seemann:
I just reread the article (I had only skimmed it before), and I'm a bit confused now. If I understand correctly, the only finding is that the authors of easyjson are based out of Moscow and employed by a Russian company (who would've thought, given that the GitHub org literally is mailru...).

Please correct me if I'm wrong, but if I understand correctly, not a single line of malicious code was found.

Of course, there's no guarantee that they won't ship a malicious update at some point in the future, but that applies to literally every dependency that you use (and many open source authors are not as transparent as the easyjson authors about their funding streams).

It also [looks](https://github.com/search?q=repo%3Afrancoispqt%2Fgojay+easyjson+language%3AGo&type=code&l=Go) like easyjson is only imported as a testing dependency, and Go prunes testing dependencies. easyjson there **is not** a transitive dependency of quic-go.

> That seems reasonable, and the article you cites is a good reason to get #3373 resolved rather sooner than later.

I therefore don't think that this article constitutes a good reason to move ahead with #3373 any faster than we initially planned (the upcoming availability of `encoding/json/v2` might be a good reason though, completely independent of this!).

---

That said, I'm ok with moving ahead with the plan to replace gojay with a fork, provided that it actually significantly slims down the dependency tree (i.e. it removes entries from `go.mod`, not only from `go.sum`).

Issue #4968, opened 2025-02-24
Connection should be a struct, not an interface


marten-seemann:
`Connection` should be a struct, not an interface. This will allow us to add methods to the `Connection` without breaking backwards-compatibility. It's also best practice in Go to return structs (and accept interfaces).

This is coming up now since connection migration (#234) will require adding one or two more methods to the connection, and it's really annoying that we have to change the interface for that.

The `EarlyConnection` can then just embed a `Connection` and implement the necessary methods:
```go
type EarlyConnection struct {
	Connection
}

func (s *EarlyConnection) HandshakeComplete() <-chan struct{} { return s.handshakeCompleteChan }
func (s *EarlyConnection) NextConnection(ctx context.Context) (*Connection, error) { // ... }
```

This will likely require some moderately annoying refactoring of the QUIC server implementation, but we already have two separate structs (`Listener` and `EarlyListener`).

Obviously, this in itself is a breaking API change, and will likely require some changes in the `http3` package, as well as in the connect-udp and connect-ip.

marten-seemann:
The challenge here is how to testing, especially testing the server. Using an interface was really convenient, since we could just mock the functions. This won't be possible once we switch to a struct.

One option would be to define an interface:
```go
type Connection struct {
     impl connectionInterface
}

func (c *Connection) AcceptStream() quic.Stream {
     return c.impl.AcceptStream()
}
```

In tests, we could then use a mock interface.

We could also have a separate implementation of `Connection` that only applies to tests. This means that we _somehow_ have to hide the production `Connection` in test code, which is a bit painful, since there's no `test` build tag (see https://github.com/golang/go/issues/21360). What we could do is introduce our own tag, but this means that one has to use that tag when running tests:
```go
// connection.go
//go:build !quicgotest

type Connection struct {}
```

This could be made slightly less non-obvious by defining a test file:
```go
// tag_test.go
//go:build !quicgotest
func init() {
    panic("go test must be run with -tags quicgotest")
}
```

marten-seemann:
We should also consider renaming `Connection` to `Conn`. This is more in line with standard library naming (for example, it's a `net.Conn`, not a `net.Connection`). It's also a "more breaking" change, which is a good thing in this case, since it makes it really obvious that users need to adapt their code to the new API.

PR #5126, opened 2025-05-07
Refactor: Split Datagram and Multipath APIs into Dedicated Interfaces

OkutaniDaichi0106:
## Refactor: Split Datagram and Multipath APIs into Dedicated Interfaces

This pull request proposes a structural refactoring of the `Connection` interface by moving Datagram and Multipath related methods into separate interfaces:

- `DatagramTransmitter`: encapsulates `SendDatagram` and `ReceiveDatagram` for RFC 9221-compliant datagram support.
- `Multipather`: encapsulates `AddPath` for multipath support.

### Motivation

The `Connection` interface in its current form aggregates a wide range of features—some of which are optional, such as Datagram and Multipath. This tight coupling results in bloated interfaces and makes it difficult for downstream users and libraries to rely on a minimal set of required APIs.

By splitting these into narrowly-scoped interfaces, we:

- Improve separation of concerns and interface clarity.
- Align with Go interface design idioms (i.e., small, focused interfaces).
- Make it easier for users to opt into optional features via interface assertion, similar to patterns used in the standard library (e.g., `http.Flusher`).
- Reduce the need for external consumers to redefine their own interfaces to isolate these features.

### Example Usage

~~~go
var conn quic.Connection

if dgram, ok := conn.(DatagramTransmitter); ok {
    _ = dgram.SendDatagram([]byte("hello"))
}

if mp, ok := conn.(Multipather); ok {
    _, _ = mp.AddPath(transport)
}
~~~

### Notes

- This change is backwards-compatible.
- It does not affect the existing behavior or performance of the `Connection` interface.
- No methods were removed; only reorganized into logical extension points.

Feedback and suggestions are very welcome. Thank you for maintaining `quic-go`.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -351,7 +351,7 @@ type ConnectionState struct {
 	GSO bool
 }
 
-type DatagramExtension interface {
+type DatagramTransmitter interface {
 	// SendDatagram sends a message using a QUIC datagram, as specified in RFC 9221.
 	// There is no delivery guarantee for DATAGRAM frames, they are not retransmitted if lost.
 	// The payload of the datagram needs to fit into a single QUIC packet.
@@ -362,7 +362,7 @@ type DatagramExtension interface {
 	ReceiveDatagram(context.Context) ([]byte, error)
 }
 
-type MultipathExtension interface {
+type Multipather interface {
 	// AddPath adds a new path to the connection.
 	AddPath(*Transport) (*Path, error)
 }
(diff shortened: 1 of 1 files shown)


marten-seemann:
We're aware of this problem, but the solution will likely look different: https://github.com/quic-go/quic-go/issues/4968

OkutaniDaichi0106:
Thank you for your reply! I believe that exposing only core QUIC functionality in the `quic` package keeps the API simple and makes it easier to build extensions on top of QUIC. Therefore, I think it’s a good approach to access extension features like Datagram and Multipath via type assertions—just as you cast an `http.ResponseWriter` to an `http.Flusher` when flushing is needed. This way, the extensions are always available but only used explicitly when necessary.

marten-seemann:
As I said, I agree with your problem analysis, but I disagree with the proposed solution.

PR #5133, opened 2025-05-10
http3: migrate the connection tests away from Ginkgo

marten-seemann:
Part of #3652.


marten-seemann:
The tests are currently failing due to #5135.

Issue #5083, opened 2025-04-28
http3: client-side GOAWAY handling


marten-seemann:

marten-seemann:
This is currently blocked on https://github.com/quic-go/quic-go/pull/5096.

PR #5141, opened 2025-05-13
http3: only retry requests for which it is safe to do so

### LINKED ISSUE #4612: http3: re-dialing logic is potentially unsafe
By an error not impacting an existing stream, I mean an error that occurred before the stream was created. If the client discovers it later (i.e., at the next stream creation), it may retry without surfacing the error to the caller, but only if it is guaranteed that the request wasn’t written on the previous connection.

The existing condition to retry the request on `net.Timeout` is, in my opinion, dangerous. You have no way of knowing if the request was received by the server and then timed out, or if the connection had already timed out. This runs the risk of performing the request twice, which in some cases might be unexpected and dangerous if the request is not idempotent.

_Originally posted by @rs in https://github.com/quic-go/quic-go/issues/4587#issuecomment-2264862145_

marten-seemann:
Fixes #4612.

This PR implements the logic outlined in https://github.com/quic-go/quic-go/issues/4612#issuecomment-2875107997. It is safe to retry a request if:
* an error that happen while opening the stream: if we can't open the stream, we obviously can't have sent the request
* the server reset the request stream using a H3_REQUEST_REJECTED error

Once we implement the client-side of graceful shutdown (#5083), we can also apply this logic to all streams with stream IDs higher than the stream ID sent in the GOAWAY frame.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -32,6 +32,11 @@ const (
 	defaultMaxResponseHeaderBytes = 10 * 1 << 20 // 10 MB
 )
 
+type errConnUnusable struct{ e error }
+
+func (e *errConnUnusable) Unwrap() error { return e.e }
+func (e *errConnUnusable) Error() string { return fmt.Sprintf("http3: conn unusable: %s", e.e.Error()) }
+
 const max1xxResponses = 5 // arbitrary bound on number of informational responses
 
 var defaultQuicConfig = &quic.Config{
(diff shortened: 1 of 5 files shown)


laserproall (approved):

laserproall:
I am sorry for my mistake of misoperation. Plz ignore my approval and dismiss it. Thanks!

marten-seemann:
@laserproall Your review was neither requested nor required.

PR #5147, opened 2025-05-21
internal/handshake: avoid dependency to testing in prod build

egonelbre:


egonelbre:
I didn't measure effective impact, but I've been in general avoiding bringing in testing related packages to production code.

marten-seemann:
This PR is wrong. It breaks the integration tests.

egonelbre:
Ah, right, I didn't consider other packages depending on this behaviour. 🤦

marten-seemann:
Why does it matter anyway?

integrationtests/self/key_update_test.go
```diff
@@ -7,14 +7,18 @@ import (
...
 
 	"github.com/stretchr/testify/require"
 )
 
 func TestKeyUpdates(t *testing.T) {
-	t.Setenv("QUIC_GO_TEST_KEY_UPDATE_INTERVAL", "1") // update keys as frequently as possible
+	// update keys as frequently as possible
+	origKeyUpdateInterval := handshake.KeyUpdateInterval
+	handshake.KeyUpdateInterval = 1
+	t.Cleanup(func() { handshake.KeyUpdateInterval = origKeyUpdateInterval })
```
marten-seemann, line 21: And now you just introduced a race condition, see #5121. What are you even trying to achieve?!

egonelbre: Ok, I think I don't understand the codebase properly.

If that test doesn't run in parallel, then it shouldn't race with changing the value. Or are there leftover goroutines that aren't waited at the end of tests that race with the values?

egonelbre:
@marten-seemann yes, I'm aware that's a small difference. For me it's more of a general principle to not import testing related package into production code. The fix seemed initially trivial, but I was completely wrong and I didn't properly run the testsuite. If I'd known there are significant problems with it I probably wouldn't have started with the PR.

Yes, the binary is unrealistic, more of a basic demonstration of the size difference. The `example/echo` has a smaller difference, i.e. 54KB (at 8MB), but should be more in line.

Anyways, I'm fine to drop this.

integrationtests/self/key_update_test.go
egonelbre: Hmm, wouldn't that imply that there's still a logical race between setting the value via environment variables? But, I guess it doesn't really break anything to have it.

marten-seemann:
I'm not opposed to a different solution here, but it's complicated. Let me know if you have any other ideas.

egonelbre:
@marten-seemann the only thing I was able to think of was to use atomic.Uint64 instead of uint64. Since the logical race is there even with t.Setenv, using an atomic shouldn't make a difference to the outcome, but it'll avoid triggering the race detector.

egonelbre:
Also it's possible to use `test` build tag to conditionally compile while under testing... as far as I recall. I initially thought about this, but it adds a few two files for the different declarations.

marten-seemann:
I don't think there's a `test` build tag. I'd absolutely love to have that for a different issue, so please let me know if you find anything about that.

Issue #5148, opened 2025-05-23
Stream, SendStream, ReceiveStream should be structs, not interfaces


marten-seemann:
Similar to #4968, the stream interfaces should be converted to structs. This will allow us to add methods without breaking users.

Obviously, this change will be very painful for quic-go users (less so for HTTP/3 users), but this is a one-time cost that we'll unfortunately have to impose to arrive at a more idiomatic API.

PR #5162, opened 2025-05-29
http3: use actual QUIC connection and stream in client tests

marten-seemann:

This reduces our dependency on extensive mocking of Connection and Stream, which is needed when we convert these to structs (https://github.com/quic-go/quic-go/issues/5148 and https://github.com/quic-go/quic-go/issues/4968).

### DIFF
--- a/http3/client_test.go
+++ b/http3/client_test.go
@@ -7,12 +7,12 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
 	"github.com/quic-go/qpack"
 	"github.com/quic-go/quic-go"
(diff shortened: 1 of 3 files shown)


http3/client_test.go
```diff
@@ -473,46 +411,50 @@ func testClientGzip(t *testing.T,
...
+		t.Fatal("timeout")
+	}
+
+	require.Equal(t, http.StatusOK, rsp.StatusCode)
+	body, err := io.ReadAll(rsp.Body)
+	require.NoError(t, err)
+	require.Equal(t, expectedRsp, body)
 }
 
 func TestClientRequestCancellationBeforeHandshakeCompletion(t *testing.T) {
```
marten-seemann, line 460: Not really sure how to transform this test.

PR #5194, opened 2025-06-01
ci: stop using Ginkgo test command

### LINKED ISSUE #3652: migrate away from Ginkgo

marten-seemann:
Fixes #3652.


.github/workflows/unit.yml
```diff
@@ -17,33 +17,37 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
       - run: go version
+      - name: Remove integrationtests
+        shell: bash
+        run: rm -rf integrationtests
```
marten-seemann, line 22: This is kind of fishy, but I don't see a better way.

Issue #4405, opened 2024-04-02
http3: alternative API idea for stream hijacking


marten-seemann:
This is just a quick thought I had, and might not work at all. Instead of adding stream hijacker callbacks (both for unidirectional and bidirectional) streams, we could introduce an `Server.ServeStream` method.

The WebTransport layer would then be responsible for spawning a QUIC listener, accepting a QUIC connection, and accepting streams from that connection. It would then decide where a stream belongs (i.e. is it an HTTP/3 stream or a WebTransport stream?) and then pass HTTP/3 streams to the `Server`.
Analogously for the `RoundTripper`.

Open question: How do we send SETTINGS, and how do we handle the peer's SETTINGS and QPACK unidirectional streams?

marten-seemann:
Alternatively, we could have the http3 layer wrap a `quic.Connection`: It would grab all streams that are actual HTTP/3 streams (as identified by the signalling frame / unidirectional stream type). All other streams could then be at the upper layer via the `quic.Connection` interface.

Not sure what to do with datagrams, since the HTTP layer takes total control of both receiving and sending. This is less of a problem for `ReceiveDatagram` (would never return anything), but more so for `SendDatagram`. We could define a smaller interface (`http3.Connection`?), or just add `panic`s to these functions.

marten-seemann:
Yet another alternative, and what I believe will work best: We'll have the WebTransport layer wrap the `quic.Connection`, and grab all the streams that are WebTransport streams. We then pass this connection to the HTTP/3 layer, which handles all remaining streams. This allows us to get rid of the `StreamHijacker` and the `UniStreamHijacker`, both on the `http3.Server` and the `http3.RoundTripper`. We can also remove the `StreamCreator` interface, as well as the `FrameType`. Big API cleanup!

For the server we don't need to anything, since we already have a `ServeQUICConn`: https://github.com/quic-go/quic-go/blob/183d42a729adb8b53a57755989e07982f403c9fa/http3/server.go#L260

For the client, we'll introduce a new struct, the `SingleOriginRoundTripper`. It behaves like a normal `RoundTripper`, but it only handles a single QUIC connection. It exposes a `OpenHTTPStream(context.Context) (HTTPStream, error)` method, that, with #4409, is used to send HTTP requests.
It also exposes a function to get access to the server's SETTINGS. This mean we can also remove the `RoundTripOpt.CheckSettings` that we introduced in #4355.

[...]

PR #5219, opened 2025-06-09
http3: add ClientConn.Context, CloseWithError and Conn

marten-seemann:


http3/client.go
```diff
@@ -367,3 +378,10 @@ func (c *ClientConn) doRequest(req *http.Request, str *RequestStream) (*http.Res
 	res.Request = req
 	return res, nil
 }
+
+// Conn returns the underlying HTTP/3 connection.
+// This method is only useful for advanced use cases, such as when the application needs to
+// open streams on the HTTP/3 connection (e.g. WebTransport).
+func (c *ClientConn) Conn() *Conn {
+	return c.conn
+}
```
marten-seemann, line 387: I'm not a big fan of this API, but it's (currently) needed by WebTransport.

When we overhaul the stream hijacking API (#4405), we should revisit if this is still necessary, or if there are any better options.

PR #5224, opened 2025-06-12
ci: add Go 1.25rc1 to tested Go versions

marten-seemann:


marten-seemann:
Everything compiles, and everything seems to work as expected. Now that's unspectacular 🙃

Issue #5223, opened 2025-06-10
wire: avoid returning a wire.Frame interface


marten-seemann:
When parsing a frame, we first return a `wire.Frame` from the `wire.FrameParser`, then we immediately do a type assertion in the `Conn` when we handle the frame.

Instead, we could define a `type FrameType uint8`, and return that from the `FrameParser`. The `Conn` would then implement a `switch` on the frame type and call the `wire.ParseXXXFrame` function directly.

We should benchmark this. This is in the hot path when receiving packets, so this might actually make a difference (or it might not, who knows).

jannis-seemann:
https://github.com/quic-go/quic-go/pull/5225

I have performaned a preliminary benchmark, and it looks like this change can improve the performance by 4-5%. I will as a next step refactor the original `BenchmarkParseOtherFrames` function to switch by the type of the parsed frame (it's currently just assuming that the frame is of a specific type), and then create another larger pull request for the comprehensive changes. 

**Benchmark result:**
```
cpu: Apple M4 Max
BenchmarkParseOtherFrames-16     	11095802	       108.3 ns/op	     232 B/op	       6 allocs/op
BenchmarkParseOtherFrames-16     	11139337	       106.6 ns/op	     232 B/op	       6 allocs/op
BenchmarkParseOtherFrames2-16    	11641682	       102.1 ns/op	     232 B/op	       6 allocs/op
BenchmarkParseOtherFrames2-16    	11965707	       100.9 ns/op	     232 B/op	       6 allocs/op
PASS
```

**Benchmark code:** 

```
type FrameType uint8

func (p *FrameParser) GetNextTyp(data []byte) (FrameType, int, error) {
	typ, l, err := quicvarint.Parse(data)
	if err != nil {
		return 0, l, &qerr.TransportError{
			ErrorCode:    qerr.FrameEncodingError,
			ErrorMessage: err.Error(),
		}
	}
	return FrameType(typ), l, nil
}


func BenchmarkParseOtherFrames2(b *testing.B) {
	maxDataFrame := &MaxDataFrame{MaximumData: 123456}
	maxStreamsFrame := &MaxStreamsFrame{MaxStreamNum: 10}
	maxStreamDataFrame := &MaxStreamDataFrame{StreamID: 1337, MaximumStreamData: 1e6}
	cryptoFrame := &CryptoFrame{Offset: 1000, Data: make([]byte, 128)}
[...]

PR #5225, opened 2025-06-12
Changing the benchmark for frame parsing to use interfaces

jannis-seemann:
In `Conn`.`handleFrame`, the type of the frame is switched based on the type of the frame, and not by the index of the frame. Thus, to have this benchmark align more closely with the actual connection logic, we should refactor this benchmark.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,7 @@ linters:
     - unconvert
     - unparam
     - unused
+    - usetesting
   settings:
     depguard:
       rules:
@@ -48,6 +49,10 @@ linters:
     misspell:
       ignore-rules:
         - ect
+    # see https://github.com/ldez/usetesting/issues/10
+    usetesting:
+      context-background: false
+      context-todo: false
   exclusions:
     generated: lax
     presets:
--- a/fuzzing/internal/helper/helper_test.go
+++ b/fuzzing/internal/helper/helper_test.go
@@ -8,19 +8,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func createTempDir(t *testing.T) string {
-	t.Helper()
-	dir, err := os.MkdirTemp("", "fuzzing-helper")
-	require.NoError(t, err)
-	t.Cleanup(func() { os.RemoveAll(dir) })
-	return dir
-}
-
 func TestWriteCorpusFile(t *testing.T) {
 	const data = "lorem ipsum"
 	const expectedShaSum = "bfb7759a67daeb65410490b4d98bb9da7d1ea2ce"
 
-	dir := createTempDir(t)
+	dir := t.TempDir()
 	require.NoError(t, WriteCorpusFile(dir, []byte(data)))
 
 	path := filepath.Join(dir, expectedShaSum)
@@ -36,7 +28,7 @@ func TestWriteCorpusFileWithPrefix(t *testing.T) {
 	const expectedShaSum = "523f5cab80fab0c7889dbf50dd310ab8c8879f9c"
 	const prefixLen = 7
 
-	dir := createTempDir(t)
+	dir := t.TempDir()
 	require.NoError(t, WriteCorpusFileWithPrefix(dir, []byte(data), prefixLen))
 
 	path := filepath.Join(dir, expectedShaSum)
@@ -49,7 +41,7 @@ func TestWriteCorpusFileWithPrefix(t *testing.T) {
 }
 
 func TestCreateDirectoryIfNotExists(t *testing.T) {
-	dir := createTempDir(t)
+	dir := t.TempDir()
 	subdir := filepath.Join(dir, "corpus")
 	require.NoDirExists(t, subdir)
 
--- a/internal/utils/log_test.go
+++ b/internal/utils/log_test.go
@@ -123,8 +123,6 @@ func TestLogAddPrefixes(t *testing.T) {
 }
 
 func TestLogLevelFromEnv(t *testing.T) {
-	defer os.Unsetenv(logEnv)
-
 	testCases := []struct {
 		envValue string
 		expected LogLevel
@@ -136,13 +134,13 @@ func TestLogLevelFromEnv(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		os.Setenv(logEnv, tc.envValue)
+		t.Setenv(logEnv, tc.envValue)
 		require.Equal(t, tc.expected, readLoggingEnv())
 	}
 
 	// invalid values
-	os.Setenv(logEnv, "")
+	t.Setenv(logEnv, "")
 	require.Equal(t, LogLevelNothing, readLoggingEnv())
-	os.Setenv(logEnv, "asdf")
+	t.Setenv(logEnv, "asdf")
 	require.Equal(t, LogLevelNothing, readLoggingEnv())
 }


jannis-seemann:
I could, but now that I'm implementing the logic into the `Conn`, I'm realising the whole frame handling logic now has to live in the Conn object anyway. So to allow benchmarking of before/after https://github.com/quic-go/quic-go/issues/5223, the whole test case would need to be rewritten and run on the `Conn` instead. I'll do this now. 

In addition: We also need a comprehensive benchmark of `Conn`.`handleFrames` anyway. In my current implementation (https://github.com/quic-go/quic-go/compare/master...jannis-seemann:quic-go:FrameParsingChanges), we receive a frame that we store in a variable of the type `wire.Frame`, we just dont switch based on the interface in the function `handleFrames` anymore. However, this we then pass down to `Conn`.`handleFrame`, which then performs a switch based on the type. I'm thinking of streamlining this even more, and call the appropriate handle function directly, but it would make this function feel even more spaghetti-like, meaning I would like to run a benchmark to see if its worth it.

jannis-seemann:
I have now adjusted the benchmark, looks like the advantage is now a little bit smaller - but still 2.9% faster. 

```
goos: darwin
goarch: arm64
pkg: github.com/quic-go/quic-go/internal/wire
cpu: Apple M4 Max
                    │   old.txt   │              new.txt               │
                    │   sec/op    │   sec/op     vs base               │
ParseOtherFrames-16   1.045µ ± 0%   1.015µ ± 0%  -2.87% (p=0.000 n=50)

                    │   old.txt    │             new.txt              │
                    │     B/op     │     B/op      vs base            │
ParseOtherFrames-16   7.967Ki ± 0%   7.967Ki ± 0%  ~ (p=1.000 n=50) ¹
¹ all samples are equal

                    │  old.txt   │            new.txt             │
                    │ allocs/op  │ allocs/op   vs base            │
ParseOtherFrames-16   16.00 ± 0%   16.00 ± 0%  ~ (p=1.000 n=50) ¹
¹ all samples are equal
```


```
type FrameType uint8

func (p *FrameParser) GetNextTyp(data []byte) (FrameType, int, error) {
	typ, l, err := quicvarint.Parse(data)
	if err != nil {
		return 0, l, &qerr.TransportError{
			ErrorCode:    qerr.FrameEncodingError,
			ErrorMessage: err.Error(),
		}
	}
	return FrameType(typ), l, nil
}

func BenchmarkParseOtherFrames(b *testing.B) {
	maxDataFrame := &MaxDataFrame{MaximumData: 123456}
	maxStreamsFrame := &MaxStreamsFrame{MaxStreamNum: 10}
	maxStreamDataFrame := &MaxStreamDataFrame{StreamID: 1337, MaximumStreamData: 1e6}
	cryptoFrame := &CryptoFrame{Offset: 1000, Data: make([]byte, 128)}
	resetStreamFrame := &ResetStreamFrame{StreamID: 87654, ErrorCode: 1234, FinalSize: 1e8}
	streamFrame := &StreamFrame{
		StreamID:       1337,
		Offset:         1e7,
		Data:           make([]byte, 200),
		DataLenPresent: true,
	}
	ackFrame := &AckFrame{
		AckRanges: []AckRange{
			{Smallest: 5000, Largest: 5200},
		},
	}
	rand.Read(cryptoFrame.Data)
	frames := []Frame{
		maxDataFrame,
		maxStreamsFrame,
		maxStreamDataFrame,
		cryptoFrame,
		streamFrame,
		streamFrame,
		streamFrame,
		ackFrame,
		&PingFrame{},
		resetStreamFrame,
		streamFrame,
		streamFrame,
		ackFrame,
		ackFrame,
	}
	var buf []byte
	for i, frame := range frames {
		var err error
		buf, err = frame.Append(buf, protocol.Version1)
		if err != nil {
			b.Fatal(err)
		}

		if i == len(frames)/2 {
			// add 3 PADDING frames
			buf = append(buf, 0)
			buf = append(buf, 0)
			buf = append(buf, 0)
		}
	}

	parser := NewFrameParser(false, false)

	b.ResetTimer()
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		data := buf
		for j := 0; j < len(frames)+3; j++ {
			typ, l, err := parser.GetNextTyp(data)
			if err != nil {
				b.Fatal(err)
			}
			data = data[l:]

			if typ&0xf8 == 0x8 {
				frame, l, err := parseStreamFrame(data, uint64(typ), protocol.Version1)
				if err != nil {
					b.Fatal(err)
				}
				if frame.StreamID != streamFrame.StreamID || frame.Offset != streamFrame.Offset {
					b.Fatalf("STREAM frame does not match: %v vs %v", streamFrame, frame)
				}
				data = data[l:]
			} else {
				switch typ {
				case 0x0: // Padding frames
					continue
				case 0x1: // Ping frame
					_ = &PingFrame{}
				case 0x2, 0x3:
					ackDelayExponent := parser.ackDelayExponent
					parser.ackFrame.Reset()
					l, err = parseAckFrame(parser.ackFrame, data, uint64(typ), ackDelayExponent, protocol.Version1)
					if err != nil {
						b.Fatal(err)
					}
					data = data[l:]
					frame := parser.ackFrame
					if len(frame.AckRanges) != 1 {
						b.Fatalf("ACK frame does not match, len(AckRanges) not equal: %v vs %v", streamFrame, frame)
					}
					if frame.AckRanges[0].Smallest != ackFrame.AckRanges[0].Smallest ||
						frame.AckRanges[0].Largest != ackFrame.AckRanges[0].Largest {
						b.Fatalf("ACK frame does not match: %v vs %v", streamFrame, frame)
					}
				case 0x10:
					frame, l, err := parseMaxDataFrame(data, protocol.Version1)
					if err != nil {
						b.Fatal(err)
					}
					data = data[l:]
					if frame.MaximumData != maxDataFrame.MaximumData {
						b.Fatalf("MAX_DATA frame does not match: %v vs %v", frame, maxDataFrame)
					}
				case 0x12, 0x13:
					frame, l, err := parseMaxStreamsFrame(data, uint64(typ), protocol.Version1)
					if err != nil {
						b.Fatal(err)
					}
					data = data[l:]
					if frame.MaxStreamNum != maxStreamsFrame.MaxStreamNum {
						b.Fatalf("MAX_STREAMS frame does not match: %v vs %v", frame, maxStreamsFrame)
					}

				case 0x11:
					frame, l, err := parseMaxStreamDataFrame(data, protocol.Version1)
					if err != nil {
						b.Fatal(err)
					}
					data = data[l:]
					if frame.StreamID != maxStreamDataFrame.StreamID ||
						frame.MaximumStreamData != frame.MaximumStreamData {
						b.Fatalf("MAX_STREAM_DATA frame does not match: %v vs %v", frame, maxStreamDataFrame)
					}
				case 0x6:
					frame, l, err := parseCryptoFrame(data, protocol.Version1)
					if err != nil {
						b.Fatal(err)
					}
					data = data[l:]
					if frame.Offset != cryptoFrame.Offset || !bytes.Equal(frame.Data, cryptoFrame.Data) {
						b.Fatalf("CRYPTO frame does not match: %v vs %v", frame, cryptoFrame)
					}
				case 0x4:
					frame, l, err := parseResetStreamFrame(data, false, protocol.Version1)
					if err != nil {
						b.Fatal(err)
					}
					data = data[l:]
					if frame.StreamID != resetStreamFrame.StreamID || frame.ErrorCode != resetStreamFrame.ErrorCode ||
						frame.FinalSize != resetStreamFrame.FinalSize {
						b.Fatalf("RESET_STREAM frame does not match: %v vs %v", frame, resetStreamFrame)
					}
				default:
					b.Fatal(fmt.Errorf("unknown frame type: %v", typ))
				}
			}
		}
	}
}
```

internal/wire/frame_parser_test.go
```diff
@@ -358,32 +368,46 @@ func BenchmarkParseOtherFrames(b *testing.B) {
...
-				if f.(*MaxDataFrame).MaximumData != maxDataFrame.MaximumData {
+
+			switch frame := f.(type) {
+			case *StreamFrame:
+				if frame.StreamID != streamFrame.StreamID || frame.Offset != streamFrame.Offset {
+					b.Fatalf("STREAM frame does not match: %v vs %v", streamFrame, frame)
+				}
+			case *AckFrame:
+				if len(frame.AckRanges) != 1 {
+					b.Fatalf("ACK frame does not match, len(AckRanges) not equal: %v vs %v", streamFrame, frame)
```
```suggestion
					b.Fatalf("ACK frame does not match, len(AckRanges) not equal: %v vs %v", ackFrame, frame)
```

internal/wire/frame_parser_test.go
```diff
@@ -358,32 +368,46 @@ func BenchmarkParseOtherFrames(b *testing.B) {
...
+				}
+				if frame.AckRanges[0].Smallest != ackFrame.AckRanges[0].Smallest ||
+					frame.AckRanges[0].Largest != ackFrame.AckRanges[0].Largest {
```
marten-seemann: The struct is comparable, isn't it?
```suggestion
				if frame.AckRanges[0] != ackFrame.AckRanges[0] {
```

internal/wire/frame_parser_test.go
```diff
@@ -358,32 +368,46 @@ func BenchmarkParseOtherFrames(b *testing.B) {
...
+					b.Fatalf("ACK frame does not match: %v vs %v", streamFrame, frame)
```
```suggestion
					b.Fatalf("ACK frame does not match: %v vs %v", ackFrame, frame)
```

marten-seemann (commented):
Looks good to me, just a few minor comments.

I suggest we keep this PR open until you have the fix for #5223 ready. Please base this fix on top of this PR.

marten-seemann:
Suggestion: introduce a `benchmarkFrameParser(b *testing.B, frames []wire.Frame)`, and call that for different frame types (STREAM, ACK, DATAGRAM, other frames). Then we can exactly see which frame types #5227 improves.

You can also measure the number of allocs and assert that we really don't allocate for ACK and STREAM.

PR #5227, opened 2025-06-13
optimize parsing logic for STREAM, DATAGRAM and ACK frames

jannis-seemann:
I’m not fully finished cleaning up my code yet, but I’d appreciate an early opinion on a tradeoff I've made.

In Conn.handleFrames, I intentionally avoided using the wire.Frame interface, especially on the hot path, to improve performance. This decision required duplicating or even triplicating the frame handling logic—now there's similar logic separately for StreamFrame, AckFrame, and all other frames handled through the interface.

Given that the frame handling logic (approximately lines 1475–1499 of `connection.go`) has multiple outcomes (continue, break, error), extracting it into a separate function seems complicated and might hurt readability or performance (we would need to handle these cases outside of it anyway).

I'm not particularly happy with the resulting duplication, but other solutions seem to require using interfaces again, impacting performance negatively.

Could you take a quick preliminary look at Conn.handleFrames and share if you notice something I might have missed?


connection.go
```diff
@@ -1431,21 +1430,175 @@ func (c *Conn) handleFrames(
...
 			break
 		}
-		if ackhandler.IsFrameAckEliciting(frame) {
+
+		if !frameType.IsAllowedAtEncLevel(encLevel) {
+			return false, false, nil, &qerr.TransportError{
+				ErrorCode:    qerr.FrameEncodingError,
+				ErrorMessage: fmt.Sprintf("%d not allowed at encryption level %s", frameType, encLevel),
+			}
+		}
```
marten-seemann: This should be in the frame parser.

connection.go
```diff
@@ -1431,21 +1430,175 @@ func (c *Conn) handleFrames(
...
+			err = c.streamsMap.HandleStreamFrame(streamFrame, rcvTime)
+			if err != nil {
+				if log == nil {
+					return false, false, nil, err
+				}
+				// If we're logging, we need to keep parsing (but not handling) all frames.
+				handleErr = err
+			}
+			continue
+		} else {
```
marten-seemann: No need for an `else`, you're already `continue`ing in the `if`.

connection.go
```diff
@@ -1431,21 +1430,175 @@ func (c *Conn) handleFrames(
...
+			switch frameType {
+			case wire.AckFrameType, wire.AckECNFrameType:
+				ackDelayExponent := c.frameParser.AckDelayExponent
```
marten-seemann: I don't like accessing `AckDelayExponent` here. What about adding a `wire.FrameParser.ParseAckFrame` method?

connection.go
```diff
@@ -1431,21 +1430,175 @@ func (c *Conn) handleFrames(
...
+				if encLevel != protocol.Encryption1RTT {
+					ackDelayExponent = protocol.DefaultAckDelayExponent
+				}
+				c.frameParser.AckFrame.Reset()
```
marten-seemann: Don't expose the `AckFrame`. We can just make this part of the API contract that the consumer of the ACK frame must not hold onto it the next time `ParseAckFrame` is called. This is what we already have.

I know it's not the most intuitive API and there's potential for misuse, but 1. this is an internal package, so we can be sure that it's only use using that API and 2. the performance benefits to not allocating for every ACK (which is sent frequently!) justify that.

connection.go
```diff
@@ -1431,21 +1430,175 @@ func (c *Conn) handleFrames(
...
+				if err != nil {
+					return wrapError(err)
+				}
+
+				data = data[l:]
+				/*
+					if ackFrame == nil {
+						break
+					}
+				*/
```
marten-seemann: no commented code

connection.go
```diff
@@ -1431,21 +1430,175 @@ func (c *Conn) handleFrames(
...
+				frame = &wire.PingFrame{}
+				l = 0
+			case wire.ResetStreamFrameType:
+				frame, l, err = wire.ParseResetStreamFrame(data, false, c.version)
+			case wire.StopSendingFrameType:
+				frame, l, err = wire.ParseStopSendingFrame(data, c.version)
+			case wire.CryptoFrameType:
+				frame, l, err = wire.ParseCryptoFrame(data, c.version)
+			case wire.NewTokenFrameType:
+				frame, l, err = wire.ParseNewTokenFrame(data, c.version)
```
marten-seemann: You can just call `handleNewTokenFrame` here (and not in `handleFrame`). This comes at the cost of adding an error check to every `case`, but that's a small price to pay to have only a single `switch` statement.

internal/wire/frame.go
```diff
@@ -19,3 +19,15 @@ func IsProbingFrame(f Frame) bool {
 	}
 	return false
 }
+
+// IsProbingFrameType returns true if the FrameType is a probing frame.
+// See section 9.1 of RFC 9000.
+// TODO: Add tests
+func IsProbingFrameType(f FrameType) bool {
+	switch f {
```
marten-seemann, line 27:
```suggestion
        //nolint:exhaustive // PATH_CHALLENGE, PATH_RESPONSE and NEW_CONNECTION_ID are the only probing frames
	switch f {
```

connection.go
```diff
@@ -1431,21 +1430,175 @@ func (c *Conn) handleFrames(
...
-		l, frame, err := c.frameParser.ParseNext(data, encLevel, c.version)
+		frameType, l, err = c.frameParser.ParseType(data)
 		if err != nil {
 			return false, false, nil, err
 		}
 		data = data[l:]
-		if frame == nil {
+
+		// Here, frameType == 0 does not equal the padding frame, ParseType only returns a 0 if it did not find a frame.
+		// This is in the hot path, I did not want to introduce another variable for this.
```
marten-seemann: This happens because you chose to use the frame type defined in RFC 9000 for the `FrameType` enum. What about just using an `iota`?

Also, why is this check needed in the first place?

jannis-seemann: In that case, we would need a separate function to turn the byte into a proper FrameType. This can be avoided now. However, it now may happen that FrameType contains a value that is not listed in the enum. 

Also, this is needed, as if the FrameParser only parses padding frames, or no frames at all, frameType would be equal to 0.

jannis-seemann:
So far, it already looks quite good. I like how it turned out with 3 different hot paths, and a fallback path for all other frame types. I believe this is a good tradeoff between performance and readability. 

**TODO for me:**
- remove ParseFrame and parseFrame
- Adjusts tests
- Write tests for FrameType
- Evaluate if we want to turn the FrameType into an iota, so that it doesn't map the RFC9000 types for frames anymore. But then, we need to specify all the types for all possible StreamFrames, they are currently not listed in the enum at all (so it's not really an enum yet). Also, byte(frameType) may still be used in the code, we need to check and refactor this then accordingly. 
- Evaluate what we do with `ErrUnknownFrameType`, where to place it
- Go through all the changes, and double check everything

jannis-seemann:
@marten-seemann Please review.

The final results are in:

```
goos: darwin
goarch: arm64
pkg: github.com/quic-go/quic-go/internal/wire
cpu: Apple M4 Max
               │   old.out   │              new.out               │
               │   sec/op    │   sec/op     vs base               │
ParseFrames-16   1.042µ ± 2%   1.020µ ± 2%  -2.11% (p=0.000 n=10)

               │   old.out    │             new.out              │
               │     B/op     │     B/op      vs base            │
ParseFrames-16   7.967Ki ± 0%   7.967Ki ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

               │  old.out   │            new.out             │
               │ allocs/op  │ allocs/op   vs base            │
ParseFrames-16   16.00 ± 0%   16.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

connection.go
```diff
@@ -1431,39 +1430,152 @@ func (c *Conn) handleFrames(
...
+	var err error
+	var l int
+	var frameType wire.FrameType
+
+	wrapError := func(err error) (isAckEliciting, isNonProbing bool, pathChallenge *wire.PathChallengeFrame, _ error) {
+		return false, false, nil, &qerr.TransportError{
+			ErrorCode:    qerr.FrameEncodingError,
+			ErrorMessage: err.Error(),
+		}
+	}
```
marten-seemann: 1. What's the deal with these many return values? I would've expected a `wrapError` function to have the signature `func(error) error`.
2. I'd prefer to not deal with this in the connection at all.

Instead, how about not exporting `wire.ParseStreamFrame`, but instead exposing a `FrameParser.ParseStreamFrame`. This function would then take care to return the correct `qerr.TransportError`. Makes it a lot easier to test, removes some complexity from the `Conn`, and (probably) compiles down to the exact same thing.

internal/ackhandler/ack_eliciting.go
```diff
@@ -2,6 +2,20 @@ package ackhandler
 
 import "github.com/quic-go/quic-go/internal/wire"
 
+// IsFrameTypeAckEliciting returns true if the frame is ack-eliciting.
+// TODO: Add tests
```
marten-seemann: Yes please :)

marten-seemann (commented):
Good progress!

I added some comments. If you introduce a `wire.FrameParser.Parse{Stream,Ack,Datagram}Frame` function, you can avoid exporting _any_ functions to parse individual frame (i.e. you can revert all your changes to `wire/*_frame.go` files. What do you think?

jannis-seemann:
So, I performed these changes, but we now lack the performance improvement that we were looking in the first place.

Thus, unless you see something that I'm missing, I need to conclude that the new version is not faster than the old, and we should not merge this PR. 

```
goos: darwin
goarch: arm64
pkg: github.com/quic-go/quic-go/internal/wire
cpu: Apple M4 Max
               │   old.out   │            new.out            │
               │   sec/op    │   sec/op     vs base          │
ParseFrames-16   1.060µ ± 1%   1.061µ ± 0%  ~ (p=0.868 n=10)

               │   old.out    │             new.out              │
               │     B/op     │     B/op      vs base            │
ParseFrames-16   7.967Ki ± 0%   7.967Ki ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

               │  old.out   │            new.out             │
               │ allocs/op  │ allocs/op   vs base            │
ParseFrames-16   16.00 ± 0%   16.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

jannis-seemann:
Turns out we didn't benchmark our code, most of the time was spend in the garbage collector, as we didn't put back the StreamFrame to be released. I'll now further investigate this.

jannis-seemann:
So, I have now debugged this, and added a few more benchmarks. The idea is that now we have benchmarks for each type, and also another test to ensure that we have 0 allocations for AckFrame and StreamFrame. 

Here are the results. While the new code is 5% slower for everything but StreamFrame / AckFrame / DatagramFrame, it is 5% faster for StreamFrame and for the AckFrame. Given the high occurence of these frame types, I believe the PR is worth it :)

```
goos: darwin
goarch: arm64
pkg: github.com/quic-go/quic-go/internal/wire
cpu: Apple M4 Max
                              │   old.out   │              new.out               │
                              │   sec/op    │   sec/op     vs base               │
ParseExtendedHeader-16          42.45n ± 0%   42.52n ± 1%       ~ (p=0.119 n=20)
ParseStreamAndACK-16            46.89n ± 0%   44.34n ± 0%  -5.43% (p=0.000 n=20)
ParseOtherFrames-16             112.8n ± 0%   117.8n ± 1%  +4.43% (p=0.000 n=20)
ParseAckFrame-16                263.2n ± 0%   249.3n ± 0%  -5.30% (p=0.000 n=20)
ParseStreamFrame-16             189.4n ± 0%   180.4n ± 0%  -4.73% (p=0.000 n=20)
ParseDatagramFrame-16           440.1n ± 1%   437.1n ± 1%  -0.66% (p=0.003 n=20)
ParseInitial/without_token-16   30.43n ± 0%   30.35n ± 1%       ~ (p=0.673 n=20)
ParseInitial/with_token-16      38.27n ± 0%   38.29n ± 0%       ~ (p=0.995 n=20)
ParseRetry-16                   36.61n ± 0%   36.80n ± 0%  +0.51% (p=0.015 n=20)
geomean                         84.08n        82.99n       -1.30%

                              │    old.out     │                new.out                │
                              │      B/op      │     B/op      vs base                 │
ParseExtendedHeader-16            224.0 ± 0%       224.0 ± 0%       ~ (p=1.000 n=20) ¹
ParseStreamAndACK-16              0.000 ± 0%       0.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseOtherFrames-16               232.0 ± 0%       232.0 ± 0%       ~ (p=1.000 n=20) ¹
ParseAckFrame-16                  0.000 ± 0%       0.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseStreamFrame-16               0.000 ± 0%       0.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseDatagramFrame-16           2.344Ki ± 0%     2.344Ki ± 0%       ~ (p=1.000 n=20) ¹
ParseInitial/without_token-16     96.00 ± 0%       96.00 ± 0%       ~ (p=1.000 n=20) ¹
ParseInitial/with_token-16        128.0 ± 0%       128.0 ± 0%       ~ (p=1.000 n=20) ¹
ParseRetry-16                     144.0 ± 0%       144.0 ± 0%       ~ (p=1.000 n=20) ¹
geomean                                      ²                 +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                              │   old.out    │               new.out               │
                              │  allocs/op   │ allocs/op   vs base                 │
ParseExtendedHeader-16          2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseStreamAndACK-16            0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseOtherFrames-16             6.000 ± 0%     6.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseAckFrame-16                0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseStreamFrame-16             0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseDatagramFrame-16           20.00 ± 0%     20.00 ± 0%       ~ (p=1.000 n=20) ¹
ParseInitial/without_token-16   1.000 ± 0%     1.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseInitial/with_token-16      2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=20) ¹
ParseRetry-16                   2.000 ± 0%     2.000 ± 0%       ~ (p=1.000 n=20) ¹
geomean                                    ²               +0.00%                ²
```

marten-seemann:
> While the new code is 5% slower for everything but StreamFrame / AckFrame / DatagramFrame, it is 5% faster for StreamFrame and for the AckFrame

Totally! Optimizing for the common frames is exactly what we should do!

Have you figured out why parsing the other frames got slower though? What is the additional operation that's costs 5% CPU cyles?

internal/wire/frame_type_test.go
```diff
@@ -0,0 +1,60 @@
...
+			t.Errorf("FrameType(0x%x).IsStreamFrameType() = false, want true", i)
+		}
+	}
+
+	if FrameType(0x1).IsStreamFrameType() {
+		t.Errorf("FrameType(0x1).IsStreamFrameType() = true, want false")
+	}
+}
+
+func TestIsAllowedAtEncLevel(t *testing.T) {
```
marten-seemann: Don't test private methods. They're an implementation details. This should be tested by giving the frame parser various frames to parse, and observing what the public API returns.

internal/wire/frame_type_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"testing"
+
+	"github.com/quic-go/quic-go/internal/protocol"
+)
+
+func TestIsStreamFrameType(t *testing.T) {
+	for i := 0x08; i <= 0x0f; i++ {
+		if !FrameType(i).IsStreamFrameType() {
+			t.Errorf("FrameType(0x%x).IsStreamFrameType() = false, want true", i)
```
marten-seemann: Use `require`.

jannis-seemann:
I have added test cases on the frame parser, to check if a frame is allowed at specific enc levels: `TestFrameAllowedAtEncLevelMatrix`. This test is quite large though, and we may even argue that this is partly already tested on the connection, but I believe, testing it directly on the FrameParser may still be quite beneficial. Let me know what you think.

Also, the Benchmark for ParseOtherFrames is meaningless, I believe we're mostly benchmarking the garbage collector:this is alright, but still, the benchmark itself then at least somewhat looses its meaning:

Screenshot 2025-07-08 at 13 50 18

internal/wire/frame_parser.go
```diff
@@ -59,20 +32,16 @@ func NewFrameParser(supportsDatagrams, supportsResetStreamAt bool) *FrameParser
...
-// It skips PADDING frames.
-func (p *FrameParser) ParseNext(data []byte, encLevel protocol.EncryptionLevel, v protocol.Version) (int, Frame, error) {
-	frame, l, err := p.parseNext(data, encLevel, v)
-	return l, frame, err
-}
-
-func (p *FrameParser) parseNext(b []byte, encLevel protocol.EncryptionLevel, v protocol.Version) (Frame, int, error) {
+// ParseType retrieves the next non-null byte, assuming it represents a frame type.
+// For performance reasons (1–2% faster parsing), it skips validation.
+// Validation is expected to be handled at the connection level.
```
marten-seemann: No!

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
 	}
 	handshakeWasComplete := c.handshakeComplete
 	var handleErr error
+	var frame wire.Frame
+	var err error
+	var l int
+	var frameType wire.FrameType
```
marten-seemann: I don't think these need to be declared here.

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
 			isNonProbing = true
 		}
-		if log != nil {
-			frames = append(frames, toLoggingFrame(frame))
-		}
-		// An error occurred handling a previous frame.
-		// Don't handle the current frame.
-		if handleErr != nil {
+
+		// Fast path: We're inlining common cases, to avoid using interfaces
```
marten-seemann:
```suggestion
		// Fast path: STREAM, DATAGRAM and ACK
```

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+		if frameType.IsStreamFrameType() {
+			streamFrame, l, err := wire.ParseStreamFrame(data, frameType, c.version)
+			if err != nil {
+				return false, false, nil, err
+			}
+
+			data = data[l:]
+			if streamFrame == nil {
+				break
+			}
```
marten-seemann: Why would that happen? `ParseStreamFrame` should always return a non-nil frame if `err == nil`.

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+			if log != nil {
+				frames = append(frames, toLoggingFrame(streamFrame))
+			}
+			// An error occurred handling a previous frame.
+			// Don't handle the current frame.
+			if handleErr != nil {
+				continue
+			}
+			err = c.streamsMap.HandleStreamFrame(streamFrame, rcvTime)
+			if err != nil {
```
marten-seemann: Nit:
```suggestion
			if err ;= c.streamsMap.HandleStreamFrame(streamFrame, rcvTime); err != nil {
```

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
 		}
-		pc, err := c.handleFrame(frame, encLevel, destConnID, rcvTime)
-		if err != nil {
-			if log == nil {
+		//nolint:exhaustive // StreamFrame has already been handled. We now need to handle a few more hot paths,
+		// the rest will be handled by the default clause.
+		switch frameType {
+		case wire.AckFrameType, wire.AckECNFrameType:
+			ackFrame, l, err := c.frameParser.ParseAckFrame(frameType, data, encLevel, c.version)
+			// Fast path: We inline the frame handling logic, to avoid using interfaces
```
marten-seemann:
```suggestion
```

fuzzing/frames/fuzz.go
```diff
@@ -41,15 +42,22 @@ func Fuzz(data []byte) int {
 	var b []byte
 	for len(data) > 0 {
 		initialLen := len(data)
-		l, f, err := parser.ParseNext(data, encLevel, version)
+		frameType, l, err := parser.ParseType(data, encLevel)
 		if err != nil {
+			if err == io.EOF { // PADDING frame
```
marten-seemann: `EOF` doesn't mean PADDING frame, it means we've reached the end of the slice!

```suggestion
			if err == io.EOF {
```

fuzzing/frames/fuzz.go
```diff
@@ -41,15 +42,22 @@ func Fuzz(data []byte) int {
...
+			}
 			break
 		}
+
 		data = data[l:]
 		numFrames++
-		if f == nil { // PADDING frame
-			continue
+
+		f, l, err := parser.ParseLessCommonFrame(frameType, data, protocol.Version1)
```
marten-seemann: Do you need logic here to parse STREAM frames?

internal/wire/frame_parser.go
```diff
@@ -82,114 +51,115 @@ func (p *FrameParser) parseNext(b []byte, encLevel protocol.EncryptionLevel, v p
...
-		return f, parsed, nil
+
+		return FrameType(typ), parsed, nil
 	}
-	return nil, parsed, nil
+	return 0, parsed, io.EOF
 }
 
-func (p *FrameParser) parseFrame(b []byte, typ uint64, encLevel protocol.EncryptionLevel, v protocol.Version) (Frame, int, error) {
+// ParseLessCommonFrame parses everything except StreamFrame, AckFrame or DatagramFrame.
```
marten-seemann:
```suggestion
// ParseLessCommonFrame parses everything except STREAM, ACK or DATAGRAM.
```

internal/wire/frame_parser.go
```diff
@@ -82,114 +51,115 @@ func (p *FrameParser) parseNext(b []byte, encLevel protocol.EncryptionLevel, v p
...
-	}
-	if !p.isAllowedAtEncLevel(frame, encLevel) {
-		return nil, l, fmt.Errorf("%s not allowed at encryption level %s", reflect.TypeOf(frame).Elem().Name(), encLevel)
+		return frame, l, &qerr.TransportError{
+			ErrorCode:    qerr.FrameEncodingError,
+			FrameType:    uint64(frameType),
+			ErrorMessage: err.Error(),
+		}
 	}
-	return frame, l, nil
```
marten-seemann, line 169: You need to explicitly handle the error here. For example, parsing a STREAM frame will return `io.EOF` if the STREAM frame contains less data than the `Length` value indicates. This `io.EOF` must not be confused with reaching the end of the packet.

It should be sufficient to add a `if err != nil` here.

internal/wire/frame_parser.go
```diff
@@ -82,114 +51,115 @@ func (p *FrameParser) parseNext(b []byte, encLevel protocol.EncryptionLevel, v p
...
+	ackDelayExponent := p.ackDelayExponent
+	if encLevel != protocol.Encryption1RTT {
+		ackDelayExponent = protocol.DefaultAckDelayExponent
+	}
+	p.ackFrame.Reset()
+	l, err := parseAckFrame(p.ackFrame, data, frameType, ackDelayExponent, v)
+	ackFrame := p.ackFrame
+
+	if err != nil {
+		return ackFrame, l, &qerr.TransportError{
```
marten-seemann:
```suggestion
		return nil, l, &qerr.TransportError{
```

internal/wire/frame_parser.go
```diff
@@ -82,114 +51,115 @@ func (p *FrameParser) parseNext(b []byte, encLevel protocol.EncryptionLevel, v p
...
 		}
-	case protocol.Encryption0RTT:
-		switch f.(type) {
-		case *CryptoFrame, *AckFrame, *ConnectionCloseFrame, *NewTokenFrame, *PathResponseFrame, *RetireConnectionIDFrame:
-			return false
-		default:
-			return true
+	}
+
+	return ackFrame, l, err
```
marten-seemann:
```suggestion
	return ackFrame, l, nil
```

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+			datagramFrame, l, err := c.frameParser.ParseDatagramFrame(frameType, data, c.version)
+			// Fast path: We inline the frame handling logic, to avoid using interfaces
+			if err != nil {
+				return false, false, nil, err
+			}
+
+			data = data[l:]
+			if datagramFrame == nil {
+				break
+			}
```
marten-seemann: Same here.

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+			if log != nil {
+				frames = append(frames, toLoggingFrame(datagramFrame))
+			}
+			// An error occurred handling a previous frame.
+			// Don't handle the current frame.
+			if handleErr != nil {
+				continue
+			}
+			err = c.handleDatagramFrame(datagramFrame)
+			if err != nil {
```
marten-seemann:
```suggestion
			if err := c.handleDatagramFrame(datagramFrame); err != nil {
```

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+		default:
+			frame, l, err = c.frameParser.ParseLessCommonFrame(frameType, data, c.version)
+			if err != nil {
+				return false, false, nil, err
+			}
+
+			data = data[l:]
+			if frame == nil {
+				break
+			}
```
marten-seemann: And here.

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
 			continue
 		}
-		pc, err := c.handleFrame(frame, encLevel, destConnID, rcvTime)
-		if err != nil {
-			if log == nil {
+		//nolint:exhaustive // StreamFrame has already been handled. We now need to handle a few more hot paths,
+		// the rest will be handled by the default clause.
+		switch frameType {
+		case wire.AckFrameType, wire.AckECNFrameType:
```
marten-seemann: It's probably nicer to introduce a `wire.IsAckFrame`, similar to the `wire.IsStreamFrame` you defined.

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+			err = c.handleAckFrame(ackFrame, encLevel, rcvTime)
+			if err != nil {
+				if log == nil {
+					return false, false, nil, err
+				}
+				// If we're logging, we need to keep parsing (but not handling) all frames.
+				handleErr = err
+			}
+			continue
+		case wire.DatagramNoLengthFrameType, wire.DatagramWithLengthFrameType:
```
marten-seemann: It's probably nicer to introduce a `wire.IsDatagramFrame`, similar to the `wire.IsStreamFrame` you defined.

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+			datagramFrame, l, err := c.frameParser.ParseDatagramFrame(frameType, data, c.version)
+			// Fast path: We inline the frame handling logic, to avoid using interfaces
```
marten-seemann:
```suggestion
```

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+			if log != nil {
+				frames = append(frames, toLoggingFrame(ackFrame))
+			}
+			// An error occurred handling a previous frame.
+			// Don't handle the current frame.
+			if handleErr != nil {
+				continue
+			}
+			err = c.handleAckFrame(ackFrame, encLevel, rcvTime)
+			if err != nil {
```
marten-seemann:
```suggestion
			if err := c.handleAckFrame(ackFrame, encLevel, rcvTime); err != nil {
```

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
+	var frame wire.Frame
+	var err error
+	var l int
+	var frameType wire.FrameType
+
 	for len(data) > 0 {
-		l, frame, err := c.frameParser.ParseNext(data, encLevel, c.version)
+		frameType, l, err = c.frameParser.ParseType(data, encLevel)
 		if err != nil {
+			if err == io.EOF {
```
marten-seemann, line 1447:
```suggestion
                        // The frame parser skips over PADDING frames, and returns an io.EOF if the PADDING
                        // frames were the last frames in this packet.
			if err == io.EOF {
```

marten-seemann (commented):
Good work, just a few comments.

connection.go
```diff
@@ -1437,39 +1436,145 @@ func (c *Conn) handleFrames(
...
-		if log != nil {
-			frames = append(frames, toLoggingFrame(frame))
-		}
-		// An error occurred handling a previous frame.
-		// Don't handle the current frame.
-		if handleErr != nil {
+
+		// Fast path: We're inlining common cases, to avoid using interfaces
+		if frameType.IsStreamFrameType() {
+			streamFrame, l, err := wire.ParseStreamFrame(data, frameType, c.version)
```
marten-seemann: This is problematic, since the error is not properly wrapped in a `TransportError`. What you could do instead is define a `wire.FrameParser.ParseStreamFrame` function that performs that error wrapping.

internal/wire/frame_type.go
```diff
@@ -0,0 +1,65 @@
...
+	PathChallengeFrameType      FrameType = 0x1a
+	PathResponseFrameType       FrameType = 0x1b
+	ConnectionCloseFrameType    FrameType = 0x1c
+	ApplicationCloseFrameType   FrameType = 0x1d
+	HandshakeDoneFrameType      FrameType = 0x1e
+	ResetStreamAtFrameType      FrameType = 0x24 // https://datatracker.ietf.org/doc/draft-ietf-quic-reliable-stream-reset/06/
+
+	DatagramNoLengthFrameType   FrameType = 0x30
+	DatagramWithLengthFrameType FrameType = 0x31
+)
```
marten-seemann: If you introduce the `IsDatagramFrame` and `IsAckFrame` functions as I suggested in my last review, you won't need to expose these.

connection.go
jannis-seemann: Yes, but this would be a function call, that would add CPU overhead. Here, having this case for the AckFrameType / AckECNFRameType does (should) not slow down even the other cases in any way.

connection.go
jannis-seemann: as seen above - this would introduce a function call and slow down the switching

internal/wire/frame_type.go
jannis-seemann: As seen above

connection.go
marten-seemann: I don't think so. This function would be small enough to be inlined.

fuzzing/frames/fuzz.go
marten-seemann: Actually, it does, so this comment might be more helpful.
```suggestion
			if err == io.EOF { // the last frame was a PADDING frame
```

internal/wire/frame_parser.go
marten-seemann:
```suggestion
// ParseType parses the frame type of the next frame.
// It skips over PADDING frames.
```

internal/wire/frame_type.go
```diff
@@ -0,0 +1,111 @@
...
+		RetireConnectionIDFrameType,
+		PathChallengeFrameType,
+		PathResponseFrameType,
+		ConnectionCloseFrameType,
+		ApplicationCloseFrameType,
+		HandshakeDoneFrameType,
+		ResetStreamAtFrameType,
+		DatagramNoLengthFrameType,
+		DatagramWithLengthFrameType:
+		return true
```
marten-seemann: This can be optimized by observing that the QUIC frame types are a contiguous range from 0x0 to 0x1e (see https://datatracker.ietf.org/doc/html/rfc9000#frame-types). The logic would then be:
```go
// frames defined in RFC 9000 use all values between 0x0 and 0x1e
return ft < 0x1e || ft == DatagramNoLengthFrameType || ft == DatagramWithLengthFrameType || ft == ResetStreamAtFrameType
```
This will be blazing fast.

internal/wire/frame_type_test.go
```diff
@@ -0,0 +1,29 @@
...
+	require.True(t, AckECNFrameType.IsAckFrameType(), "AckECNFrameType should be recognized as ACK")
+	require.False(t, PingFrameType.IsAckFrameType(), "PingFrameType should not be recognized as ACK")
+	require.False(t, FrameType(0x10).IsAckFrameType(), "MaxDataFrameType should not be recognized as ACK")
+}
+
+func TestIsDatagramFrameType(t *testing.T) {
+	require.True(t, DatagramNoLengthFrameType.IsDatagramFrameType(), "DatagramNoLengthFrameType should be recognized as datagram")
+	require.True(t, DatagramWithLengthFrameType.IsDatagramFrameType(), "DatagramWithLengthFrameType should be recognized as datagram")
+	require.False(t, PingFrameType.IsDatagramFrameType(), "PingFrameType should not be recognized as datagram")
+	require.False(t, FrameType(0x1e).IsDatagramFrameType(), "HandshakeDoneFrameType should not be recognized as datagram")
```
marten-seemann:
```suggestion
	require.True(t, DatagramNoLengthFrameType.IsDatagramFrameType(), "DatagramNoLengthFrameType should be recognized as DATAGRAM")
	require.True(t, DatagramWithLengthFrameType.IsDatagramFrameType(), "DatagramWithLengthFrameType should be recognized as DATAGRAM")
	require.False(t, PingFrameType.IsDatagramFrameType(), "PingFrameType should not be recognized as DATAGRAM")
	require.False(t, FrameType(0x1e).IsDatagramFrameType(), "HandshakeDoneFrameType should not be recognized as DATAGRAM")
```

packet_packer_test.go
```diff
@@ -774,10 +781,15 @@ func TestPackShortHeaderPadToAtLeast4Bytes(t *testing.T) {
 
 	// ... followed by the STREAM frame
 	frameParser := wire.NewFrameParser(false, false)
-	frameLen, frame, err := frameParser.ParseNext(payload[1:], protocol.Encryption1RTT, protocol.Version1)
+	frameType, l, err := frameParser.ParseType(payload[1:], protocol.Encryption1RTT)
+	require.NoError(t, err)
+	require.Equal(t, 1, l)
+	require.Equal(t, wire.FrameType(0x9), frameType)
```
marten-seemann:
```suggestion
	require.True(t, wire.IsStreamFrameType(frameType))
```

packet_packer_test.go
```diff
@@ -736,10 +736,17 @@ func TestPackLongHeaderPadToAtLeast4Bytes(t *testing.T) {
 	require.Equal(t, []byte{0, 0}, data[:2])
 	// ...followed by the PING frame
 	frameParser := wire.NewFrameParser(false, false)
-	l, frame, err := frameParser.ParseNext(data[2:], protocol.EncryptionHandshake, protocol.Version1)
+
+	frameType, lt, err := frameParser.ParseType(data[2:], protocol.EncryptionHandshake)
+	require.NoError(t, err)
+	require.Equal(t, 1, lt)
+	require.Equal(t, wire.PingFrameType, frameType)
+
```
marten-seemann: Don't really need to assert that here, we're already asserting that we're getting a PING frame just below.
```suggestion
```

packet_packer_test.go
```diff
@@ -774,10 +781,15 @@ func TestPackShortHeaderPadToAtLeast4Bytes(t *testing.T) {
...
+	frameType, l, err := frameParser.ParseType(payload[1:], protocol.Encryption1RTT)
+	require.NoError(t, err)
+	require.Equal(t, 1, l)
+	require.Equal(t, wire.FrameType(0x9), frameType)
+
+	frame, frameLen, err := wire.ParseStreamFrame(payload[1+l:], frameType, protocol.Version1)
 	require.NoError(t, err)
 	require.Equal(t, f, frame)
-	require.Equal(t, len(payload)-1, frameLen)
+	require.Equal(t, len(payload)-2, frameLen)
```
marten-seemann, line 790: Why is this changing?

internal/wire/frame_parser_test.go
```diff
@@ -202,30 +325,378 @@ func TestFrameParserFrames(t *testing.T) {
...
 
-func checkFrameUnsupported(t *testing.T, err error, expectedFrameType uint64) {
-	t.Helper()
-	require.ErrorContains(t, err, errUnknownFrameType.Error())
-	var transportErr *qerr.TransportError
-	require.ErrorAs(t, err, &transportErr)
-	require.Equal(t, qerr.FrameEncodingError, transportErr.ErrorCode)
-	require.Equal(t, expectedFrameType, transportErr.FrameType)
-	require.Equal(t, "unknown frame type", transportErr.ErrorMessage)
+func TestFrameAllowedAtEncLevelMatrix(t *testing.T) {
```
marten-seemann: This test looks AI-generated, and definitely not worth 300 LOC.

internal/wire/frame_parser.go
```diff
@@ -82,114 +51,134 @@ func (p *FrameParser) parseNext(b []byte, encLevel protocol.EncryptionLevel, v p
 			continue
 		}
 
-		f, l, err := p.parseFrame(b, typ, encLevel, v)
-		parsed += l
-		if err != nil {
-			return nil, parsed, &qerr.TransportError{
+		frameType := FrameType(typ)
```
marten-seemann: With the improved validation logic, you can verify that the frame type is valid here.

internal/wire/frame_parser_test.go
```diff
@@ -252,139 +740,302 @@ func TestFrameParsingErrorsOnInvalidFrames(t *testing.T) {
...
-	data, err := ack.Append([]byte{}, protocol.Version1)
-	if err != nil {
-		b.Fatal(err)
-	}
-	data, err = sf.Append(data, protocol.Version1)
-	if err != nil {
-		b.Fatal(err)
+func evaluateFrames(tb testing.TB, parser *FrameParser, buf []byte, frames ...Frame) {
+	data := buf
+	for j := 0; j < len(frames); j++ {
```
marten-seemann: `for f := range frames`?

internal/wire/frame_parser_test.go
```diff
@@ -252,139 +740,302 @@ func TestFrameParsingErrorsOnInvalidFrames(t *testing.T) {
...
+}
 
-	data, err := ack.Append([]byte{}, protocol.Version1)
-	if err != nil {
-		b.Fatal(err)
-	}
-	data, err = sf.Append(data, protocol.Version1)
-	if err != nil {
-		b.Fatal(err)
+func evaluateFrames(tb testing.TB, parser *FrameParser, buf []byte, frames ...Frame) {
```
marten-seemann: Is there a reason this function doesn't use `require`?

internal/wire/frame_parser_test.go
```diff
@@ -252,139 +740,302 @@ func TestFrameParsingErrorsOnInvalidFrames(t *testing.T) {
...
+			case MaxDataFrameType:
+				maxDataFrame, ok := expectedFrame.(*MaxDataFrame)
+				if !ok {
+					tb.Fatalf("Expected MaxDataFrame but got %v", expectedFrame)
+				}
+
+				frame := f.(*MaxDataFrame)
+				if frame.MaximumData != maxDataFrame.MaximumData {
+					tb.Fatalf("MAX_DATA frame does not match: %v vs %v", f, maxDataFrame)
+				}
```
marten-seemann: Can you just compare frames with `==`?

internal/wire/frame_test.go
```diff
@@ -27,3 +27,43 @@ func TestProbingFrames(t *testing.T) {
 		require.Equal(t, expected, IsProbingFrame(f))
 	}
 }
+
+func TestIsProbingFrameType(t *testing.T) {
```
marten-seemann, line 31: This also looks too verbose, have a look at the tests in the ackhandler package for how to make this more compact.

internal/wire/frame_parser_test.go
```diff
@@ -202,30 +325,378 @@ func TestFrameParserFrames(t *testing.T) {
...
+	frameType, l, err := parser.ParseType(b, protocol.Encryption1RTT)
+	require.NoError(t, err)
+	require.Equal(t, DatagramNoLengthFrameType, frameType)
+	require.Equal(t, 1, l)
+
+	// ParseLessCommonFrame should not be used to handle Datagram Frames
+	frame, lw, err := parser.ParseLessCommonFrame(frameType, b[l:], protocol.Version1)
+	checkFrameUnsupported(t, err, 0x30)
+	require.Nil(t, frame)
+	require.Zero(t, lw)
```
marten-seemann: No need to check any other return value than `err`.

internal/wire/frame_parser_test.go
```diff
@@ -252,139 +740,302 @@ func TestFrameParsingErrorsOnInvalidFrames(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, MaxStreamDataFrameType, frameType)
+	require.Equal(t, 1, l)
+
+	frame, _, err := parser.ParseLessCommonFrame(frameType, b[1:len(b)-2], protocol.Version1)
 	require.Error(t, err)
 	var transportErr *qerr.TransportError
 	require.ErrorAs(t, err, &transportErr)
 	require.Equal(t, qerr.FrameEncodingError, transportErr.ErrorCode)
+	require.Nil(t, frame)
```
marten-seemann: Same here, no need to check `frame`.

internal/wire/frame_parser_test.go
```diff
@@ -252,139 +740,302 @@ func TestFrameParsingErrorsOnInvalidFrames(t *testing.T) {
...
-		Offset:         1e7,
-		Data:           make([]byte, 200),
-		DataLenPresent: true,
+func framesToBuffer(tb testing.TB, frames ...Frame) []byte {
+	buf := []byte{}
+	for _, frame := range frames {
+		var err error
+		buf, err = frame.Append(buf, protocol.Version1)
+		if err != nil {
+			tb.Error(err)
```
marten-seemann: Use `require`.

internal/wire/frame_parser_test.go
```diff
@@ -234,14 +705,31 @@ func TestFrameParserResetStreamAtUnsupported(t *testing.T) {
...
+	frameType, l, err := parser.ParseType(encodeVarInt(0x42), protocol.Encryption1RTT)
+
+	// Expected: No validation being handled by parser.ParseType
+	require.Equal(t, 2, l)
+	require.Equal(t, FrameType(0), frameType)
+
+	require.Error(t, err)
+	var transportErr *qerr.TransportError
+	require.ErrorAs(t, err, &transportErr)
+	require.Equal(t, qerr.FrameEncodingError, transportErr.ErrorCode)
```
marten-seemann: I don't understand this test. This looks like validation is in fact performed.

marten-seemann (commented):
Review of the tests.

jannis-seemann:
Here we go :)

internal/wire/frame_parser_test.go
```diff
@@ -2,41 +2,57 @@ package wire
...
 	parser := NewFrameParser(true, true)
-	l, f, err := parser.ParseNext([]byte{0, 0, 0}, protocol.Encryption1RTT, protocol.Version1)
-	require.NoError(t, err)
-	require.Nil(t, f)
+	b := []byte{0, 0, 0}
+
+	frameType, l, err := parser.ParseType(b, protocol.Encryption1RTT)
+	require.Equal(t, io.EOF, err)
 	require.Equal(t, 3, l)
+	require.Equal(t, FrameType(0), frameType)
```
```suggestion
	_, l, err := parser.ParseType(b, protocol.Encryption1RTT)
	require.Equal(t, io.EOF, err)
	require.Equal(t, 3, l)
```

internal/wire/frame_parser_test.go
```diff
@@ -2,41 +2,57 @@ package wire
...
+	frameType, l, err := parser.ParseType(b, protocol.Encryption1RTT)
 	require.NoError(t, err)
-	require.Equal(t, &PingFrame{}, f)
-	require.Equal(t, 2+1, l)
+	require.Equal(t, 3, l)
+	require.Equal(t, PingFrameType, frameType)
+
+	frame, l, err := parser.ParseLessCommonFrame(frameType, b[1:], protocol.Version1)
+	require.NoError(t, err)
+	require.Equal(t, 0, l)
```
```suggestion
	require.Zero(t, l)
```

internal/wire/frame_parser_test.go
```diff
@@ -47,23 +63,32 @@ func TestFrameParsingParsesSingleFrame(t *testing.T) {
...
-	l, f, err := parser.ParseNext(b, protocol.Encryption1RTT, protocol.Version1)
+	frameType, l, err := parser.ParseType(b, protocol.Encryption1RTT)
 	require.NoError(t, err)
-	require.IsType(t, &PingFrame{}, f)
+	require.Equal(t, PingFrameType, frameType)
 	require.Equal(t, 1, l)
+
+	frame, l, err := parser.ParseLessCommonFrame(frameType, b, protocol.Version1)
+	require.NoError(t, err)
+	require.Equal(t, 0, l)
```
```suggestion
	require.Zero(t, l)
```

packet_packer_test.go
```diff
@@ -736,10 +736,17 @@ func TestPackLongHeaderPadToAtLeast4Bytes(t *testing.T) {
...
+	frameType, lt, err := frameParser.ParseType(data[2:], protocol.EncryptionHandshake)
+	require.NoError(t, err)
+	require.Equal(t, 1, lt)
+	require.Equal(t, wire.PingFrameType, frameType)
+
+	frame, l, err := frameParser.ParseLessCommonFrame(frameType, data[2+lt:], protocol.Version1)
 	require.NoError(t, err)
 	require.IsType(t, &wire.PingFrame{}, frame)
-	require.Equal(t, sealer.Overhead(), len(data)-2-l)
+	require.Equal(t, 0, l)
```
marten-seemann:
```suggestion
	require.Zero(t, l)
```

packet_packer_test.go
```diff
@@ -774,10 +779,15 @@ func TestPackShortHeaderPadToAtLeast4Bytes(t *testing.T) {
 
 	// ... followed by the STREAM frame
 	frameParser := wire.NewFrameParser(false, false)
-	frameLen, frame, err := frameParser.ParseNext(payload[1:], protocol.Encryption1RTT, protocol.Version1)
+	frameType, l, err := frameParser.ParseType(payload[1:], protocol.Encryption1RTT)
+	require.NoError(t, err)
+	require.Equal(t, 1, l)
+	require.True(t, wire.IsStreamFrameType(frameType))
```
marten-seemann:
```suggestion
	require.True(t, frameType.IsStreamFrameType())
```

internal/wire/frame_type.go
```diff
@@ -0,0 +1,77 @@
...
+	FrameTypeApplicationClose   FrameType = 0x1d
+	FrameTypeHandshakeDone      FrameType = 0x1e
+	FrameTypeResetStreamAt      FrameType = 0x24 // https://datatracker.ietf.org/doc/draft-ietf-quic-reliable-stream-reset/06/
+
+	FrameTypeDatagramNoLength   FrameType = 0x30
+	FrameTypeDatagramWithLength FrameType = 0x31
+)
+
+func (t FrameType) IsStreamFrameType() bool {
+	return byte(t)&0xf8 == 0x8
```
marten-seemann: This is wrong, we can't just assume that the first 7 bytes are 0.

marten-seemann: This is very easy: https://datatracker.ietf.org/doc/html/rfc9000#frame-types
```suggestion
	return t >= 0x8 && t <= 0xf
```

internal/wire/wire.test
marten-seemann, line 1: No!

internal/wire/frame_parser_test.go
marten-seemann: I improved it.

marten-seemann (approved):

PR #5229, opened 2025-06-17
quicvarint: speed up parsing of 1, 2 and 4 byte varints 

jannis-seemann:
I would like to propose the following changes. quicvarint.Parse is one of the most called functions during parsing, and it is responsible for about 40% of time spend during parsing, at least for this benchmark. Maybe (depending on the number of ACK frames), it may be a bit lower, but even it if was 10%, it would be significant. 

`go test -bench=BenchmarkParseStreamAndACK -cpuprofile=cpu.pprof`

Screenshot 2025-06-17 at 22 10 20

Thus, if we were to improve the performance of quicvarint.Parse, we could significantly improve the performance during parsing, and I believe adjustments here could be worth the effort. 

I propose the following changes: 

- We can further stop converting the length into a number, to further improve the performance
- We can use a bitwise OR pipe operator instead of the addition - this seems to be slightly faster than the addition, at least minimally. This also seems to be the way to go in the go standard library, see binary.BigEndian.Uint64. 
- We can use the binary.BigEndian module to convert 8 byte directly into an uint64. This seems to be faster for the 8 byte integer only though (even though it uses pretty much the same code, just in the standard library). Adding the bounds check from the standard library directly to our code also didn't made a difference. 

Overall, I believe this change could improve the overall stream parsing by another few percentage points.

```
goos: darwin
goarch: arm64
pkg: github.com/quic-go/quic-go/quicvarint
cpu: Apple M4 Max
                │ original.out │         original-pipe.out          │              plus.out               │                pipe.out                │               binary.out               │                mix.out                 │
                │    sec/op    │   sec/op     vs base               │   sec/op     vs base                │   sec/op     vs base                   │   sec/op     vs base                   │   sec/op     vs base                   │
Parse/1-byte-16    1.947n ± 1%   1.973n ± 1%  +1.34% (p=0.001 n=10)   1.723n ± 0%  -11.50% (p=0.000 n=10)   1.689n ± 1%  -13.25% (p=0.000 n=10+60)   1.705n ± 1%  -12.43% (p=0.000 n=10+60)   1.669n ± 0%  -14.30% (p=0.000 n=10+60)
Parse/2-byte-16    2.030n ± 1%   2.039n ± 0%  +0.44% (p=0.022 n=10)   1.775n ± 0%  -12.54% (p=0.000 n=10)   1.768n ± 0%  -12.88% (p=0.000 n=10+60)   1.769n ± 0%  -12.86% (p=0.000 n=10+60)   1.748n ± 0%  -13.89% (p=0.000 n=10+60)
Parse/4-byte-16    2.012n ± 1%   1.998n ± 1%       ~ (p=0.054 n=10)   1.958n ± 0%   -2.71% (p=0.000 n=10)   1.972n ± 0%   -1.99% (p=0.000 n=10+60)   1.956n ± 0%   -2.78% (p=0.000 n=10+60)   1.948n ± 0%   -3.16% (p=0.000 n=10+60)
Parse/8-byte-16    2.014n ± 0%   2.002n ± 1%  -0.60% (p=0.003 n=10)   2.444n ± 1%  +21.32% (p=0.000 n=10)   2.414n ± 2%  +19.83% (p=0.000 n=10+60)   2.201n ± 0%   +9.28% (p=0.000 n=10+60)   2.193n ± 0%   +8.89% (p=0.000 n=10+60)
geomean            2.001n        2.003n       +0.12%                  1.956n        -2.23%                  1.942n        -2.94%                     1.898n        -5.11%                     1.879n        -6.08%
```

I ran the following benchmarks: 

- original.out: The original code from before
- original-pipe.out: The original code, just with a bitwise OR instead of an addition
- pipe.out: The code attached below
- plus.out: The code for pipe.out, just replaced the | with an +
- binary.out: Attached below, it uses the binary module for the 4 byte / 8 byte varint
- mix.out: Using binary.BigEndian seems to be faster for 8 byte varints only. Thus, we're using this module here, and this is the version of the PR. I'm not fully sure why it is faster, as in the standard library, it is implemented the same way as if we wrote it here (and the bounds check should be omitted due to checking the length before anyway). still, it's faster this way. 

pipe.out:
```
func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
	if len(b) == 0 {
		return 0, 0, io.EOF
	}

	first := b[0]
	switch first >> 6 {
	case 0: // 1-byte encoding: 00xxxxxx
		return uint64(first & 0x3F), 1, nil
	case 1: // 2-byte encoding: 01xxxxxx
		if len(b) < 2 {
			return 0, 0, io.ErrUnexpectedEOF
		}
		return uint64(first&0x3F)<<8 | uint64(b[1]), 2, nil
	case 2: // 4-byte encoding: 10xxxxxx
		if len(b) < 4 {
			return 0, 0, io.ErrUnexpectedEOF
		}
		return uint64(first&0x3F)<<24 |
			uint64(b[1])<<16 |
			uint64(b[2])<<8 |
			uint64(b[3]), 4, nil
	case 3: // 8-byte encoding: 11xxxxxx
		if len(b) < 8 {
			return 0, 0, io.ErrUnexpectedEOF
		}
		return uint64(first&0x3F)<<56 |
			uint64(b[1])<<48 |
			uint64(b[2])<<40 |
			uint64(b[3])<<32 |
			uint64(b[4])<<24 |
			uint64(b[5])<<16 |
			uint64(b[6])<<8 |
			uint64(b[7]), 8, nil
	}

	// Should never be reached
	return 0, 0, io.ErrUnexpectedEOF
}
```

binary.out:
```
func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
	if len(b) == 0 {
		return 0, 0, io.EOF
	}

	first := b[0]
	switch first >> 6 {
	case 0: // 1-byte encoding: 00xxxxxx
		return uint64(first & 0x3F), 1, nil
	case 1: // 2-byte encoding: 01xxxxxx
		if len(b) < 2 {
			return 0, 0, io.ErrUnexpectedEOF
		}
		return uint64(first&0x3F)<<8 | uint64(b[1]), 2, nil
	case 2: // 4-byte encoding
		if len(b) < 4 {
			return 0, 0, io.ErrUnexpectedEOF
		}
		// inject the top 6 bits into the highest byte
		tmp := binary.BigEndian.Uint32(b[0:4])
		tmp &= 0x3FFFFFFF // clear top 2 bits (in first byte)
		return uint64(tmp), 4, nil

	case 3: // 8-byte encoding
		if len(b) < 8 {
			return 0, 0, io.ErrUnexpectedEOF
		}
		tmp := binary.BigEndian.Uint64(b[0:8])
		tmp &= 0x3FFFFFFFFFFFFFFF // clear top 2 bits
		return tmp, 8, nil
	}

	// Should never be reached
	return 0, 0, io.ErrUnexpectedEOF
}
```

### DIFF
--- a/quicvarint/varint.go
+++ b/quicvarint/varint.go
@@ -97,7 +97,8 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
 		if len(b) < 8 {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
-		tmp := binary.BigEndian.Uint64(b[0:8])
+		// binary.BigEndian.Uint64 only reads the first 8 bytes, so passing the full slice avoids slicing overhead.
+		tmp := binary.BigEndian.Uint64(b)
 		tmp &= 0x3FFFFFFFFFFFFFFF
 		return tmp, 8, nil
 	}
--- a/quicvarint/varint.go
+++ b/quicvarint/varint.go
@@ -79,32 +79,26 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
 	first := b[0]
 	switch first >> 6 {
 	case 0: // 1-byte encoding: 00xxxxxx
-		return uint64(first & 0x3F), 1, nil
+		return uint64(first & 0b00111111), 1, nil
 	case 1: // 2-byte encoding: 01xxxxxx
 		if len(b) < 2 {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
-		return uint64(b[1]) | uint64(first&0x3F)<<8, 2, nil
+		return uint64(b[1]) | uint64(first&0b00111111)<<8, 2, nil
 	case 2: // 4-byte encoding: 10xxxxxx
 		if len(b) < 4 {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
-		return uint64(b[3]) |
-			uint64(b[2])<<8 |
-			uint64(b[1])<<16 |
-			uint64(first&0x3F)<<24, 4, nil
+		return uint64(b[3]) | uint64(b[2])<<8 | uint64(b[1])<<16 | uint64(first&0b00111111)<<24, 4, nil
 	case 3: // 8-byte encoding: 00xxxxxx
 		if len(b) < 8 {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
-		// binary.BigEndian.Uint64 only reads the first 8 bytes, so passing the full slice avoids slicing overhead.
-		tmp := binary.BigEndian.Uint64(b)
-		tmp &= 0x3FFFFFFFFFFFFFFF
-		return tmp, 8, nil
+		// binary.BigEndian.Uint64 only reads the first 8 bytes. Passing the full slice avoids slicing overhead.
+		return binary.BigEndian.Uint64(b) & 0x3fffffffffffffff, 8, nil
 	}
 
-	// Should never be reached
-	return 0, 0, io.ErrUnexpectedEOF
+	panic("unreachable")
 }
 
 // Append appends i in the QUIC varint format.
--- a/quicvarint/varint.go
+++ b/quicvarint/varint.go
@@ -84,15 +84,15 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
 		if len(b) < 2 {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
-		return uint64(first&0x3F)<<8 | uint64(b[1]), 2, nil
+		return uint64(b[1]) | uint64(first&0x3F)<<8, 2, nil
 	case 2: // 4-byte encoding: 10xxxxxx
 		if len(b) < 4 {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
-		return uint64(first&0x3F)<<24 |
-			uint64(b[1])<<16 |
+		return uint64(b[3]) |
 			uint64(b[2])<<8 |
-			uint64(b[3]), 4, nil
+			uint64(b[1])<<16 |
+			uint64(first&0x3F)<<24, 4, nil
 	case 3: // 8-byte encoding: 00xxxxxx
 		if len(b) < 8 {
 			return 0, 0, io.ErrUnexpectedEOF
--- a/quicvarint/varint.go
+++ b/quicvarint/varint.go
@@ -93,7 +93,7 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
 			uint64(b[1])<<16 |
 			uint64(b[2])<<8 |
 			uint64(b[3]), 4, nil
-	case 3: // 8-byte encoding: 00xxxxxx 
+	case 3: // 8-byte encoding: 00xxxxxx
 		if len(b) < 8 {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
--- a/quicvarint/varint.go
+++ b/quicvarint/varint.go
@@ -98,7 +98,7 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
 			return 0, 0, io.ErrUnexpectedEOF
 		}
 		tmp := binary.BigEndian.Uint64(b[0:8])
-		tmp &= 0x3FFFFFFFFFFFFFFF // clear top 2 bits
+		tmp &= 0x3FFFFFFFFFFFFFFF
 		return tmp, 8, nil
 	}
 
--- a/quicvarint/varint.go
+++ b/quicvarint/varint.go
@@ -1,6 +1,7 @@
 package quicvarint
 
 import (
+	"encoding/binary"
 	"fmt"
 	"io"
 )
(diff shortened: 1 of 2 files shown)


jannis-seemann:
@marten-seemann ready

quicvarint/varint.go
```diff
@@ -74,23 +75,36 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
...
+		}
+		// binary.BigEndian.Uint64 only reads the first 8 bytes, so passing the full slice avoids slicing overhead.
+		tmp := binary.BigEndian.Uint64(b)
+		tmp &= 0x3FFFFFFFFFFFFFFF
+		return tmp, 8, nil
 	}
-	return uint64(b[7]) + uint64(b[6])<<8 + uint64(b[5])<<16 + uint64(b[4])<<24 + uint64(b[3])<<32 + uint64(b[2])<<40 + uint64(b[1])<<48 + uint64(b0)<<56, 8, nil
+
+	// Should never be reached
+	return 0, 0, io.ErrUnexpectedEOF
```
marten-seemann: Let's better make this explicit. This way, the fuzzer will find it if we got it wrong.
```suggestion
	panic("unreachable")
```

quicvarint/varint.go
```diff
@@ -74,23 +75,36 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
...
+			uint64(b[1])<<16 |
+			uint64(first&0x3F)<<24, 4, nil
+	case 3: // 8-byte encoding: 00xxxxxx
+		if len(b) < 8 {
+			return 0, 0, io.ErrUnexpectedEOF
+		}
+		// binary.BigEndian.Uint64 only reads the first 8 bytes, so passing the full slice avoids slicing overhead.
+		tmp := binary.BigEndian.Uint64(b)
+		tmp &= 0x3FFFFFFFFFFFFFFF
+		return tmp, 8, nil
```
marten-seemann:
```suggestion
		return binary.BigEndian.Uint64(b) & 0x3fffffffffffffff, 8, nil
```

quicvarint/varint.go
```diff
@@ -74,23 +75,36 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
...
+
+	first := b[0]
+	switch first >> 6 {
+	case 0: // 1-byte encoding: 00xxxxxx
+		return uint64(first & 0x3F), 1, nil
+	case 1: // 2-byte encoding: 01xxxxxx
+		if len(b) < 2 {
+			return 0, 0, io.ErrUnexpectedEOF
+		}
+		return uint64(b[1]) | uint64(first&0x3F)<<8, 2, nil
```
marten-seemann: Nit:
```suggestion
		return uint64(b[1]) | uint64(first&0x3f)<<8, 2, nil
```

quicvarint/varint.go
```diff
@@ -74,23 +75,36 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
...
-	if l == 2 {
-		return uint64(b[1]) + uint64(b0)<<8, 2, nil
-	}
-	if l == 4 {
-		return uint64(b[3]) + uint64(b[2])<<8 + uint64(b[1])<<16 + uint64(b0)<<24, 4, nil
+
+	first := b[0]
+	switch first >> 6 {
+	case 0: // 1-byte encoding: 00xxxxxx
+		return uint64(first & 0x3F), 1, nil
```
marten-seemann: Nit:
```suggestion
		return uint64(first & 0x3f), 1, nil
```

quicvarint/varint.go
```diff
@@ -74,23 +75,36 @@ func Parse(b []byte) (uint64 /* value */, int /* bytes consumed */, error) {
...
+		}
+		return uint64(b[1]) | uint64(first&0x3F)<<8, 2, nil
+	case 2: // 4-byte encoding: 10xxxxxx
+		if len(b) < 4 {
+			return 0, 0, io.ErrUnexpectedEOF
+		}
+		return uint64(b[3]) |
+			uint64(b[2])<<8 |
+			uint64(b[1])<<16 |
+			uint64(first&0x3F)<<24, 4, nil
```
marten-seemann:
```suggestion
		return uint64(b[3]) | uint64(b[2])<<8 | uint64(b[1])<<16 | uint64(first&0x3f)<<24, 4, nil
```

marten-seemann (commented):
LGTM, just a few nits.

quicvarint/varint.go
marten-seemann: Actually, for better readability, you should use `0b00111111`. Makes it crystal clear what's going on. That feature was not available when we first implement varint parsing.

marten-seemann (approved):

marten-seemann:
Here's the benchmark of the change that we merged:
```
name                     old time/op  new time/op  delta
Read/1-byte-16           3.32ns ± 3%  3.27ns ± 2%   -1.53%  (p=0.000 n=20+20)
Read/2-byte-16           4.81ns ± 3%  4.86ns ± 2%   +0.88%  (p=0.032 n=20+20)
Read/4-byte-16           7.22ns ± 2%  7.44ns ± 1%   +2.95%  (p=0.000 n=18+20)
Read/8-byte-16           12.6ns ± 1%  12.7ns ± 2%   +0.68%  (p=0.001 n=20+19)
Parse/1-byte-16          2.23ns ± 2%  1.95ns ± 2%  -12.66%  (p=0.000 n=19+20)
Parse/2-byte-16          2.31ns ± 2%  2.01ns ± 2%  -12.87%  (p=0.000 n=20+19)
Parse/4-byte-16          2.24ns ± 2%  2.21ns ± 3%   -1.26%  (p=0.020 n=20+20)
Parse/8-byte-16          2.56ns ± 1%  2.74ns ± 2%   +7.16%  (p=0.000 n=19+19)
Append/1-byte-16         2.55ns ± 2%  2.54ns ± 3%     ~     (p=0.101 n=20+20)
Append/2-byte-16         3.67ns ± 2%  3.64ns ± 1%   -0.78%  (p=0.002 n=20+20)
Append/4-byte-16         3.62ns ± 2%  3.62ns ± 2%     ~     (p=0.931 n=20+20)
Append/8-byte-16         3.91ns ± 1%  3.86ns ± 1%   -1.28%  (p=0.000 n=18+20)
AppendWithLen/1-byte-16  3.89ns ± 2%  3.85ns ± 1%   -1.15%  (p=0.000 n=20+20)
AppendWithLen/2-byte-16  5.31ns ± 2%  5.25ns ± 1%   -1.15%  (p=0.000 n=20+18)
AppendWithLen/4-byte-16  5.56ns ± 2%  5.49ns ± 1%   -1.24%  (p=0.000 n=20+19)
AppendWithLen/8-byte-16  5.85ns ± 1%  5.77ns ± 1%   -1.39%  (p=0.000 n=20+19)
```

Issue #4674, opened 2024-09-14
move HTTP/3 code to a separate repo


marten-seemann:
http3-go?

rthellend:
After https://github.com/quic-go/quic-go/pull/5256, creating a new http3 repo should be pretty straight forward:

```bash
git clone https://github.com/quic-go/quic-go.git
cd quic-go/
git filter-repo --subdirectory-filter http3
cp <otherdir>/quic-go/{LICENSE,SECURITY.md} .
sed -i -re 's:github.com/quic-go/quic-go/http3:github.com/quic-go/http3-go:g' server_test.go
go mod init github.com/quic-go/http3-go
go mod tidy
git add .
```
Then commit the changes and push to the new repo.

PR #5256, opened 2025-07-03
http3: Remove dependency on quic internal package

rthellend:
Remove the dependency on internal/protocol from the http3 package. This makes it possible for a forked http3 to use the mainline quic-go package.


interface.go
```diff
@@ -25,6 +28,22 @@ const (
...
+
+const (
+	// A perspective of acting as a server.
+	PerspectiveServer = protocol.PerspectiveServer
+	// A perspective of acting as a client.
+	PerspectiveClient = protocol.PerspectiveClient
+)
+
+// MaxByteCount is the maximum value of a ByteCount
+const MaxByteCount = protocol.MaxByteCount
```
marten-seemann: We definitely don't need to expose this, it's only used in tests.

interface.go
```diff
@@ -25,6 +28,22 @@ const (
...
+// Perspective determines if we're acting as a server or a client
+type Perspective = protocol.Perspective
+
+const (
+	// A perspective of acting as a server.
+	PerspectiveServer = protocol.PerspectiveServer
+	// A perspective of acting as a client.
+	PerspectiveClient = protocol.PerspectiveClient
+)
+
```
marten-seemann: Not a fan of exposing this. It's not used in the API anywhere. We could either redefine it in the `HTTP/3` package, or just use an `isServer bool` flag.

interface.go
```diff
@@ -25,6 +28,22 @@ const (
 	Version2 = protocol.Version2
 )
 
+// SupportedVersions lists the versions that the server supports must be in sorted descending order
```
marten-seemann: This doesn't only apply to the server.

interface.go
```diff
@@ -15,6 +15,9 @@ import (
 // The StreamID is the ID of a QUIC stream.
 type StreamID = protocol.StreamID
 
+// InvalidStreamID is a stream ID that is invalid. The first valid stream ID in QUIC is 0.
+const InvalidStreamID = protocol.InvalidStreamID
```
marten-seemann: Can we get away with not exposing this?

marten-seemann (commented):
Although I'm not a fan of your plan to fork HTTP/3, I think this is a worthwhile change, especially due to #4674.

I'd like to avoid exposing too much additional API surface in the `quic` package, see my comments that I added.

Can you also add a `depguard` rule in the golangci-lint config, so that we don't accidentally reintroduce a dependency on an internal package?

interface.go
marten-seemann: Maybe also add that this value shouldn't be modified? Or maybe we should make this a function, so it's actually not possible to modify it?

rthellend:
I addressed your comments. I'm not 100% sure about the depguard syntax. I'll see what golangci says.

There's one dependency left that I missed earlier:

http3/server_test.go:   "github.com/quic-go/quic-go/internal/testdata"

How would you like to handle it? It would be trivial to copy the data.

Thanks!

rthellend:
The depguard rule is working now.

```
  Error: http3/server_test.go:18:2: import 'github.com/quic-go/quic-go/internal/testdata' is not allowed from list 'http3-internal': no dependency on quic-go/internal (depguard)
```

How would you like to handle this last dependency?

rthellend:
The integration test failure looks unrelated to this change.

marten-seemann:
Thank you! I'll do a review later, but this will definitely go into the next release.

http3/conn.go
```diff
@@ -51,11 +53,22 @@ type Conn struct {
...
+type perspective int
+
+func (p perspective) opposite() perspective {
+	return 3 - p
+}
+
+const (
+	perspectiveServer perspective = 1
+	perspectiveClient perspective = 2
+)
```
marten-seemann: I think it's easier to just use a bool.

PR #5259, opened 2025-07-12
fix: close transport when dial error

rbqvq:
Close the transport after dial fails to avoid memory leaks.

---

I noticed that `DialAddrEarly` has `tr.Close()` but `DialAddr` does not, if this is intentional then please close this PR


marten-seemann (approved):
Yes, this looks correct. Thank you!

PR #5263, opened 2025-07-18
wire: improve frame parsing benchmarks

jannis-seemann:
In this PR, we are doing some major changes to how frames are being handeled: 
=> https://github.com/quic-go/quic-go/pull/5227

We should however first adjust the benchmarks, and only then merge this other pull request, so we can see how the performance improves. This is the PR for changing the benchmarks.

### DIFF
--- a/internal/wire/frame_parser_test.go
+++ b/internal/wire/frame_parser_test.go
@@ -270,8 +270,8 @@ func writeFrames(tb testing.TB, frames ...Frame) []byte {
 	return b
 }
 
-// This function is used in a benchmark, and also to ensure 0 allocations for StreamFrames
-// require causes allocations, we thus need to test manually.
+// This function is used in benchmarks, and also to ensure zero allocation for STREAM frame parsing.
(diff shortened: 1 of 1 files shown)


internal/wire/frame_parser_test.go
```diff
@@ -259,132 +260,171 @@ func TestFrameParsingErrorsOnInvalidFrames(t *testing.T) {
...
 
-	data, err := ack.Append([]byte{}, protocol.Version1)
-	if err != nil {
-		b.Fatal(err)
-	}
-	data, err = sf.Append(data, protocol.Version1)
-	if err != nil {
-		b.Fatal(err)
+// This function is used in a benchmark, and also to ensure 0 allocations for StreamFrames
+// require causes allocations, we thus need to test manually.
```
```suggestion
// This function is used in benchmarks, and also to ensure zero allocation for STREAM frame parsing.
// We can therefore not use the require framework, as it allocates.
```

marten-seemann (approved):

PR #5271, opened 2025-07-24
Add ConnectionStats

### LINKED ISSUE #5262: Get RTT from Conn
It is helpful for certain [application protocols](https://ethresear.ch/t/the-paths-of-least-resistance-introducing-wfr-gossip/22671/3) to know what the RTT estimate is of the underlying connection.

Note this is complementary to https://github.com/quic-go/quic-go/issues/5255. That would give you the probe results, this would give you the connection's current view of the RTT stats (including smoothedRTT).

MarcoPolo:
closes #5262

### DIFF
--- a/internal/congestion/cubic_sender.go
+++ b/internal/congestion/cubic_sender.go
@@ -22,7 +22,6 @@ const (
 type cubicSender struct {
 	hybridSlowStart HybridSlowStart
 	rttStats        *utils.RTTStats
-	connStats       *utils.ConnectionStats
 	cubic           *Cubic
 	pacer           *pacer
 	clock           Clock
@@ -69,7 +68,6 @@ var (
 func NewCubicSender(
 	clock Clock,
 	rttStats *utils.RTTStats,
-	connStats *utils.ConnectionStats,
 	initialMaxDatagramSize protocol.ByteCount,
 	reno bool,
 	tracer *logging.ConnectionTracer,
@@ -77,7 +75,6 @@ func NewCubicSender(
 	return newCubicSender(
 		clock,
 		rttStats,
-		connStats,
 		reno,
 		initialMaxDatagramSize,
 		initialCongestionWindow*initialMaxDatagramSize,
@@ -89,7 +86,6 @@ func NewCubicSender(
 func newCubicSender(
 	clock Clock,
 	rttStats *utils.RTTStats,
-	connStats *utils.ConnectionStats,
 	reno bool,
 	initialMaxDatagramSize,
 	initialCongestionWindow,
@@ -98,7 +94,6 @@ func newCubicSender(
 ) *cubicSender {
 	c := &cubicSender{
 		rttStats:                   rttStats,
-		connStats:                  connStats,
 		largestSentPacketNumber:    protocol.InvalidPacketNumber,
 		largestAckedPacketNumber:   protocol.InvalidPacketNumber,
 		largestSentAtLastCutback:   protocol.InvalidPacketNumber,
@@ -194,9 +189,6 @@ func (c *cubicSender) OnPacketAcked(
 }
 
 func (c *cubicSender) OnCongestionEvent(packetNumber protocol.PacketNumber, lostBytes, priorInFlight protocol.ByteCount) {
-	c.connStats.PacketsLost.Add(1)
-	c.connStats.BytesLost.Add(uint64(lostBytes))
-
 	// TCP NewReno (RFC6582) says that once a loss occurs, any losses in packets
 	// already sent should be treated as a single loss event, since it's expected.
 	if packetNumber <= c.largestSentAtLastCutback {
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -132,7 +132,6 @@ func newSentPacketHandler(
 	congestion := congestion.NewCubicSender(
 		congestion.DefaultClock{},
 		rttStats,
-		connStats,
 		initialMaxDatagramSize,
 		true, // use Reno
 		tracer,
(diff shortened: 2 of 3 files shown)


connection.go
```diff
@@ -726,6 +729,36 @@ func (c *Conn) ConnectionState() ConnectionState {
...
+	SmoothedRTT   time.Duration
+	MeanDeviation time.Duration
+
+	BytesSent       uint64
+	PacketsSent     uint64
+	BytesReceived   uint64
+	PacketsReceived uint64
+	BytesLost       uint64
+	PacketsLost     uint64
+}
```
marten-seemann: Can you add documentation for the struct and its members? Totally fine to copy-paste from the W3C WebTransport spec.

internal/utils/rtt_stats.go
```diff
@@ -67,32 +78,42 @@ func (r *RTTStats) UpdateRTT(sendDelta, ackDelay time.Duration) {
 	// ackDelay but the raw observed sendDelta, since poor clock granularity at
 	// the client may cause a high ackDelay to result in underestimation of the
 	// r.minRTT.
-	if r.minRTT == 0 || r.minRTT > sendDelta {
-		r.minRTT = sendDelta
+	for {
```
marten-seemann: Why is this a loop now? This is in the hot path, as it gets updated every time an ACK is received.

MarcoPolo: A general way to protect against concurrent writes. On second thought this isn't necessary because we only have a single writer with a possibly concurrent reader.

connection.go
```diff
@@ -726,6 +729,59 @@ func (c *Conn) ConnectionState() ConnectionState {
 	return c.connState
 }
 
+type ConnectionStats struct {
+	// MinRTT is the minimum round-trip time observed on the entire connection.
```
marten-seemann: Is this a per-path instead of a per-connection property? See also https://github.com/w3c/webtransport/issues/683.

connection.go
```diff
@@ -726,6 +729,59 @@ func (c *Conn) ConnectionState() ConnectionState {
 	return c.connState
 }
 
+type ConnectionStats struct {
```
marten-seemann, line 733: This needs a comment as well.

MarcoPolo: done in 3720ddb

connection.go
MarcoPolo: Yes it's per path. Updated the text in 3720ddb

internal/utils/rtt_stats.go
```diff
@@ -105,15 +121,26 @@ func (r *RTTStats) SetInitialRTT(t time.Duration) {
...
-	r.smoothedRTT = 0
-	r.meanDeviation = 0
+	r.minRTT.Store(0)
+	r.latestRTT.Store(0)
+	r.smoothedRTT.Store(0)
+	r.meanDeviation.Store(0)
 	// max_ack_delay remains valid
 }
+
+func (r *RTTStats) Clone() *RTTStats {
```
marten-seemann, line 137: Why is this needed? I only see it used in a test.

MarcoPolo: It was to keep this test semantically the same: https://github.com/quic-go/quic-go/pull/5271/files#diff-4ba4925225c635731cd9ae756417f332d424697942518bc6cb89396c7f8ae6f3L256.

It's awkward to implement this outside of this package as we need to copy the private fields.

Previously we could simply copy the RTTStats struct, but now the struct contains atomics which can't be copied.

marten-seemann: Why is it needed at all? Aren't `atomic.Int64`s copied the same way that `int64`s are?

MarcoPolo: No, atomic.Int64 holds a [nocopy](https://cs.opensource.google/go/go/+/refs/tags/go1.24.6:src/sync/atomic/type.go;l=224-231) https://cs.opensource.google/go/go/+/refs/tags/go1.24.6:src/sync/atomic/type.go;l=100

marten-seemann: Makes sense, let's just keep `Clone` then.

internal/congestion/cubic_sender.go
```diff
@@ -189,6 +194,9 @@ func (c *cubicSender) OnPacketAcked(
 }
 
 func (c *cubicSender) OnCongestionEvent(packetNumber protocol.PacketNumber, lostBytes, priorInFlight protocol.ByteCount) {
+	c.connStats.PacketsLost.Add(1)
+	c.connStats.BytesLost.Add(uint64(lostBytes))
```
marten-seemann: This is not correct, is it? When using ECN, we also call `OnCongestionEvent` when we receive a CE mark.

I think this counter belongs in the `sentPacketHandler`.

MarcoPolo: >  When using ECN, we also call OnCongestionEvent when we receive a CE mark.

Do you mean [here](https://github.com/MarcoPolo/quic-go/blob/3720ddb93308873a3fd627bf05a8ce158e184d68/internal/ackhandler/sent_packet_handler.go#L384)? Would you expect an ECN mark to call `OnCongestionEvent` with non-zero `lostBytes`?

I agree we should put it in sentPacketHandler. [a5fe427](https://github.com/quic-go/quic-go/pull/5271/commits/a5fe42724335cfc7cb968b05db282be49f1a96b1) makes that change.

marten-seemann: A CE-marked packet is not lost. It's just a signal from the bottleneck router that a queue is building up. We treat it as loss for congestion purposes (to shrink the cwnd), but the packet was definitely not actually lost (it must have been received for the CE count to be reported).

marten-seemann:
Removed from master via force-push. Recreating in #5281.

PR #5273, opened 2025-07-25
fix missing log statement for STREAM, DATAGRAM and ACK

jannis-seemann:
Logging had been missing for these 3 cases, this PR re-adds them back. This was lost in #5227.


marten-seemann:
This PR is wrong. Force-pushing to remove it from master.

PR #5274, opened 2025-07-29
s/go 1.23/go/1.23.0/

bemasc:
This avoids an error on my dev machine:

```
go: download go1.23 for darwin/arm64: toolchain not available
```

See https://github.com/golang/go/issues/66175


marten-seemann:
As we discussed in the other repo, this was caused by using an outdated Go version.

PR #5279, opened 2025-08-06
connection: Call newPathManagerOutgoing only once per conn

MarcoPolo:
Noticed this in passing. The compare and swap operation does not do what the comment suggests. Example https://play.golang.com/p/S7dTJ9Pek4W

This is one solution that does what the comment wanted, an alternative is to do something like:
```go
old := ptr.Load()
if old == nil {
  v := init()
  ptr.CompareAndSwap(nil, v)
}
```

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -2645,16 +2645,26 @@ func (c *Conn) LocalAddr() net.Addr { return c.conn.LocalAddr() }
 // RemoteAddr returns the remote address of the QUIC connection.
 func (c *Conn) RemoteAddr() net.Addr { return c.conn.RemoteAddr() }
 
+// getPathManager lazily initializes the Conn's pathManagerOutgoing.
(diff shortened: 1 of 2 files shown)


marten-seemann:
That's a nice find @MarcoPolo!

> This is one solution that does what the comment wanted, an alternative is to do something like:
> 
> ```go
> old := ptr.Load()
> if old == nil {
>   v := init()
>   ptr.CompareAndSwap(nil, v)
> }
> ```

I think I'd prefer that, if we document that this might call the constructor twice if `AddPath` is called concurrently (which is fairly unlikely to begin with). That way, we don't have to increase the memory size of the `Connection` struct.

marten-seemann (approved):
Thank you @MarcoPolo!

marten-seemann:
Removed from master via force-push. Recreating in #5280.

PR #5291, opened 2025-08-16
use synctest to make receive stream tests fully deterministc

marten-seemann:


internal/synctest/synctest_go124.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"testing"
+	"testing/synctest"
+)
+
+func Test(t *testing.T, f func(t *testing.T)) {
+	synctest.Run(func() {
+		f(t)
+	})
+}
```
marten-seemann, line 14: For some reason, `t.Helper` doesn't seem to have any effect here. Not sure how much I care though, since this is just a shim for the old Go version anyway.

MarcoPolo: Maybe because it's being inlined?

marten-seemann:
> Will be nice to get rid of `scaleDuration`

`grep`ping for `scaleDuration` is a great way to find tests that would benefit from `synctest` :)

internal/synctest/synctest_go124.go
marten-seemann: I don't know, I thought they're doing some magic such that inline doesn't mess with the stack trace.

But as I said, I don't really care. We need to continue support Go 1.24 since downstream users might be slow to upgrade their Go versions, but new development work will only happen on Go 1.25.

PR #5295, opened 2025-08-20
Add an accessor for the local EnabledDatagrams setting

bemasc:
Currently, there is no way to detect whether Datagrams are enabled on an http3.Conn or http3.ClientConn after it has been created. This means that any application that needs this information must pass this flag in parallel to any component that requires it, and risks bugs if the value is set incorrectly.

This change exposes the EnableDatagrams setting from each connection, allowing applications to determine if Datagrams can be used by inspecting the connection.

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -47,10 +47,6 @@ var defaultQuicConfig = &quic.Config{
 type ClientConn struct {
 	conn *Conn
 
-	// Enable support for HTTP/3 datagrams (RFC 9297).
-	// If a QUICConfig is set, datagram support also needs to be enabled on the QUIC layer by setting enableDatagrams.
-	enableDatagrams bool
-
 	// Additional HTTP/3 settings.
 	// It is invalid to specify any settings defined by RFC 9114 (HTTP/3) and RFC 9297 (HTTP Datagrams).
 	additionalSettings map[uint64]uint64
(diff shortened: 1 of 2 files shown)


marten-seemann:
I read RFC 9297 differently:
> An endpoint can indicate to its peer that it is willing to receive HTTP/3 Datagrams by sending the SETTINGS_H3_DATAGRAM (0x33) setting with a value of 1.

It's a unilateral declaration of the peer, indicating its support for receiving H3 datagrams. You can inspect the peer's SETTINGS using `http3.Settings`: https://pkg.go.dev/github.com/quic-go/quic-go@v0.54.0/http3#Settings

For obtaining the local value of the config flag, the application should know what config value it passed to the http3 package, so I don't think we need any API for that (of course, the application is free to wrap the `http3.Conn` and add a getter function, if it desires to expose that information to higher layers). As a point of reference, we don't have any way to inspect the QUIC Transport Parameters (the `quic.Config`) used to create the `quic.Conn` either.

bemasc:
The declaration is unilateral, but the peer also needs to know its own setting:

> QUIC DATAGRAM frames MUST NOT be sent until the SETTINGS_H3_DATAGRAM setting has been both sent and received with a value of 1.

I'll close this for now, and we can reopen if needed.

Issue #3846, opened 2023-05-18
ackhandler.Packet memory optimization


marten-seemann:
This is how `ackhandler.Packet` currently looks (with #3844 applied):
```
type Packet struct {
	PacketNumber    protocol.PacketNumber
	Frames          []*Frame
	LargestAcked    protocol.PacketNumber // InvalidPacketNumber if the packet doesn't contain an ACK
	Length          protocol.ByteCount
	SendTime        time.Time

	EncryptionLevel protocol.EncryptionLevel // that's a uint8

	IsPathMTUProbePacket bool // We don't report the loss of Path MTU probe packets to the congestion controller.

	includedInBytesInFlight bool
	declaredLost            bool
	skippedPacket           bool
}
```

It now consumes 80 bytes. Relatively straightforward wins:
1. `LargestAcked` is only rarely set. We should refactor the code such that the `Packet` is not exported any more. Then we can create a fake "frame" that we include in the `Frames` slice, which saves the largest acknowledged packet number. Down to 72 bytes.
2. `Length` should be a uint16 instead of a uint64 (see #3843). Down to 64 bytes.
3. The `SendTime` should be a (u)int64, not a `time.Time` (24 bytes). Unfortunately, this will require some `unsafe` operations to extract the monotonic time from a `time.Time`. Down to 48 bytes.

More involved optimizations:
1. Encode the packet number into 31 bits, relative to one of two offsets (referenced by the 32nd bit).
2. Use a microsecond resolution timestamp to encode the send time. That allows encoding of time differences of up to 70 minutes.

marten-seemann:
More option:
1. Don't save the packet number. The packet number can be calculated from the slice index plus a start offset.
2. Use a flags uint8 to replace all the bools.

marten-seemann:
The [`monotime`](https://github.com/quic-go/quic-go/tree/monotime) branch contains some initial implementation, but we need to think more carefully about the timestamps exported to the logger / qlog. Maybe make it all durations and have the qlog package deal with converting it back to a timestamp?

marten-seemann:
We might even be able to _store_ the timestamp as an uint32 in the packet history, with microsecond (not nanosecond) precision. When dealing with timestamps, we'd still use the uint64 definition from the `monotime` branch, but in the packet history, we convert to uint32, i.e. the time difference from a base timestamp.

2^23 microseconds are 71 minutes, so that's definitely too short for a general-purpose QUIC stack.

[...]

PR #5312, opened 2025-08-28
ackhandler: avoid storing packet number in packet struct

marten-seemann:
Part of #3846.

The packet number can be derived from the position that this packet is stored at in the packets slice in the sent packet history. There is no need to store the packet number, saving 8 bytes per packet. This slice contains all in-flight packets, and therefore can consume significant amounts of memory on high-BDP connection. 

By going from an `iter.Seq[*packet]` to an `iter.Seq2[protocol.PacketNumber, *packet]`, the changes to the logic calling into the packet history are fairly minimal (although they're somewhat verbose).

cc @MarcoPolo, since we talked about this a long time ago

### DIFF
--- a/internal/ackhandler/sent_packet_history.go
+++ b/internal/ackhandler/sent_packet_history.go
@@ -9,7 +9,7 @@ import (
 
 type sentPacketHistory struct {
 	packets          []*packet
-	pathProbePackets []*packetWithPacketNumber
+	pathProbePackets []packetWithPacketNumber
 
 	numOutstanding int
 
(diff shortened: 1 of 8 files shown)


internal/ackhandler/sent_packet_history.go
```diff
@@ -52,38 +54,38 @@ func (h *sentPacketHistory) SentNonAckElicitingPacket(pn protocol.PacketNumber)
...
+	return func(yield func(protocol.PacketNumber, *packet) bool) {
+		// h.firstPacketNumber might be updated in the yield function,
+		// so we need to save it here.
+		firstPacketNumber := h.firstPacketNumber
+		for i, p := range h.packets {
 			if p == nil {
 				continue
 			}
-			if !yield(p) {
+			if !yield(firstPacketNumber+protocol.PacketNumber(i), p) {
```
MarcoPolo, line 80: Nice, I was going to suggest using an iterator with the packet number as the first item.

internal/ackhandler/sent_packet_history.go
```diff
@@ -9,16 +9,18 @@ import (
 
 type sentPacketHistory struct {
 	packets          []*packet
```
MarcoPolo, line 11: I wonder if you would see a perf improvement by using a `[]packet` here. That way when iterating over packets located contiguously in memory.

marten-seemann: That's a good idea. I was concerned about copying the struct over and over again, but then we're shrinking the size of this struct...

We should benchmark that! I'll open an issue.

PR #5313, opened 2025-08-28
ackhandler: immediately clear ackedPacket slice after processing ACK

marten-seemann:
I don't think the current code was wrong or racy, but it's weird to put a buffer back into the pool, and still hold a pointer to it (even if it's just in a slice that won't be accessed anymore). Clearing and immediately re-slicing seems like the safer way to go, and it probably allows the GC to clean up elements from the `sync.Pool` earlier as well.


internal/ackhandler/sent_packet_handler.go
```diff
@@ -432,7 +434,6 @@ func (h *sentPacketHandler) GetLowestPacketNotConfirmedAcked() protocol.PacketNu
 // Packets are returned in ascending packet number order.
 func (h *sentPacketHandler) detectAndRemoveAckedPackets(ack *wire.AckFrame, encLevel protocol.EncryptionLevel) ([]*packet, error) {
 	pnSpace := h.getPacketNumberSpace(encLevel)
-	h.ackedPackets = h.ackedPackets[:0]
```
marten-seemann, line 435: Maybe this should be a bug check instead.

marten-seemann: Done.

Issue #5307, opened 2025-08-25
Data race in http3.Transport


keliramu:
When multiple goroutines use the same `http3.Transport` instance to make concurrent requests to different hosts, a data race occurs in the `dial()` method:

### In v0.48.2:
- **Read** at line 308: `if t.transport == nil`
- **Write** at line 313: `t.transport = &quic.Transport{Conn: udpConn}`

### In v0.54.0 (latest):
- **Read** at line 353: `if t.transport == nil`
- **Write** at line 358: `t.transport = &quic.Transport{Conn: udpConn}`

This happens because the `transport` field is accessed without proper synchronization.

WARNING: DATA RACE
Write at 0x... by goroutine X:
  github.com/quic-go/quic-go/http3.(*Transport).dial()
    .../http3/transport.go:313
Previous read at 0x... by goroutine Y:
  github.com/quic-go/quic-go/http3.(*Transport).dial()
    .../http3/transport.go:308

The race occurs because:
1. Multiple goroutines call `transport.RoundTrip()` concurrently
2. Each request to a different host triggers a new dial operation
3. In `dial()`, the code checks if `t.transport == nil`
4. If nil, it creates and assigns a new transport
5. No mutex protects these operations

This race condition can lead to:
- Multiple UDP connections being created
- Resource leaks
- Unpredictable behavior in production

marten-seemann:
This logic probably needs to be moved into the initialization function.

PR #5323, opened 2025-08-30
http3: fix data race in Transport

Glonee:
Fixes https://github.com/quic-go/quic-go/issues/5307

### DIFF
--- a/http3/transport.go
+++ b/http3/transport.go
@@ -448,6 +448,7 @@ func (t *Transport) Close() error {
 			return err
 		}
 		t.transport = nil
+		t.initOnce = sync.Once{}
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


http3/transport.go
```diff
@@ -448,6 +448,7 @@ func (t *Transport) Close() error {
 			return err
 		}
 		t.transport = nil
+		t.initOnce = sync.Once{}
```
marten-seemann, line 451: Why is this needed?

Glonee: After close, the transport will be set to nil. If we don't reset `initOnce` here, the transport will not be initialized next time we use it and result in panic.
My first commit failed on tests because of this.

marten-seemann: We should prevent this. It's not idiomatic to reuse a struct that you have closed. Or is there any reason to allow reuse?

Doesn't need to be in this PR, I'm fine merging this as is and fixing it in a follow-up PR.

Glonee: Will be fixed in https://github.com/quic-go/quic-go/pull/5324

PR #5324, opened 2025-08-31
http3: don't allow usage of closed Transport

Glonee:


marten-seemann:
I just merged #5323. Can you rebase this PR please?

Glonee:
I have rebased it, can you take a look?

http3/transport.go
```diff
@@ -431,6 +439,7 @@ func (t *Transport) NewClientConn(conn *quic.Conn) *ClientConn {
 }
 
 // Close closes the QUIC connections that this Transport has used.
+// A Transport should not be used after it has been closed.
```
marten-seemann:
```suggestion
// A Transport cannot be used after it has been closed.
```

integrationtests/self/http_test.go
marten-seemann, line 1: Can you make this a unit test instead? Otherwise this will not count for test coverage.

Glonee: OK

integrationtests/self/http_hotswap_test.go
marten-seemann, line 1: I think all changes to this file can be reverted.

Glonee: I am afraid not, the original test closed the transport at
https://github.com/quic-go/quic-go/blob/6dcac15a1274c3c9bc250665a4a831eaaf9562a2/integrationtests/self/http_hotswap_test.go#L85
then use transport again, which will cause error if we don't allow the usage of closed transport.

marten-seemann: Got it, thank you!

PR #5327, opened 2025-09-07
fix: return stream frames to pool on error paths

lidel:
Unsure if I got this correctly, but the aim of this PR is to fix potential memory leak where StreamFrames were not returned to sync.Pool in some of error/cancellation paths.

leak surface: frames are allocated via GetStreamFrame() but never returned via PutBack() when streams enter terminal states without going through normal OnAcked() cleanup.

fixed paths:
- CancelWrite(): return frames when reliableOffset == 0
- handleStopSendingFrame(): return frames on STOP_SENDING
- closeForShutdown(): return frames on connection shutdown
- OnLost(): return frame when stream reset with no data sent

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -454,6 +454,18 @@ func (s *SendStream) SetReliableBoundary() {
 	}
 }
 
+// returnFramesToPool returns all queued frames to the sync.Pool
+func (s *SendStream) returnFramesToPool() {
+	for _, f := range s.retransmissionQueue {
+		f.PutBack()
+	}
+	s.retransmissionQueue = nil
+	if s.nextFrame != nil {
+		s.nextFrame.PutBack()
+		s.nextFrame = nil
+	}
+}
+
 // CancelWrite aborts sending on this stream.
 // Data already written, but not yet delivered to the peer is not guaranteed to be delivered reliably.
 // Write will unblock immediately, and future calls to Write will fail.
@@ -485,7 +497,7 @@ func (s *SendStream) CancelWrite(errorCode StreamErrorCode) {
 	reliableOffset := s.reliableOffset()
 	if reliableOffset == 0 {
 		s.numOutstandingFrames = 0
-		s.retransmissionQueue = nil
+		s.returnFramesToPool()
 	}
 	s.queuedResetStreamFrame = &wire.ResetStreamFrame{
 		StreamID:  s.streamID,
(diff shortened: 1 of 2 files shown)


send_stream.go
```diff
@@ -454,6 +454,18 @@ func (s *SendStream) SetReliableBoundary() {
 	}
 }
 
+// returnFramesToPool returns all queued frames to the sync.Pool
+func (s *SendStream) returnFramesToPool() {
+	for _, f := range s.retransmissionQueue {
+		f.PutBack()
+	}
+	s.retransmissionQueue = nil
```
marten-seemann, line 463: You should `clear` the slice as well, otherwise there will still be a pointer to that frame, and the GC won't be able to garbage collect this frame.

send_stream_test.go
marten-seemann, line 1: I don't think these tests add a lot of value, and I'm hesitant to increase the size of the test suite by that much. I'm not sure if pooling needs to be tested in a unit test, since it's just a performance optimization. Can you remove these?

send_stream.go
```diff
@@ -673,6 +686,8 @@ func (s *sendStreamAckHandler) OnLost(f wire.Frame) {
 	// If the reliable size was 0 when the stream was cancelled,
 	// the number of outstanding frames was immediately set to 0, and the retransmission queue was dropped.
 	if s.resetErr != nil && (*SendStream)(s).reliableOffset() == 0 {
+		// Return the frame to pool since it won't be retransmitted
+		sf.PutBack()
```
marten-seemann, line 691: This should be really rare, and is one of the cases where I'd opt for simplicity.

marten-seemann (commented):
Not returning frames to the pool is not a memory leak, they'll simply get picked up by the GC, as any other struct that's not reference anymore would. Therefore using the pool exclusively in the hot path should be fine.

That said, there might be applications that frequently cancel streams, so returning frames to the pool in those cases doesn't seem unreasonable.

send_stream.go
marten-seemann: You don't need to remove it though, this is pretty straight-forward.

send_stream_test.go
lidel: np, done: https://github.com/quic-go/quic-go/pull/5327/commits/44c9e2e6e13e0f7f6090d086773001694a392101

send_stream.go
lidel: my bad, done: https://github.com/quic-go/quic-go/pull/5327/commits/44c9e2e6e13e0f7f6090d086773001694a392101

marten-seemann (approved):
Thank you!

PR #5339, opened 2025-09-16
improve connection timer logic

### LINKED ISSUE #5340: timer not properly reset after idle
The connection timer prevents itself from being set to an earlier timestamp. This is a problem after an idle period, where the timer will be set to the idle timeout (e.g. 30s in the future). When we start sending packets, we definitely need to be able to set a timer for loss recovery purposes.

This logic was introduced to prevent busy-looping, see https://github.com/quic-go/quic-go/issues/3567 and https://github.com/quic-go/quic-go/pull/4943.

marten-seemann:
Fixes #5340.

1. As of Go 1.23, timer resets work as expected, and there’s no need to read a (potentially) stale value from the timer channel. This means we don't need the `utils.Timer` wrapper anymore.
2. We should be able to set the timer to a timestamp earlier than the most recent timestamp it was armed to. This is especially important when resuming from an idle connection.
3. To prevent busy-looping in cases where we're blocked (e.g. because the send queue is full, or because we're congestion limited), we track if we're blocked. If we are, the timer is sent to the idle timeout (we still need to be able to shut down the connection!).


connection.go
```diff
@@ -2118,6 +2145,7 @@ func (c *Conn) triggerSending(now monotime.Time) error {
 	case ackhandler.SendAny:
 		return c.sendPackets(now)
 	case ackhandler.SendNone:
+		c.isBlocked = true
 		return nil
 	case ackhandler.SendPacingLimited:
 		deadline := c.sentPacketHandler.TimeUntilSend()
```
sukunrt, line 2169: This block has a fallthrough at the end. That means when we're pacing limited, we will mark the connection as blocked. Now in `maybeResetTimer` we'll reset to idletimeout as opposed to the pacing deadline.

connection.go
```diff
@@ -2133,6 +2161,7 @@ func (c *Conn) triggerSending(now monotime.Time) error {
 		// We can at most send a single ACK only packet.
 		// There will only be a new ACK after receiving new packets.
 		// SendAck is only returned when we're congestion limited, so we don't need to set the pacing timer.
+		c.isBlocked = true
```
sukunrt: This feels like a No Op to me. 
In the connection run loop. On line 694:

```
		if err := c.triggerSending(now); err != nil {
			c.setCloseError(&closeError{err: err})
			break runLoop
		}
		if c.sendQueue.WouldBlock() {
			// The send queue is still busy sending out packets. Wait until there's space to enqueue new packets.
			sendQueueAvailable = c.sendQueue.Available()
			// Cancel the pacing timer, as we can't send any more packets until the send queue is available again.
			c.pacingDeadline = 0
			c.isBlocked = true
		} else {
			sendQueueAvailable = nil
			c.isBlocked = false
		}
```

Both the if conditions update the blocked state so any update in `triggerSending` is overwritten.

connection.go
```diff
@@ -614,9 +622,9 @@ runLoop:
 			select {
 			case <-c.closeChan:
 				break runLoop
-			case <-c.timer.Chan():
-				c.timer.SetRead()
+			case <-c.timer.C:
 			case <-c.sendingScheduled:
+				c.isBlocked = false
```
sukunrt: Do you need this? This value will be overwritten later in the method anyway.

connection.go
```diff
@@ -1055,6 +1081,7 @@ func (c *Conn) handleOnePacket(rp receivedPacket) (wasProcessed bool, _ error) {
 	}
 
 	p.buffer.MaybeRelease()
+	c.isBlocked = false
```
sukunrt: Do you need this? On receiving a packet, we'll try to send something. That sending logic(`triggerSending` And `sendQueue.WouldBlock()`) checks will set this value correctly.

connection.go
```diff
@@ -675,6 +683,7 @@ runLoop:
 			sendQueueAvailable = c.sendQueue.Available()
 			// Cancel the pacing timer, as we can't send any more packets until the send queue is available again.
 			c.pacingDeadline = 0
```
sukunrt, line 695: You can remove the resetting of `pacingDeadline` here. When we set isBlocked, we'll ignore the pacingDeadline anyway.

connection.go
```diff
@@ -691,8 +700,10 @@ runLoop:
...
```
sukunrt, line 713: Same comment here. You can remove this specific `pacingDeadline` reset as isBlocked will override any pacingDeadline.

connection.go
marten-seemann: What if we're still (or again, after sending out as many bytes as were acknowledged) congestion limited? In that case, we shouldn't rearm the timer.

connection.go
marten-seemann: This one can probably be removed.

marten-seemann: Done in c45ac056c.

connection.go
marten-seemann: Good catch. I think we need to set `isBlocked = false` before `triggerSending`. Then sending might set it to `true` when we're congestion limited, and we can set it to `true` if `sendQueue.WouldBlock()`.

marten-seemann: Done in f51c93ef4.

connection.go
marten-seemann: Oh right, that's smart! Thanks for catching this!

marten-seemann: Actually, we might want to keep it here, just to be on the safe side. I'm worried about cases where we become unblocked and then have a stale deadline.

connection.go
marten-seemann: Done in 2e82cc722.

connection.go
sukunrt: Not sure if that can happen, but this is fine too. Adding this as a comment would be nice.

connection.go
sukunrt: This looks right to me. :heavy_check_mark:

connection.go
sukunrt: That'll be taken care of by the `triggerSending` method, no? 

In any case, this setting seems like a no-op to me, just code structure wise. After we handle any packets, we update c.isBlocked based on `triggerSending` or `sendQueue` unavailability. So this update to isBlocked has no effect.

connection.go
```diff
@@ -816,20 +826,35 @@ func (c *Conn) maybeResetTimer() {
...
+				deadline = keepAliveTime
+			} else {
+				deadline = c.nextIdleTimeoutTime()
+			}
 		}
 	}
+	if c.isBlocked {
+		c.timer.Reset(monotime.Until(deadline))
+		return
+	}
```
sukunrt, line 857: I think this logic is incorrect for the congestion limited case. 

Say we are blocked because of congestion: 
Now the timer should be reset to PTO and not to idletimeout. It should also be set to before keepAliveTimeout not to idletimeout. 
In our case that means we need to consider the following line for setting the timer.
```
if t := c.sentPacketHandler.GetLossDetectionTimeout(); !t.IsZero() && t.Before(deadline) {
		deadline = t
	}
```

Imagine both the sides have 200 packets to send and 100packets of congestion window. Now both the sides immediately send 100 packets. They're both congestion limited and the network drops all these 100 packets from both sides. Now both sides of the connection will just wait till idle time out and then close the connection.

sukunrt (changes requested):

connection.go
```diff
@@ -178,6 +194,12 @@ type Conn struct {
 	versionNegotiated   bool
 	receivedFirstPacket bool
 
+	// isBlocked is set to true when the connection is blocked. This can happen when:
```
sukunrt: stale comment.

marten-seemann: Done.

PR #5352, opened 2025-10-02
ci: remove 386 (32bit x86) from CI

MarcoPolo:
The last 386 cpu was from 20 years ago (Pentium 4)


.github/workflows/cross-compile.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: go build example/main.go
       - name: Run cross compilation
         # run in parallel on as many cores as are available on the machine
-        run: go tool dist list | xargs -I % -P "$(nproc)" .github/workflows/cross-compile.sh %
+        run: go tool dist list | grep -v "386" | xargs -I % -P "$(nproc)" .github/workflows/cross-compile.sh %
```
marten-seemann: Let's keep the compilation step. This doesn't cost us anything, now that the Go build cache is cached and restored.

.github/workflows/integration.yml
```diff
@@ -8,14 +8,12 @@ jobs:
...
         go: [ "1.24.x", "1.25.x" ]
         race: [ false ]
-        use32bit: [ false ]
         include:
           - os: "ubuntu"
             go: "1.25.x"
             race: true
           - os: "ubuntu"
             go: "1.25.x"
-            use32bit: true
```
marten-seemann, line 18: I think this entire configuration can be deleted, otherwise we'll be testing on Ubuntu twice right?

marten-seemann (changes requested):
What about the unit tests (`unit.yml`). There's a 32 bit step there as well.

marten-seemann:
@MarcoPolo Are you planning to address the remaining review comments?

MarcoPolo:
ah sorry, I thought I had pushed

marten-seemann (approved):
Thank you!

PR #5400, opened 2025-10-24
fix: `TestAddTimestamp` parse time in local timezone

### LINKED ISSUE #5399: flaky `TestAddTimestamp` due to timezone difference
My current test run are failing on `TestAddTimestamp`:

```sh
$ go test -v -shuffle on ./...

...
=== RUN   TestAddTimestamp
    log_test.go:98: 
        	Error Trace:	/home/zq2/quic-go/internal/utils/log_test.go:98
        	Error:      	Max difference between 2025-10-23 21:37:27.233847387 -0500 CDT m=+0.007483601 and 2025-10-23 00:00:00 +0000 UTC allowed is 25h0m0s, but difference was 26h37m27.233847387s
        	Test:       	TestAddTimestamp
--- FAIL: TestAddTimestamp (0.00s)
...

$ timedatectl
               Local time: Thu 2025-10-23 21:50:30 CDT
           Universal time: Fri 2025-10-24 02:50:30 UTC
                 RTC time: Fri 2025-10-24 02:50:30
                Time zone: America/Chicago (CDT, -0500)
System clock synchronized: yes
              NTP service: active
          RTC in local TZ: no
```

I believe the root cause is that `time.Now()` is picking up my current timezone while the parser assumes UTC timezone. It's late night here already but technically it's still on the same date. Due to truncation, the hours diff is more than 25 hours.

rMaxiQp:
Current `TestAddTimestamp` parse the given timestamp in UTC. In rare cases (see #5399), the hours diff can be more than 25 hours. As the fix, we should respect the local timezone as we parse the timestamp.

Fix #5399

### DIFF
--- a/internal/utils/log_test.go
+++ b/internal/utils/log_test.go
@@ -93,7 +93,7 @@ func TestAddTimestamp(t *testing.T) {
 	DefaultLogger.SetLogLevel(LogLevelInfo)
 	DefaultLogger.Infof("info")
 	timestamp := b.String()[:b.Len()-6]
-	parsedTime, err := time.Parse(format, timestamp)
+	parsedTime, err := time.ParseInLocation(format, timestamp, time.Local)
 	require.NoError(t, err)
 	require.WithinDuration(t, time.Now(), parsedTime, 25*time.Hour)
 }


marten-seemann:
Not sure if this is worth fixing. Long term, we'll replace the logger with slog anyway.

rMaxiQp:
Hi @marten-seemann, thank you for the quick reply and sharing the extra context. Migrating to `slog` definitely simplify things in the long run.

This PR isn't meant to extend the lifespan of the current logger, but to improve the reliability of existing tests to surface more critical issues and other developers living in similar timezones could help to address. I believe my current change should've fixed the issue already. Thus, there's no future work needed.

PR #5418, opened 2025-11-02
ackhandler: fix qlogging of RTT values

marten-seemann:


internal/ackhandler/sent_packet_handler.go
```diff
@@ -333,6 +333,28 @@ func (h *sentPacketHandler) SentPacket(
 func (h *sentPacketHandler) qlogMetricsUpdated() {
 	var metricsUpdatedEvent qlog.MetricsUpdated
 	var updated bool
+	if h.rttStats.HasMeasurement() {
+		if h.lastMetrics.MinRTT == nil || *h.lastMetrics.MinRTT != h.rttStats.MinRTT() {
+			metricsUpdatedEvent.MinRTT = pointer(h.rttStats.MinRTT())
```
MarcoPolo, line 338: why are these pointers?

marten-seemann: We only want to log them if they're set. `metrics_updated` is also used when sending packets (because we need to update the packets in flight / bytes in flight counter).

marten-seemann: We'll change this in a separate PR.

PR #5427, opened 2025-11-12
add a way for the user to get current datagram limit

### LINKED ISSUE #4259: add a way for the user to get current datagram limit
Right now there's no way for the user to know how big they can make their datagrams before they're [dropped during packing]( https://github.com/quic-go/quic-go/blob/4407c60f046bdf851833df3338bae965a6350292/packet_packer.go#L609) and thus have to size their datagrams to something conservative. Having an API to get current datagram size limit would allow users to utilize available packet space better, when they're not also sending stream data.

solomon-os:
Resolves #4259 

expose maximum datagram size in ConnectionState

Add MaxDatagramSize field to ConnectionState that reports the maximum datagram payload size that can be sent. This value is calculated as the minimum of:
- The peer's advertised datagram frame size limit
- The maximum frame size according to current MTU, connection ID length, and encryption overhead

The calculation uses the actual packet number length (via PeekPacketNumber) rather than the maximum possible length for better accuracy.

Also add an integration test that verifies the reported value works and that value+1 does not work.

cc @marten-seemann @chungthuang

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -764,6 +764,66 @@ func (c *Conn) supportsDatagrams() bool {
 	return c.peerParams.MaxDatagramFrameSize > 0
 }
 
+// calculateMaxDatagramSize calculates the maximum datagram payload size that can be sent.
+// This is the minimum of the peer's advertised limit and the maximum frame size
+// according to the current MTU, connection ID length, and encryption overhead.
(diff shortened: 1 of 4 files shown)


marten-seemann:
Looks like AI slop.

PR #5510, opened 2025-12-29
ci: ignore integrationtests directory for Codecov

marten-seemann:
Recently, the Codecov coverage report uploader has been failing since it was trying (for reasons I don't understand, but might be related to https://docs.codecov.com/docs/fixing-reports) to access files in the `integrationtests` directory.


marten-seemann:
Doesn't work.

PR #5513, opened 2026-01-01
http3: respect context cancellation in http.Response.Body.Read (fix #4205)

### LINKED ISSUE #4205: http3: respect request context while reading from the response body
The Go documentation for the request context says:
> For an outgoing client request, the context controls the entire lifetime of a request and its response: obtaining a connection, sending the request, and reading the response headers and body.

The `http.Response.Body` therefore needs to return the context error as well. Currently, we don't even pass the context to the body, and it's not clear to me how to correctly respect the context without spawning a new Go routine, since the `Stream.Read` call is blocking call.

_Originally posted by @marten-seemann in https://github.com/quic-go/quic-go/pull/4203#discussion_r1426131710_

ismailtrm:
Fixes #4205.

## Problem
Currently, `http.Response.Body.Read` can block indefinitely for HTTP/3 if the request context is canceled while waiting for data. The underlying QUIC stream is not promptly aborted on context cancellation after headers are received, which deviates from expected `net/http` behavior.

## Solution
This PR introduces a context-aware wrapper around the response body stream:

1. **Immediate return**: `Read` checks `ctx.Done()` before blocking calls. If data is available (`n > 0`), it returns the data (preserving partial reads) even if the context is canceled. If no data is read (`n == 0`), the stream signals an error, and the context is canceled, it returns `ctx.Err()`.
2. **Unblock pending reads**: A lightweight goroutine (managed by `newResponseBody` and `setContext`) monitors `ctx.Done()` and calls `CancelRead` on the underlying stream with `H3_REQUEST_CANCELLED` (0x010c). This ensures that any blocked `Read` call on the underlying stream unblocks immediately.
   - **Single Monitor**: Uses `sync.Once` to ensure only one monitoring goroutine is started per body, even if `setContext` is called multiple times (e.g., for 1xx responses).
3. **No leaks**: The monitoring goroutine exits deterministically when `Body.Close()` is called.
4. **Thread Safety**: Access to the context and cancellation options is protected by a mutex.

## Semantics & API Changes
- **RoundTripOpt**: Added `DontCloseRequestStream` field to `RoundTripOpt`.
- **DontCloseRequestStream**: When `true`, the cancellation behavior is disabled. The response body ignores the request context (full detach) so callers can manage stream lifetime manually.
- **Idempotency**: `Body.Close()` remains safe to call multiple times.

## Verification
- `TestBodyReadRespectsContext`: Verifies a hung read unblocks immediately via `CancelRead` when the context is canceled.
- `TestBodyRead_GoroutineLeak`: Verifies that the monitoring goroutine exits when `Body.Close()` is called.
- `TestBodyRead_PartialData`: Verifies that if `Read` retrieves data (`n > 0`), it returns the data without error even if the context is canceled during the call.
- `TestBodyRead_DontCloseRequestStream`: Verifies cancellation is ignored when the option is enabled.
- `TestBodyRead_MultipleSetContext`: Verifies that calling `setContext` multiple times does not leak goroutines or cause multiple cancellations.
- Validated with `go test -v -race ./http3/...`.

### DIFF
--- a/http3/body.go
+++ b/http3/body.go
@@ -95,321 +95,97 @@ func newRequestBody(str *Stream, contentLength int64, connCtx context.Context, r
 }
 
 type hijackableBody struct {
-
 	body body
 
-
-
 	// only set for the http.Response
-
 	// The channel is closed when the user is done with this response:
-
 	// either when Read() errors, or when Close() is called.
-
 	reqDone     chan<- struct{}
-
 	reqDoneOnce sync.Once
 
-
-
 	mu                     sync.Mutex
-
 	ctx                    context.Context
-
 	dontCloseRequestStream bool
 
-
-
-		// closed when Close() is called
-
-
-
-		bodyClosed     chan struct{}
-
-
-
-		bodyClosedOnce sync.Once
-
-
-
-		monitorOnce    sync.Once
-
-
-
-	}
-
-
-
-	
-
-
-
-	var _ io.ReadCloser = &hijackableBody{}
-
-
-
-	
-
-
-
-	func newResponseBody(str *Stream, contentLength int64, done chan<- struct{}) *hijackableBody {
-
-
-
-		return &hijackableBody{
-
-
-
-			body:       *newBody(str, contentLength),
-
-
-
-			reqDone:    done,
-
-
-
-			bodyClosed: make(chan struct{}),
-
-
-
-		}
-
-
-
-	}
-
-
-
-	
-
-
-
-	func (r *hijackableBody) setContext(ctx context.Context, dontClose bool) {
-
-
-
-		r.mu.Lock()
-
-
-
-		r.ctx = ctx
-
-
-
-		r.dontCloseRequestStream = dontClose
-
-
-
-		r.mu.Unlock()
-
-
-
-	
-
-
-
-		if dontClose {
-
-
-
-			return
-
-
-
-		}
-
-
-
-	
-
-
-
-		r.monitorOnce.Do(func() {
-
-
-
-			go func() {
-
-
-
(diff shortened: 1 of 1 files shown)


ismailtrm:
Applied gofmt / formatting cleanup as suggested. No functional changes.

marten-seemann:
Is this vibe-coded? Which AI was used?

ismailtrm:
Yes; I used AI assistance during development (Gemini CLI / Gemini 3 Pro with SpecKit for spec + iterative coding, GPT-5.2 for edge-case reasoning, and Claude for an independent review).

I wanted to validate that this fix is feasible and implementable without regressions, so I kept the final ownership on my side: I applied the changes, ran the http3 tests (incl. -race), and I’m happy to walk through any part of the approach or adjust semantics if you prefer different behavior (e.g., around DontCloseRequestStream).

marten-seemann:
Thank you, this explains why it's trying to reintroducing a questionable API choice from an ancient quic-go version.

ismailtrm:
Got it, thanks. I likely overextended DontCloseRequestStream into a “full detach” (i.e. also ignoring ctx for Response.Body), which I understand you consider a questionable legacy API direction.

My goal was purely to address #4205: ensure Response.Body.Read unblocks promptly on context cancellation and returns ctx.Err().
I can send a follow-up PR that always makes response body reads respect the request context, and treats DontCloseRequestStream strictly as “don’t close the request/upload stream” (no effect on response cancellation).

Would that align better with the intended API direction?

(Also: I’m new to the quic-go community, if there are any help-wanted / good-first issues you’d recommend, I’d love to pick one up.)

PR #5527, opened 2026-01-06
http3: reduce memory allocation in writeHeader

qiulaidongfeng:
goos: windows
goarch: amd64
pkg: github.com/quic-go/quic-go/http3
cpu: AMD Ryzen 7 7840HS w/ Radeon 780M Graphics
               │     old     │              new              │
               │   sec/op    │   sec/op     vs base          │
WriteHeader-16   2.254µ ± 1%   2.245µ ± 1%  ~ (p=0.470 n=10)

               │     old      │                 new                 │
               │     B/op     │     B/op      vs base               │
WriteHeader-16   5.821Ki ± 0%   5.691Ki ± 0%  -2.23% (p=0.000 n=10)

               │    old     │                new                │
               │ allocs/op  │ allocs/op   vs base               │
WriteHeader-16   103.0 ± 0%   100.0 ± 0%  -2.91% (p=0.000 n=10)

### DIFF
--- a/http3/response_writer.go
+++ b/http3/response_writer.go
@@ -8,6 +8,7 @@ import (
 	"net/textproto"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/quic-go/qpack"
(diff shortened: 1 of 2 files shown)


marten-seemann:
This doesn't look wrong in isolation, but I think we'll need to start a larger effort to reduce allocations in the HTTP/3 package. There are a lot of unnecessary allocs here to optimize, and we might need a global buffer pool here, instead of the `bytes.Buffer` pool introduced here.

marten-seemann:
Opened a tracking issue: https://github.com/quic-go/quic-go/issues/5528

PR #5550, opened 2026-01-24
handshake: refactor setupConfigForServer

### LINKED ISSUE #5543: Go 1.25.6 breaks session resumption
Session resumption breaks with Go 1.25.6 (and the various backports performed for the other release branches), unless the application explicitly configured session resumption keys.

See https://github.com/golang/go/issues/77113#issuecomment-3762921882.

It looks like there’s nothing we can do in quic-go, we’ll need the standard library to revert this so-called fix.

rbqvq:
This commit aims to remove the workaround for [tls.Config.Clone]

In addition, the potential infinite loop issue of GetConfigForClient has been resolved.

Fixes #5543

### DIFF
--- a/internal/handshake/tls_config.go
+++ b/internal/handshake/tls_config.go
@@ -3,37 +3,78 @@ package handshake
 import (
 	"crypto/tls"
 	"net"
+	"reflect"
 )
 
-func setupConfigForServer(conf *tls.Config, localAddr, remoteAddr net.Addr) *tls.Config {
-	// Workaround for https://github.com/golang/go/issues/60506.
-	// This initializes the session tickets _before_ cloning the config.
-	_, _ = conf.DecryptTicket(nil, tls.ConnectionState{})
+func setupBaseConfigForServer(parent *tls.Config, fakeConn net.Conn) *tls.Config {
+	child := parent.Clone()
+	child.MinVersion = tls.VersionTLS13
 
-	conf = conf.Clone()
-	conf.MinVersion = tls.VersionTLS13
+	if child.GetCertificate != nil {
+		child.GetCertificate = func(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			info.Conn = fakeConn
+			return parent.GetCertificate(info)
+		}
+	}
+	return child
+}
+
+func setupChildConfigForServer(originalConfig, parent, child *tls.Config) {
+	if child.SessionTicketsDisabled {
+		return
+	}
+
+	if child.WrapSession != nil && child.UnwrapSession != nil {
+		return
+	}
+
+	if originalConfig != nil {
+		sessionTicketKeys := reflect.ValueOf(child).Elem().FieldByName("sessionTicketKeys")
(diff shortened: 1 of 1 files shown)


wwqgtxx:
I don't think it's a good thing to use reflection to access the internal structure of crypto/tls as a relatively basic network library, as this will increase the probability of incompatibility during future Go updates.

rbqvq:
> I don't think it's a good thing to use reflection to access the internal structure of crypto/tls as a relatively basic network library, as this will increase the probability of incompatibility during future Go updates.

So I was thinking that the deleted workaround should be added back as a guarantee scheme.

So we have two scenarios at the same time, and even if either change is modified upstream, we have another method as a fallback.

In addition, struct changes are generally added in RC, and in most cases are not modified, which I think is OK.

wwqgtxx:
Instead, we should actively push for the merge of https://go-review.googlesource.com/c/go/+/737700, avoiding this kind of "security update" that claims to minimize the impact on existing code while simultaneously breaking widely used use cases.

rbqvq:
> Instead, we should actively push for the merge of https://go-review.googlesource.com/c/go/+/737700, avoiding this kind of "security update" that claims to minimize the impact on existing code while simultaneously breaking widely used use cases.

That's true. I'm dealing with these at the same time. Please don't worry.

In fact, we're back to Dec 29, where I submitted the initial patch. 😆

Actually the reflection part is to handle the SessionTicketKeys fallback behavior of GetConfigForClient, which was not handled before.

It's not bad for us to add (

wwqgtxx:
As a library, it should not rely on the internal private structure of crypto/tls unless absolutely necessary, even if this private structure has remained unchanged for many years. After all, this is not a personal used CLI toy, but a QUIC library used by many large projects.

rbqvq:
As I said in https://github.com/golang/go/issues/77113#issuecomment-3794400419

I don't have to other way to resolve the keys fallback behavior, its only one resolution.

Let marten decide.

wwqgtxx:
In reality, quic-go only needs to set the remote and local addresses, which can be accomplished in the future by extending tls.QUICConfig (although this may not happen until go 1.27). (Ref: https://github.com/golang/go/issues/77113#issuecomment-3780707199)

Clearly, the existing quic-go code will work fine once that "security update" is actually fixed, so there's no need to add more strange code just to fix this minor version regression.

marten-seemann:
I don't like using reflect and I don't like accessing private member variables. We've done that before and it has caused a huge amount of pain for everyone involved.

rbqvq:
I don't like it either, but this is how Get ConfigForClient behaves.

They also don't provide any API that doesn't rely on cfg.Clone to solve the problem.

Unless we continue to use cfg.Clone.

:(

---

Actually, if I remember correctly, the early quic-go was using unsafe and linkname.

There seems to be no problem using reflect here.

rbqvq:
If you decide whether to continue to rely on cfg.Clone, please let me know and I will update this PR.

marten-seemann:
#5543 was fixed by Go 1.25.7.

PR #5553, opened 2026-01-27
http3: Add Date and Content-Type headers like std

### LINKED ISSUE #5546: Date response header getting added too early
For http3 server,
The date response header is added prior to the handler.

In net/http and fasthttp, the date response header gets added at the end, after processing the handler.

qiulaidongfeng:
Fixes #5546


marten-seemann (changes requested):
This breaks content type sniffing, doesn’t it?

Issue #5551, opened 2026-01-25
http.Request.Context().Done() and http.Request.Context().Err() not working


pjebs:
When a connection is closed by the client, the http.Request.Context().Done() channel is not closed
and http.Request.Context().Err() is still nil.

In net/http server, that is not the case.

pjebs:
This seems to be what I''m after (untested): https://pkg.go.dev/github.com/quic-go/quic-go/http3#ClientConn.Context

But no idea how to access it from a `*http.Request`, which is all I have from the handler.

pjebs:
The ultimate aim is to detect when the client has closed and the request is no longer relevant to process and write responses (for a server side events endpoint)

pjebs:
As a intermediate solution, can the response writer's `Write()` function return an error if the connection is lost.
Right now a handler serving a server-sent-events endpoint leaks forever.

marten-seemann:
This issue seems to be invalid as well. The request context is derived from the connection context and the following test shows that the context is properly cancelled:

```go
func TestServerRequestContextCanceledOnConnectionClose(t *testing.T) {
	clientConn, serverConn := newConnPair(t)
	str, err := clientConn.OpenStream()
	require.NoError(t, err)
	_, err = str.Write(encodeRequest(t, httptest.NewRequest(http.MethodHead, "https://www.example.com", nil)))
	require.NoError(t, err)

	done := make(chan struct{})
	s := &Server{
		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			<-r.Context().Done()
			close(done)
		}),
	}

	go s.ServeQUICConn(serverConn)

	select {
	case <-done:
		t.Fatal("request context was canceled prematurely")
	case <-time.After(scaleDuration(10 * time.Millisecond)):
	}

	clientConn.CloseWithError(0, "client disconnect")

	select {
	case <-done:
	case <-time.After(time.Second):
		t.Fatal("timeout: request context should be cancelled when connection closes")
	}
}
```

pjebs:
It's not invalid. I'll submit a demo project

PR #5556, opened 2026-02-01
cancel stream context when the connection is closed

zvdy:
fix for #5551 

https://datatracker.ietf.org/doc/html/rfc9000#name-connection-termination not implicitly documented, but, I like the idea of matching `net/http` as well as it makes sense to me to not leak.

Also:
- Close
- CancelWrite
- handleStopSendingFrame

Already cancel ctx.

When a connection is closed abruptly, `closeForShutdown()` is called on all streams but was not cancelling the context. Thus making it impossible to detect those disconnections. The fix simply cancels the context and adds a test to cover this scenario

> [!NOTE]
> ### Cancel `SendStream.closeForShutdown` context with the shutdown error to fix leak in [send_stream.go](https://github.com/quic-go/quic-go/pull/5556/files#diff-0683df4767cffeaa4b1127b3e6027900d06dac7973a80454782e810871af742e)
> Add `s.ctxCancel(err)` after unlocking in `SendStream.closeForShutdown` and assert context cancellation and `context.Cause(str.Context())` in [send_stream_test.go](https://github.com/quic-go/quic-go/pull/5556/files#diff-a4ea0a3098177eeb5ef44b94781f438cbaea993839bce8765fc44139c3820d62).
>
> #### 📍Where to Start
> Start with `SendStream.closeForShutdown` in [send_stream.go](https://github.com/quic-go/quic-go/pull/5556/files#diff-0683df4767cffeaa4b1127b3e6027900d06dac7973a80454782e810871af742e).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3ddb08c.</sup>
> 
>

### DIFF
--- a/http3/server_test.go
+++ b/http3/server_test.go
@@ -465,45 +465,6 @@ func TestServerRequestContext(t *testing.T) {
 	close(block)
 }
 
-func TestServerRequestContextCanceledOnConnectionClose(t *testing.T) {
-	clientConn, serverConn := newConnPair(t)
-	str, err := clientConn.OpenStream()
-	require.NoError(t, err)
-	_, err = str.Write(encodeRequest(t, httptest.NewRequest(http.MethodHead, "https://www.example.com", nil)))
-	require.NoError(t, err)
-
-	ctxChan := make(chan context.Context, 1)
-	s := &Server{
-		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			ctxChan <- r.Context()
-			<-r.Context().Done()
-		}),
-	}
-
-	go s.ServeQUICConn(serverConn)
-
-	var requestContext context.Context
-	select {
-	case requestContext = <-ctxChan:
-	case <-time.After(time.Second):
-		t.Fatal("timeout waiting for request context")
-	}
-
-	select {
-	case <-requestContext.Done():
-		t.Fatal("request context was canceled prematurely")
-	case <-time.After(scaleDuration(10 * time.Millisecond)):
-	}
-
-	clientConn.CloseWithError(0, "client disconnect")
-
-	select {
-	case <-requestContext.Done():
-	case <-time.After(time.Second):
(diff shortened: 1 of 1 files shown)


http3/server_test.go
```diff
@@ -465,6 +465,47 @@ func TestServerRequestContext(t *testing.T) {
...
+	require.NoError(t, err)
+	_, err = str.Write(encodeRequest(t, httptest.NewRequest(http.MethodHead, "https://www.example.com", nil)))
+	require.NoError(t, err)
+
+	ctxChan := make(chan context.Context, 1)
+	block := make(chan struct{})
+	s := &Server{
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctxChan <- r.Context()
+			<-block
```
marten-seemann: Why are you blocking here?

marten-seemann (commented):
The fix looks correct to me, but I need to fix CI first (see #5543).

zvdy:
Will rebase once a fix is merged, thanks for letting me know

marten-seemann:
@zvdy I just updated the CI config to the new Go version. Can you please rebase your PR?

http3/server_test.go
marten-seemann: @zvdy Can you address this comment please?

zvdy: rebased and addressed this, honestly I dont even remember why I've done it like this, perhaps wanted to keep the handler alive during the test, but you're right - I've updated it to Context Done instead to handle it properly.

zvdy:
@marten-seemann rebased and CI passes, also addressed the test, please take a look

marten-seemann:
Despite #5551 being invalid, this fix is correct. We just need to remove the HTTP/3 test.

zvdy:
@marten-seemann I suspect flaky test, change is unrelated with MITM test, a re run should do the job

PR #5557, opened 2026-02-04
datagram: use RingBuffer for receive queue to prevent memory leak

ocean2811:
Previously, the datagram receive queue (`rcvQueue`) was implemented as a slice of byte slices (`[][]byte`). Elements were removed using the slicing operation `queue = queue[1:]`.

In Go, re-slicing from the front does not shrink the underlying array's capacity. As new datagrams are appended and old ones are removed, the capacity of the underlying array grows indefinitely, potentially leading to a memory leak even if the queue length (`len`) stays within the limit.

This commit replaces the slice implementation with `ringbuffer.RingBuffer`, which handles memory recycling correctly and prevents unbounded capacity growth.

### DIFF
--- a/datagram_queue.go
+++ b/datagram_queue.go
@@ -20,7 +20,7 @@ type datagramQueue struct {
 	sent      chan struct{} // used to notify Add that a datagram was dequeued
 
 	rcvMx    sync.Mutex
-	rcvQueue [][]byte
+	rcvQueue ringbuffer.RingBuffer[[]byte]
 	rcvd     chan struct{} // used to notify Receive that a new datagram was received
 
 	closeErr error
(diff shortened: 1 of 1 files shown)


marten-seemann:
> In Go, re-slicing from the front does not shrink the underlying array's capacity. As new datagrams are appended and old ones are removed, the capacity of the underlying array grows indefinitely, potentially leading to a memory leak even if the queue length (len) stays within the limit.

This statement is not correct. When the slice capacity is reached on append, the old array is gc-ed. There’s no memory leak here.

ocean2811:
You are right that the backing array is eventually GC. However, the q = q[1:] idiom has two specific downsides compared to the RingBuffer approach used in sendQueue:
﻿
Ghost References (Delayed GC): The slicing operation doesn't nil out the pointer in the underlying array. This pins the consumed []byte payloads in memory until the next reallocation event, causing unnecessary memory retention.
﻿
Capacity Hysteresis: The slice's backing array doesn't shrink automatically. A temporary traffic burst that expands cap will cause that large memory block to linger even after len drops back to 1.
﻿
The RingBuffer implementation solves the reference retention issue (by overwriting slots `r.ring[r.headPos] = *new(T)`) and ensures stable memory usage without O(N) copying overheads.

PR #5564, opened 2026-02-11
handshake: use the tls.QUICErrorEvent added in Go 1.26

marten-seemann:


internal/handshake/quic_event_go125.go
```diff
@@ -0,0 +1,11 @@
+//go:build go1.25 && !go1.26
+
+package handshake
+
+import "crypto/tls"
+
+const quicErrorEvent tls.QUICEventKind = -1
+
+func extractQUICEventError(ev tls.QUICEvent) error {
```
marten-seemann:
```suggestion
func extractQUICEventError(tls.QUICEvent) error {
```

PR #5644, opened 2026-05-12
http3: don't negotiate gzip for CONNECT requests

### LINKED ISSUE #4410: http3: only apply Content Encoding for responses that have content
In particular, Content Encoding MUST NOT be applied to CONNECT (and Extended CONNECT) requests: Capsules are not content, and therefore Content Encoding doesn't apply (i.e. they're not gzipped even if gzip is negotiated).

SAY-5:
Content-Encoding does not apply to CONNECT or Extended CONNECT requests, since capsules are not content, so the client should not send `Accept-Encoding: gzip` on those requests even when compression is enabled. This skips gzip negotiation when the request method is CONNECT. Closes #4410.

> [!NOTE]
> ### Fix gzip negotiation to skip `Accept-Encoding` for CONNECT requests in http3
> CONNECT and Extended CONNECT requests do not use Content-Encoding, so requesting gzip via `Accept-Encoding` is incorrect. In [stream.go](https://github.com/quic-go/quic-go/pull/5644/files#diff-78b9dd15996d79e5718e0b1a031fd73b67983fceb406305fb50ec797dbb558de), `s.isConnect` is now set before the gzip negotiation check, and the condition skips `requestedGzip` when the method is CONNECT. A test in [stream_test.go](https://github.com/quic-go/quic-go/pull/5644/files#diff-543e33e12efd3a6cd27e7cb27b437c4c5533906aee74e8f71a069a0959f6beef) verifies `Accept-Encoding` is present for GET and absent for CONNECT.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 55a940a.</sup>
> 
>

### DIFF
--- a/http3/stream.go
+++ b/http3/stream.go
@@ -296,11 +296,13 @@ func (s *RequestStream) sendRequestHeader(req *http.Request) error {
 	if s.sentRequest {
 		return errors.New("http3: invalid duplicate use of RequestStream.SendRequestHeader")
 	}
-	if !s.disableCompression && req.Method != http.MethodHead &&
+	s.isConnect = req.Method == http.MethodConnect
+	// Content-Encoding does not apply to CONNECT (and Extended CONNECT) requests:
+	// capsules are not content, so they are never gzipped even if gzip is negotiated.
+	if !s.disableCompression && req.Method != http.MethodHead && !s.isConnect &&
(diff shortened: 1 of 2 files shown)


marten-seemann (changes requested):
This PR is wrong. Offering gzip on a CONNECT steam is totally valid, and even beneficial, in case the server rejects the CONNECT request.

PR #5647, opened 2026-05-17
add Conn.MaxDatagramPayloadSize to query current datagram limit

### LINKED ISSUE #4259: add a way for the user to get current datagram limit

z9z:
Closes #4259.
Exposes the maximum datagram payload size that SendDatagram will accept, so callers can size payloads without guessing. The value was already computed internally; this just makes it queryable.

> [!NOTE]
> ### Add `Conn.MaxDatagramPayloadSize` to query the current datagram payload limit
> Adds a `MaxDatagramPayloadSize()` method to `Conn` that returns the current maximum sendable datagram payload size as `int64`, or `0` if datagrams are not supported by the peer. The computation is extracted into a new `currentMaxDatagramPayloadSize()` helper, which is also used by `SendDatagram` to replace its previous inline calculation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e12e70.</sup>
> 
>

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -3026,19 +3026,30 @@ func (c *Conn) SendDatagram(p []byte) error {
 		return errors.New("datagram support disabled")
 	}
 
-	f := &wire.DatagramFrame{DataLenPresent: true}
-	// The payload size estimate is conservative.
-	// Under many circumstances we could send a few more bytes.
-	maxDataLen := min(
-		f.MaxDataLen(c.peerParams.MaxDatagramFrameSize, c.version),
-		protocol.ByteCount(c.currentMTUEstimate.Load()),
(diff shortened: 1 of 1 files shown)


marten-seemann:
Making an API change is never just a "just". Please don't submit code without agreeing on the API first.

PR #5649, opened 2026-05-21
integrationtests: add test for datagram size limit including overhead

jinq0123:
## Summary

- Adds a test `TestDatagramSizeLimitIncludesPacketOverhead` that verifies `MaxDatagramPayloadSize` accounts for QUIC packet overhead (1 byte type + dest connID + packet number + AEAD authentication tag)
- The test sends a datagram at the reported `MaxDatagramPayloadSize` limit and confirms it is actually deliverable

## Test plan

- [x] Run the new test locally: `go test -v -run TestDatagramSizeLimitIncludesPacketOverhead ./integrationtests/self/`

> [!NOTE]
> ### Add integration test for datagram size limit including packet overhead
> Adds `TestDatagramSizeLimitIncludesPacketOverhead` in [datagram_test.go](https://github.com/quic-go/quic-go/pull/5649/files#diff-dcd6bfe78bf00f3b683ab9cd006bed2a3cea29b31efbacf249706a138c802bfb) to verify that `MaxDatagramPayloadSize` already accounts for packet and AEAD overhead. The test sends an oversized datagram to retrieve the limit via `DatagramTooLargeError`, then sends a datagram exactly at that size and confirms the server receives it intact.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1a63226.</sup>
> 
>

PR #5650, opened 2026-05-21
fix maximum datagram size estimation after MTU discovery

### LINKED ISSUE #4983: Large egress datagrams are silently being dropped

jinq0123:
Fixes #4983.

## Summary

- Fixes a bug where `MaxDatagramPayloadSize` incorrectly calculated the maximum datagram size after MTU discovery
- The bug was introduced in PR #4941 which stored the raw MTU value instead of the payload size (MTU minus header and AEAD overhead)
- Also renames `currentMTUEstimate` back to `maxPayloadSizeEstimate` for clarity

## Changes

- In `handleAckFrame`: use `estimateMaxPayloadSize(mtu)` when updating after MTU discovery instead of raw MTU
- Rename `currentMTUEstimate` to `maxPayloadSizeEstimate` to accurately reflect what it stores

## Test plan

- [x] Run `TestDatagramSizeLimitIncludesPacketOverhead` locally: PASS
- [x] Run all datagram tests: PASS

## Related

- Fixes the test added in PR #5649
- Reverts behavior to match what existed before PR #4941

> [!NOTE]
> ### Fix maximum datagram size estimation after MTU discovery in `Conn`
> - Renames `currentMTUEstimate` to `maxPayloadSizeEstimate` in [connection.go](https://github.com/quic-go/quic-go/pull/5650/files#diff-25a7cf08cc8fc8eb57475835cc380524de2fd1215cda9c6915132ae86dbd80a7) and stores the result of `estimateMaxPayloadSize(mtu)` rather than the raw MTU value when a path MTU probe is acknowledged.
> - `SendDatagram` now caps payload size against the estimated payload size (MTU minus overhead) instead of the raw MTU, preventing overestimation after MTU discovery.
> - Adds `TestDatagramSizeLimitWithMTUDiscovery` in [datagram_test.go](https://github.com/quic-go/quic-go/pull/5650/files#diff-dcd6bfe78bf00f3b683ab9cd006bed2a3cea29b31efbacf249706a138c802bfb) to verify that datagram size limits track MTU discovery events and that payloads above the limit produce `DatagramTooLargeError`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e0a416b.</sup>
> 
>

### DIFF
--- a/integrationtests/self/datagram_test.go
+++ b/integrationtests/self/datagram_test.go
@@ -147,7 +147,7 @@ func TestDatagramSizeLimitIncludesPacketOverhead(t *testing.T) {
 	require.NoError(t, err)
 	defer server.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), scaleDuration(5*time.Second))
 	defer cancel()
 	clientConn, err := quic.Dial(
 		ctx,
@@ -167,7 +167,7 @@ func TestDatagramSizeLimitIncludesPacketOverhead(t *testing.T) {
 	defer serverConn.CloseWithError(0, "")
 
  	// wait for the MTU discovery to complete	
-	time.Sleep(time.Millisecond * 500)
+	time.Sleep(scaleDuration(time.Millisecond * 500))
 
 	// Find the maximum payload size reported by DatagramTooLargeError
 	var maxPayloadSize protocol.ByteCount
@@ -183,7 +183,7 @@ func TestDatagramSizeLimitIncludesPacketOverhead(t *testing.T) {
 	require.NoError(t, err)
 
 	// Verify the datagram was actually delivered and data matches
-	serverCtx, serverCancel := context.WithTimeout(ctx, 500*time.Millisecond)
+	serverCtx, serverCancel := context.WithTimeout(ctx, scaleDuration(500*time.Millisecond))
 	defer serverCancel()
 	datagram, err := serverConn.ReceiveDatagram(serverCtx)
 	require.NoError(t, err, "datagram should be actually deliverable when respecting MaxDatagramPayloadSize")
--- a/integrationtests/self/datagram_test.go
+++ b/integrationtests/self/datagram_test.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/quic-go/quic-go"
+	"github.com/quic-go/quic-go/internal/protocol"
 	"github.com/quic-go/quic-go/internal/wire"
 	"github.com/quic-go/quic-go/testutils/simnet"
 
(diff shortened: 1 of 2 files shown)


integrationtests/self/datagram_test.go
```diff
@@ -127,6 +128,69 @@ func TestDatagramSizeLimit(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+	defer clientConn.CloseWithError(0, "")
+
+	serverConn, err := server.Accept(ctx)
+	require.NoError(t, err)
+	defer serverConn.CloseWithError(0, "")
+
+ 	// wait for the MTU discovery to complete	
+	time.Sleep(time.Millisecond * 500)
```
marten-seemann: Sleeping is not an appropriate way to ensure that MTU discovery has concluded.

marten-seemann (commented):
Can you (you, not an AI) explain this change?

I don't understand the proposed fix. The MTU discoverer deals with packet sizes, not with payload sizes, so renaming `currentMTUEstimate` to `maxPayloadSizeEstimate` seems wrong.

Also, I might be missing something, but does this PR do anything more than just rename the variable?

integrationtests/self/datagram_test.go
jinq0123: Add scaleDuration() to make it more sure. But there is not 100% sure. Any suggestions?

jinq0123: Simple explaination:

Ignoring the renaming and the unit test, the fix is only one line:
```go
-			c.currentMTUEstimate.Store(uint32(mtu))
+			c.maxPayloadSizeEstimate.Store(uint32(estimateMaxPayloadSize(mtu)))
```

Previously, DatagramTooLargeError is initialized with currentMTUEstimate, but should be the maxPayloadSizeEstimate. currentMTUEstimate is 37 bytes larger than maxPayloadSizeEstimate.
```go
type DatagramTooLargeError struct {
	MaxDatagramPayloadSize int64
}
```

TestDatagramSizeLimitIncludesPacketOverhead() shows the error fails the sending of MaxDatagramPayloadSize bytes datagram.

Renaming currentMTUEstimate to maxPayloadSizeEstimate is correct, because it is storing the value of estimateMaxPayloadSize(mtu) not MTU, and is setting to DatagramTooLargeError.MaxDatagramPayloadSize.

marten-seemann: Sleeping is inherently racy. What you should do instead is use an event recorder and wait for the `MTUUpdated` event. It also makes sense to use the `synctest` package.

marten-seemann:
@jinq0123 Can you please resolve the remaining comments? I'd like to include this in the v0.60.0 release.

PR #5679, opened 2026-06-13
feat: add TriggerPTO API to force immediate PTO handling

### LINKED ISSUE #5481: Proposal: API to skip current PTO wait and trigger immediate probe
I noticed that after a successful reconnection, quic-go does not immediately resume communication. Instead, it waits for the PTO (Probe Timeout) to expire. As a result, even after the network is restored, it can take a considerable amount of time before the connection is fully reestablished.

It would be very helpful if quic-go could provide an API that allows the application to explicitly skip the current PTO wait and trigger an immediate probe (i.e., behave as if the PTO just fired). This could significantly improve recovery latency in scenarios where the application is able to detect that network connectivity has already been restored.

Is this something that could be considered, or is there an existing or recommended way to achieve this?

lonnywong:
Fixes #5481.

Add a `Conn.TriggerPTO()` API that allows applications to request immediate PTO handling instead of waiting for the current PTO timer to expire.

This is intended for cases where the application has external knowledge that connectivity may be available again and wants to expedite QUIC's recovery and probing logic.

> [!NOTE]
> ### Add `TriggerPTO` method to `Conn` to force immediate PTO handling
> - Adds a new exported `Conn.TriggerPTO` method that sets an atomic `ptoRequested` flag and calls `scheduleSending` to wake the run loop.
> - The run loop in `Conn.run` now checks `ptoRequested` in addition to the scheduled timeout, invoking `sentPacketHandler.OnLossDetectionTimeout` immediately when the flag is set.
> - Behavioral Change: the loss detection path can now be triggered externally without waiting for the PTO timer to expire.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1308e6f.</sup>
> 
>

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -223,6 +223,10 @@ type Conn struct {
 	connStateMutex sync.Mutex
 	connState      ConnectionState
 
+	// ptoRequested requests an immediate execution of PTO handling,
+	// regardless of the current PTO timer state.
+	ptoRequested atomic.Bool
+
 	logID     string
 	qlogTrace qlogwriter.Trace
 	qlogger   qlogwriter.Recorder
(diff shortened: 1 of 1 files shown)


marten-seemann:
This feels deeply, deeply wrong. We are not going to expose any API to mess with the internal QUIC PTO logic.

PR #5692, opened 2026-06-27
http3: add MaxHeaderValueCount

### LINKED ISSUE #5689: http3: add support for Server.MaxHeaderValueCount
See https://github.com/golang/go/issues/79936.

We don't need to wait for support from the standard library. We can already implement this.

FirstSS-Sub:
Fixes #5689.

This adds `Server.MaxHeaderValueCount` and applies it when parsing request headers and request trailers.

This adds a limit on the number of decoded header fields, in addition to the existing `MaxHeaderBytes` limit. The default is local for now because `net/http` doesn't expose a value for this yet.

If this approach looks right, I'll add tests for both request headers and request trailers.

> [!NOTE]
> ### Add `MaxHeaderValueCount` limit to HTTP/3 server header parsing
> - Adds a `MaxHeaderValueCount` field to the `Server` struct, defaulting to 500, which limits the number of header fields allowed per request (initial headers and trailers).
> - Propagates the limit through `requestFromHeaders` and `decodeTrailers` in [http3/headers.go](https://github.com/quic-go/quic-go/pull/5692/files#diff-fd2fc9252b467e3ef8b4b4a45eafc66de3fab721b4de75b705da663b3e0f3160); exceeding it returns `errHeaderTooLarge`.
> - Client-side trailer decoding passes `math.MaxInt` to disable the limit for responses.
> - Behavioral Change: servers without explicit configuration now enforce a 500 header value limit per request by default, where previously no such limit existed.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized aa66760. 4 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/http3/client.go
+++ b/http3/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"math"
 	"net/http"
 	"net/http/httptrace"
 	"net/textproto"
@@ -188,7 +189,7 @@ func (c *ClientConn) openRequestStream(
 	trace := httptrace.ContextClientTrace(ctx)
 	return newRequestStream(
 		newStream(hstr, c.rawConn, trace, func(r io.Reader, hf *headersFrame) error {
-			hdr, err := decodeTrailers(r, hf, maxHeaderBytes, c.decoder, c.qlogger, str.StreamID())
+			hdr, err := decodeTrailers(r, hf, maxHeaderBytes, math.MaxInt, c.decoder, c.qlogger, str.StreamID())
 			if err != nil {
 				return err
 			}
--- a/http3/headers.go
+++ b/http3/headers.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"net/textproto"
 	"net/url"
@@ -51,7 +52,7 @@ var invalidHeaderFields = [...]string{
 	"upgrade",
 }
 
-func parseHeaders(decodeFn qpack.DecodeFunc, isRequest bool, sizeLimit int, headerFields *[]qpack.HeaderField) (header, error) {
+func parseHeaders(decodeFn qpack.DecodeFunc, isRequest bool, sizeLimit, maxHeaderValueCount int, headerFields *[]qpack.HeaderField) (header, error) {
 	hdr := header{Headers: make(http.Header)}
 	var readFirstRegularHeader, readContentLength bool
 	var contentLengthStr string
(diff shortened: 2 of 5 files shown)


marten-seemann:
Vibe-coded?

FirstSS-Sub:
@marten-seemann
What do you mean? That is not the case.

I have reviewed both the issue in this repository and the corresponding Go issue, and I understand this to be a problem the maintainers want addressed. I wrote 100% of the code myself.

I am well aware of the concerns regarding AI-generated code, but I would appreciate your feedback on the changes themselves. If this project only accepts contributions from trusted contributors, please state that clearly.

marten-seemann:
Producing code these days is cheap.

Picking up random issues and contributing a fix provides (and this is different from pre-LLM times) close to zero value to a project. Quite the opposite, it shifts the burden of reviewing the PR onto maintainers.

If you're using quic-go in production, and you're actually experiencing an issue, I'm happy to review a patch (I've done this a lot of times in the past).

FirstSS-Sub:
@marten-seemann 
Of course, I understand what you are saying. I did not do this because I wanted to take up your time when you are busy.

However, am I understanding this correctly: even though this is OSS, and even though the issue was opened by you as the maintainer, that does not mean that PRs from external contributors are accepted for it? In other words, PRs for this issue were not particularly being accepted?

If that is the case, I would still like that to be stated clearly. Since this was an issue opened by a maintainer, I seriously worked on it with my own coding and tried to solve it. So I was very sad about this.

Thank you for taking the time to reply.

marten-seemann:
OSS has changed with LLMs. Writing code is not the hard part anymore. Reviewing code is. Deciding what to implement and what not to implement is.

For for example, it makes zero sense to implement to to resolve this issue right now, given that Go 1.27 will contain this functionality (there was a freeze exception). It makes a lot of sense to just wait for for a couple of weeks and see what Go implements, so that this new option behaves equally on all HTTP versions.

FirstSS-Sub:
@marten-seemann 
> it makes zero sense to implement to to resolve this issue right now, given that Go 1.27 will contain this functionality (there was a freeze exception)

I understand that. But if that is the case, I still do not understand why this issue was opened in the first place.
(It was because the comment stated, it stated: "We don't need to wait for support from the standard library. We can already implement this.")

As I would with other OSS projects, I simply worked on an issue opened by a maintainer. However, it seems that this was either a joke or just a personal note, and that in quic-go it was not actually an "Issue" in the sense I understood it. It seems I failed to understand that.

At the very least, I now understand that external contributors working on opened issues are not welcome here, so I would like to end this here. I also do not want to take any more of your valuable time.

I wish you the best. Thank you.

marten-seemann:
Not every issue needs immediate resolution 🤷‍♂️ Doesn't mean that there's no value in tracking it. Seems like you're very new to OSS...

PR #5734, opened 2026-07-13
ackhandler: adapt loss detection thresholds to observed packet reorderin

reusu:
### The bug

Packet-threshold loss detection uses a fixed reordering threshold of 3 (`packetThreshold`,
the initial value suggested by RFC 9002). On network paths that reorder UDP datagrams —
per-packet ECMP load balancing, WiFi/LTE, TUN/VPN stacks — every packet displaced by 3 or
more is declared lost **even though it is delivered**, and every one of those spurious
losses puts the connection into recovery and halves the congestion window. The connection
never learns: the threshold is a constant, so the same misjudgment repeats forever and the
congestion window stays pinned near the minimum.

This is not a corner case; it makes QUIC lose badly to TCP on the same path, because
kernel TCP stacks adapt their reordering tolerance (RACK's `reo_wnd`, RFC 8985) and
therefore don't collapse.

Measured on a real path (consumer broadband → cloud VM, tunnel proxying a bulk transfer,
qlog on both endpoints):

* 17.5% of packets arrived out of order, with a displacement of 1–23 packet numbers.
* **Zero real loss**: cross-referencing both endpoints' qlogs, 200 out of 200 packets the
  sender declared lost (~20/s, 90% trigger `reordering_threshold`) were actually received.
* The congestion window was pinned at ~10 KB with the connection in recovery almost
  continuously; throughput was stuck at ~1.4 Mbit/s of a 3.4 Mbit/s link (~40%), while a
  TCP connection saturated the very same link.

RFC 9002, section 6.1.1 explicitly anticipates this failure mode and its fix:

> Implementations can detect spurious retransmissions and increase the reordering
> threshold in packets or time to reduce future spurious retransmissions and loss events.
> [...] Algorithms that increase the reordering threshold after spuriously detecting
> losses, such as RACK [RFC8985], have proven to be useful in TCP and are expected to be
> at least as useful in QUIC.

### The fix

`detectSpuriousLosses` already computes the reordering depth (`maxPacketReordering`) and
extra delay (`maxTimeReordering`) of every spurious loss, but only reports them via qlog.
This PR feeds those observations back into loss detection:

* the packet threshold becomes a per-connection value: it starts at 3 and is raised past
  the deepest observed spurious reordering, capped at 64;
* the time threshold gains a cushion covering the observed extra delay, capped at one
  smoothed RTT.

Both values only ever increase over the lifetime of a connection, and paths that don't
reorder keep exactly the RFC-default behavior. Real loss remains detected by the time
threshold and PTOs even if deep reordering effectively disables packet-based detection.

With the adaptation active, the transfer described above saturates the link
(3.2–3.6 Mbit/s, on par with TCP), and spurious loss declarations drop by an order of
magnitude.

### Tests

* New `TestSentPacketHandlerAdaptivePacketThreshold`: a spurious loss with reordering
  depth 9 raises the threshold to 10; the same reordering pattern afterwards no longer
  declares any packet lost.
* One expectation in `TestSentPacketHandlerSpuriousLoss` updated (commented in the test):
  after the second ACK reveals a reordering depth of 15, the third ACK no longer declares
  two shallowly-reordered packets (depth 3–4) lost.

Happy to split the time-threshold cushion into a separate PR if you'd prefer to take the
packet-threshold adaptation on its own.

> [!NOTE]
> ### Adapt loss detection thresholds to observed packet reordering in `sentPacketHandler`
> - Replaces the fixed `packetThreshold=3` constant with an adaptive field, starting at 3 and capping at 64, which increases when spurious losses reveal deeper reordering.
> - Adds a `timeThresholdCushion` field that widens the time-based loss detection window (up to one smoothed RTT) when observed time reordering exceeds the computed loss delay.
> - Both thresholds are updated in `detectSpuriousLosses` and consumed in `detectLostPackets`, following RFC 9002 §6.1.1 recommendations.
> - Behavioral Change: previously lost packets may no longer be declared lost once thresholds adapt, and future loss declarations can be delayed relative to prior behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 112fa0c.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loss detection for 1-RTT application data under heavy packet reordering.
  * Reduced false packet loss reports by dynamically adjusting both reordering and timing thresholds based on observed behavior.
  * Updated timing-based detection to better account for variable delivery delays and avoid premature loss marking.
* **Tests**
  * Added coverage for adaptive threshold growth, including proper upper-bound limiting and time-cushion behavior after spurious loss scenarios.

### DIFF
--- a/internal/ackhandler/sent_packet_handler.go
+++ b/internal/ackhandler/sent_packet_handler.go
@@ -20,7 +20,14 @@ const (
 	// Specified as an RTT multiplier.
 	timeThreshold = 9.0 / 8
 	// Maximum reordering in packets before packet threshold loss detection considers a packet lost.
-	packetThreshold = 3
+	// This is the initial value: when spurious losses reveal that the path reorders more deeply,
+	// the per-connection threshold adapts upward (see detectSpuriousLosses), as recommended by
+	// RFC 9002, section 6.1.1.
+	initialPacketThreshold = 3
+	// The adaptive packet threshold is never raised beyond this value. Even if packet-based
+	// loss detection is effectively disabled by deep reordering, time-based detection and
+	// PTOs still detect real losses.
+	maxPacketThreshold = 64
 	// Before validating the client's address, the server won't send more than 3x bytes than it received.
 	amplificationFactor = 3
 	// We use Retry packets to derive an RTT estimate. Make sure we don't set the RTT to a super low value yet.
@@ -93,6 +100,15 @@ type sentPacketHandler struct {
 	rttStats   *utils.RTTStats
 	connStats  *utils.ConnectionStats
 
+	// packetThreshold is the packet-based loss detection threshold (RFC 9002, section 6.1.1).
+	// It starts at initialPacketThreshold and is raised (never lowered) when spurious losses
+	// show that the path reorders packets more deeply, similar to TCP RACK's reordering
+	// window (RFC 8985).
+	packetThreshold protocol.PacketNumber
+	// timeThresholdCushion widens the time-based loss detection delay when spurious losses
+	// show that packets arrive later than timeThreshold * RTT. It is capped at one smoothed RTT.
+	timeThresholdCushion time.Duration
+
 	// The number of times a PTO has been sent without receiving an ack.
 	ptoCount uint32
 	ptoMode  SendMode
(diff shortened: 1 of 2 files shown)


marten-seemann:
This will require some more careful testing and benchmarking.

PR #5753, opened 2026-07-18
add SendStream.WriteWithLimit for higher-level flow control

marten-seemann:
Closes #5736. Closes #5740.

> [!NOTE]
> ### Add `SendStream.WriteWithLimit` for limiter-driven partial writes
> - Adds `SendStream.WriteWithLimit` (and `Stream.WriteWithLimit`) accepting a caller-provided limiter function that caps how many bytes are consumed per packetization cycle; returns `ErrWriteLimitReached` when the limiter prevents the full write.
> - Extends flow controllers with `AddBytesSentWithLimiter` at both stream and connection level to slice send credit according to the limiter and signal when a write was partially accepted.
> - Replaces the blocking `WaitForReceiveFinalSize` API on `ReceiveStream` with `SetReceiveFinalSizeCallback`, which invokes the callback immediately if the final size is already known or on the first FIN/RESET_STREAM frame.
> - `SendStream.Write` is unchanged in behavior, delegating to `WriteWithLimit` with a nil limiter.
> - Risk: `canBufferStreamFrame` now returns false during any limited write, which changes pre-buffering behavior for concurrent writers using the new API.
>
> 
> #### Changes since #5753 opened
>
> - Updated `SendStream.WriteWithLimit` method documentation to clarify limiter semantics [ab8adc9]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fa20b46.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added `WriteWithLimit` to `SendStream` and the higher-level `Stream` API to cap per-call enqueued data.
  * Added `ErrWriteLimitReached` when the limiter prevents the full payload from being accepted.
* **Behavior Improvements**
  * When limiting is hit, `WriteWithLimit` returns `ErrWriteLimitReached` while still reporting the number of bytes successfully queued/written.
  * Existing `Write` behavior remains unchanged when no limiter is provided.
* **Tests**
  * Added unit and integration coverage for stream and connection flow-control scenarios involving `WriteWithLimit`.

### DIFF
--- a/send_stream.go
+++ b/send_stream.go
@@ -40,8 +40,8 @@ type SendStream struct {
 	dataForWriting []byte // during a Write() call, this slice is the part of p that still needs to be sent out
 
 	writeLimiter func(int) int
-	// Set by the packetizer when writeLimiter shortens a reservation. It makes the blocked
-	// WriteWithLimit return ErrWouldBlock and prevents another dequeue before it wakes up.
+	// Set by the packetizer when writeLimiter reduces the allowed byte count. It makes the blocked
+	// WriteWithLimit return ErrWriteLimitReached and prevents another dequeue before it wakes up.
 	writeLimited bool
 
 	nextFrame *wire.StreamFrame
@@ -104,7 +104,7 @@ func (s *SendStream) Write(p []byte) (int, error) {
 // WriteWithLimit writes data to the stream, subject to an additional send limit.
 // During packetization, limiter receives the bytes allowed for the next STREAM frame after
 // QUIC flow control and returns how many may be sent. Values outside [0, maxBytes] are clamped.
-// A short result returns the accepted prefix and [ErrWouldBlock]; the caller can wait
+// A short result returns the accepted prefix and [ErrWriteLimitReached]; the caller can wait
 // for external credit and retry the suffix. QUIC blocking behaves like [SendStream.Write].
 // limiter can run multiple times on another goroutine while QUIC send flow-control accounting
 // is locked. It must be concurrency-safe and must not block or call QUIC methods.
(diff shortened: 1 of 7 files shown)


MarcoPolo:
what do you think about a new method on the Stream that sets the limiter? Then you can also use this with `TryWriteAll`

send_stream.go
```diff
@@ -92,13 +98,25 @@ func (s *SendStream) StreamID() StreamID {
...
 func (s *SendStream) Write(p []byte) (int, error) {
+	return s.WriteWithLimit(p, nil)
+}
+
+// WriteWithLimit writes data to the stream, subject to an additional send limit.
+// During packetization, limiter receives the bytes allowed for the next STREAM frame after
+// QUIC flow control and returns how many may be sent. Values outside [0, maxBytes] are clamped.
+// A short result returns the accepted prefix and [ErrWriteLimitReached]; the caller can wait
+// for external credit and retry the suffix. QUIC blocking behaves like [SendStream.Write].
+// limiter can run multiple times on another goroutine while QUIC send flow-control accounting
```
MarcoPolo, line 111: "run multiple times"

When you return x bytes from the limiter, can you assume x bytes will be sent and you can lower your credits by x? In other words, when you return `x` have you commited to using that flow control credit?

MarcoPolo (approved):

send_stream.go
marten-seemann: I updated the comment

marten-seemann:
> what do you think about a new method on the Stream that sets the limiter? Then you can also use this with `TryWriteAll`

I don't think I want that? Would be nice if we don't have to store the callback?

Issue #5018, opened 2025-04-02
iOS (including macOS) socket buffer size


lzy201314:
I am writing an APP that integrates the go SDK and uses the quic-go framework. The following error is reported:
failed to sufficiently increase receive buffer size (was: 768 kiB, wanted: 7168 kiB, got: 6144 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for details.\

How do I set the maximum buffer size of the socket in iOS APP development? It seems that the iOS system has a maximum limit of 6MB.

marten-seemann:
You might want to ask that question in some iOS developer forum.

xf-yang:
`
quicConfig := &quic.Config{
		EnableDatagrams: true,
	}

	if runtime.GOOS == "ios" {
		quicConfig.MaxConnectionReceiveWindow = 1024 * 1024 * 6
		quicConfig.MaxStreamReceiveWindow = 1024 * 1024 * 6
	}`

marten-seemann:
@xf-yang This is wrong. The UDP has absolutely nothing to do with the OS UDP receive buffer.

Issue #3476, opened 2022-07-12
UDP receive buffer size on OpenBSD?


mrusme:
Hey there,

I'm running into an issue on OpenBSD where I get the following output:

```
failed to increase receive buffer size (wanted: 2048 kiB, got 0 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
```

Unfortunately the linked documentation doesn't describe how to do so. Also, resources that I found (e.g. https://comp.unix.bsd.openbsd.misc.narkive.com/X8fbptCR/how-to-raise-the-maxsockbuf-in-openbsd) doesn't seem to provide a feasible solution to this.

Is there a way quic-go can circumvent this issue?

Thank you!

marten-seemann:
I don’t have any experience with OpenBSD. A PR would be appreciated.

shassard:
I just tested this on OpenBSD 7.2 and have taken the following notes:

The correct sysctl to change is:
```
sysctl net.inet.udp.recvspace=...
```

While the maximum value supported by the kernel is 2147483647, resolving DNS hostnames via dig fails when the value is set beyond 2048000.

The maximum value doesn't seem to be affected by allocating more memory to the system (8GB vs 16GB available to the host at boot time).

o-alquimista:
> While the maximum value supported by the kernel is 2147483647, resolving DNS hostnames via dig fails when the value is set beyond 2048000.

On OpenBSD 7.8 I get "Invalid argument" for anything larger than 2097152 (~2 MB).

Anyway, that seems to be the right variable, based on `sysctl(2)`:

```
udp.recvspace (net.inet.udp.recvspace)
    Returns the default UDP receive buffer size.
```

Should we [update the wiki](https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes) with this information?

---

EDIT: Considering that my warning message is `(wanted: 7168 kiB, got 0 kiB)`, is quic-go even able to work with this? (with a maximum of 2 MB of buffer space). Also, what exactly does it mean when it says "got 0 kiB"? That it's got nothing to use at all? Or that it couldn't automatically increase the size (not one byte)?

o-alquimista:
[...]

PR #5785, opened 2026-08-12
probe for the largest supported receive and send buffer size

the-sarge:
Addresses #3476.

Some kernels reject `SO_RCVBUF` / `SO_SNDBUF` requests exceeding their limit outright instead of clamping, leaving the buffer at its (much smaller) default. `setReceiveBuffer` ignores the `SetReadBuffer` error and verifies via readback — correct on Linux, where oversized requests clamp to `rmem_max`, but on fail-atomic kernels the socket silently stays at the default.

Measured on OpenBSD 7.9/amd64: the kernel caps both buffer directions at exactly 2,097,152 bytes (`SB_MAX`, compile-time); larger requests fail with `ENOBUFS` and keep the previous value. Defaults are 41,600 B receive / 9,216 B send — so quic-go currently runs at ~41 kB receive and ~9 kB send there, with the OS willing to grant 2 MB of each.

This PR:

1. **Probes for the largest size the kernel accepts** when the desired size didn't take effect (after the `RCVBUFFORCE` attempt on Linux): a binary search verified by reading the value back, so Linux's silent clamping is handled and Linux/Windows behavior is unchanged. The helper lives in a separate file so the `go:generate`'d send-buffer twin shares it. Unit tests cover fail-atomic kernels, silently-clamping kernels, and inspection failure.
2. **Implements `inspectReadBuffer` / `inspectWriteBuffer` on OpenBSD** (same `GetsockoptInt` pattern as FreeBSD/Windows). Previously the stubs returned 0, which is why the warning on OpenBSD reported the fabricated "got 0 kiB".

Validated on OpenBSD 7.9 (no OpenBSD CI exists, so measurements below are from a clean OpenBSD 7.9/amd64 VM, Go 1.26.5):

- before: `failed to increase receive buffer size (wanted: 7168 kiB, got 0 kiB)` — socket left at 41,600 B
- after: `failed to sufficiently increase receive buffer size (was: 40 kiB, wanted: 7168 kiB, got: 2048 kiB)` — socket at exactly 2,097,152 B

To be clear about intent: this does **not** silence the buffer warning. On OpenBSD it still fires — now with accurate numbers — since 2 MB remains below the desired size. The change just stops leaving the OS maximum on the table. The same mechanism should help other fail-atomic platforms, e.g. Darwin at `kern.ipc.maxsockbuf` (#5018).

Happy to also add an OpenBSD section to the [UDP Buffer Sizes wiki page](https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes) (`SB_MAX` is compile-time; there is no sysctl to raise it).

> [!NOTE]
> ### Probe for largest supported socket receive and send buffer size via binary search
> - When `setReceiveBuffer` or `setSendBuffer` successfully sets a buffer but the result is below the desired size, a binary search is now performed between the current and desired sizes to find the largest value the kernel will accept.
> - The search is implemented in a new `largestBufferSize` helper in [sys_conn_buffers_common.go](https://github.com/quic-go/quic-go/pull/5785/files#diff-f5c8486c5f6099d81250f0d98e337f074ddb414e2e65b9b27a5d6b82cf376274), which uses a setter and an inspector (via `RawConn.Control` getsockopt) to determine the effective buffer size after each attempt.
> - Adds OpenBSD-specific implementations of `inspectReadBuffer` and `inspectWriteBuffer` in [sys_conn_openbsd.go](https://github.com/quic-go/quic-go/pull/5785/files#diff-d425a765b802e8a32d5adef3ca590c69699e5deafca3727593092f51bf3ffafa), along with a stub `newConn` and `packetInfo` to support the platform.
> - The build constraint in [sys_conn_no_oob.go](https://github.com/quic-go/quic-go/pull/5785/files#diff-e5560981ca2a90c26bde7300a04ea096bc86dc7640148ce3a9c17e5a3b2bbe51) is updated to exclude OpenBSD now that it has its own file.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b84d824.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added connection support for OpenBSD.
  - Improved socket buffer sizing to identify the largest value accepted by the operating system when requested sizes are capped or rejected.
  - Applied the improved sizing behavior to both receive and send buffers.

- **Bug Fixes**
  - Prevented buffer configuration from settling below the requested size when larger values are supported.

- **Tests**
  - Added coverage for capped, clamped, already-sufficient, and inspection-failure scenarios.

### DIFF
--- a/sys_conn_buffers_common.go
+++ b/sys_conn_buffers_common.go
@@ -8,9 +8,9 @@ package quic
 // its (much smaller) default size.
 // Success is verified by reading the size back, not by the error returned from
 // setting it: Linux, for example, silently clamps instead of failing.
-func largestBufferSize(current, desired int, set func(int) error, inspect func() (int, error)) int {
+func largestBufferSize(current, desired int, set func(int) error, inspect func() (int, error)) (int, error) {
 	if desired <= current {
-		return current
+		return current, nil
 	}
 	known := current // largest size read back after a successful set
 	// Invariant: lo took effect (or is the starting size), hi didn't.
@@ -20,7 +20,7 @@ func largestBufferSize(current, desired int, set func(int) error, inspect func()
 		_ = set(mid)
 		size, err := inspect()
 		if err != nil {
-			return known
+			return known, err
 		}
 		if size >= mid {
 			lo, known = mid, size
@@ -28,5 +28,5 @@ func largestBufferSize(current, desired int, set func(int) error, inspect func()
 			hi = mid
 		}
 	}
-	return known
+	return known, nil
 }
--- a/sys_conn_buffers_common.go
+++ b/sys_conn_buffers_common.go
@@ -1,0 +1,32 @@
+package quic
+
+// largestBufferSize finds the largest socket buffer size in (current, desired)
+// that the kernel accepts, and leaves the socket buffer set to it.
+// It is called after an attempt to set desired directly didn't take effect.
+// Some kernels (e.g. OpenBSD, Darwin) reject requests exceeding their limit
+// outright instead of clamping, so without probing the buffer would stay at
+// its (much smaller) default size.
+// Success is verified by reading the size back, not by the error returned from
+// setting it: Linux, for example, silently clamps instead of failing.
+func largestBufferSize(current, desired int, set func(int) error, inspect func() (int, error)) int {
+	if desired <= current {
+		return current
+	}
+	known := current // largest size read back after a successful set
+	// Invariant: lo took effect (or is the starting size), hi didn't.
+	lo, hi := current, desired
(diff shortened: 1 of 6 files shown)


marten-seemann (commented):
Is this AI-generated code?

sys_conn_buffers_common.go
marten-seemann, line 1: Why is this a separate file?

marten-seemann (commented):
I'm not sure I like this approach. Hammering the system with a bunch of syscalls to increase the buffer size for every QUIC connection seems suboptimal.

PR #5856, opened 2026-09-13
Fix PMTU discovery gating after path migration

### LINKED ISSUE #5844: Path migration ignores DisablePathMTUDiscovery
In quic-go v0.62.0, path migration starts MTU probing even when `DisablePathMTUDiscovery` is set. Discovery should remain disabled after migration.

Observed on IPv4 loopback with Go 1.26.0 on macOS arm64:

1. Establish a connection with `DisablePathMTUDiscovery: true` and `InitialPacketSize: 1200` on both endpoints.
2. Exchange application data. Neither endpoint sends MTU probes.
3. Migrate the client to a second UDP socket using `AddPath`, `Probe`, and `Switch`, then continue exchanging data.

Both endpoints then record acknowledged MTU probes of 1326, 1389, 1420, and 1436 bytes. The configuration is unchanged.

[Handshake confirmation](https://github.com/quic-go/quic-go/blob/v0.62.0/connection.go#L990-L992) checks `DisablePathMTUDiscovery` and the socket's DF capability before starting discovery. The migration paths instead call [`mtuFinder.Reset`](https://github.com/quic-go/quic-go/blob/v0.62.0/mtu_discoverer.go#L171-L177), which arms probing without either check.

Migration should always discard the old path's MTU estimate, but restart probing only when both checks allow it. For client migration, the DF check must use the new socket.

herefindalex:
## Summary

Fixes #5844.

Path migration currently restarts MTU probing even when `DisablePathMTUDiscovery` is set or the socket does not support DF.

Separate resetting the MTU state from starting discovery:

- Always discard the previous path’s MTU estimate and outstanding probe state on migration.
- Restart probing only when discovery is enabled and the socket supports DF.
- For client migration, check the **new transport’s socket**. Server migration continues checking its existing local socket.
- Preserve protection against stale probe acknowledgments and losses from the previous path.

## Potential impact

When discovery is disabled or the new socket lacks DF support, packet sizing remains at `InitialPacketSize` after migration. Compared with the buggy behavior, this may increase packet overhead and reduce throughput on paths that support larger packets. This is the expected consequence of honoring the discovery policy.

When discovery is allowed, migration still resets the learned MTU and probes again. The temporary use of smaller packets and additional probe traffic are existing migration behavior.

There are no public API changes.

## Validation

- Added client regression tests covering discovery settings and DF transitions in both directions, including old-socket DF support differing from the new socket.
- Added server migration and MTU reset assertions.
- Extended the migration integration test to run with discovery both enabled and disabled.
- Targeted unit and integration tests passed, including race checks.
- `go test -short ./...` passed.

Validation ran on Linux amd64 with Go 1.27.1. The reported macOS arm64 / Go 1.26.0 environment was not tested natively. The integration test verifies data transfer across migration; deterministic unit tests verify whether probing is armed.

> [!NOTE]
> ### Fix PMTU discovery gating after path migration
> - Changes `mtuFinder.Reset` to clear scheduling, loss, and in-flight state without making the finder eligible to probe; callers must now call `Start` explicitly to resume probing
> - Updates `Conn.switchToNewPath` and `Conn.handleShortHeaderPacket` to start MTU discovery only when path MTU discovery is enabled and the replacement socket is DF-capable
> - Adds unit and integration tests covering migration under both enabled and disabled discovery configurations with DF-capable and non-DF sockets
> - Risk: any out-of-tree caller of `mtuFinder.Reset` using the old signature (with timestamp) will fail to compile; the finder will also remain inactive after reset until `Start` is called, which could leave probing dormant if a caller relied on the old implicit-scheduling behavior
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1a2963c.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved path MTU discovery during connection migration.
  - MTU probing now resumes only when enabled and supported by the new connection path.
  - Prevented unnecessary or immediate MTU probes after moving to a new path.
  - Ensured MTU discovery remains disabled when configured, while resetting path-specific discovery state correctly.


marten-seemann:
This doesn't look like a clean fix.

PR #5859, opened 2026-09-14
http3: qlog SETTINGS in wire order and send them deterministically

itnovalv:
Settings that the `http3` package doesn't interpret are stored in a map, and the map is ranged over both when sending the SETTINGS frame and when logging it to qlog. Since Go randomizes map iteration order:

* the SETTINGS frame sent on the wire lists its additional settings in a random order,
* the qlog event for a received SETTINGS frame doesn't reflect the order of the settings on the wire, and
* the qlog event for a sent SETTINGS frame can list the settings in a different order than the one that was actually sent, since the map is ranged over a second time.

This also meant that qlog output for the same SETTINGS frame was not deterministic, which is why the existing tests only covered a single unknown setting.

The qlog HTTP/3 event schema models the settings of a SETTINGS frame as an ordered array. This PR changes `qlog.SettingsFrame` to carry the settings as an ordered `[]Setting`:

* When parsing, settings are recorded in the order in which they appear on the wire.
* When sending, `settingsFrame.Append` and the qlog event are derived from the same list, so the logged order always matches the order on the wire. Additional settings are now sent sorted by ID.

The JSON encoding is unchanged, apart from the order of the entries in the `settings` array.

Note that this changes the exported `qlog.SettingsFrame` type: the `MaxFieldSectionSize`, `Datagram`, `ExtendedConnect` and `Other` fields are replaced by `Settings []Setting`.

Tests cover parsing a SETTINGS frame with an order different from the one `Append` produces, the determinism of `Append` with multiple additional settings, and that the logged settings of a sent frame match the wire. `TestSettingsFrame` now also covers multiple unknown settings interleaved with known ones.

## Summary by CodeRabbit

* **Improvements**
  * SETTINGS events in qlog now preserve the order received on the wire.
  * Known and unknown settings are recorded consistently, improving protocol diagnostics.
  * SETTINGS frame serialization is deterministic, producing stable output across repeated operations.
  * Additional settings are included in connection-level qlog output instead of being omitted.

* **Compatibility**
  * qlog settings are now represented uniformly as ordered ID/value pairs, including recognized and unknown settings.

> [!NOTE]
> ### Make HTTP/3 SETTINGS qlog events preserve wire order and serialize deterministically
> - Replaces the qlog SETTINGS representation (separate named fields plus a map for unknowns) with an ordered slice of `qlog.Setting` values so both incoming and outgoing events reflect the actual wire order
> - Outgoing SETTINGS frames now serialize additional settings sorted by ascending ID instead of relying on map iteration order, via `settingsFrame.settings` and `settingsFrame.Append`
> - Incoming `parseSettingsFrame` collects settings in the order they appear on the wire and passes that ordered list to the `FrameParsed` qlog event
> - `rawConn.openControlStream` now uses `settingsFrame.qlogSettings` so outgoing `FrameCreated` events match the serialized byte order
> - Risk: `qlog.SettingsFrame` struct shape changes (named fields replaced by `Settings` slice) — any out-of-tree consumers reading the old named fields (`MaxFieldSectionSize`, `QpackBlockedStreams`, `Datagram`) will break
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 30e64aa.</sup>
> 
>


marten-seemann:
But why?

Issue #5854, opened 2026-09-13
http3: distinguish stream errors from connection errors


marten-seemann:
We currently convert both `*quic.StreamError` and `*quic.ApplicationError` to `*http3.Error`, losing the distinction between stream cancellation and connection closure. We should expose this distinction to applications, either through a field on `http3.Error` or by preserving the underlying error with `Unwrap()`.

Rayan-and-beyond:
I'd like to take this. I'll keep it focused on preserving the distinction between stream-level cancellation and connection-level closure through the http3 error layer, with compatibility-focused tests around errors.Is/errors.As (or the smallest API surface the current error model supports).

PR #5862, opened 2026-09-17
http3: preserve underlying QUIC error type

Rayan-and-beyond:
Follow-up to #5852. Addresses #5854.

`maybeReplaceError` currently converts both `*quic.StreamError` and `*quic.ApplicationError` into `*http3.Error`, which loses the distinction between stream cancellation and connection closure. This keeps the original QUIC error as the wrapped cause and exposes it through `http3.Error.Unwrap()`, so existing `errors.As` checks can still distinguish the two while callers retain the normalized HTTP/3 code and `Remote` flag.

The conversion tests verify both underlying error types are preserved, and the datagram integration assertions use the existing `errors.Is` contract rather than deep equality now that `http3.Error` carries cause state.

Validation on top of `http3-consistent-errors` (`b0db92a0`):
- `GOMAXPROCS=1 go test -p=1 -parallel=1 ./http3`
- `GOMAXPROCS=1 go test -p=1 -parallel=1 ./integrationtests/self -run 'TestHTTPStreamErrors|TestHTTPDatagramStreamReset' -count=1`
- `GOMAXPROCS=1 go vet ./http3`
- `git diff --check b0db92a0..HEAD`

This is opened as a draft against #5852's branch so the review diff contains only the #5854 follow-up. It can be retargeted to `master` once #5852 lands.

> [!NOTE]
> ### Preserve underlying QUIC error type in `http3` error conversion
> - `maybeReplaceError` now wraps the converted HTTP/3 Error in a new `errorWithCause` type that retains the original QUIC stream or application error as an unwrap-able cause
> - The wrapper delegates `Error` and `Is` to the HTTP/3 Error, exposes the QUIC error via `Unwrap`, and exposes the HTTP/3 Error via `As` for standard Go error-chain inspection
> - Tests and the datagram reset integration test switch from direct equality to `errors.Is` assertions to accommodate the wrapper
> - Behavioral Change: errors returned by the conversion path are now `*errorWithCause` instead of `*Error`; callers using `errors.As` or `errors.Is` are unaffected, but direct type assertions on `*Error` will no longer match — check `maybeReplaceError` in [error.go](https://github.com/quic-go/quic-go/pull/5862/files#diff-9ad4b0639a0bd9e505e4e2fe0d59dd9633c0942c39471582e025d8ee22183dd3)
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c72344f.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP/3 error handling so wrapped errors retain their original QUIC error types.
  * Preserved HTTP/3 error codes and remote/local indicators when identifying errors through wrapping.
  * Error matching now reliably recognizes both converted HTTP/3 errors and their underlying QUIC errors.

* **Tests**
  * Expanded coverage for error conversion, error matching, QUIC error types, and datagram stream resets.


Rayan-and-beyond:
fixed the compatibility concern. the exported Error stays a three-field struct, and the quic cause now lives in a private wrapper that preserves errors.As for both *http3.Error and the underlying quic error. focused http3 tests + go vet are green.

PR #5875, opened 2026-09-21
fix TestSysConnPacketInfo on macOS by dialing the loopback address

tsuna:
`runSysConnServer` listens on the wildcard address, so the address it returns is `0.0.0.0:<port>` (or `[::]:<port>`). `TestSysConnPacketInfoIPv4` and `TestSysConnPacketInfoIPv6` dial that address directly, then assert that `conn.LocalAddr()` matches the source address the server observed.

Two things combine to break this on macOS:
1. `connect()`ing a UDP socket to the wildcard address does not assign a local port. `getsockname(2)` reports `127.0.0.1:0` immediately after `connect()`, while connecting to `127.0.0.1` instead assigns a port right away. A port is only assigned once a datagram has been sent.
2. `net.DialUDP` caches the local address at dial time, so `conn.LocalAddr()` keeps reporting port 0 for the lifetime of the socket, even after the kernel has assigned one.

The assertion therefore compares a cached `127.0.0.1:0` against the real source port the server saw:

```
    Error:      Not equal:
                expected: &net.UDPAddr{IP:..., Port:0}
                actual  : &net.UDPAddr{IP:..., Port:58285}
```

This does not happen on Linux (`connect()` to the wildcard address remaps to loopback), which is why CI has stayed green.

Dial the loopback address explicitly, keeping the server bound to the wildcard address (which is required for the packet info to be returned). This matches what `TestSysConnPacketInfoDualStack` already does.

> [!NOTE]
> ### Fix `TestSysConnPacketInfo` on macOS by using loopback addresses
> Changes the UDP client destination in `TestSysConnPacketInfoIPv4` and `TestSysConnPacketInfoIPv6` from the server's wildcard address to the explicit loopback address (`127.0.0.1` / IPv6 loopback), keeping the dynamically assigned server port. On macOS, dialing the wildcard address does not yield a usable local address, which causes the packet metadata comparison to fail.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e3e283d.</sup>
> 
>

## Summary by CodeRabbit

* **Tests**
  * Improved IPv4 and IPv6 socket tests by connecting through loopback addresses, ensuring reliable local port detection on macOS.


tsuna:
To observe the behavior difference between Linux and macOS, I used this test program:

```c
#include <stdio.h>
#include <string.h>
#include <arpa/inet.h>
#include <sys/socket.h>
#include <unistd.h>

static void show(const char *label, int fd) {
    struct sockaddr_in sa; socklen_t l = sizeof(sa);
    getsockname(fd, (struct sockaddr *)&sa, &l);
    char ip[64]; inet_ntop(AF_INET, &sa.sin_addr, ip, sizeof ip);
    printf("    %-24s %s:%d\n", label, ip, ntohs(sa.sin_port));
}

static void run(const char *label, const char *dial_ip, int srvport, int srvfd) {
    printf("  %s\n", label);
    int fd = socket(AF_INET, SOCK_DGRAM, 0);
    struct sockaddr_in to; memset(&to, 0, sizeof to);
    to.sin_family = AF_INET;
    to.sin_port = htons(srvport);
    inet_pton(AF_INET, dial_ip, &to.sin_addr);
    if (connect(fd, (struct sockaddr *)&to, sizeof to) < 0) { perror("connect"); return; }
    show("after connect:", fd);
    for (int i = 0; i < 2; i++) {
        send(fd, "x", 1, 0);
        char buf[64]; struct sockaddr_in from; socklen_t fl = sizeof from;
        recvfrom(srvfd, buf, sizeof buf, 0, (struct sockaddr *)&from, &fl);
        char ip[64]; inet_ntop(AF_INET, &from.sin_addr, ip, sizeof ip);
        printf("    server saw src #%d:       %s:%d\n", i + 1, ip, ntohs(from.sin_port));
        show("      getsockname now:", fd);
    }
    close(fd);
}

int main(void) {
    int srv = socket(AF_INET, SOCK_DGRAM, 0);
    struct sockaddr_in sa; memset(&sa, 0, sizeof sa);
    sa.sin_family = AF_INET;
    sa.sin_addr.s_addr = htonl(INADDR_ANY);
    sa.sin_port = 0;
    bind(srv, (struct sockaddr *)&sa, sizeof sa);
    socklen_t l = sizeof sa;
    getsockname(srv, (struct sockaddr *)&sa, &l);
    int port = ntohs(sa.sin_port);
    printf("server bound to 0.0.0.0:%d\n", port);
    run("connect() to 0.0.0.0 :", "0.0.0.0", port, srv);
    run("connect() to 127.0.0.1 :", "127.0.0.1", port, srv);
    return 0;
}
```

macOS output:
```
server bound to 0.0.0.0:57639
  connect() to 0.0.0.0 :
    after connect:           127.0.0.1:0      <----- !!
    server saw src #1:       127.0.0.1:54736
          getsockname now:   127.0.0.1:54736
    server saw src #2:       127.0.0.1:54736
          getsockname now:   127.0.0.1:54736
  connect() to 127.0.0.1 :
    after connect:           127.0.0.1:57163
    server saw src #1:       127.0.0.1:57163
```

Linux output:
```
server bound to 0.0.0.0:58681
  connect() to 0.0.0.0 :
    after connect:           127.0.0.1:51749
    server saw src #1:       127.0.0.1:51749
          getsockname now:   127.0.0.1:51749
    server saw src #2:       127.0.0.1:51749
  connect() to 127.0.0.1 :
    after connect:           127.0.0.1:60332
    server saw src #1:       127.0.0.1:60332
```

marten-seemann:
I was not aware it was broken in the first place.

benoit-nexthop:
> I was not aware it was broken in the first place.

Yeah, no worries!

marten-seemann:
Maybe you can explain? Did it fail on your machine? Flaky?

benoit-nexthop:
> Maybe you can explain? Did it fail on your machine? Flaky?

No it's not flaky, it just doesn't pass on macOS due to aforementioned behavior difference with Linux.

marten-seemann:
That’s weird because it works on my Mac, and it works on the Mac that we are running in CI.

tsuna:
That's weird. Can you run the above C program and see if it prints anything different? I didn't mention it before but I'm on Go 1.27.1, macOS Tahoe 26.6.2 (`uname -a` = `Darwin damogran 25.6.0 Darwin Kernel Version 25.6.0: Fri Jul 31 19:11:03 PDT 2026; root:xnu-12377.161.14~5/RELEASE_ARM64_T8132 arm64`)

marten-seemann:
First of all, what’s the problem you’re trying to solve? Just fixing a test case doesn’t seem to be worth the hassle.

tsuna:
I have a fork of quic-go where I added BRRv1 support for congestion control, the tests have always passed until I rebased again earlier today. I was just trying to help by fixing what seemed to me like a broken test on macOS.

marten-seemann:
Thanks for the explanation. I'm going to close this, since I don't see a problem that needs to be resolved.

